PR #11, opened 2022-07-06
update init description and remove template flag temporarily

yoav-lavi:
# Description

## Features

- Updates `init` command description

## Refactoring

- Removes the `init` `--template` flag as it's not implemented yet

# Type of change

- [ ] 💔 Breaking
- [x] 🚀 Feature
- [ ] 🐛 Fix
- [ ] 🛠️ Tooling
- [x] 🔨 Refactoring
- [ ] 🧪 Test
- [ ] 📦 Dependency

### DIFF
--- a/cli/crates/cli/src/cli_input.rs
+++ b/cli/crates/cli/src/cli_input.rs
@@ -33,10 +33,10 @@ pub fn build_cli() -> Command<'static> {
             Command::new("init")
                 .args(&[
                     arg!([name] "the name of the project to create"),
-                    arg!(-t --template "the name of the template to use for the new project"),
+                    // arg!(-t --template "the name of the template to use for the new project"),
                 ])
                 .about(indoc! {"
-                Creates a grafbase/schema.graphql file and folder in the current directory
+                Sets up the current or a new project for Grafbase
             "}),
         )
     // .subcommand(Command::new("login").about("TBD"))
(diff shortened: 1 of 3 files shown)


Miaxos (approved):

cli/crates/cli/src/cli_input.rs
```diff
@@ -33,10 +33,10 @@ pub fn build_cli() -> Command<'static> {
             Command::new("init")
                 .args(&[
                     arg!([name] "the name of the project to create"),
-                    arg!(-t --template "the name of the template to use for the new project"),
+                    // arg!(-t --template "the name of the template to use for the new project"),
```
mlafeldt, line 36: Not a fan of commenting out code. That's what the git history is for. 😄

mlafeldt (approved):

cli/crates/cli/src/cli_input.rs
yoav-lavi: Normally I'd be the one saying this, the idea here specifically is to outline the future interface (as the commented code below is doing). It's less of "I'm temporarily disabling this since it doesn't work" and more "this is the planned interface"

yoav-lavi: We could still remove it, but I saw it as more of a description of the future additions

PR #44, opened 2022-08-05
Added unique directive to todo template titles.

DougAtGrafbase:
# Description

I don't see anyone wanting two todo lists with the same title or two todo list items with the same title in the same list (I'm not sure whether adding unique to the latter type accomplishes that goal).

# Type of change

- [ ] 💔 Breaking
- [ ] 🚀 Feature
- [ ] 🐛 Fix
- [ ] 🛠️ Tooling
- [ ] 🔨 Refactoring
- [ ] 🧪 Test
- [ ] 📦 Dependency

### DIFF
--- a/templates/todo/grafbase/schema.graphql
+++ b/templates/todo/grafbase/schema.graphql
@@ -1,12 +1,12 @@
 type TodoList @model {
   id: ID!
-  title: String!
+  title: String! @unique
   todos: [Todo]
 }
 
 type Todo @model {
   id: ID!
   list: TodoList
-  title: String!
+  title: String! @unique
   complete: Boolean!
 }


yoav-lavi:
@DougAtGrafbase I agree regarding the list names, but something like a `Todo` could have a non unique name in an application (e.g. for recurring TODOs)

yoav-lavi:
Although as a counter example to the point on lists 😄 
Screen Shot 2022-08-05 at 19 28 05

DougAtGrafbase:
Yeah, being able to prevent a todo list from having multiple instances of the same item is probably going to have to wait until we have custom resolvers.

But you see my point. If I have two items in the same list, but one is complete and the other isn't, which is correct?

yoav-lavi:
`@unique` would work for a Todo as well as far as I'm aware, I meant that as a template I'm not sure unique by default would necessarily make sense, since the average TODO app would let you duplicate items (since they can have different dates / be recurring / be assigned to different people etc.)

Although again since this is an example it's not necessarily critical

yoav-lavi:
If you mean unique on the `Todo` item in the `TodoList` - IIRC `@unique` is only for primitives

yoav-lavi:
Anyhow since it's a template and to showcase the feature it may make sense, just a thought rather than a blocker

vitalikda:
Do we need to update examples with this change?

templates/todo/grafbase/schema.graphql
```diff
@@ -1,12 +1,12 @@
 type TodoList @model {
   id: ID!
-  title: String!
+  title: String! @unique
```
fbjork, line 3: Why restrict the title to be unique? I have multiple todo lists with the same name in my Notes app

fbjork:
Closing since we don't want to restrict users from creating toto lists or todos with the same title

PR #46, opened 2022-08-08
Fix TS errors for GRAFBASE_API_KEY env var

mlafeldt:
Which might be undefined, but we require it.

# Type of change

- [ ] 💔 Breaking
- [ ] 🚀 Feature
- [x] 🐛 Fix
- [ ] 🛠️ Tooling
- [ ] 🔨 Refactoring
- [ ] 🧪 Test
- [ ] 📦 Dependency

### DIFF
--- a/examples/nextjs/src/utils/server-side-fetch.ts
+++ b/examples/nextjs/src/utils/server-side-fetch.ts
@@ -1,10 +1,9 @@
 const serverSideFetch = async (query: string) =>
-  await fetch(process.env.GRAFBASE_API_URL as string, {
+  await fetch(process.env.GRAFBASE_API_URL!, {
     method: 'POST',
-    // @ts-ignore
     headers: {
       'Content-Type': 'application/json',
-      'x-api-key': process.env.GRAFBASE_API_KEY
+      'x-api-key': process.env.GRAFBASE_API_KEY!
     },
     body: JSON.stringify(query)
   }).then((response) => response.json())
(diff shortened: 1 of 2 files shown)


examples/nextjs/src/utils/server-side-fetch.ts
```diff
@@ -1,10 +1,9 @@
 const serverSideFetch = async (query: string) =>
-  await fetch(process.env.GRAFBASE_API_URL as string, {
+  await fetch(process.env.GRAFBASE_API_URL!, {
     method: 'POST',
-    // @ts-ignore
     headers: {
       'Content-Type': 'application/json',
-      'x-api-key': process.env.GRAFBASE_API_KEY
+      'x-api-key': process.env.GRAFBASE_API_KEY!
```
fbjork, line 6: How will this work for local dev?

yoav-lavi: It's not needed in local dev AFAIK

vitalikda (approved):

examples/nextjs/src/utils/server-side-fetch.ts
vitalikda: Yes, you don't need it, just set `GRAFBASE_API_URL=http://localhost:4000` in .env file

mlafeldt: The examples will set the API key to an empty string via the env file, e.g. https://github.com/grafbase/grafbase/blob/main/examples/nextjs/.env.example#L1

So this should work OOTB.

PR #76, opened 2022-09-07
use nextest, rename tests

yoav-lavi:
# Description

## Testing

- Uses `nextest`
- Renames 2 test files

# Type of change

- [ ] 💔 Breaking
- [ ] 🚀 Feature
- [ ] 🐛 Fix
- [ ] 🛠️ Tooling
- [ ] 🔨 Refactoring
- [x] 🧪 Test
- [ ] 📦 Dependency


.github/workflows/cli.yml
```diff
@@ -125,17 +125,30 @@ jobs:
...
+        with:
+          command: binstall
+          args: 'cargo-nextest --secure'
+
       - name: test
         uses: actions-rs/cargo@v1
         with:
-          command: test
+          command: nextest
+          args: '--test-threads 1'
```
mlafeldt, line 144: What's the reason for that? Doesn't it defeat one of the main advantages of using nextest?

mlafeldt (approved):
LGTM! Just one question

mlafeldt:
> https://github.com/grafbase/grafbase/runs/8223686460?check_suite_focus=true

There seems to be an issue with installing where it's expecting an interactive environment.

yoav-lavi:
@mlafeldt thanks for reviewing! I actually meant to open this as a draft but I opened it from VSCode and so it ended up being  ready, looking into that issue

.github/workflows/cli.yml
yoav-lavi: The CLI currently can't be run more than once in the exact same time due to finding a port not being an atomic operation (two CLIs find the same bridge port and one of them errors), so it can't run in parallel yet, will look into it soon. I tried creating a struct which holds a listener and a port number to be able to drop it before using the port (effectively "locking" it), but that has problems with the OS freeing the port in time and probably isn't the most efficient

mlafeldt: Ah, in that case you might want to add a [config file](https://github.com/grafbase/api/blob/main/.config/nextest.toml) (with a comment) so that nextest will work locally and in CI.

yoav-lavi:
Closing this for now, will circle back when I have more time to look into it

PR #77, opened 2022-09-08
GB-2050: Infer environment variables from the standard environment and `dotenv`

jakubadamw:
# Description

This enables support for environment variables inferred from the environment of the process + .env files as supported by the [`dotenv`](https://crates.io/crates/dotenv) package.

# Type of change

- [ ] 💔 Breaking
- [X] 🚀 Feature
- [ ] 🐛 Fix
- [ ] 🛠️ Tooling
- [ ] 🔨 Refactoring
- [ ] 🧪 Test
- [ ] 📦 Dependency

### DIFF
--- a/cli/crates/server/src/servers.rs
+++ b/cli/crates/server/src/servers.rs
@@ -291,9 +291,9 @@ async fn run_schema_parser() -> Result<(), ServerError> {
             .spawn()
             .map_err(ServerError::SchemaParserError)?;
 
-        let node_command_stdin = node_command.stdin.as_mut().unwrap();
+        let node_command_stdin = node_command.stdin.as_mut().expect("stdin must be available");
         node_command_stdin
-            .write_all(&serde_json::to_vec(&environment_variables).unwrap())
+            .write_all(&serde_json::to_vec(&environment_variables).expect("must serialise to JSON just fine"))
             .await
             .map_err(ServerError::SchemaParserError)?;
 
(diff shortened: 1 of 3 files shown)


cli/crates/server/src/servers.rs
```diff
@@ -265,29 +274,40 @@ async fn run_schema_parser() -> Result<(), ServerError> {
...
+                    .project_grafbase_schema_path
+                    .to_str()
+                    .ok_or(ServerError::ProjectPath)?,
+            ])
+            .current_dir(&environment.project_dot_grafbase_path)
+            .stderr(Stdio::piped())
+            .spawn()
+            .map_err(ServerError::SchemaParserError)?;
+
+        let node_command_stdin = node_command.stdin.as_mut().unwrap();
```
yoav-lavi, line 293: Is there a valid case this could happen? I'd use `expect` if it shouldn't happen or handle the error if it should

cli/crates/server/src/servers.rs
```diff
@@ -265,29 +274,40 @@ async fn run_schema_parser() -> Result<(), ServerError> {
...
+        node_command_stdin
+            .write_all(&serde_json::to_vec(&environment_variables).unwrap())
```
yoav-lavi, line 295: See above

cli/crates/server/src/servers.rs
jakubadamw: Good point, changed to `expect()`.

yoav-lavi (approved):
Looks good!

PR #95, opened 2022-10-06
Add updatedAt and createdAt to scalars test

cnixon-gb:
# Description

This adds an additional test to the cli to check that updatedAt and createdAt are returned on models and that they are in the correct format

# Type of change

- [ ] 💔 Breaking
- [ ] 🚀 Feature
- [ ] 🐛 Fix
- [ ] 🛠️ Tooling
- [X] 🔨 Refactoring
- [X] 🧪 Test
- [ ] 📦 Dependency

### DIFF
--- a/cli/Cargo.lock
+++ b/cli/Cargo.lock
@@ -231,9 +231,9 @@ dependencies = [
 
 [[package]]
 name = "clap"
-version = "4.0.10"
+version = "4.0.12"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "3b1a0a4208c6c483b952ad35c6eed505fc13b46f08f631b81e828084a9318d74"
+checksum = "385007cbbed899260395a4107435fead4cad80684461b3cc78238bdcb0bad58f"
 dependencies = [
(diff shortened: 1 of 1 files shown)


fbjork (approved):

cli/crates/server/src/bridge/types.rs
```diff
@@ -57,6 +65,8 @@ pub struct Record {
     pub entity_type: String,
     pub relation_names: Value,
     pub document: Value,
-    pub created_at: String,
-    pub updated_at: String,
+    #[serde(serialize_with = "serialize_dt_to_rfc3339")]
+    pub created_at: DateTime<Utc>,
```
yoav-lavi, line 69: @cnixon-gb Is this already done in the worker side? We originally had it like this but removed it both due to the warning in `chrono` and due to it basically adding an unneeded operation

yoav-lavi: Minus the tag of course

yoav-lavi (approved):
Nice!

PR #130, opened 2022-11-29
feat(live): add apollo-link example

vitalikda:
# Description

# Type of change

- [ ] 💔 Breaking
- [x] 🚀 Feature
- [ ] 🐛 Fix
- [ ] 🛠️ Tooling
- [ ] 🔨 Refactoring
- [ ] 🧪 Test
- [ ] 📦 Dependency


notrab:
I notice here we're using pnpm and other examples Yarn.

Do we want to default to pnpm for external users too? Or stick with NPM to less to setup/configure. I feel like if someone knows how to run pnpm, they will instead of NPM.

examples/react-apollo-live/.env.example
```diff
@@ -0,0 +1,3 @@
+# Refer to the README on how to fill these variables
+VITE_GRAFBASE_API_URL=
+VITE_CLERK_FRONTEND_API=
```
notrab, line 3: We should maybe include a comment at the top of this file that the Clerk URL must be the full absolute URL, so we can remove the protocol from the `schema.graphql` and only rely on the true ENV value threre.

examples/react-apollo-live/grafbase/schema.graphql
```diff
@@ -0,0 +1,13 @@
+schema
+  @auth(
+    providers: [{ type: oidc, issuer: "https://{{ env.CLERK_ISSUER_URL }}" }]
```
notrab, line 3:
```suggestion
    providers: [{ type: oidc, issuer: "{{ env.CLERK_ISSUER_URL }}" }]
```

We don't tend to do this in other examples or documentation. Instead we rely on using environment variables for absolute values.

notrab (changes requested):
Hey @vitalikda 

This is awesome. Nice job!

I've left a few comments which we should also apply to the URQL example too.

Having to login to view the guestbook is quite annoying but I know we've a product limitation that prevents up from making it public. So here's what I suggest we could potentially do to make it a smoother experience for new users:

1. Create a default JWT that expires way in the future. With a group that has access to "Read and Write".
2. Create another JWT with a group with operation permissions to "Read only"
3. Add some buttons to the top of the page that switches the current active token &mdash; this showcases group-based auth too!
4. Done

An example of the buttons to switch "current user" token can be seen in this video: https://youtu.be/vwUGlwbdfvI?t=47

examples/react-apollo-live/README.md
```diff
@@ -0,0 +1,13 @@
+# Grafbook: Server sent events (SSE) x Clerk ⨯ React
+
+## Getting Started
+
+1. Run `npx degit grafbase/grafbase/examples/react-clerk-live grafbase-with-react-clerk-live` to clone this example
+1. Change directory into the new folder `cd grafbase-with-react-clerk-live`
+1. Run `cp .env.example .env` to copy the example `.env.example` file to `.env` in the root of your project
+1. Open `.env` in your code editor, and provide your Grafbase API endpoint and [Clerk frontend and backend API keys](https://dashboard.clerk.dev/last-active?path=api-keys)
```
notrab, line 8: We could remove this step entirely for demo purposes if we used the JWT provider. Let me know what you think about the suggestion below in the review comments.

examples/react-apollo-live/.env.example
vitalikda: Should we? I find it easier to copy-paste exactly how Clerk provides URL in their dashboard

Also, note that `VITE_CLERK_FRONTEND_API` is used in ClerkProvider without URL protocol, so we would have two versions of the same thing 
- `VITE_CLERK_FRONTEND_API` without protocol 
- `CLERK_ISSUER_URL` with protocol

notrab: I think we should yeah. They are separate `.env` files after all, so you'd have the same duplication anyways.

examples/react-apollo-live/README.md
vitalikda: done

examples/react-apollo-live/package.json
```diff
@@ -0,0 +1,26 @@
+{
+  "name": "apollo-link-ts",
+  "private": true,
+  "version": "0.0.0",
```
notrab, line 4: maybe remove `name` and `version`?

examples/react-apollo-live/README.md
```diff
@@ -0,0 +1,12 @@
+# Grafbook: Server sent events (SSE) x JWT ⨯ React
```
vitalikda, line 1:
```suggestion
# Grafbook: Server sent events (SSE) x Apollo ⨯ React
```

examples/react-apollo-live/package.json
```diff
@@ -0,0 +1,26 @@
+{
+  "name": "apollo-link-ts",
```
vitalikda, line 2:
```suggestion
  "name": "apollo-live-example",
```

examples/react-apollo-live/package.json
vitalikda: updated name

PR #131, opened 2022-12-06
0.14.0

yoav-lavi:
# Description

## Features

- Adds an event listener which is fed by sqlite triggers writing into an additional table. This allows us to send modification events to the stream-router and simulate production.

## Fixes

- Reports errors in the bridge rather than restarting it

## Dependencies

- Adds `reqwest`, `chrono` and `uuid`

## Misc.

- Bumps versions before release and adds changelog

# Type of change

- [ ] 💔 Breaking
- [x] 🚀 Feature
- [x] 🐛 Fix
- [ ] 🛠️ Tooling
- [ ] 🔨 Refactoring
- [ ] 🧪 Test
- [x] 📦 Dependency


cli/crates/server/src/bridge/listener/server.rs
```diff
@@ -0,0 +1,89 @@
...
+        if !results.is_empty() {
+            let dynamo_events = results
+                .iter()
+                .map(|result| EventRecord {
+                    aws_region: "us-east-1".to_owned(),
+                    change: StreamRecord {
+                        // as we poll 10 times per second and this is rounded to seconds,
+                        // using the current time is accurate enough.
+                        // using the record update time would be inaccurate for deletions
+                        approximate_creation_date_time: Utc::now().timestamp_millis() / 1000,
```
Finistere, line 53: OPT: It's ok, but couldn't just add the current timestamp in the SQLite trigger instead?

Finistere (approved):

cli/crates/server/src/bridge/listener/server.rs
yoav-lavi: We could probably just move this out of the loop for the same effect

yoav-lavi: I'm not sure how accurate we need to be here

yoav-lavi: But if SQLite has a nice way to generate a timestamp we could do that as well and add an extra column

yoav-lavi: @Finistere done

Finistere: Ah missed the update! Thanks for the change, looks good to me!

PR #137, opened 2022-12-14
add qwik example

hhthacker:
# Description

# Type of change

- [ ] 💔 Breaking
- [ ] 🚀 Feature
- [ ] 🐛 Fix
- [ ] 🛠️ Tooling
- [ ] 🔨 Refactoring
- [ ] 🧪 Test
- [ ] 📦 Dependency


examples/qwik/src/utils/grafbase.ts
```diff
@@ -0,0 +1,18 @@
...
+  query: string
+  variables: Record<string, any>
+}): Promise<T> => {
+  const response = await fetch(
+    import.meta.env.VITE_GRAFBASE_API_URL as string,
+    {
+      method: 'POST',
+      headers: {
+        'content-type': 'application/json',
+        'x-api-key': import.meta.env.VITE_GRAFBASE_API_KEY as string
```
notrab, line 11: The `x-api-key` shouldn't be used on the client.

Is there a way with Qwik to process requests on the server?

notrab (changes requested):
Looks great @hhthacker!

Most of the comments I left in the old PR at `grafbase/examples`.

One more thing I didn't realise before was that the `/graphql` endpoint was called on the client. We should consider removing the mutation example from this and instead just read data from the backend to simplify the example.

examples/qwik/src/routes/index.tsx
```diff
@@ -0,0 +1,109 @@
...
+          id
+          name
+          description
+        }
+      }
+    }
+  }
+`
+
+export const AddNewPlantMutation = `
```
notrab, line 30: I'd get rid of this for the example. We can just read from Grafbase and add to the README details on how to run the mutation using the playground.

examples/qwik/src/routes/index.tsx
```diff
@@ -0,0 +1,109 @@
...
+  }
+}
+
+interface Plant {
+  id: string
+  name: string
+  description: string
+}
+
+export const GetAllPlantsQuery = `
```
notrab, line 16: I'd update this to be:

```ts
type Message = {
  id: string
  author: string
  message: string
  createdAt: string
}

export const GetAllMessagesQuery = /* GraphQL */ `
  query GetAllMessages($first: Int!) @live {
    messageCollection(first: $first) {
      edges {
        node {
          id
          author
          message
          createdAt
        }
      }
    }
  }
`
```

examples/qwik/src/routes/index.tsx
```diff
@@ -0,0 +1,109 @@
...
+        description
+      }
+    }
+  }
+`
+
+export default component$(() => {
+  const newPlant = useSignal('')
+  const newPlantDescription = useSignal('')
+  const allPlants = useSignal<Plants>()
```
notrab, line 45: We could change this to be something like:

```ts
const messages = useSignal<Message[]>([])
```

examples/qwik/src/routes/index.tsx
```diff
@@ -0,0 +1,109 @@
...
+
+  useServerMount$(async () => {
+    const plants: Plants = await grafbaseClient({
```
notrab, line 48: We could for the purposes of this example ditch the util and just use `fetch` here since we only use it once:

```ts
 const {
      data: { messageCollection }
    }: {
      data: { messageCollection: { edges: { node: Message }[] } }
    } = await fetch(import.meta.env.VITE_GRAFBASE_API_URL as string, {
      method: 'POST',
      headers: {
        'content-type': 'application/json'
      },
      body: JSON.stringify({
        query: GetAllMessagesQuery,
        variables: {
          first: 100
        }
      })
    }).then((res) => res.json())
```

examples/qwik/src/routes/index.tsx
```diff
@@ -0,0 +1,109 @@
...
+      query: GetAllPlantsQuery,
+      variables: { first: 100 }
+    })
+    allPlants.value = plants
```
notrab, line 52: I don't know much about `.value` but we could satisfy the response type above in my example to we just return the edges as the value, making the types simpler for the signal:

```ts
messages.value = messageCollection?.edges || []
```

examples/qwik/src/routes/index.tsx
```diff
@@ -0,0 +1,109 @@
...
+      <h1>Plants</h1>
+      {allPlants.value?.plantCollection.edges.map(({ node }) => (
+        <>
+          <div>
+            {node?.name} : {node?.description}
+          </div>
+        </>
+      ))}
+
+      <h2>New plant</h2>
```
notrab, line 66: We can get rid of this for the example as these will run on client. We can add a separate "auth" example with Qwik later that lets users pass the `authorization` header.

examples/qwik/src/routes/index.tsx
```diff
@@ -0,0 +1,109 @@
...
+    }
+  }
+`
+
+export default component$(() => {
+  const newPlant = useSignal('')
+  const newPlantDescription = useSignal('')
+  const allPlants = useSignal<Plants>()
+
+  useServerMount$(async () => {
```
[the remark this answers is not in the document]
hhthacker: searched through the discord, `useServerMount()` was deprecated while i was out of town, replaced with useTask() - release notes: https://github.com/BuilderIO/qwik/releases/tag/v0.16.0

examples/qwik/src/root.tsx
```diff
@@ -0,0 +1,29 @@
+import { component$, useStyles$ } from '@builder.io/qwik';
+import { QwikCityProvider, RouterOutlet, ServiceWorkerRegister } from '@builder.io/qwik-city';
+import { RouterHead } from './components/router-head/router-head';
+
+import globalStyles from './global.css?inline';
```
notrab, line 5: @hhthacker we can for the purpose of this example remove the styles and follow the semantics of the react-apollo-live examples.

Later this month we should create a shared stylesheet across examples so they follow the same theme and link to other examples. Will need @designhorf to pitch in.

examples/qwik/src/routes/index.tsx
```diff
@@ -0,0 +1,58 @@
...
+  const messages = useSignal<Message[]>([])
+
+  const {
+    data: { messageCollection }
+  }: {
+    data: { messageCollection: { edges: { node: Message }[] } }
+  } = await fetch(import.meta.env.VITE_GRAFBASE_API_URL as string, {
+    method: 'POST',
+    headers: {
+      'content-type': 'application/json'
```
notrab, line 35: We'll need to pass the `authorization` header here. Does `component$` run on the server?

examples/qwik/src/root.tsx
hhthacker: removed styles!

notrab (approved):
Looks great @hhthacker

notrab:
I'll push an update to use the `.env` endpoint instead of the hardcoded `127.0.0.1`.

PR #138, opened 2022-12-14
gb2697 fix pagination

Finistere:
- Add Nix Flake for dev shell (linux)
- fix(cli/tests): Fix Pagination tests

# Description

Add Nix for local development shell (linux-only for now!). 
Fix pagination tests.

# Type of change

- [ ] 💔 Breaking
- [ ] 🚀 Feature
- [X] 🐛 Fix
- [ ] 🛠️ Tooling
- [ ] 🔨 Refactoring
- [ ] 🧪 Test
- [ ] 📦 Dependency

Fixes: GB-2697


cli/crates/cli/tests/graphql/pagination/paginate-todo-lists.graphql
```diff
@@ -1,15 +1,21 @@
 query PaginateTodoLists($first: Int, $last: Int, $before: String, $after: String) {
-  todoListCollection(first: $first, last: $last, before: $before, after: $after) {
```
yoav-lavi, line 2: Why remove the arguments here? Makes sense to leave it as a possibility

flake.nix
```diff
@@ -0,0 +1,57 @@
+{
```
yoav-lavi, line 1: Could we split the flake changes into a separate PR?

cli/crates/cli/tests/utils/macros.rs
```diff
@@ -2,22 +2,24 @@ pub use json_dotpath::DotPaths;
 
 #[macro_export]
 macro_rules! dot_get_opt {
-    ($item: ident, $dotpath: literal, $ty: ty) => {{
+    ($item: expr, $dotpath: expr, $ty: ty) => {{
```
yoav-lavi, line 5: Why did these change?
We originally debated allowing an expression but there wasn't consensus, plus we want to keep parity with the API version.

We could discuss this again

cli/crates/cli/tests/utils/macros.rs
```diff
@@ -2,22 +2,24 @@ pub use json_dotpath::DotPaths;
...
-    ($item: ident, $dotpath: literal) => {{
+    ($item: expr, $dotpath: expr) => {{
         use json_dotpath::DotPaths;
         $item.dot_get::<_>($dotpath).ok().flatten()
     }};
 }
 
 #[macro_export]
 macro_rules! dot_get {
-    ($item: ident, $dotpath: literal, $ty: ty) => {{
```
yoav-lavi, line 17: Why did these change?

cli/crates/cli/tests/graphql/pagination/paginate-todo-lists.graphql
Finistere: Cursors aren't the same so I removed them. I kept `first` & `last` because they could eventually help find a bug and it's free even though it should be doing the same thing.
Otherwise adding specific cursors wouldn't bring that much IMHO. We're already testing root pagination. We could also test root pagination in conjunction with nested one, but that's not what we're doing currently. Should we add it eventually, we could just add the parameters then.

flake.nix
Finistere: Yeah, had it lying around and got lazy. Will separate it. :)

cli/crates/cli/tests/utils/macros.rs
Finistere: Because literal doesn't work with the `format!("{prefix}.edges...")` that I'm using in my loop to test nested & root pagination. It stays consistent in terms of usage with the API, a literal is still an expression.

If I remove it, I have to unroll my loop. So I'm not sure to see any benefit to not support expression currently. ¯\\_(ツ)_/¯

flake.nix
Finistere: Well and its was in a different commit, so it felt OK to me. But I have no problem creating another PR.

cli/crates/cli/tests/utils/macros.rs
yoav-lavi: I see, ok let's change it then

yoav-lavi (approved):
Looks good!

cli/crates/cli/tests/pagination.rs
```diff
@@ -4,301 +4,171 @@ mod utils;
...
-    PAGINATION_MUTATION, PAGINATION_PAGINATE_TODOS, PAGINATION_PAGINATE_TODO_LISTS, PAGINATION_SCHEMA,
+    PAGINATION_CREATE_TODO, PAGINATION_CREATE_TODO_LIST, PAGINATION_PAGINATE_TODOS, PAGINATION_PAGINATE_TODO_LISTS,
+    PAGINATION_SCHEMA,
 };
 use utils::environment::Environment;
 
 #[test]
-
-fn pagination() {
+fn test_pagination() {
```
yoav-lavi, line 13: Why was the name here changed? This makes the output a bit less clean since it shows up in certain locations (and you're running tests to begin with)

Finistere: Ah force of habit, changing it

yoav-lavi:
@Finistere can we merge this?

Finistere:
Yes

PR #140, opened 2022-12-14
Add tests for JWT auth provider

mlafeldt:
# Description

This adds some basic auth tests.

# Type of change

- [ ] 💔 Breaking
- [ ] 🚀 Feature
- [ ] 🐛 Fix
- [ ] 🛠️ Tooling
- [ ] 🔨 Refactoring
- [x] 🧪 Test
- [ ] 📦 Dependency


cli/crates/cli/tests/auth.rs
```diff
@@ -0,0 +1,95 @@
...
+use serde_json::{json, Value};
+use utils::environment::Environment;
+
+const ISSUER_URL: &str = "https://some.issuer.test";
+const JWT_SECRET: &str = "topsecret";
+
+#[test]
+fn jwt_provider() {
+    const GROUP: &str = "backend";
+    const SCHEMA: &str = const_format::formatcp!(
```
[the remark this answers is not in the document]
mlafeldt: That's in fact how I started, but it would break interpolation and make the test harder to understand IMO.

cli/crates/cli/tests/auth.rs
```diff
@@ -0,0 +1,95 @@
...
+    env.write_schema(SCHEMA);
+    env.grafbase_dev();
+
+    let client = env.create_client();
+    client.poll_endpoint(30, 300);
+
+    // No auth header -> no authorization done in CLI
+    let resp = client.gql::<Value>(json!({ "query": QUERY }).to_string());
+    let errors: Option<Value> = dot_get_opt!(resp, "errors");
+    assert!(errors.is_none(), "errors: {errors:#?}");
```
mlafeldt, line 44: Those error comparisons look like common operations to me that should be shared between tests. Adding something to utils maybe?

cli/crates/cli/tests/auth.rs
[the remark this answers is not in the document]
yoav-lavi: Not sure I understand why we're formatting here at all, aren't all of these static for the entire test?

cli/crates/cli/tests/auth.rs
yoav-lavi: We could yes

cli/crates/cli/tests/auth.rs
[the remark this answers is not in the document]
mlafeldt: I moved the schema to a file and used env variables to avoid duplication of the issuer URL and JWT secret.

I still think that the test is less clear now and that using files didn't add anything other than maybe proper code formatting in the IDE. But on the other hand, I don't have strong feelings about it either, so let's keep it consistent.

cli/crates/cli/tests/utils/environment.rs
```diff
@@ -104,7 +104,10 @@ impl Environment {
     pub fn set_variables(&mut self, variables: HashMap<String, String>) {
         let env_file = variables
             .into_iter()
-            .fold(String::new(), |_, (key, value)| format!(r#"{key}="{value}""#));
```
mlafeldt, line 107: This had a bug where only one var would be written to the file.

yoav-lavi: Seems like it was missing appending the accumulator in the fold

yoav-lavi (approved):
Looks great

PR #142, opened 2022-12-16
fix(live queries): Buf fix & more tracing

Finistere:
# Description

# Type of change

- [ ] 💔 Breaking
- [ ] 🚀 Feature
- [ ] 🐛 Fix
- [ ] 🛠️ Tooling
- [ ] 🔨 Refactoring
- [ ] 🧪 Test
- [ ] 📦 Dependency


cli/crates/server/src/bridge/listener/server.rs
```diff
@@ -71,16 +72,34 @@ async fn event_listener(worker_port: u16) -> Result<(), ServerError> {
...
+                    "Failed to send event to stream-router. Responded: {:?}\n {}",
+                    response.headers().clone(),
+                    match response.text().await {
+                        Ok(text) => text,
+                        Err(err) => err.to_string(),
+                    }
+                );
+            } else {
+                trace!(
+                    "Successfully sent event to stream-router. Responded: {}",
```
yoav-lavi, line 96: I'd maybe split this into a trace before sending the event and a trace with the response rather than differentiating based on the code.

Since we don't need the response itself we should consider whether we want to await it for tracing purposes

yoav-lavi: @Finistere

PR #156, opened 2022-12-21
fix(ci): Fix missing cargo-nextest for MacOS

Finistere:
CI for MacOS didn't install `cargo-nextest` anymore, it only worked because we're re-using previous jobs cache. I also removed what seemed unnecessary for Linux. Not sure how it worked before without `musl-tools` though... ¯\\_(ツ)_/¯


hfgbarrigas (approved):
Nice

PR #161, opened 2023-01-03
GB-1504 - Init templates

yoav-lavi:
# Description

## Features

- Supports named and URL based init templates

## Refactoring

- Cleans up dependencies

# Type of change

- [ ] 💔 Breaking
- [x] 🚀 Feature
- [ ] 🐛 Fix
- [ ] 🛠️ Tooling
- [x] 🔨 Refactoring
- [ ] 🧪 Test
- [ ] 📦 Dependency

### DIFF
--- a/cli/crates/backend/src/project.rs
+++ b/cli/crates/backend/src/project.rs
@@ -165,8 +165,6 @@ async fn download_github_template(template: GitHubTemplate<'_>) -> Result<(), Ba
         branch,
     } = template.into_external_github_template();
 
-    let path = path.map(PathBuf::from);
-
     let org_and_repo = format!("{org}/{repo}");
 
     let extraction_dir = PathBuf::from(format!("{repo}-{branch}"));
--- a/cli/crates/backend/src/project.rs
+++ b/cli/crates/backend/src/project.rs
@@ -62,14 +62,18 @@ pub fn init(name: Option<&str>, template: Option<&str>) -> Result<(), BackendErr
     if grafbase_path.exists() {
         Err(BackendError::AlreadyAProject(grafbase_path))
     } else if let Some(template) = template {
-        if let Ok(repo_url) = Url::parse(template) {
-            match repo_url.host_str() {
-                Some("github.com") => handle_github_repo_url(&repo_url),
-                _ => Err(BackendError::UnsupportedTemplateURL(template.to_string())),
+        // as directory names cannot contain slashes, and URLs with no scheme or path cannot
+        // be differentiated from a valid template name,
+        // anything with a slash is treated as a URL
+        if template.contains('/') {
+            if let Ok(repo_url) = Url::parse(template) {
+                match repo_url.host_str() {
+                    Some("github.com") => handle_github_repo_url(&repo_url),
+                    _ => Err(BackendError::UnsupportedTemplateURL(template.to_string())),
+                }
+            } else {
+                return Err(BackendError::MalformedTemplateURL(template.to_owned()));
             }
-        // as directory names cannot contain slashes, names with slashes are considered malformed URLs
-        } else if template.contains('/') {
-            return Err(BackendError::UnsupportedTemplateURL(template.to_owned()));
         } else {
             download_github_template(GitHubTemplate::Grafbase(GrafbaseGithubTemplate { path: template }))
         }
--- a/cli/crates/backend/src/project.rs
+++ b/cli/crates/backend/src/project.rs
@@ -11,7 +11,7 @@ use std::env;
 use std::fs;
 use std::io::{Error as IoError, ErrorKind as IoErrorKind};
 use std::iter::Iterator;
-use std::path::{Path, PathBuf};
+use std::path::PathBuf;
 use tokio_stream::StreamExt;
 use tokio_util::compat::TokioAsyncReadCompatExt;
 use tokio_util::io::StreamReader;
--- a/cli/crates/backend/src/project.rs
+++ b/cli/crates/backend/src/project.rs
@@ -119,7 +119,7 @@ async fn download_github_template(template_info: &TemplateInfo<'_>) -> Result<()
     let (_, repo_without_org) = repo.split_once('/').expect("must have a slash");
 
     // not using the common environment since it's not initialized here
-    // if the OS does not have a cache path or it is not utf-8, we don't cache the download
+    // if the OS does not have a cache path or it is not UTF-8, we don't cache the download
     let cache_directory = dirs::cache_dir().and_then(|path| path.join("grafbase").to_str().map(ToOwned::to_owned));
 
     let mut client_builder = ClientBuilder::new(Client::new());
--- a/cli/crates/backend/src/project.rs
+++ b/cli/crates/backend/src/project.rs
@@ -63,18 +63,17 @@ pub fn init(name: Option<&str>, template: Option<&str>) -> Result<(), BackendErr
     if grafbase_path.exists() {
         Err(BackendError::AlreadyAProject(grafbase_path))
     } else if let Some(template) = template {
-        match Url::parse(template) {
-            Ok(repo_url) => match repo_url.host_str() {
+        if let Ok(repo_url) = Url::parse(template) {
+            match repo_url.host_str() {
                 Some("github.com") => handle_github_repo_url(&repo_url),
                 _ => Err(BackendError::UnsupportedTemplateURL(template.to_string())),
-            },
-            Err(_) => {
-                if template.contains("/") {
-                    return Err(BackendError::UnsupportedTemplate(template.to_owned()));
-                }
-
-                download_github_template(&TemplateInfo::Grafbase { path: template })
             }
+        } else {
+            if template.contains('/') {
+                return Err(BackendError::UnsupportedTemplate(template.to_owned()));
+            }
+
+            download_github_template(&TemplateInfo::Grafbase { path: template })
         }
     } else {
         fs::create_dir_all(&grafbase_path).map_err(BackendError::CreateGrafbaseDirectory)?;
--- a/cli/crates/backend/src/project.rs
+++ b/cli/crates/backend/src/project.rs
@@ -45,7 +45,7 @@ use url::Url;
 ///
 /// - returns [`BackendError::StoreRepoArchive`] if a repo tar could not be stored
 ///
-/// - returns [`BackendError::NoFilesExtracted`] if no files were extracted from the repo archive
+/// - returns [`BackendError::TemplateNotFound`] if no files matching the template path were extracted (excluding extraction errors)
 ///
 /// - returns [`BackendError::MoveExtractedFiles`] if the extracted files from the template repository could not be moved
 ///
--- a/cli/crates/backend/src/project.rs
+++ b/cli/crates/backend/src/project.rs
@@ -12,7 +12,6 @@ use std::fs;
 use std::io::{Error as IoError, ErrorKind as IoErrorKind};
 use std::iter::Iterator;
 use std::path::{Path, PathBuf};
-use std::str::FromStr;
 use tokio_stream::StreamExt;
 use tokio_util::compat::TokioAsyncReadCompatExt;
 use tokio_util::io::StreamReader;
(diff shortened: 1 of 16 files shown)


yoav-lavi:
Fixing a platform specific issue on Windows

cli/crates/backend/src/project.rs
```diff
@@ -65,3 +125,129 @@ fn to_project_path(name: Option<&str>) -> Result<PathBuf, BackendError> {
...
+        tokio::fs::remove_dir_all(extraction_dir)
+            .await
+            .map_err(BackendError::CleanExtractedFiles)?;
+    }
+
+    extraction_result
+}
+
+async fn stream_github_archive<'a>(
+    repo: &'a str,
```
Finistere, line 163: OPT: I wouldn't pass both `path` and `extraction_dir`, they're defining the path within the archive as far as I understand. This path is defined in two steps:
```
    let extraction_dir = PathBuf::from_str(&format!("{repo_without_org}-{branch}")).expect("must succeed");
```
and later
```
    let mut template_path = PathBuf::new();

    template_path.push(extraction_dir);

    if let Some(path) = path {
        template_path.push(path);
    }

    template_path.push("grafbase");
```
I would have combined both to define it in a single place.

cli/crates/backend/src/project.rs
```diff
@@ -65,3 +125,129 @@ fn to_project_path(name: Option<&str>) -> Result<PathBuf, BackendError> {
...
+            entry.unpack_in(".").await.map_err(BackendError::ExtractArchiveEntry)?;
+        }
+    }
+
+    // FIXME: an incorrect template name errors here rather than initially
+    if !template_path.exists() {
+        return Err(BackendError::NoFilesExtracted);
+    }
+
+    tokio::fs::rename(template_path, "grafbase")
```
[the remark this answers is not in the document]
yoav-lavi: We can't init `Environment` here since the folder doesn't exist yet, but we can use a const

cli/crates/backend/src/project.rs
```diff
@@ -19,40 +33,81 @@ use std::path::PathBuf;
...
-pub fn reset() -> Result<(), BackendError> {
-    let environment = Environment::get();
+fn handle_github_repo_url(repo_url: &Url) -> Result<(), BackendError> {
+    if let Some(segments) = repo_url
+        .path_segments()
+        .map(Iterator::collect::<Vec<_>>)
+        // TODO: allow URLs without 'tree/branch' by checking the default branch via the API
+        .filter(|segments| segments.len() >= 4 && segments[2] == "tree")
+    {
+        let repo = &segments[..=1].join("/");
```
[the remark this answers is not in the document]
yoav-lavi: Where specifically?

cli/crates/backend/src/project.rs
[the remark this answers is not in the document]
Finistere: Yes, that's what I meant, with the re-use of the constant it's explicit that those paths are really expected to be the same.

cli/crates/backend/src/project.rs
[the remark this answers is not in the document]
Finistere: Well thinking about it it's probably more error-prone. I'm not sure what characters are allowed for project/ organizations in GitHub. So taking back what I said, your implementation is clear as it is IMHO.

cli/crates/backend/src/project.rs
```diff
@@ -65,3 +120,125 @@ fn to_project_path(name: Option<&str>) -> Result<PathBuf, BackendError> {
...
+    External {
+        repo: &'a str,
+        path: Option<&'a str>,
+        branch: &'a str,
+    },
+}
+
+#[tokio::main]
+async fn download_github_template(template_info: &TemplateInfo<'_>) -> Result<(), BackendError> {
+    let (repo, path, branch) = match template_info {
```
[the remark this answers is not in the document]
yoav-lavi: I don't think you can type a function as returning an enum variant

cli/crates/backend/src/project.rs
```diff
@@ -65,3 +120,125 @@ fn to_project_path(name: Option<&str>) -> Result<PathBuf, BackendError> {
...
+        TemplateInfo::Grafbase { path } => ("grafbase/grafbase", Some(PathBuf::from("templates").join(path)), "main"),
+        TemplateInfo::External { repo, path, branch } => (*repo, path.map(PathBuf::from), *branch),
+    };
+
+    let (_, repo_without_org) = repo.split_once('/').expect("must have a slash");
```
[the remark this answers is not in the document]
yoav-lavi: This would mean reconnecting them later for the URL, seems like it results in the same thing

cli/crates/backend/src/project.rs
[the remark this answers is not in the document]
Finistere: No indeed my bad, for this you would need to add an intermediate struct which would be what the enum holds like:
```
struct TemplateInfoExternal {}
enum TemplateInfo {
  External(TemplateInfoExternal)
}
```

cli/crates/backend/src/project.rs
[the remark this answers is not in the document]
Finistere: Sure, but IMHO it's more explicit. I find the naming `repo_without_org` a bit weird and it's easier to get issues with  `split_once('/')`, as shown with the `expect`, than when joining two strings later.

cli/crates/backend/src/project.rs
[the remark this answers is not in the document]
Finistere: By the way, I probably would have used the following naming in this case
```
struct GitHubTemplateUrl {}
enum TemplateUrl {
  GitHub(GitHubTemplateUrl),
  Grafbase { ... }
}
```

cli/crates/backend/src/project.rs
```diff
@@ -19,40 +33,83 @@ use std::path::PathBuf;
...
     let project_path = to_project_path(name)?;
     let grafbase_path = project_path.join(GRAFBASE_DIRECTORY);
     let schema_path = grafbase_path.join(GRAFBASE_SCHEMA);
 
-    if schema_path.exists() {
-        Err(BackendError::AlreadyAProject(schema_path))
+    if grafbase_path.exists() {
+        Err(BackendError::AlreadyAProject(grafbase_path))
+    } else if let Some(template) = template {
+        if let Ok(repo_url) = Url::parse(template) {
```
jakubadamw, line 66: As discussed, I think we should flip the logic to first probe for a simple template name (`[a-zA-Z-]+`) and then proceed to treating the input as an URL as a fallback. 🙂

jakubadamw (changes requested):
All great comments by @Finistere. Just one thing from me, as already discussed on Slack.

jakubadamw:
And great work, of course!

cli/crates/backend/src/project.rs
yoav-lavi: I'm debating whether this would be a bit limiting, we can't assume templates will only use alphanumeric characters and dashes

cli/crates/backend/src/project.rs
yoav-lavi: Done

cli/crates/backend/src/project.rs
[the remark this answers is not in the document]
yoav-lavi: Done

cli/crates/backend/src/project.rs
```diff
@@ -65,3 +117,129 @@ fn to_project_path(name: Option<&str>) -> Result<PathBuf, BackendError> {
...
+
+async fn stream_github_archive<'a>(
+    repo: &'a str,
+    path: Option<PathBuf>,
+    branch: &'a str,
+    extraction_dir: &'a Path,
+) -> Result<(), BackendError> {
+    // not using the common environment since it's not initialized here
+    // if the OS does not have a cache path or it is not UTF-8, we don't cache the download
+    let cache_directory = dirs::cache_dir().and_then(|path| path.join("grafbase").to_str().map(ToOwned::to_owned));
```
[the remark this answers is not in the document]
yoav-lavi: Done

cli/crates/backend/src/project.rs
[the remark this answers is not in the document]
yoav-lavi: Done

cli/crates/backend/src/project.rs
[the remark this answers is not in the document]
yoav-lavi: Done

Finistere (approved):
Nice work!

cli/crates/backend/src/project.rs
jakubadamw: IMO it's definitely better than what we have currently. If you pass something URL-like that fails to parse, it should result in an appropriate error.

jakubadamw: Even something like:
```rust
if template.starts_with("https://") {
    // Treat as an URL.
} else {
    // Treat as a template name.
}
```

would be better than this "fall back on error" thing.

yoav-lavi: Semantically it's "fall back if the input is not a proper URL", which is more or less the same as you're doing with the protocol check

yoav-lavi: As in - All cases of urls without `http://` and `https://` will currently be treated as a template name, exactly like the example

yoav-lavi: Regardless we can add a check for the specific scheme

jakubadamw: URL parsing can fail in various other ways, so no. “Does it start with a particular scheme” is a heuristic check for how to treat the input, and then whether that input is valid in light of the distinction made is determined separately. How to treat the input isn't determined by the initial guess of what it is, in other words. Okay, this isn't important enough to be arguing over too much, so I won't be insistent over this if you're set on doing it this way. That said, you do have a FIXME related to it, so later it'll need to be addressed one way or another.

yoav-lavi: I don't see this as an argument, this is a discussion. My goal isn't to advocate for this solution specifically, it's to find one that would be robust.

I'm not particularly married to checking if it's a URL first, my goal is to avoid cases where we'd need to redeploy because we used a type of character we didn't account for in a template name (e.g. `todo-2`).

My point before is that checking for the protocol wouldn't add anything extra since it's already being done by the parser, if your goal is presenting a better error message in the case of an invalid URL with a scheme we could do this check first, but that'd rule out trying to append a scheme if one doesn't exist which we also discussed earlier (although that'd also have some edge cases in case of templates containing dots, which is why it's not included here)

The `FIXME` isn't specifically related to this issue IMO, it derives from knowing whether a template name is correct only after downloading the archive, for which we'll need to check the archive in advance (pre-extraction) and tweak the error message to cover both cases.

All the above being said I still think we should consider separate arguments for URLs and template names since they'd rule these cases out and be more discoverable (with not much of a DX price), but given what we're doing currently there's tradeoffs to any sort of heuristic we choose

yoav-lavi: Since we're streaming the tar, checking the contents before extraction is tricky. 

We could treat the absence of extracted files as definite evidence that the files didn't exist in the tar but there are other cases that could cause that.

We could set a variable after first extracting a file but that isn't as clean as I would like

jakubadamw: I didn't mean “argument” in a negative sense of the word. 🙂 Nothing wrong with civil arguments. 🙂  I meant to say that I would rather we kept iterating on this, hey, even making mistakes and perhaps improving in future versions rather than spending too much time on overthinking this.

My general point is this: this is tying the initial guess of “is this an URL or is this template name” to the validation informed by one of the two possible outcomes. To me it doesn't seem right on a purely logical basis, but incidentally it also causes inaccurate error reporting in a particular scenario. The FIXME seems related because that branch of code will specifically kick in for the case of an invalid URL, and that's what would be avoided by decoupling these two concerns.

Anyway, as I said, not a blocker.

yoav-lavi: I see, I actually added a check for `/` at some point in the template name since URLs would almost always contain a slash and directories cannot have one, so most of these cases don't reach the `FIXME` anymore, hence me not associating it with this part of the code.

The one case that'd still trigger it is something like `mysite.com` which we wouldn't cover by checking the scheme (but also can't be ruled out as a template name)

Checking for a scheme would provide a better error message in a case like `https://%` though because it would consider it a URL rather than an invalid template name (before downloading the archive). Although we could also just consider anything with a slash as a malformed URL due to templates not being able to have a slash at all

jakubadamw: @yoav-lavi oh, yes, that's much better, thanks! 🙂

yoav-lavi: @jakubadamw could you please re-review?

jakubadamw (approved):
Great!

PR #173, opened 2023-01-05
tests: Add various bugfix test cases

Finistere:
Adding tests cases for:
- scalar coercion
- default directive applied in nested inputs
- consistent handling of createdAt/updatedAt

# Description

# Type of change

- [ ] 💔 Breaking
- [ ] 🚀 Feature
- [ ] 🐛 Fix
- [ ] 🛠️ Tooling
- [ ] 🔨 Refactoring
- [X] 🧪 Test
- [ ] 📦 Dependency


cli/crates/cli/Cargo.toml
```diff
@@ -45,6 +45,7 @@ sysinfo = "0.27"
 tempfile = "3"
 json_dotpath = "1"
 tokio = { version = "1", features = ["full"] }
+regex = "1.7"
```
yoav-lavi, line 48: Our scheme is normally:

`dependency = "x"`
`dependency = "0.x"`

Depending on whether the major is over `0`

yoav-lavi: I have a different PR fixing this for other dependencies and sorting them alphabetically

cli/crates/cli/tests/reserved_dates.rs
```diff
@@ -0,0 +1,100 @@
...
+    RESERVED_DATES_CREATE_TODO, RESERVED_DATES_CREATE_TODO_LIST, RESERVED_DATES_NESTED_CREATION, RESERVED_DATES_SCHEMA,
+};
+use chrono::{DateTime, Duration, Utc};
+use serde_json::{json, Value};
+use utils::environment::Environment;
+
+#[test]
+fn reserved_dates() {
+    // TODO: Create simpler client setup (one-line)
+    let mut env = Environment::init(4016);
```
yoav-lavi, line 13: The init templates PR adds a test on `4016`, until we have a better solution for this, would it be possible to start with `4017`?

cli/crates/cli/tests/scalars.rs
```diff
@@ -1,35 +1,197 @@
 mod utils;
 
 use serde_json::{json, Value};
-use utils::consts::{SCALARS_MUTATION, SCALARS_QUERY, SCALARS_SCHEMA};
+use utils::client::Client;
+use utils::consts::{SCALARS_CREATE_OPTIONAL, SCALARS_CREATE_REQUIRED, SCALARS_SCHEMA};
 use utils::environment::Environment;
 
+trait ClientScalarExt {
```
yoav-lavi, line 8: Is this used anywhere else? A bit confused as to why it's a trait

cli/crates/cli/Cargo.toml
Finistere: I don't mind changing it, but why? If I'm not mistaken `"1.7"` means a version in `[1.7.0, 2.0.0[`. So I don't see why allowing a larger internal `[1.0.0, 2.0.0[` is useful in any way.

cli/crates/cli/tests/reserved_dates.rs
Finistere: Yeah, I'll fix those once the gateway PRs have been merged when rebasing. If you haven't merged your PR by then, I'll start from 4017.

cli/crates/cli/Cargo.toml
yoav-lavi: AFAIK `1.7` means `1.7.0`-`1.7.n`.
In both the CLI and the workers we limit changes to breaking versions, allowing minor versions automatically

cli/crates/cli/tests/scalars.rs
Finistere: I'm just adding some methods to the `client`. For this, I need a trait & impl. But I agree that just using functions shouldn't hurt readability, I guess I'm just too used to this pattern from Kotlin. I don't mind changing if you prefer functions in this case.

cli/crates/cli/Cargo.toml
Finistere: I think you may be confusing it with "0.x" releases, from the cargo book:
```
1.2.3  :=  >=1.2.3, <2.0.0
1.2    :=  >=1.2.0, <2.0.0
1      :=  >=1.0.0, <2.0.0
0.2.3  :=  >=0.2.3, <0.3.0
0.2    :=  >=0.2.0, <0.3.0
0.0.3  :=  >=0.0.3, <0.0.4
0.0    :=  >=0.0.0, <0.1.0
0      :=  >=0.0.0, <1.0.0
```
https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html

cli/crates/cli/tests/scalars.rs
yoav-lavi: I see, it's fine, I thought you were using a trait for re-use which didn't really happen here

Finistere: But to be honest I'm not a big fan of the API I came up with for this. I'll probably try to change it. `transform` and `error_msg` are just poor method names.

cli/crates/cli/Cargo.toml
yoav-lavi: I see, I recalled it working differently, perhaps an NPM holdover.
@jakubadamw why do we use this convention?

jakubadamw: @yoav-lavi mostly because we rarely ever want to denote that we need a version of a dependency higher than a particular minimum minor version. Having that ability would be more useful if we were publishing libraries for other to use (in that case the dependent leaf package could be dependent on the same dependency as one among ours, but with an upper bound, like `<=1.6`, so having a minimum of 1.7 allows us to ensure Cargo won't resolve that dependency to 1.6 if we depend on a feature from 1.7).

However, because all packages are all contained in one workspace (with a single Cargo.lock) and never escape the workspace (i.e. don't get published to crates.io), there's no benefit to denoting lower-bounds on minor or patch versions. If we were following the `1.x` or `0.x.y` scheme, our Cargo.toml would effectively become a mirror of Cargo.lock. The practical downside is that our Renovate PRs would have to update both Cargo.toml *and* Cargo.lock on each update of a dependency (most likely in separate PRs, by the way).

I don't know if this thinking applies to grafbase/grafbase as we do publish the `grafbase` package to crates.io. However, the package is still a leaf executable-only package (not a library to be consumed by others), so I'd say it's a similar situation, actually.

yoav-lavi (approved):
Looks good!

PR #174, opened 2023-01-06
feat(templates): add twitter template

notrab:
# Description

# Type of change

- [ ] 💔 Breaking
- [x] 🚀 Feature
- [ ] 🐛 Fix
- [ ] 🛠️ Tooling
- [ ] 🔨 Refactoring
- [ ] 🧪 Test
- [ ] 📦 Dependency


templates/twitter/grafbase/schema.graphql
```diff
@@ -0,0 +1,18 @@
+type User @model {
+  name: String!
```
fbjork, line 2: also needs `username` that is unique

notrab: Good spot!

hhthacker (approved):
looks good! wonder if `handle` is more universal than `username`? 
i would guess `name` is the display text of the user
`username` or `handle` is the user's @ id

also wonder if `replies` or `media` is relevant for including in this schema

templates/twitter/grafbase/schema.graphql
```diff
@@ -0,0 +1,18 @@
...
+  email: Email! @unique
+  image: String
```
fbjork, line 4: perhaps URL?

notrab:
@fbjork @hhthacker I've updated the schema now to include a few more fields, with the correct directives, and a relation to `replies` and `media`.

templates/twitter/grafbase/schema.graphql
```diff
@@ -0,0 +1,31 @@
+type User @model {
+  username: String! @unique
+  email: Email! @unique
+  url: URL
```
fbjork, line 4: What is url?

templates/twitter/grafbase/schema.graphql
```diff
@@ -0,0 +1,31 @@
...
+  avatar: String
```
fbjork, line 5: Use URL scalar here?

templates/twitter/grafbase/schema.graphql
notrab: better named `website` perhaps

templates/twitter/grafbase/schema.graphql
notrab: I opted not to use `URL` here since the image could be relative to the application using this schema.

templates/twitter/grafbase/schema.graphql
fbjork: Let's stick to url since that's what the Twitter API uses.

templates/twitter/grafbase/schema.graphql
fbjork: I would stick to URL since that's what the Twitter API uses.

fbjork (approved):

templates/twitter/grafbase/schema.graphql
notrab, line 5:
```suggestion
  avatar: URL
```

PR #176, opened 2023-01-09
tests: Improve e2e dev test case

Finistere:
The update mutation was returning stale data before, improving a e2e
test to check for it.

# Description

# Type of change

- [ ] 💔 Breaking
- [ ] 🚀 Feature
- [ ] 🐛 Fix
- [ ] 🛠️ Tooling
- [ ] 🔨 Refactoring
- [ ] 🧪 Test
- [ ] 📦 Dependency

### DIFF
--- a/cli/crates/cli/tests/graphql/default/create.graphql
+++ b/cli/crates/cli/tests/graphql/default/create.graphql
@@ -3,17 +3,20 @@ mutation {
     input: {
       title: "My todo list"
       todos: [
-        { create: { title: "My first todo!", complete: true } }
-        { create: { title: "My second todo!", complete: false } }
+        {create: {title: "My first todo!", complete: true}}
+        {create: {title: "My second todo!", complete: false}}
       ]
     }
   ) {
     todoList {
+      id
       title
       todos(last: 100) {
         edges {
           node {
+            id
             title
+            complete
           }
         }
       }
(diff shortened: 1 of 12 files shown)


cli/crates/cli/tests/graphql/default/create.graphql
```diff
@@ -3,17 +3,20 @@ mutation {
     input: {
       title: "My todo list"
       todos: [
-        { create: { title: "My first todo!", complete: true } }
-        { create: { title: "My second todo!", complete: false } }
+        {create: {title: "My first todo!", complete: true}}
+        {create: {title: "My second todo!", complete: false}}
```
mlafeldt, line 7: Those extra spaces are added by prettier (there's a config in the root folder). Let's stick to the formatter.

mlafeldt (approved):
LGTM. Just a formatting nit.

cli/crates/cli/tests/graphql/default/create.graphql
yoav-lavi: We need to add it to the CI but yes, we should be using prettier

Finistere: Ok, fixing it! thx

yoav-lavi (approved):
Looks good!

PR #194, opened 2023-01-24
feat(templates): add chatroom template

notrab:
# Description

This adds a schema template for a chatroom.

# Type of change

- [ ] 💔 Breaking
- [x] 🚀 Feature
- [ ] 🐛 Fix
- [ ] 🛠️ Tooling
- [ ] 🔨 Refactoring
- [ ] 🧪 Test
- [ ] 📦 Dependency


templates/chatroom/grafbase/schema.graphql
```diff
@@ -0,0 +1,29 @@
+type Room @model {
+  title: String!
+  private: Boolean @default(value: false)
```
fbjork, line 3: don't booleans have a default value already?

fbjork (approved):
Just added a small comment.

templates/chatroom/grafbase/schema.graphql
notrab: If it's nullable then it's `null`. Otherwise if it's non-nullable `Boolean!` then a value is required on create.

PR #210, opened 2023-02-10
features(search): Add local search with basic API

Finistere:
- Create an in-memory Tantivy index for search requests
- Added a `./scripts/dev.sh` to run grafbase with custom assets pushed by the api CI.

Resolves: GB-2982

# Description

# Type of change

- [ ] 💔 Breaking
- [ ] 🚀 Feature
- [ ] 🐛 Fix
- [ ] 🛠️ Tooling
- [ ] 🔨 Refactoring
- [ ] 🧪 Test
- [ ] 📦 Dependency

### DIFF
--- a/cli/crates/server/src/bridge/search.rs
+++ b/cli/crates/server/src/bridge/search.rs
@@ -1,0 +1,400 @@
+use std::borrow::Borrow;
+use std::net::{IpAddr, Ipv6Addr};
+
+use chrono::{DateTime, NaiveDate, NaiveTime, TimeZone, Utc};
+use futures_util::TryStreamExt;
+use once_cell::sync::OnceCell;
+use regex::Regex;
+use serde_json::Value;
+use sqlx::SqlitePool;
+
+use tantivy::collector::TopDocs;
+use tantivy::directory::RamDirectory;
+use tantivy::query::{QueryParser, QueryParserError};
+use tantivy::schema::{Field, Schema, INDEXED, STORED, STRING, TEXT};
+use tantivy::store::Compressor;
+use tantivy::{DocAddress, Document, IndexSettings, Score, TantivyError};
+
+use super::errors::ApiError;
+use super::types::{RecordDocument, SearchField, SearchSchema};
+
+const DATE_FORMAT: &str = "%Y-%m-%d";
+static BASE_TYPE_RE: OnceCell<regex::Regex> = OnceCell::new();
+
+mod scalars {
+    pub const URL: &str = "URL";
+    pub const EMAIL: &str = "Email";
+    pub const PHONE: &str = "PhoneNumber";
+    pub const STRING: &str = "String";
+    pub const DATE: &str = "Date";
+    pub const DATE_TIME: &str = "DateTime";
+    pub const TIMESTAMP: &str = "Timestamp";
+    pub const INT: &str = "Int";
+    pub const FLOAT: &str = "Float";
+    pub const BOOL: &str = "Boolean";
+    pub const IP: &str = "IPAddress";
+}
+
+impl From<TantivyError> for ApiError {
+    fn from(error: TantivyError) -> Self {
+        error!("TantivyError: {error}");
+        Self::ServerError
+    }
+}
+
+impl From<QueryParserError> for ApiError {
+    fn from(error: QueryParserError) -> Self {
+        error!("Tantivy QueryParserError: {error}");
+        Self::ServerError
+    }
+}
+
+pub struct Index {
+    inner: tantivy::Index,
+    default_fields: Vec<Field>,
+    id_field: Field,
+}
+
+impl Index {
+    pub fn search_top_records(&self, query: &str, limit: usize) -> Result<Vec<String>, ApiError> {
+        let searcher = self.inner.reader()?.searcher();
+        let query_parser = QueryParser::for_index(&self.inner, self.default_fields.clone());
+        let tantivy_query = query_parser.parse_query(query)?;
+
+        let top_docs: Vec<(Score, DocAddress)> = searcher.search(&tantivy_query, &TopDocs::with_limit(limit))?;
+
+        let mut matching_records = vec![];
+        for (_score, doc_address) in top_docs {
+            // Retrieve the actual content of documents given its `doc_address`.
+            let retrieved_doc = searcher.doc(doc_address)?;
+            matching_records.push(
+                String::from_utf8(
+                    retrieved_doc
+                        .get_first(self.id_field)
+                        .and_then(tantivy::schema::Value::as_bytes)
+                        .expect("Id is always present")
+                        .to_vec(),
+                )
+                .expect("Id is a valid string"),
+            );
+        }
+
+        trace!(
+            "Found {} top documents:\n{:?}",
+            matching_records.len(),
+            matching_records
+        );
+        Ok(matching_records)
+    }
+
+    pub async fn build(pool: &SqlitePool, entity_type: &str, search_schema: &SearchSchema) -> Result<Index, ApiError> {
+        trace!("Building index for {entity_type} and schema:\n{search_schema:?}");
+
+        let mut schema_builder = Schema::builder();
+        let id_field = schema_builder.add_bytes_field("id", STORED);
+        let (fields, default_fields) = {
+            let mut all = vec![];
+            let mut defaults = vec![];
+            let mut search = vec![];
+
+            for field in &search_schema.fields {
+                let base_type = BASE_TYPE_RE
+                    .get_or_init(|| Regex::new(r"[\[\]!]").unwrap())
+                    .replace_all(&field.r#type, "");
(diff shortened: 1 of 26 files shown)


scripts/dev.sh
```diff
@@ -0,0 +1,104 @@
+#!/usr/bin/env bash
```
fbjork, line 1: This should live in the cli folder since it's not global to the mono repo.

Finistere: Done!

cli/crates/server/src/servers.rs
```diff
@@ -372,10 +373,10 @@ async fn validate_dependencies() -> Result<(), ServerError> {
 // to avoid issues when starting multiple CLIs simultainiously,
 // we segment the ephemeral port range into 100 segments and select a segment based on the last two digits of the process ID.
 // this allows for simultainious start of up to 100 CLIs
-fn get_bridge_port() -> Result<u16, ServerError> {
+fn get_bridge_port(http_port: u16) -> Result<u16, ServerError> {
     // must be 0-99, will fit in u16
     #[allow(clippy::cast_possible_truncation)]
-    let segment = (process::id() % 100) as u16;
+    let segment = http_port % 100;
```
yoav-lavi, line 379: This works for serial ports like we use in tests, but it wouldn't work if e.g. you use `5000`, `6000`

Finistere: Wow, I was convinced that PR was merged...
In which case would you use 5000/6000 ports? I had regular issues with that on my Linux to be honest, tests complaining about port being already used. Otherwise maybe `http_port + process::id()`?

yoav-lavi: I mean if a user were manually to use 5000 and 6000

Finistere: It can also happen that the two process ids have the same modulo. `http_port` is at least deterministic IMHO. In practice for most users, I would expect the cli to search for an available http_port and do 5000, 5001 etc. for the user. Now if you have a strong opinion against it, I'll change it back.

yoav-lavi: The idea here is that processes that are launched one after another will have consecutive IDs, so the modulo colliding happens only after 100 processes

Finistere: Well, I'm not sure what you can assume from process ids, but what I'm sure is that running `cargo nextest run` I always/very often ended up with errors like: `error binding to 127.0.0.1:51763: error creating server listener: Address already in use (os error 98)`. But I didn't anymore with this change. `cargo test` seems ok though.

yoav-lavi: That's because we use serial port IDs, but we can try this out and see if anyone runs into issues

Finistere: What do you mean by "serial port IDs"? searching sequentially for a port? If so I don't see how we could do differently when searching for a port, we just need a good offset

cli/crates/server/src/bridge/search.rs
```diff
@@ -0,0 +1,401 @@
...
+use tantivy::schema::{Field, Schema, INDEXED, STORED, STRING, TEXT};
+use tantivy::store::Compressor;
+use tantivy::{DocAddress, Document, IndexSettings, Score, TantivyError};
+
+use super::errors::ApiError;
+use super::types::{RecordDocument, SearchField, SearchSchema};
+
+const DATE_FORMAT: &str = "%Y-%m-%d";
+
+lazy_static::lazy_static! {
```
yoav-lavi, line 22: Better to use `once_cell` nowdays

cli/crates/server/src/bridge/search.rs
```diff
@@ -0,0 +1,401 @@
...
+                    scalars::DATE | scalars::DATE_TIME | scalars::TIMESTAMP => {
+                        schema_builder.add_date_field(&field.name, INDEXED)
+                    }
+                    scalars::INT => schema_builder.add_i64_field(&field.name, INDEXED),
+                    scalars::FLOAT => schema_builder.add_f64_field(&field.name, INDEXED),
+                    scalars::BOOL => schema_builder.add_bool_field(&field.name, INDEXED),
+                    scalars::IP => schema_builder.add_ip_addr_field(&field.name, INDEXED),
+                    unknown_type => {
+                        error!("Unknown field type: '{unknown_type}'");
+                        return Err(ApiError::ServerError);
```
yoav-lavi, line 122: Wouldn't we want to return a specific error so a user can see it?

cli/crates/server/src/bridge/search.rs
```diff
@@ -0,0 +1,401 @@
...
+            .schema(schema.clone())
+            .settings(IndexSettings {
+                docstore_compression: Compressor::None,
+                ..Default::default()
+            });
+
+        let index = builder.open_or_create(RamDirectory::create())?;
+        let mut index_writer = index.writer_with_num_threads(1, 20_000_000)?;
+
+        let mut fut = sqlx::query_as(
```
yoav-lavi, line 144: Could we add this to the SQL type?

cli/crates/server/src/bridge/search.rs
```diff
@@ -0,0 +1,401 @@
...
+        SELECT pk AS id, document
+        FROM records WHERE entity_type = $1 AND pk = sk
+        "#,
+        )
+        .bind(entity_type)
+        .fetch(pool);
+
+        let mut record_count: usize = 0;
+        while let Some::<RecordDocument>(record) = fut.try_next().await? {
+            record_count += 1;
```
yoav-lavi, line 155: If this is just for the trace I'd leave it out personally

cli/crates/server/src/bridge/search.rs
```diff
@@ -0,0 +1,401 @@
...
+                }
+            }
+            _ => unreachable!("Field could not have been added to the schema."),
+        };
+    }
+    Ok(())
+}
+
+#[derive(Debug)]
+pub enum DynamoItemFieldError {
```
yoav-lavi, line 252: Not sure we should explicitly name these Dynamo, the datastore being Dynamo compatible is for convenience

cli/scripts/dev.sh
```diff
@@ -0,0 +1,104 @@
+#!/usr/bin/env bash
```
yoav-lavi, line 1: Ideally we'd do this with something like xtask or cargo make, but we can use bash for now

yoav-lavi (changes requested):

cli/scripts/dev.sh
yoav-lavi: Looks like a great dev tool by the way

cli/crates/server/src/bridge/search.rs
Finistere: That's really an internal error, it meant we had an issue with our parser. With my next PR I'm replacing the strings with a proper enum, so this will go away

cli/crates/server/src/bridge/search.rs
Finistere: Which SQL type? There's only one I know of in api.

yoav-lavi: Ah right, I'm used to SQL queries being formed there. Nevermind

cli/crates/server/src/bridge/search.rs
Finistere: Well, the whole point of the trace is to show that number. I found it somewhat useful initially to know whether anything was indexed at all and how much. I'd tend to keep it as it can still help, but I'm not sure if it'll be ever used again.

cli/crates/server/src/bridge/search.rs
Finistere: I've changed the error & result name, but I'd keep the `DynamoItemExt` name. Whether by convenience or by necessity, the documents are constructed as DynamoDB items.

cli/crates/server/src/bridge/search.rs
```diff
@@ -0,0 +1,400 @@
...
+        .fetch(pool);
+
+        let mut record_count: usize = 0;
+        while let Some::<RecordDocument>(record) = fut.try_next().await? {
+            record_count += 1;
+            let mut doc = Document::default();
+            doc.add_bytes(id_field, record.id.as_bytes());
+            for (search_field, field) in &fields {
+                add_field(&mut doc, *field, search_field, &record).map_err(|err| {
+                    error!(
```
yoav-lavi, line 160: Would you mind checking that these aren't user visible without `--trace`?

yoav-lavi (approved):
Looks good, see the one comment

cli/crates/server/src/bridge/search.rs
Finistere: Oh they certainly are, good catch!

Finistere: I was wrong, it doesn't actually. The query/mutation endpoints actually use those as well.

PR #211, opened 2023-02-11
GB-3086: CLI: Extend the bridge API with support for custom resolver invocations that runs them with v8 in a straight-forward non-cached way

jakubadamw:
# Description

CLI: Extend the bridge API with support for custom resolver invocations.
For now, run them with V8 in a straight-forward non-cached way.
We'll shortly optimise it and add proper error handling.

# Type of change

- [ ] 💔 Breaking
- [X] 🚀 Feature
- [ ] 🐛 Fix
- [ ] 🛠️ Tooling
- [ ] 🔨 Refactoring
- [ ] 🧪 Test
- [ ] 📦 Dependency

### DIFF
--- a/.github/workflows/cli.yml
+++ b/.github/workflows/cli.yml
@@ -17,7 +17,7 @@ permissions:
 
 env:
   GRAFBASE_SKIP_ASSET_VERSION_CHECK: 'true'
-  ASSETS_VERSION: 'release/949c313-2023-02-06'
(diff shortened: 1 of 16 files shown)


jakubadamw:
Okay, so rusty_v8 [doesn't provide packages](https://github.com/denoland/rusty_v8/pull/999) for Linux musl. That may actually be a blocker for us using v8 in-process.

PR #219, opened 2023-02-26
add react example

hhthacker:
# Description

- add react example

# Type of change

- [ ] 💔 Breaking
- [ ] 🚀 Feature
- [ ] 🐛 Fix
- [ ] 🛠️ Tooling
- [ ] 🔨 Refactoring
- [ ] 🧪 Test
- [ ] 📦 Dependency


examples/react/src/index.tsx
```diff
@@ -0,0 +1,75 @@
...
+          'Content-Type': 'application/json',
+        },
+        body: JSON.stringify({ query: GetAllMessagesQuery,
+          variables: {
+            first: 100
+          } }),
+      });
+
+      const result = await response.json() as Data;
+      console.log(result.data)
```
jvidalv, line 49: This console log should be removed.

notrab (approved):
Great job @hhthacker!

For others:
* Heather and I discussed this PR earlier.
* I suggested we move this into two examples. One with Vite and another with CRA.
* CRA is the official way to create a React app from the docs. So it makes sense the "vanilla" example uses this for anyone new to React AND Grafbase without getting into the weeds of custom setups (ala Vite).

PR #230, opened 2023-03-09
GB-2798: Add owned_by column

tomasol:
# Description

Add `owned_by` column to the Records table as a prerequisite for owner-based authorization.

# Type of change

- [ ] 💔 Breaking
- [x] 🚀 Feature
- [ ] 🐛 Fix
- [ ] 🛠️ Tooling
- [ ] 🔨 Refactoring
- [ ] 🧪 Test
- [ ] 📦 Dependency


cli/crates/server/src/sql/001_add_owned_by.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE records ADD COLUMN owned_by TEXT;
```
Finistere, line 1: I'm sad sqlite doesn't support `IF NOT EXISTS` :'(

yoav-lavi: It does support that

yoav-lavi: Ah, it doesn't support it in this specific case :(
But it does work for other cases, we use it in most places

cli/crates/server/src/bridge/server.rs
```diff
@@ -116,6 +117,7 @@ pub async fn start(port: u16, worker_port: u16, event_bus: Sender<Event>) -> Res
     let pool = SqlitePoolOptions::new().connect(&db_url).await?;
 
     query(PREPARE).execute(&pool).await?;
+    let _ = query(MIGRATION_001).execute(&pool).await; // Ignore error when `owned_by` column already exists.
```
yoav-lavi, line 120: Since this may error in other ways, we may want to check for the column beforehand / check for the specific error.
We could also consider a breaking change and instruct people to reset beforehand since we're not in public beta yet

yoav-lavi (changes requested):
See comment

cli/crates/server/src/sql/001_add_owned_by.sql
yoav-lavi: @Finistere it does support it, but not for this specific case for some reason

cli/crates/server/src/bridge/server.rs
yoav-lavi: > We could track migrations in a separate table
That should work and allow us to handle errors directly

tomasol:
I'm closing this because after refactoring https://github.com/grafbase/api/pull/1817 the owner is now an array and is tracked in the `records.document` JSON attribute. If needed, I can extract [f98fcda](https://github.com/grafbase/grafbase/pull/230/commits/f98fcda9157e4670c8bd13f69ccb717f3ed744d4) into a separate PR.

PR #231, opened 2023-03-09
Print miniflare logs when tracing is enabled

tomasol:
# Description

Use the inherited stdout, stderr for the miniflare process when the CLI is started with `--trace 2`.
To preserve current behavior (debug messages of CLI), use `--trace 1` . Passing `2` will print miniflare logs as well.

# Type of change

- [ ] 💔 Breaking
- [ ] 🚀 Feature
- [x] 🐛 Fix
- [ ] 🛠️ Tooling
- [ ] 🔨 Refactoring
- [ ] 🧪 Test
- [ ] 📦 Dependency

### DIFF
--- a/cli/crates/backend/src/server_api.rs
+++ b/cli/crates/backend/src/server_api.rs
@@ -15,10 +15,10 @@ type ServerInfo = (thread::JoinHandle<Result<(), ServerError>>, Receiver<ServerM
 /// returns [`BackendError::AvailablePort`] if no available port can  be found
 ///
 /// returns [`BackendError::PortInUse`] if search is off and the supplied port is in use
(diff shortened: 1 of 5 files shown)


Finistere (approved):
That was long overdue, thanks!

yoav-lavi:
Might be best to be able to separate these, perhaps an optional argument for `--trace`?

yoav-lavi (changes requested):
I'd consider an argument for `--trace` that enables this, but let's discuss if others disagree

yoav-lavi:
Another thing to consider here is that the tracing flag doesn't actually exist when we compile for release, but I'm not sure that distinction is still meaningful now that we plan to open source the worker.

yoav-lavi (approved):
Nice work!

PR #232, opened 2023-03-09
Fail on bridge startup error instead of hanging forever

tomasol:
# Description

If an error occurs e.g. during sqlite initialization, `Event::BridgeReady` is never sent. 

# Type of change

- [ ] 💔 Breaking
- [ ] 🚀 Feature
- [x] 🐛 Fix
- [ ] 🛠️ Tooling
- [ ] 🔨 Refactoring
- [ ] 🧪 Test
- [ ] 📦 Dependency

### DIFF
--- a/cli/crates/server/src/servers.rs
+++ b/cli/crates/server/src/servers.rs
@@ -124,9 +124,9 @@ async fn spawn_servers(
 
     trace!("waiting for bridge ready");
     tokio::select! {
-        _result = wait_for_event(receiver, Event::BridgeReady) => (),
-        result = &mut bridge_handle => result??
-    }
+        _ = wait_for_event(receiver, Event::BridgeReady) => (),
+        result = &mut bridge_handle => {result??; return Ok(());}
+    };
     trace!("bridge ready");
 
     let registry_path = environment
--- a/cli/crates/server/src/servers.rs
+++ b/cli/crates/server/src/servers.rs
@@ -10,6 +10,7 @@ use common::consts::EPHEMERAL_PORT_RANGE;
 use common::environment::Environment;
 use common::types::LocalAddressType;
 use common::utils::find_available_port_in_range;
+use futures_util::FutureExt;
 use std::env;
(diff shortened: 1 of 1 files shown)


cli/crates/server/src/servers.rs
```diff
@@ -118,12 +119,14 @@ async fn spawn_servers(
 
     let environment = Environment::get();
 
-    let bridge_handle = tokio::spawn(async move { bridge::start(bridge_port, worker_port, bridge_sender).await });
+    let mut bridge_handle =
+        tokio::spawn(async move { bridge::start(bridge_port, worker_port, bridge_sender).await }).fuse();
```
Finistere, line 123: Why is the fuse necessary? in the `tokio::select`, `poll` will only be called once, no?

tomasol: I'm not entirely sure it is needed, I added it to avoid possible panics in the `select!` macro since we are polling the same future.
Looking at the code it looks like this could happen if `Event::Reload` is received by `listener::start` before `Event::BridgeReady` is received by the first `select!`. The `bridge_handle` branch would win and the second `select!` would try to poll `bridge_handle` again. Not entirely sure if this race can happen, but if yes, even this fix is not 100% correct. Instead of a panic we could end with miniflare running, bridge server down and ignoring future `Event::Reload` events.

Finistere: ah good point, didn't think of other events when reading the review. Thanks!

Finistere (approved):

cli/crates/server/src/servers.rs
tomasol: To address my last sentence, `Event::Reload` will be received by `server_loop` so it is not an issue, but for clarity I've added [8ecd9f7](https://github.com/grafbase/grafbase/pull/232/commits/8ecd9f7bec83fcb46992a43c5e0da7d2c1c0430f)

PR #239, opened 2023-03-17
GB-3323: Add initial support for building resolvers

jakubadamw:
# Description

Add initial support for building resolvers using `npm`.

# Type of change

- [ ] 💔 Breaking
- [X] 🚀 Feature
- [ ] 🐛 Fix
- [ ] 🛠️ Tooling
- [ ] 🔨 Refactoring
- [ ] 🧪 Test
- [ ] 📦 Dependency


cli/crates/server/src/bridge/server.rs
```diff
@@ -14,14 +14,15 @@ use common::environment::Environment;
...
+use std::collections::HashMap;
 use std::net::{Ipv4Addr, SocketAddr};
 use std::sync::Arc;
 
 use tokio::sync::broadcast::Sender;
 use tower_http::trace::TraceLayer;
 
 async fn query_endpoint(
-    State(pool): State<Arc<SqlitePool>>,
+    State((_port, _environment_variables, pool)): State<(u16, HashMap<String, String>, Arc<SqlitePool>)>,
```
obmarg, line 25: Not important, but a `State<SomeStruct>` might have been more concise here?

cli/crates/server/src/custom_resolvers.rs
```diff
@@ -0,0 +1,217 @@
...
+
+    let artifact_directory_path_string = artifact_directory_path
+        .as_ref()
+        .to_str()
+        .ok_or(ServerError::CachePath)?;
+
+    let mut arguments = vec!["--prefix", artifact_directory_path_string];
+    arguments.extend(extra_arguments);
+
+    trace!("running 'npm {}'", arguments.iter().format(" "));
```
obmarg, line 34: Should this use `CommandType` rather than assuming it's always `npm`?

obmarg (approved):

cli/crates/server/src/bridge/server.rs
jakubadamw: Absolutely. But, in fact, these changes weren't supposed to be part of this PR, I failed at splitting my branch. So for now I took this out altogether. In the future PR we'll have a struct for this.

cli/crates/server/src/custom_resolvers.rs
jakubadamw: Definitely. Fixed.

PR #264, opened 2023-03-31
Escape environment variables that get added to wrangler.toml

jakubadamw:
# Description

# Type of change

- [ ] 💔 Breaking
- [ ] 🚀 Feature
- [X] 🐛 Fix
- [ ] 🛠️ Tooling
- [ ] 🔨 Refactoring
- [ ] 🧪 Test
- [ ] 📦 Dependency

### DIFF
--- a/cli/crates/server/src/custom_resolvers.rs
+++ b/cli/crates/server/src/custom_resolvers.rs
@@ -56,6 +56,25 @@ async fn run_npm_command<P: AsRef<Path>>(
     }
 }
 
+// https://toml.io/en/v1.0.0#string
+// > Any Unicode character may be used except those that must be escaped:
+// > quotation mark, backslash, and the control characters other than tab
+// (U+0000 to U+0008, U+000A to U+001F, U+007F).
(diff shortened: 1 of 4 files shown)


cli/crates/server/src/custom_resolvers.rs
```diff
@@ -187,7 +208,10 @@ async fn build_resolver(
             "#,
             vars = environment_variables
                 .iter()
-                .format_with("\n", |(key, value), f| f(&std::format_args!("{key} = \"{value}\"")))
+                .format_with("\n", |(key, value), f| f(&std::format_args!(
```
Finistere, line 211: Add `FIXME`, you should rely on `TOML` serialization and not escape by hand IMO

Finistere (approved):

cli/crates/server/src/custom_resolvers.rs
jakubadamw: Agreed.

jakubadamw: Will address in another PR.

PR #269, opened 2023-04-03
Link & Unlink

yoav-lavi:
# Description

## Features

- Implements `link` and `unlink`

## Fixes

- Uses paginated APIs in place of non paginated ones, limits results to 100 (5 for API keys)

# Type of change

- [ ] 💔 Breaking
- [x] 🚀 Feature
- [x] 🐛 Fix
- [ ] 🛠️ Tooling
- [ ] 🔨 Refactoring
- [ ] 🧪 Test
- [ ] 📦 Dependency


cli/crates/backend/src/api/link.rs
```diff
@@ -0,0 +1,96 @@
...
+use cynic::{http::ReqwestExt, QueryBuilder};
+use std::iter;
+
+/// # Errors
+///
+/// see [`ApiError`]
+#[allow(clippy::module_name_repetitions)]
+pub async fn get_viewer_data_for_link() -> Result<Vec<AccountWithProjects>, ApiError> {
+    // TODO consider if we want to do this elsewhere
+    if project_linked() {
```
jakubadamw, line 19: We seem to be mixing synchronous APIs with async. This checks if a file exists with a synchronous API within an `async` function. I know the CLi is a mix match but whenever we're in an `async` scope already, it'd be desirable for all I/O within to play nicely as well. Or we wrap it with `tokio::task::spawn_blocking()` if there's no async API to do the same thing.

jakubadamw: I guess all I meant to say is that we should use [`tokio::fs::try_exists()`](https://docs.rs/tokio/latest/tokio/fs/fn.try_exists.html) here. 🙂

jakubadamw (approved):
One issue around I/O API mixing, looks great otherwise! Great work!

cli/crates/backend/src/api/link.rs
yoav-lavi: Makes sense, fixed!

PR #284, opened 2023-04-07
Keep reloading on compilation errors

yoav-lavi:
# Description

## Features

- Continues the reload cycle on compilation errors, displays an error page instead of the playground and a GQL response instead of the endpoint containing the error

# Type of change

- [ ] 💔 Breaking
- [x] 🚀 Feature
- [ ] 🐛 Fix
- [ ] 🛠️ Tooling
- [ ] 🔨 Refactoring
- [ ] 🧪 Test
- [ ] 📦 Dependency

### DIFF
--- a/cli/crates/server/src/servers.rs
+++ b/cli/crates/server/src/servers.rs
@@ -147,8 +147,7 @@ async fn spawn_servers(
                 .ok()
                 .and_then(|stripped| String::from_utf8(stripped).ok())
                 .unwrap_or(error.to_string());
-            tokio::spawn(async move { error_server::start(worker_port, error.to_string(), bridge_event_bus).await })
-                .await??;
+            tokio::spawn(async move { error_server::start(worker_port, error, bridge_event_bus).await }).await??;
             return Ok(());
         }
     };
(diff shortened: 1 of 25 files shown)


cli/crates/server/src/servers.rs
```diff
@@ -123,7 +124,15 @@ async fn spawn_servers(
 
     let environment_variables: std::collections::HashMap<_, _> = crate::environment::variables().collect();
 
-    let resolvers = run_schema_parser(&environment_variables).await?;
+    let resolvers = match run_schema_parser(&environment_variables).await {
+        Ok(resolvers) => resolvers,
+        Err(error) => {
+            let _ = sender.send(ServerMessage::CompilationError(error.to_string()));
+            tokio::spawn(async move { error_server::start(worker_port, error.to_string(), bridge_event_bus).await })
```
jakubadamw, line 131: We could do the same for failed resolver compilation?

yoav-lavi: Adding this as well

Miaxos (approved):

PR #291, opened 2023-04-13
add database folder

yoav-lavi:
# Description

## Breaking

- Moves the database into its own subfolder to prevent the deletion of other files by `reset`

# Type of change

- [x] 💔 Breaking
- [ ] 🚀 Feature
- [ ] 🐛 Fix
- [ ] 🛠️ Tooling
- [ ] 🔨 Refactoring
- [ ] 🧪 Test
- [ ] 📦 Dependency


fbjork (approved):

yoav-lavi:
@fbjork reset currently removes only the database and resolvers / the registry remain (but are overwritten when the CLI is run)

The registry is still in the root of the folder, we can move it later if needed

PR #300, opened 2023-04-17
update default schema

notrab:
# Description

# Type of change

- [ ] 💔 Breaking
- [ ] 🚀 Feature
- [ ] 🐛 Fix
- [ ] 🛠️ Tooling
- [ ] 🔨 Refactoring
- [ ] 🧪 Test
- [ ] 📦 Dependency

### DIFF
--- a/cli/crates/backend/assets/default-schema.graphql
+++ b/cli/crates/backend/assets/default-schema.graphql
@@ -1,11 +1,41 @@
-type TodoList @model {
-  id: ID!
+# Welcome to Grafbase!
+# Build your own database, integrate auth, permission rules, custom resolvers, search, and more with Grafbase.
+
+# Configure Auth
+# https://grafbase.com/docs/auth
+#
+# schema @auth(providers: [{ type: oidc, issuer: "{{ env.ISSUER_URL }}" }], rules: [{ allow: private }]) {
+#   query: Query
+# }
+
+# Define Data Model(s)
+# https://grafbase.com/docs/database
+type Post @model @search {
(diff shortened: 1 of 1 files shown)


cli/crates/backend/assets/default-schema.graphql
```diff
@@ -1,11 +1,43 @@
...
+  name: String!
+  email: Email
+  posts: [Post]
+  comments: [Comment]
+  # Extend models with resolvers
+  # https://grafbase.com/docs/edge-gateway/resolvers
+  # gravatar: URL @resolver(name: "user/gravatar")
+}
+
+# Start your backend
```
yoav-lavi, line 41: Could we somehow format this so it doesn't all seem on the same level? The URL is essentially a comment and the header could probably be distinguished

cli/crates/backend/assets/default-schema.graphql
```diff
@@ -1,11 +1,43 @@
-type TodoList @model {
-  id: ID!
+# Welcome to Grafbase!
+# Build your own database, integrate auth, permission rules, custom resolvers, search, and more with Grafbase.
```
fbjork, line 2: "Define your data models" rather than "Build your own database"?

cli/crates/backend/assets/default-schema.graphql
```diff
@@ -1,11 +1,43 @@
...
+
+# Configure Auth
```
fbjork, line 4:
```suggestion
# Integrate Auth
```

PR #305, opened 2023-04-17
Try to fix the npm prerelease conditional

obmarg:
We appear to always be marking releases as pre-release.  Not entirely sure why, but this is an attempt at a fix.


yoav-lavi (approved):
Nice work!

PR #313, opened 2023-04-21
See if rust-cache is better than plain actions cache

obmarg:
# Description

We've been seeing pretty slow builds in CI.  This PR updates our rust builds to use the specialised rust-cache action rather than the plain actions cache, which _might_ be more effective.  Going to do a CI run or two and see.


obmarg:
Answer: not significantly

PR #316, opened 2023-04-24
GB-3599: Make --trace available in release mode

tomasol:
# Description

Allow using `--trace` flag in release builds to ease debugging.

# Type of change

- [ ] 💔 Breaking
- [ ] 🚀 Feature
- [x] 🐛 Fix
- [ ] 🛠️ Tooling
- [ ] 🔨 Refactoring
- [ ] 🧪 Test
- [ ] 📦 Dependency

### DIFF
--- a/cli/crates/cli/src/cli_input.rs
+++ b/cli/crates/cli/src/cli_input.rs
@@ -1,19 +1,15 @@
-use cfg_if::cfg_if;
 use clap::{arg, command, value_parser, Arg, ArgAction, Command};
 use indoc::indoc;
 
 /// creates the cli interface
 #[must_use]
 pub fn build_cli() -> Command {
-    cfg_if! {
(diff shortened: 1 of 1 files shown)


obmarg:
I'm fine with this, just wondering: are we intending for users to use the flag?  We could maybe make it a bit more user-friendly in a future PR if we are?  If it's just for us to debug then whatever

fbjork:
I think we should log errors in the terminal by default, but have different levels of verbosity.

tomasol:
> I'm fine with this, just wondering: are we intending for users to use the flag? We could maybe make it a bit more user-friendly in a future PR if we are? If it's just for us to debug then whatever

The main motivation is to be able to see things like `panic` in the worker. I don't expect the users to understand or fix the issue, just make bug reports more actionable.

Miaxos (approved):

PR #320, opened 2023-04-24
GB-3605: Allow opting out of using ~/.grafbase folder

tomasol:
# Description
For debug builds only, add a switch to disable using `~/.grafbase` folder. 
As the dynamodb configuration comes from wrangler.toml, we need to store it next to the project. This is needed for running multiple tests in parallel.

# Type of change

- [ ] 💔 Breaking
- [x] 🚀 Feature
- [ ] 🐛 Fix
- [ ] 🛠️ Tooling
- [ ] 🔨 Refactoring
- [ ] 🧪 Test
- [ ] 📦 Dependency


cli/crates/cli/src/init.rs
```diff
@@ -1,8 +1,11 @@
 use crate::{errors::CliError, output::report};
 use backend::project;
+use common::environment::Environment;
 
-pub fn init(name: Option<&str>, template: Option<&str>) -> Result<(), CliError> {
-    project::init(name, template).map_err(CliError::BackendError)?;
+pub fn init(name: Option<&str>, template: Option<&str>, no_home: bool) -> Result<(), CliError> {
+    let project_path = project::init(name, template).map_err(CliError::BackendError)?;
+    Environment::try_init(no_home, Some(project_path)).map_err(CliError::CommonError)?;
+    server::export_embedded_files(&Environment::get().user_dot_grafbase_path).map_err(CliError::ServerError)?;
```
Finistere, line 8: What was the problem with the previous location of the file export?

cli/crates/common/src/environment.rs
```diff
@@ -30,7 +30,9 @@ pub struct Environment {
     /// the path of `$PROJECT/grafbase/schema.graphql`, the Grafbase schema,
     /// in the nearest ancestor directory with said directory and file
     pub project_grafbase_schema_path: PathBuf,
-    /// the path of `$HOME/.grafbase`, the user level local developer tool cache directory
+    /// the path of `.grafbase` folder, contains local developer tool cache.
+    /// By default placed in $HOME, if that fails or `--nohome` is supplied, will be under the project's grafbase folder.
+    // TODO: rename to dot_grafbase_path
```
Finistere, line 35: IMHO I'd rather name it by its function, `shared_cache_path` or something like that.

cli/crates/server/src/servers.rs
```diff
@@ -70,8 +70,10 @@ pub fn start(port: u16, watch: bool, tracing: bool) -> (JoinHandle<Result<(), Se
 
                     tokio::select! {
                         result = start_watcher(environment.project_grafbase_path.clone(),  move |path, event_type| {
-                            let relative_path = path.strip_prefix(&environment.project_path).expect("must succeed by definition").to_owned();
-                            watch_event_bus.send(Event::Reload(relative_path, event_type)).expect("cannot fail");
+                            if !path.starts_with(&environment.user_dot_grafbase_path) {
```
Finistere, line 73: Why? Sure if the cache is updated we'll reload stuff, so we'll do more reloading than necessary, but we'll still reload properly if hot reload was expected. IMHO this is a place where specifying a different shared cache directory (home) would be useful, the new home could just be `~/.grafbase-dynamodb`.

cli/crates/server/src/servers.rs
```diff
@@ -270,12 +272,24 @@ async fn spawn_servers(
...
+/// * write otherwise
+///
+/// # Errors
+///
+/// returns [`ServerError::ReadVersion`] if the version file for the extracted worker files cannot be read
+///
+/// returns [`ServerError::CreateCacheDir`] if the `Environment::user_dot_grafbase_path` cannot be created
+///
+/// returns [`ServerError::WriteFile`] if a file cannot be written into `Environment.user_dot_grafbase_path`
+pub fn export_embedded_files(user_dot_grafbase_path: &Path) -> Result<(), ServerError> {
```
Finistere, line 289: Is there a reason not to rely on the environment anymore? I mean, don't get me wrong, I have nothing against not using globals. :D Just wondering. :)

cli/crates/cli/src/cli_input.rs
```diff
@@ -4,7 +4,20 @@ use indoc::indoc;
 /// creates the cli interface
 #[must_use]
 pub fn build_cli() -> Command {
-    command!()
+    cfg_if::cfg_if! {
+        if #[cfg(debug_assertions)] {
+            let command_builder = command!()
+                .arg(
+                    arg!(-n --nohome "Disable using .grafbase folder in user's home directory")
```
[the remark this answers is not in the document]
tomasol: Agreed, that's what the other PR is doing, thus closing this issue. Thanks for review!

PR #322, opened 2023-04-26
GB-3604: Allow running CLI integration tests with DynamoDB Local

tomasol:
# Description

Introduce feature `dynamodb` that is used for selecting between sqlite and dynamodb test harness. 
DynamoDB tests run the following steps:

1. `Environment.init` creates a new DynamoDB table
2. `grafbase dev` is called with the env var  `DYNAMODB_TABLE_NAME` and the inherited env vars.
3. After the test finishes, the table is dropped. In async context this must be done by the test.

# Type of change

- [ ] 💔 Breaking
- [ ] 🚀 Feature
- [ ] 🐛 Fix
- [ ] 🛠️ Tooling
- [ ] 🔨 Refactoring
- [x] 🧪 Test
- [ ] 📦 Dependency

### DIFF
--- a/cli/crates/cli/Cargo.toml
+++ b/cli/crates/cli/Cargo.toml
@@ -62,6 +62,8 @@ tempfile = "3"
 tokio = { version = "1", features = ["full"] }
 regex = "1"
 rstest = "0.17"
+rusoto_core = "0.48"
+rusoto_dynamodb = "0.48"
 wiremock = "0.5"
 
 [[bin]]
--- a/cli/crates/cli/Cargo.toml
+++ b/cli/crates/cli/Cargo.toml
@@ -69,7 +69,5 @@ wiremock = "0.5"
 [[bin]]
 name = "grafbase"
 
-
 [features]
-default = ["sqlite"]
-sqlite = []
+dynamodb = ["backend/dynamodb"]
--- a/cli/crates/cli/Cargo.toml
+++ b/cli/crates/cli/Cargo.toml
@@ -68,3 +68,8 @@ wiremock = "0.5"
 
 [[bin]]
 name = "grafbase"
+
+
+[features]
+default = ["sqlite"]
+sqlite = []
--- a/cli/crates/cli/tests/utils/environment.rs
+++ b/cli/crates/cli/tests/utils/environment.rs
@@ -3,7 +3,7 @@
 use super::async_client::AsyncClient;
 use super::kill_with_children::kill_with_children;
 use super::{cargo_bin::cargo_bin, client::Client};
-use common::consts::{GRAFBASE_DIRECTORY_NAME, GRAFBASE_SCHEMA_FILE_NAME};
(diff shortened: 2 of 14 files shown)


cli/crates/cli/Cargo.toml
```diff
@@ -62,7 +62,12 @@ tempfile = "3"
 tokio = { version = "1", features = ["full"] }
 regex = "1"
 rstest = "0.17"
+rusoto_core = "0.48"
+rusoto_dynamodb = "0.48"
```
obmarg, line 66: Could these be optional depending on the feature flag?

cli/crates/cli/tests/utils/environment.rs
```diff
@@ -49,42 +52,121 @@ fn get_free_port() -> u16 {
...
+                return Self {
+                    directory: other.directory.clone(),
+                    commands,
+                    endpoint,
+                    playground_endpoint,
+                    schema_path: other.schema_path.clone(),
+                    temp_dir,
+                    port,
+                };
+            }
```
obmarg, line 168: Could this just be a `#[cfg` on the `dynamodb_env` field rather than duplicating everything?

```suggestion
        Self {
            directory: other.directory.clone(),
            commands,
            endpoint,
            playground_endpoint,
            schema_path: other.schema_path.clone(),
            temp_dir,
            port,
            #[cfg(feature = "dynamodb")]
            dynamodb_env: dynamodb::DynamoDbEnvironment {
                dynamodb_client: None, // Only one dynamodb client is needed for the cleanup.
                table_name: other.dynamodb_env.table_name.clone(),
            }
        }
```

obmarg (approved):
Nice work

cli/crates/cli/Cargo.toml
tomasol: Would be nice but:
```
  dev-dependencies are not allowed to be optional: `rusoto_core`
````

File: cli/crates/cli/tests/utils/environment.rs (hunk 2 above)
Comment (tomasol) [reply cli/crates/cli/tests/utils/environment.rs]: Nice, applied!

File: cli/crates/cli/Cargo.toml (hunk 1 above)
Comment (obmarg) [reply cli/crates/cli/Cargo.toml:66]: Ah, I didn't spot these were under dev-deps (we have _a lot_ of dev deps) - never mind

File: cli/crates/cli/tests/utils/environment.rs [hunk 3]
```diff
@@ -74,16 +74,15 @@ impl Environment {
...
                 return (myself, cleanup_fut);
             } else {
                 return (Self::init_internal(port), std::future::ready(()));
             }
         );
     }
 
-    #[cfg(feature = "dynamodb")]
-    fn init_internal(dynamodb_env: dynamodb::DynamoDbEnvironment, port: u16) -> Self {
+    fn init_internal(port: u16, #[cfg(feature = "dynamodb")] dynamodb_env: dynamodb::DynamoDbEnvironment) -> Self {
```
obmarg, line 85: Nice 👌

PR #333, opened 2023-05-05
GB-3326: TypeScript Config

pimeys:
# Description

Introduce the ability to configure Grafbase projects using TypeScript to add type safety.

How to read the pull request(s):

- [New glue scripts to API](https://github.com/grafbase/api/pull/2166)
- [The @grafbase/sdk library to generate configs](https://github.com/grafbase/grafbase/tree/tsconfig/packages/grafbase-sdk)
- The changes in cli Rust code detecting are we using a TS config file or the graphql schema file.
- Running the embedded `ts-node` library for the client's TS config, which generates a new file in `~/.grafbase/generated/schema/schema.grapql`

# Type of change

- [ ] 💔 Breaking
- [x] 🚀 Feature
- [ ] 🐛 Fix
- [x] 🛠️ Tooling
- [ ] 🔨 Refactoring
- [ ] 🧪 Test
- [ ] 📦 Dependency

### DIFF
--- a/cli/crates/cli/src/output/report.rs
+++ b/cli/crates/cli/src/output/report.rs
@@ -3,8 +3,11 @@ use crate::{
     watercolor::{self, watercolor},
 };
 use colored::Colorize;
-use common::consts::{GRAFBASE_DIRECTORY_NAME, GRAFBASE_SCHEMA_FILE_NAME, LOCALHOST};
 use common::types::ResolverMessageLevel;
+use common::{
+    consts::{GRAFBASE_DIRECTORY_NAME, GRAFBASE_SCHEMA_FILE_NAME, LOCALHOST},
+    environment::Warning,
+};
 use std::path::Path;
 
 /// reports to stdout that the server has started
@@ -66,6 +69,20 @@ pub fn error(error: &CliError) {
     watercolor::output_error!("Error: {error}", @BrightRed);
     if let Some(hint) = error.to_hint() {
         watercolor::output_error!("Hint: {hint}", @BrightBlue);
+    }
+}
+
+pub fn warnings(warnings: &[Warning]) {
+    for warning in warnings {
+        let msg = warning.message();
+
+        watercolor::output!("Warning: {msg}", @BrightYellow);
+
+        if let Some(hint) = warning.hint() {
+            watercolor::output!("Hint: {hint}", @BrightBlue)
+        }
+
+        println!();
     }
 }
 
--- a/cli/crates/cli/src/output/report.rs
+++ b/cli/crates/cli/src/output/report.rs
@@ -79,7 +79,7 @@ pub fn warnings(warnings: &[Warning]) {
         watercolor::output!("Warning: {msg}", @BrightYellow);
 
         if let Some(hint) = warning.hint() {
-            watercolor::output!("Hint: {hint}", @BrightBlue)
+            watercolor::output!("Hint: {hint}", @BrightBlue);
         }
 
         println!();
--- a/cli/crates/common/src/environment.rs
+++ b/cli/crates/common/src/environment.rs
@@ -33,11 +33,13 @@ pub struct GrafbaseSchemaPath {
 
 impl GrafbaseSchemaPath {
     /// The location of the schema file.
+    #[must_use]
     pub fn location(&self) -> &SchemaLocation {
         &self.location
     }
 
     /// True, if we find both `schema.graphql` and `grafbase.config.ts`.
+    #[must_use]
     pub fn has_both_files(&self) -> bool {
         self.has_both_files
     }
--- a/cli/crates/common/src/environment.rs
+++ b/cli/crates/common/src/environment.rs
@@ -9,6 +9,7 @@ use crate::{
 };
 use once_cell::sync::OnceCell;
 use std::{
+    borrow::Cow,
     env,
     path::{Path, PathBuf},
 };
@@ -28,7 +29,6 @@ pub enum SchemaLocation {
 #[derive(Debug)]
 pub struct GrafbaseSchemaPath {
     location: SchemaLocation,
-    has_both_files: bool,
 }
 
 impl GrafbaseSchemaPath {
@@ -38,23 +38,15 @@ impl GrafbaseSchemaPath {
         &self.location
     }
 
-    /// True, if we find both `schema.graphql` and `grafbase.config.ts`.
-    #[must_use]
-    pub fn has_both_files(&self) -> bool {
-        self.has_both_files
-    }
-
     fn ts_config(path: PathBuf) -> Self {
         Self {
             location: SchemaLocation::TsConfig(path),
-            has_both_files: false,
         }
     }
 
     fn graphql(path: PathBuf) -> Self {
         Self {
             location: SchemaLocation::Graphql(path),
-            has_both_files: false,
         }
     }
 
@@ -64,6 +56,33 @@ impl GrafbaseSchemaPath {
         match self.location() {
             TsConfig(path) | Graphql(path) => path.parent(),
         }
+    }
+}
+
+#[derive(Debug)]
+pub struct Warning {
+    message: Cow<'static, str>,
+    hint: Option<Cow<'static, str>>,
+}
+
+impl Warning {
+    pub fn new(message: impl Into<Cow<'static, str>>) -> Self {
+        Self {
+            message: message.into(),
+            hint: None,
+        }
+    }
+
+    pub fn set_hint(&mut self, message: impl Into<Cow<'static, str>>) {
+        self.hint = Some(message.into());
+    }
+
+    pub fn message(&self) -> &str {
+        &self.message
+    }
+
+    pub fn hint(&self) -> Option<&str> {
+        self.hint.as_deref()
     }
 }
 
@@ -95,6 +114,8 @@ pub struct Environment {
     pub resolvers_build_artifact_path: PathBuf,
     /// the path within '$PROJECT/.grafbase' containing the database
     pub database_directory_path: PathBuf,
+    /// warnings when loading the environment
+    pub warnings: Vec<Warning>,
 }
 
 /// static singleton for the environment struct
@@ -114,8 +135,10 @@ impl Environment {
     ///
     /// returns [`CommonError::FindGrafbaseDirectory`] if the grafbase directory is not found
     pub fn try_init() -> Result<(), CommonError> {
+        let mut warnings = Vec::new();
+
         let project_grafbase_schema_path =
-            Self::get_project_grafbase_path()?.ok_or(CommonError::FindGrafbaseDirectory)?;
+            Self::get_project_grafbase_path(&mut warnings)?.ok_or(CommonError::FindGrafbaseDirectory)?;
 
         let project_grafbase_path = project_grafbase_schema_path
             .parent()
@@ -147,6 +170,7 @@ impl Environment {
                 resolvers_source_path,
                 resolvers_build_artifact_path,
                 database_directory_path,
+                warnings,
             })
             .expect("cannot set environment twice");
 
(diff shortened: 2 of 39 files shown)


packages/grafbase-sdk/package.json
```diff
@@ -0,0 +1,52 @@
+{
+  "name": "@grafbase/sdk",
+  "version": "0.18.8",
+  "description": "The Grafbase SDK",
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/grafbase/grafbase",
+    "directory": "cli/npm/sdk"
```
fbjork, line 8:
```suggestion
    "directory": "packages/grafbase-sdk"
```

packages/grafbase-sdk/package.json
```diff
@@ -0,0 +1,53 @@
+{
+  "name": "@grafbase/sdk",
+  "version": "0.18.8",
```
fbjork, line 3:
```suggestion
  "version": "0.0.1",
```

packages/grafbase-sdk/src/grafbase_schema.ts
```diff
@@ -0,0 +1,187 @@
+import { Model } from './model'
```
vitalikda, line 1: `grafbase_schema.ts` -> `grafbase-schema.ts`

packages/grafbase-sdk/README.md
```diff
@@ -0,0 +1,300 @@
...
+enum Fruits {
+  Apples,
+  Oranges
+}
+
+const fruits = g.enumType("Fruits", Fruits)
+
+// then use it e.g. in a model
+g.model("User", {
+  favoriteFruit: g.enum(fruits)
```
obmarg, line 175: Could this use `g.ref`, like when you're embedding objects?  Would free up `enum` to be used for the type.

packages/grafbase-sdk/src/grafbase_schema.ts
```diff
@@ -0,0 +1,187 @@
...
+    const iface = Object
+      .entries(types)
+      .reduce((iface, [name, definition]) => iface.field(name, definition), new Interface(name))
+
+    this.interfaces.push(iface)
+
+    return iface
+  }
+
+  public union(name: string, types: Record<string, Type>): Union {
```
obmarg, line 60: I might be missing something, but should types just be `[Type]` rather than a record - doesn't seem like the keys are doing anything

packages/grafbase-sdk/src/query.ts
```diff
@@ -0,0 +1,67 @@
...
+  public pushArgument(name: string, type: InputType): Query {
+    this.arguments.push(new QueryArgument(name, type))
+
+    return this
+  }
+  
+  public toString(): string {
+    let header = `extend type ${this.type} {`
+    let args = this.arguments.map(String).join(", ")
+    let query = `  ${this.name}(${args}): ${this.returns} @resolver(name: "${this.resolver}")`
```
obmarg, line 62: iirc you want to omit the brackets around `${args}` if there are no arguments

packages/grafbase-sdk/src/type.ts
```diff
@@ -0,0 +1,41 @@
...
+
+  public implements(i: Interface): Type {
+    this.interface = i
+
+    return this
+  }
+
+  public toString(): string {
+    const impl = this.interface ? ` implements ${this.interface?.name}` : ""
+    const header = `type ${this.name}${impl} {`
```
obmarg, line 30: types can implement multiple interfaces, so you might want to make `this.interface` a list

obmarg (commented):
Nice, I've just got a couple of comments

packages/grafbase-sdk/src/grafbase_schema.ts
pimeys: So the interface is per notion document:

```
g.union('UserOrAddress', { user, address })
```

Keys do nothing, but I think an array works here either if we want kind of a set(ish) behavior.

obmarg (approved):
LGTM

vitalikda (approved):
Looking forward to using it! 🤩

cli/crates/server/src/servers.rs
```diff
@@ -477,6 +483,64 @@ async fn run_schema_parser(
     Ok(detected_resolvers)
 }
 
+/// Parses a TypeScript Grafbase configuration and generates a GraphQL schema
+/// file to the filesystem, returning a path to the generated file.
+async fn parse_and_generate_config_from_ts(ts_config_path: &Path) -> Result<String, ServerError> {
+    let environment = Environment::get();
+
+    if environment.project_grafbase_schema_path.has_both_files() {
+        warn!("grafbase.config.ts takes precedence over schema.graphql - please choose one and delete the other to avoid conflicts");
```
yoav-lavi, line 492: These traces are used for us, for user output we should use the report module

cli/crates/server/src/servers.rs
```diff
@@ -477,6 +483,64 @@ async fn run_schema_parser(
...
+    }
+
+    let config_parser_path = environment
+        .user_dot_grafbase_path
+        .join(SCHEMA_PARSER_DIR)
+        .join(CONFIG_PARSER_SCRIPT);
+
+    let ts_node_path = environment
+        .user_dot_grafbase_path
+        .join("node_modules/ts-node/dist/bin.js");
```
yoav-lavi, line 509: Could this be in a const?

cli/crates/server/src/servers.rs
```diff
@@ -477,6 +483,64 @@ async fn run_schema_parser(
...
+        ts_node_path.as_path(),
+        config_parser_path.as_path(),
+        ts_config_path,
+        &generated_config_path,
+    ];
+
+    let node_command = Command::new("node")
+        .args(args)
+        .current_dir(&environment.user_dot_grafbase_path)
+        .stdin(Stdio::piped())
```
yoav-lavi, line 521: Do we need to pipe stdin here?

packages/grafbase-sdk/jest.config.js
```diff
@@ -0,0 +1,5 @@
+/** @type {import('ts-jest').JestConfigWithTsJest} */
+module.exports = {
+  preset: 'ts-jest',
+  testEnvironment: 'node',
+};
```
yoav-lavi, line 5: Missing a newline here

packages/grafbase-sdk/src/grafbase-schema.ts
```diff
@@ -0,0 +1,223 @@
...
+      query = q
+    }
+
+    this.queries.push(query)
+
+    return query
+  }
+
+  public mutation(name: string, definition: QueryInput): Query {
+    const q = new Query(
```
yoav-lavi, line 112: I'd avoid single letter variables personally

cli/crates/server/src/errors.rs
```diff
@@ -87,6 +87,10 @@ pub enum ServerError {
     #[error("could not parse grafbase/schema.graphql\n{0}")]
     ParseSchema(String),
 
+    /// returned if the typescript config parser command exits unsuccessfully
+    #[error("could not load grafbase/grafbase.config.ts\n{0}")]
```
yoav-lavi, line 91: on newer errors we typically do something like `caused by: {0}`, I'd copy one of those

yoav-lavi: Although it'd depend if the error output here makes sense in that context

packages/grafbase-sdk/src/field/typedefs.ts
```diff
@@ -0,0 +1,308 @@
...
+
+  public unique(scope?: string[]): GUniqueDef {
+    return new GUniqueDef(this, scope)
+  }
+
+  public search(): GSearchDef {
+    return new GSearchDef(this)
+  }
+
+  public list(): GListDef {
```
yoav-lavi, line 127: Why are we prefixing everything with G?

yoav-lavi (changes requested):
Really nice work! See comments

cli/crates/common/src/environment.rs
```diff
@@ -13,6 +13,60 @@ use std::{
...
+
+    fn graphql(path: PathBuf) -> Self {
+        Self {
+            location: SchemaLocation::Graphql(path),
+            has_both_files: false,
+        }
+    }
+
+    fn parent(&self) -> Option<&Path> {
+        use SchemaLocation::{Graphql, TsConfig};
```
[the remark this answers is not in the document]
pimeys: I don't like to have this kind of short forms available in the whole file. It's about making this one match nicer, but it is a bit weird to have `Graphql` and `TsConfig` available in the whole module.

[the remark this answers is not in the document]
yoav-lavi: I normally default to top imports but not critical if you prefer it this way

cli/crates/server/src/errors.rs
yoav-lavi: So its clear that the errors are related

cli/crates/cli/src/output/report.rs
```diff
@@ -69,6 +72,20 @@ pub fn error(error: &CliError) {
...
+    for warning in warnings {
+        let msg = warning.message();
+
+        watercolor::output!("Warning: {msg}", @BrightYellow);
+
+        if let Some(hint) = warning.hint() {
+            watercolor::output!("Hint: {hint}", @BrightBlue)
+        }
+
+        println!();
```
yoav-lavi, line 85: We could just add a `\n` after the hint

packages/grafbase-sdk/src/field/typedefs.ts
```diff
@@ -0,0 +1,298 @@
...
+  Date = 'Date',
+  DateTime = 'DateTime',
+  IPAddress = 'IPAddress',
+  Timestamp = 'Timestamp',
+  URL = 'URL',
+  JSON = 'JSON',
+  PhoneNumber = 'PhoneNumber'
+}
+
+export class SearchDef {
```
yoav-lavi, line 33: What are these `Def` suffixes? Are they meant as "directive"?

cli/crates/cli/src/output/report.rs
pimeys: but if hint is not there, we need to add it to the warning

packages/grafbase-sdk/src/field/typedefs.ts
pimeys: Definition.

cli/crates/common/src/environment.rs
```diff
@@ -109,44 +192,65 @@ impl Environment {
...
-
-                // if we're looking at a directory called `grafbase`, also check for the schema in the current directory
-                if let Some(first) = path.components().next() {
-                    if Path::new(&first) == PathBuf::from(GRAFBASE_DIRECTORY_NAME) {
-                        path.push(GRAFBASE_SCHEMA_FILE_NAME);
-                        if path.is_file() {
-                            return Some(path);
-                        }
-                        path.pop();
+    fn get_project_grafbase_path(warnings: &mut Vec<Warning>) -> Result<Option<GrafbaseSchemaPath>, CommonError> {
```
yoav-lavi, line 203: Might make more sense for this function to receive self than mutate an input variable, is it used elsewhere?

packages/grafbase-sdk/src/field/typedefs.ts
yoav-lavi: Sure, but why not just `Search`?

cli/crates/common/src/environment.rs
pimeys: We don't have an env instance yet to have it as `self`. I'd also have these outside of `Environment` instead of static methods.

yoav-lavi: Hmm, ok, could this maybe return a tuple of result and warnings?
Not sure if it'd be better but I don't love mutating input vars

pimeys: Most of them could be without, but `DateDef` overlaps with `Date`.

yoav-lavi: @pimeys I think you commented on the wrong chain

pimeys: We accumulate warnings, and this is quite often (read: rustc, rust-analyzer) done with a vector. We have more than one warning, so we collect them throughout the run to a collection, then eventually output them.

packages/grafbase-sdk/src/field/typedefs.ts
yoav-lavi: Ok we can leave it as is

yoav-lavi (approved):
Great work!

PR #345, opened 2023-05-08
Prettier

yoav-lavi:
# Description

## Tooling

- Run prettier in the CI for the CLI project

# Type of change

- [ ] 💔 Breaking
- [ ] 🚀 Feature
- [ ] 🐛 Fix
- [x] 🛠️ Tooling
- [ ] 🔨 Refactoring
- [ ] 🧪 Test
- [ ] 📦 Dependency


.github/workflows/cli.yml
```diff
@@ -43,6 +43,20 @@ jobs:
               - '.github/**/*'
               - '!.github/workflows/packages.yml'
 
+  prettier:
```
pimeys, line 46: Could we also have this for the packages?

pimeys (approved):

.github/workflows/cli.yml
yoav-lavi: Yeah, should probably be a separate PR though since I think there's some fixes to be done there

PR #347, opened 2023-05-09
Fix remote_unions_test when running with dynamodb

tomasol:
# Description

Switch the test to `multi_thread` flavor and call `Environment::init_async` instead of `init` so that dynamodb cleanup routine in `Environment::drop` works.

# Type of change

- [ ] 💔 Breaking
- [ ] 🚀 Feature
- [x] 🐛 Fix
- [ ] 🛠️ Tooling
- [ ] 🔨 Refactoring
- [ ] 🧪 Test
- [ ] 📦 Dependency

### DIFF
--- a/cli/crates/cli/tests/README.md
+++ b/cli/crates/cli/tests/README.md
@@ -1,6 +1,7 @@
 # Running with DynamoDB Local
 
 ## Assets
+
 Assets in the API repo need to be built:
 
 ```sh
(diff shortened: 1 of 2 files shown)


yoav-lavi:
@tomasol our production is single threaded though, we'd probably want to retain that

obmarg (approved):

PR #348, opened 2023-05-09
GB-3397 - Add argument support for `create`

yoav-lavi:
# Description

## Features

- Allows passing arguments to `create` rather than using interactive input|

```
Usage: grafbase create [OPTIONS]

Options:
  -n, --name <name>       The name to use for the new project
  -a, --account <slug>    The slug of the account in which the new project should be created
  -r, --regions <region>  The regions in which the database for the new project should be created
  -h, --help              Print help
```

## Fixes

- Fixes a few cases of output to STDERR instead of STDOUT

# Type of change

- [ ] 💔 Breaking
- [x] 🚀 Feature
- [x] 🐛 Fix
- [ ] 🛠️ Tooling
- [ ] 🔨 Refactoring
- [ ] 🧪 Test
- [ ] 📦 Dependency

### DIFF
--- a/cli/crates/cli/src/cli_input.rs
+++ b/cli/crates/cli/src/cli_input.rs
@@ -1,4 +1,4 @@
-use clap::{arg, command, value_parser, Arg, ArgAction, Command};
+use clap::{arg, command, value_parser, Arg, ArgAction, ArgGroup, Command};
 use indoc::indoc;
 
 /// creates the cli interface
@@ -51,7 +51,17 @@ pub fn build_cli() -> Command {
         "}))
         .subcommand(Command::new("login").about("Logs into your Grafbase account"))
         .subcommand(Command::new("logout").about("Logs out of your Grafbase account"))
-        .subcommand(Command::new("create").about("Set up and deploy a new project"))
+        .subcommand(
+            Command::new("create").about("Set up and deploy a new project").args(&[
+                arg!(-n --name <name> "The name to use for the new project"),
+                arg!(-a --account <slug> "The slug of the account in which the new project should be created"),
+                arg!(-r --regions <region> "The regions in which the database for the new project should be created")
+            ])
+            .group(ArgGroup::new("create")
+            .args(["regions", "account", "name"])
+            .multiple(true)
+            .requires_all(["regions", "account", "name"])),
+        )
         .subcommand(Command::new("deploy").about("Deploy your project"))
         .subcommand(Command::new("link").about("Connect a local project to a remote project"))
         .subcommand(Command::new("unlink").about("Disconnect a local project from a remote project"))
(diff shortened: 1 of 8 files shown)


cli/crates/cli/src/cli_input.rs
```diff
@@ -51,7 +51,17 @@ pub fn build_cli() -> Command {
         "}))
         .subcommand(Command::new("login").about("Logs into your Grafbase account"))
         .subcommand(Command::new("logout").about("Logs out of your Grafbase account"))
-        .subcommand(Command::new("create").about("Set up and deploy a new project"))
+        .subcommand(
+            Command::new("create").about("Set up and deploy a new project").args(&[
```
obmarg, line 55: Not suggesting we change it, but I've been wondering for a while: is there a reason we're defining all this by hand rather than using the derives in clap?

cli/crates/cli/src/create.rs
```diff
@@ -25,15 +25,38 @@ impl Display for RegionSelection {
...
+        let domains = create::create(&account_id, arguments.name, &arguments.regions)
+            .await
+            .map_err(CliError::BackendApiError)?;
+
+        report::created(arguments.name, &domains);
+
+        return Ok(());
+    }
+
     let options: Vec<AccountSelection> = accounts.into_iter().map(AccountSelection).collect();
```
pimeys, line 60: Should we split this to another function? Then we can do `if let ... else { interactive_flow }`...

obmarg (approved):

cli/crates/cli/src/cli_input.rs
yoav-lavi: I recall the derive API being limited in comparison, we could look into it again if it has parity

cli/crates/cli/src/main.rs
```diff
@@ -103,7 +110,22 @@ fn try_main() -> Result<(), CliError> {
         Some(("reset", _)) => reset(),
         Some(("login", _)) => login(),
         Some(("logout", _)) => logout(),
-        Some(("create", _)) => create(),
+        Some(("create", matches)) => {
+            let account_slug = matches.get_one::<String>("account").map(AsRef::as_ref);
```
pimeys, line 114: Do we really need these variables? It would also just work with something like:

```rust
let arguments = matches.get_one::<String>("account")
    .zip(matches.get_one::<String>("name"))
    .zip(matches.get_one::<String>("regions"))
    .map(...);
```

pimeys (approved):
nitpicks, but ready to go if you want.

cli/crates/cli/src/main.rs
yoav-lavi: Fair enough, fixed

PR #353, opened 2023-05-10
Update Path filter checks

vitalikda:
# Description

This aims to improve the reliability and consistency of change detection for the CLI and Packages 

# Type of change

- [x] 🐛 Fix

### DIFF
--- a/.github/workflows/cli.yml
+++ b/.github/workflows/cli.yml
@@ -40,8 +40,7 @@ jobs:
           filters: |
             build:
               - 'cli/**/*'
-              - '.github/**/*'
-              - '!.github/workflows/packages.yml'
+              - '.github/workflows/cli.yml'
 
   prettier:
     needs: [detect-change-type]
--- a/.github/workflows/cli.yml
+++ b/.github/workflows/cli.yml
@@ -41,6 +41,7 @@ jobs:
             build:
               - 'cli/**/*'
               - '.github/workflows/cli.yml'
+              - '.github/actions/**/*'
 
   prettier:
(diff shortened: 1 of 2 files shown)


.github/workflows/cli.yml
```diff
@@ -40,8 +40,7 @@ jobs:
           filters: |
             build:
               - 'cli/**/*'
-              - '.github/**/*'
-              - '!.github/workflows/packages.yml'
+              - '.github/workflows/cli.yml'
```
vitalikda, line 43: I updated the filter due to it affects all files except `packages.yml`
Do we need to run CLI workflow for anything else?

obmarg: `cli.yml` definitely pulls in a bunch of other files inside `.github/workflows` & `.github/actions`.  We probably want to list them here as well?  or add some more ignores for whatever prompted this PR in the first place?

vitalikda: @obmarg wanted to be explicit here 

Looking into cli related workflows 
- `.github/workflows/cli.yml` - included in filter 
- `.github/workflows/cli-partial-release.yml` - triggered by `cli.yml`
- `.github/workflows/sanitize-alert-schedule.yml` - triggered by cron or manually 

So I do not think workflows besides cli should be included. Actions on the other hand should - I will add them

Wdyt?

obmarg: Yeah, we can try that.

vitalikda: Added `.github/actions` 👍

PR #357, opened 2023-05-12
feat(resolvers): GB-3629: Move to lazy compilation of resolvers in local dev

jakubadamw:
# Description

Move to lazy compilation of resolvers.
We used to share a single miniflare instance with resolvers being compiled before its start and attached as separate modules under their own routes. With lazy loading we'll be building resolvers and starting separate miniflare instances for them on demand.

# Type of change

- [ ] 💔 Breaking
- [X] 🚀 Feature
- [ ] 🐛 Fix
- [ ] 🛠️ Tooling
- [ ] 🔨 Refactoring
- [ ] 🧪 Test
- [ ] 📦 Dependency


pimeys (commented):
This is a big change to a codebase not familiar to me yet. Could you explain in broad terms what are we changing and what is the strategy here? I can rubber-stamp if I must, but I'd like @yoav-lavi to take a look before doing so.

cli/crates/server/src/bridge/server.rs
```diff
@@ -121,14 +138,98 @@ async fn search_endpoint(
...
+
+    let resolver_worker_port = loop {
+        let mut resolver_builds = handler_state.resolver_builds.lock().await;
+
+        let notify = if let Some(resolver_build) = resolver_builds.get(&payload.resolver_name) {
+            match resolver_build {
+                ResolverBuild::Succeeded { worker_port, .. } => break *worker_port,
+                ResolverBuild::Failed => return Err(ApiError::ResolverSpawnError),
+                ResolverBuild::InProgress { notify } => {
+                    if Arc::strong_count(notify) == 1 {
```
Finistere, line 158: What's the purpose here of this branch? Shouldn't you always wait where it's `InProgress`?

cli/crates/server/src/bridge/resolvers.rs
```diff
@@ -31,16 +32,109 @@ struct ResolverResponse {
...
+        .map_err(|_| ApiError::CouldNotFindPortForResolverWorker)?;
+    let resolver_worker_port_string = resolver_worker_port.to_string();
+
+    let miniflare_path = environment
+        .user_dot_grafbase_path
+        .join(crate::consts::MINIFLARE_CLI_JS_PATH)
+        .canonicalize()
+        .unwrap();
+
+    let resolver_name_cloned = resolver_name.to_owned();
```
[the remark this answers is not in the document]
jakubadamw: Not really, the original `resolver_name` is used down below.

jakubadamw:
> This is a big change to a codebase not familiar to me yet. Could you explain in broad terms what are we changing and what is the strategy here? I can rubber-stamp if I must, but I'd like @yoav-lavi to take a look before doing so.

We used to share a single miniflare instance with resolvers being compiled before its start and attached as separate modules under their own routes. With lazy loading we'll be building resolvers and starting separate miniflare instances for them on demand. So the main change is the move from a single miniflare instance to multiple.

cli/crates/server/src/bridge/resolvers.rs
```diff
@@ -31,16 +32,109 @@ struct ResolverResponse {
     rest: serde_json::Value,
 }
 
+async fn wait_until_resolver_ready(resolver_worker_port: u16, resolver_name: &str) -> bool {
+    const RESOLVER_WORKER_MINIFLARE_READY_RETRY_COUNT: usize = 30 * 4;
+    const RESOLVER_WORKER_MINIFLARE_READY_RETRY_INTERVAL: std::time::Duration = std::time::Duration::from_millis(250);
```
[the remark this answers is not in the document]
jakubadamw: The total duration should be long enough to allow the build of a resolver to finish. If a resolver depends on many external packages, it'll take time to download and build them. In fact, this could just wait infinitely…

cli/crates/server/src/servers.rs
```diff
@@ -104,18 +104,25 @@ async fn server_loop(
             result = spawn_servers(worker_port, bridge_port, watch, sender.clone(), event_bus.clone(), path_changed.as_deref(), tracing) => {
                 result?;
             }
-            path = wait_for_event_and_match(receiver, |event| match event {
-                Event::Reload(path) => Some(path),
-                Event::BridgeReady => None,
+            event = wait_for_event_and_match(receiver, |event| match event {
```
Finistere, line 107: Doesn't seem really useful as a change here

cli/crates/server/src/bridge/server.rs
jakubadamw: It's to handle interrupted builds when a request invoking a resolver gets cancelled by the client. I agree this isn't a great way to handle this, but for a better approach we would need a dedicated resolver build server which I'd like to do in a future PR when I do the single HTTP server refactoring.

cli/crates/server/src/bridge/server.rs
```diff
@@ -121,14 +138,98 @@ async fn search_endpoint(
...
+                        notify.notified().await;
+                        continue;
+                    }
+                }
+            }
+        } else {
+            Arc::new(Notify::new())
+        };
+
+        resolver_builds.insert(
```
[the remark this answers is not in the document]
jakubadamw: Because we also want to `insert()` in turn after the first `if` branch. 🙂 But I see what you mean, I guess we can not remove it in the first place (line 160). Fair enough, let me see…

cli/crates/server/src/servers.rs
jakubadamw: Right, it was needed for a more substantial change that has since been removed from this PR. Reverting this too.

cli/crates/server/src/bridge/server.rs
Finistere: Ah got it, didn't realize at first `Arc::strong_count(notify) == 1` meant that the `resolver_builds` was the sole owner of notify. Might be worth a comment IMHO.

jakubadamw: Added, a comment. 🙂

Finistere: Thanks!

cli/crates/server/src/bridge/resolvers.rs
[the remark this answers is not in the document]
jakubadamw: Actually, scratch that, I'm being silly here… Of course, this isn't actually waiting for the build, but only for miniflare to start responding to requests. My bad. Yes, this needs to be lowered.

PR #358, opened 2023-05-12
chore: Switch to struct args from builder

pimeys:
# Description

Makes it simpler to use the command line arguments.

# Type of change

- [ ] 💔 Breaking
- [ ] 🚀 Feature
- [ ] 🐛 Fix
- [ ] 🛠️ Tooling
- [x] 🔨 Refactoring
- [ ] 🧪 Test
- [ ] 📦 Dependency


cli/crates/cli/src/cli_input.rs
```diff
@@ -1,75 +1,193 @@
...
+    }
+
+    fn completions_for_shell(generator: impl Generator) {
+        clap_complete::generate(generator, &mut Args::command(), "grafbase", &mut std::io::stdout());
+    }
+}
+
+#[derive(Debug, Parser)]
+pub struct CompletionsCommand {
+    /// The shell to generate completions for.
```
yoav-lavi, line 55: If we're not adding the second line here I think we can remove the dot

yoav-lavi (approved):
Really nice work!
Could we diff the help output between old and new to make sure it acts the same?
Possibly would be good to have a snapshot for it

pimeys:
# Screenshots from this branch

pimeys:
# Screenshots from this branch

PR #363, opened 2023-05-15
Speed up CI for the CLI

obmarg:
CI for the CLI seems to take 25-30 minutes _most of the time_.  It's a bit
ridiculous.  This PR has a bunch of changes that seem to improve things:

1. Release builds on my laptop are also excruciatingly slow.  From some testing
   this seems to be related to `RustEmbed` - I'm not sure whether it's the
   compression rust embed is doing or the fact that it's including every binary
   blob separately, but embedding a single `.tar.gz` manually halved release
   compile time on my laptop.
2. The size of the assets we're embedding seem to have a direct impact on
   compile times (probably due to [this rust issue](https://github.com/rust-lang/rust/issues/65818)).  A `.tar.gz` was a
   good 30% smaller than a `.zip` so I'll have a separate PR to the `api` repo
   to output one of these.  We could get some more improvements by bundling
   less things - I accidentally tried a build without `typescript` in it and
   shaved another 20-30 seconds off my build.
3. `lto` & `codegen-units = 1` were also slowing things down a bit.  Shaved
    another 25 seconds off builds on my machine.   Suspect we can live without
    them.
4. I've split the individual package builds out so they can be parallelised,
   and also to improve the `lint` cache (I _suspect_ some stuff in `target` was
   being clobbered by the individual builds, forcing any `lint` builds that use
   the cache to repeat more work than otherwise.  Not certain though, it's very
   slow to test all this stuff)
5. I've disabled the sanitize step for `aarch64` builds - it's not actually
   testing on `aarch64` so seemed a bit pointless when we've already got an
   `x86_64` build doing the same thing.
6. Disabled debug artifacts, because they go in the cache and we don't need
   them on CI.
7. Adjusted cache keys & cache prefixes a bit to try and improve things.
8. Updated the builds to record timings and push them as artifacts, so we can
   see more details about why it's slow.
9. After I'd done that, I noticed:
   - Tantivy was taking several minutes to compile.  Disabling it's default
     features improved this somewhat.
   - TOML was taking about a minute to compile.  Looking into this it was only
     used for panic reports, which could easily be JSON.  So I made that
     change.

This seems to have reduced an uncached CI run to about 20 minutes, and so far
I've been hitting caches fairly consistently and getting builds of 9-15
minutes (though it's possible that's because I was the only person doing things
at the weekend):

CleanShot 2023-05-15 at 12 27 39@2x

### DIFF
--- a/.github/workflows/cli.yml
+++ b/.github/workflows/cli.yml
@@ -17,8 +17,10 @@ permissions:
 
 env:
   GRAFBASE_SKIP_ASSET_VERSION_CHECK: 'true'
-  ASSETS_VERSION: 'release/2ad2c2a-2023-05-12'
-  PROD_ASSETS: assets.grafbase.com
+  #ASSETS_VERSION: 'release/2ad2c2a-2023-05-12'
+  ASSETS_VERSION: 'graeme-playing-around/98fa98d-2023-05-13'
+    #PROD_ASSETS: assets.grafbase.com
+  PROD_ASSETS: assets.grafbase.dev
   CARGO_TERM_COLOR: 'always'
   CARGO_INCREMENTAL: 0
 
--- a/.github/workflows/cli.yml
+++ b/.github/workflows/cli.yml
@@ -17,10 +17,8 @@ permissions:
 
 env:
   GRAFBASE_SKIP_ASSET_VERSION_CHECK: 'true'
-  #ASSETS_VERSION: 'release/2ad2c2a-2023-05-12'
-  ASSETS_VERSION: 'graeme-playing-around/98fa98d-2023-05-13'
-    #PROD_ASSETS: assets.grafbase.com
-  PROD_ASSETS: assets.grafbase.dev
+  ASSETS_VERSION: 'release/f0516a0-2023-05-16'
+  PROD_ASSETS: assets.grafbase.com
   CARGO_TERM_COLOR: 'always'
   CARGO_PROFILE_DEV_DEBUG: 0
   CARGO_PROFILE_TEST_DEBUG: 0
--- a/.github/workflows/cli.yml
+++ b/.github/workflows/cli.yml
@@ -22,6 +22,8 @@ env:
     #PROD_ASSETS: assets.grafbase.com
   PROD_ASSETS: assets.grafbase.dev
   CARGO_TERM_COLOR: 'always'
+  CARGO_PROFILE_DEV_DEBUG: 0
+  CARGO_PROFILE_TEST_DEBUG: 0
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
--- a/.github/workflows/cli.yml
+++ b/.github/workflows/cli.yml
@@ -92,10 +92,10 @@ jobs:
           # However, each package is published individually, so we're checking that all packages compile
           # with only their defined features.
           # See: https://github.com/rust-lang/cargo/issues/4463
-          (cd common && cargo check)
-          (cd server && cargo check)
-          (cd backend && cargo check)
-          (cd cli && cargo check)
+          (cd common && cargo check --release)
+          (cd server && cargo check --release)
+          (cd backend && cargo check --release)
+          (cd cli && cargo check --release)
 
   windows:
     needs: [lint]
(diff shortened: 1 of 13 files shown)


.github/workflows/cli.yml
```diff
@@ -70,6 +72,7 @@ jobs:
         with:
           platform: linux
           cache-key: ${{ runner.os }}-${{ runner.arch }}-cargo-dev-${{ hashFiles('cli/Cargo.lock') }}
+          restore-key: ${{ runner.os }}-${{ runner.arch }}-cargo-dev
```
Finistere, line 75: You're never building anything for that `restore-key`, no?

cli/Cargo.toml
```diff
@@ -1,7 +1,18 @@
 [workspace]
 members = ["crates/*"]
 
+# TODO: Probably only do these on CI if they work
+
+[profile.dev]
```
Finistere, line 6: `CARGO_PROFILE_DEV_DEBUG=0` in the CI instead? While I'm rarely debugging, I would still be surprised to not see any debug info locally.

.github/workflows/cli.yml
```diff
@@ -17,8 +17,10 @@ permissions:
 
 env:
   GRAFBASE_SKIP_ASSET_VERSION_CHECK: 'true'
-  ASSETS_VERSION: 'release/2ad2c2a-2023-05-12'
-  PROD_ASSETS: assets.grafbase.com
+  #ASSETS_VERSION: 'release/2ad2c2a-2023-05-12'
```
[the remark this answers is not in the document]
obmarg: Yeah, this is just temporary till I merge that `api` PR

.github/workflows/cli.yml
obmarg: The restore-key looks for any cache with the given prefix, so this restore-key will match an older version of the cache above.

I am actually wondering if the `restore-key`s are doing more harm than good, but that's something I want to play with in a future PR (if I get time)

Finistere: Ah right always forget it's a prefix. :D The only problem I see is that there won't be any cleanup, old artifacts will accumulate in the cache. Now how big of an issue it actually is, no idea. For the cli, it's probably fine to just try it out and see.

obmarg: > The only problem I see is that there won't be any cleanup, old artifacts will accumulate in the cache.

Yeah, that's exactly what I meant by "doing more harm than good".  But if we want to ditch them it'd be good to make that change in isolation so it's a bit clearer whether it is or isn't helping

Finistere: But... you're adding that change right now currently, no? What's there to change in later PR?

Finistere (approved):

yoav-lavi:
>lto & codegen-units = 1 were also slowing things down a bit. Shaved
another 25 seconds off builds on my machine. Suspect we can live without
them.

How much does this increase the binary by?
Does it affect performance? I'm assuming it doesn't, IIRC it just reuses code

yoav-lavi:
>I've disabled the sanitize step for aarch64 builds - it's not actually
testing on aarch64 so seemed a bit pointless when we've already got an
x86_64 build doing the same thing


Why not? It should be testing on that image

yoav-lavi:
Really nice work!

yoav-lavi:
@obmarg AFAIK we are using that image: 

```yaml
  darwin:
    needs: [lint]
    runs-on: macos-latest
    strategy:
      fail-fast: false
      matrix:
        target: [aarch64-apple-darwin, x86_64-apple-darwin]
  ```

yoav-lavi:
Oh nevermind it's cross compiling. I remembered that we did run natively

obmarg:
> How much does this increase the binary by? Does it affect performance? ~I'm assuming it doesn't, IIRC it just reuses more code~ Edit: Actually I'm assuming it could since there's probably edge cases where you can optimize things if you know how they're used outside a module

Increases the binary by 3MB (from 36 -> 39).

Performance wise: I assume it could be worse, but I'm also not sure it matters?  Are we doing anything seriously performance critical in the CLI?

fbjork (approved):

PR #378, opened 2023-05-17
Pull in tar-rs fork to work around startup issue.

obmarg:
# Description

`tar-rs` has an issue where it won't overwrite files if they're stored as hard links in the tar.  We install our asset `node_modules` with pnpm, which uses hard links to deduplicate content.  This means `grafbase dev` will refuse to overwrite some existing files on startup.

I've [fixed this and PRd it](https://github.com/alexcrichton/tar-rs/pull/319), this pulls the fix in the meantime.

# Type of change

- [ ] 💔 Breaking
- [ ] 🚀 Feature
- [x] 🐛 Fix
- [ ] 🛠️ Tooling
- [ ] 🔨 Refactoring
- [ ] 🧪 Test
- [ ] 📦 Dependency

### DIFF
--- a/cli/Cargo.lock
+++ b/cli/Cargo.lock
@@ -4069,8 +4069,7 @@ dependencies = [
 [[package]]
 name = "tar"
 version = "0.4.38"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "4b55807c0344e1e6c04d7c965f5289c39a8d94ae23ed5c0b57aabac549f871c6"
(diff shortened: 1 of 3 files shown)


pimeys (approved):
nice

.github/actions/cdn_assets/action.yml
```diff
@@ -22,7 +22,3 @@ runs:
       run: |
         curl https://${{ inputs.cdn }}/cli/${{ inputs.assets_version }}.tar.gz --output assets.tar.gz --fail
         cp assets.tar.gz crates/server
-        mkdir ~/.grafbase
```
obmarg, line 25: Actually this isn't going to trigger this particular issue...

PR #392, opened 2023-05-22
GB-3773: Support Authentication in SDK

pimeys:
# Description

Implements authentication providers and rules in all necessary parts of the Grafbase configuration. Implements a new object \`auth\` that can be used to create providers:

``` typescript
import { auth } from '@grafbase/sdk'
```

## How to Read the PR

There's one refactoring that splits stuff to modules, so it might be easier to go commit by commit.

## OpenID

[Read tests for more examples](https://github.com/grafbase/grafbase/blob/sdk/auth/packages/grafbase-sdk/tests/unit/auth.openid.test.ts)

``` typescript
// first create the provider
const clerk = auth.OpenIDConnect({
  issuer: '{{ env.ISSUER_URL }}'
})

// add it to the config with the rules
const cfg = config({
  schema: g,
  auth: {
    providers: [clerk],
    rules: (rules) => {
      rules.private()
    }
  }
})
```

## JWT

[Read tests for more examples](https://github.com/grafbase/grafbase/blob/sdk/auth/packages/grafbase-sdk/tests/unit/auth.jwt.test.ts)

``` typescript
const derp = auth.JWT({
  issuer: '{{ env.ISSUER_URL }}',
  secret: '{{ env.JWT_SECRET }}'
})
```

## JWKS

[Read tests for more examples](https://github.com/grafbase/grafbase/blob/sdk/auth/packages/grafbase-sdk/tests/unit/auth.jwks.test.ts)

``` typescript
const derp = auth.JWKS({
  issuer: '{{ env.ISSUER_URL }}'
})

// or
const derp = auth.JWKS({
  jwksEndpoint: '{{ env.JWKS_ENDPOINT }}'
})
```

## Rule Definitions

Everywhere where one can define authentication rules, it happens through a lambda with a rules builder.

``` typescript
(rules) => {
  rules.private().read()
  rules.owner().create()
  rules.group(["admin", "root"]).delete()
}
```

[See more examples in the class definition](https://github.com/grafbase/grafbase/blob/sdk/auth/packages/grafbase-sdk/src/auth.ts#LL30C6-L30C6)

### Global Rules

``` typescript
const clerk = auth.OpenIDConnect({
  issuer: '{{ env.ISSUER_URL }}'
})

const cfg = config({
  schema: g,
  auth: {
    providers: [clerk],
    rules: (rules) => {
      rules.private()
    }
  }
})
```

### Model-level Rules

``` typescript
g.model('User', {
  name: g.string()
}).auth((rules) => {
  rules.private().read()
})
```

### Field-level Rules

``` typescript
g.model('User', {
  name: g.string().auth((rules) => {
    rules.groups(['admin'])
  })
})
```

# Type of change

- [ ] 💔 Breaking
- [x] 🚀 Feature
- [ ] 🐛 Fix
- [ ] 🛠️ Tooling
- [x] 🔨 Refactoring
- [x] 🧪 Test
- [ ] 📦 Dependency


packages/grafbase-sdk/src/auth.ts
```diff
@@ -0,0 +1,157 @@
...
+
+  public update(): AuthRule {
+    return this.operation('update')
+  }
+
+  public delete(): AuthRule {
+    return this.operation('delete')
+  }
+
+  public all() {
```
notrab, line 63: Do we need this? If no operation rule is specified then it'll default to all in the backend afaik.

notrab: That's not to say I don't prefer being more explicit.

pimeys: No we don't.

fbjork (approved):
looks great

packages/grafbase-sdk/tests/unit/auth.openid.test.ts
```diff
@@ -0,0 +1,465 @@
...
+            { type: oidc, issuer: "{{ env.ISSUER_URL }}" }
+          ]
+          rules: [
+            { allow: groups, groups: ["backend", "admin"], operations: [delete] }
+          ]
+        )"
+    `)
+  })
+
+  it('renders multiple providers like a champ', () => {
```
tomasol, line 401: Although the schema is valid, we currently only support one provider, so this would cause parsing error. Would it be possible to have a constraint on the size of list == 1 until we implement multiple providers?

notrab (approved):
Looks great! @tomasol made a good point regarding the multiple auth config.

packages/grafbase-sdk/tests/unit/auth.openid.test.ts
pimeys: We can do that. I asked @fbjork about it before doing so, and he said it's possible to add multiple providers in the SDL so therefore we also do it in the typescript config. I get it that it breaks things under the surface, and it's easy enough to prevent this, so I'll do it.

pimeys: Fixed now to be an array of size one: https://github.com/grafbase/grafbase/pull/392/commits/ef0ab6b1d373b48702cd1fcc79a1b2c959879eb9

PR #395, opened 2023-05-23
GB-3811: Do not show a warning in the CLI when the TS config and schema aren't on the same level

pimeys:
# Description

If we have a project structure such as this one:

```
- grafbase
  - schema.graphql
- other project
  - grafbase
    - grafbase.config.ts
```

And we run the CLI in the `other project` directory, we got a warning of having both config files even though they were in different projects. We should check that the parent directories are the same and only warn if needed.

# Type of change

- [ ] 💔 Breaking
- [ ] 🚀 Feature
- [x] 🐛 Fix
- [ ] 🛠️ Tooling
- [ ] 🔨 Refactoring
- [ ] 🧪 Test
- [ ] 📦 Dependency

### DIFF
--- a/cli/crates/common/src/environment.rs
+++ b/cli/crates/common/src/environment.rs
@@ -144,7 +144,7 @@ impl Environment {
         let mut warnings = Vec::new();
 
         let project_grafbase_schema_path =
-            Self::get_project_grafbase_path(&mut warnings)?.ok_or(CommonError::FindGrafbaseDirectory)?;
+            get_project_grafbase_path(&mut warnings)?.ok_or(CommonError::FindGrafbaseDirectory)?;
 
         let project_grafbase_path = project_grafbase_schema_path
             .parent()
@@ -165,6 +165,7 @@ impl Environment {
         let resolvers_build_artifact_path = project_dot_grafbase_path.join(RESOLVERS_DIRECTORY_NAME);
         let database_directory_path = project_dot_grafbase_path.join(DATABASE_DIRECTORY);
         let wrangler_installation_path = user_dot_grafbase_path.join(WRANGLER_DIRECTORY_NAME);
+
         ENVIRONMENT
             .set(Self {
                 project_path,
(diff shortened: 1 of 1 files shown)


cli/crates/common/src/environment.rs
```diff
@@ -212,16 +212,18 @@ impl Environment {
...
-                    .with_hint("Delete one of them to avoid conflicts");
-
-                warnings.push(warning);
+                    warnings.push(warning);
+                }
+                _ => (),
             }
 
-            Ok(Some(path))
+            Ok(Some(GrafbaseSchemaPath::ts_config(path)))
```
obmarg, line 226: Does this not mean that if a user has

```
- grafbase
  - grafbase.config.ts
- other project
  - grafbase
    - schema.graphql
```

we'll pick the TS file even if they're clearly in other project?

pimeys: No. This utilizes the same mechanics to find the nearest `grafbase.config.ts` or `schema.graphql`. The bug happens when the closest one is the typescript config, and then somewhere else in the project you have a `schema.graphql`.

I'm rewriting this once more to make the directory scan only once.

obmarg: Really?  I may be missing something, but it looks to me like `get_path_to_file_in_project_grafbase` will go up the directory tree till it finds `grafbase/{file}`.  Which in the case above would find `grafbase/grafbase.config.ts` & `other project/grafbase/schema.graphql`.  And will then pick the TS file?

What am I missing?

pimeys: It will first find the `grafbase.config.ts` and if not found, do another search for `schema.graphql`. But... this changed a bit, check the new version: https://github.com/grafbase/grafbase/blob/fix/prevent-warning-if-tsconfig-and-schema-are-not-on-the-same-level/cli/crates/common/src/environment.rs#L211-L254

obmarg (approved):

yoav-lavi (approved):
Subscribe

PR #401, opened 2023-05-24
Anonymous analytics

yoav-lavi:
# Description

## Features 

- Adds support for privacy preserving anonymous analytics
  - Does not include or store any PII
  - Checks for `GRAFBASE_RUDDERSTACK_WRITE_KEY` and `GRAFBASE_RUDDERSTACK_DATAPLANE_URL` **at compile time** (CI secrets) to enable analytics
  - Checks for `DO_NOT_TRACK`<sup>[1](https://do-not-track.dev),[2](https://consoledonottrack.com)</sup> **at runtime** and disables analytics if found with any value
  - Added context: 
    ```json
     {
       "anonymousId": "..." // anonymous ULID, stored between sessions. not linked to anything.
       "sessionId": "...", // anonymous ULID, generated once per session. not linked to anything.
       "startTime": "...", // ISO 8601 timestamp
       "version": "..." // semantic version
     }
    ```
  - Current events:
    - "Command Executed" - sends the name and arguments (**argument names only, omits values!**) of the currently used command. We use this to understand which features are being used.
  - Opting out (may be supported directly later on):
    - Option 1: Write / overwrite `$HOME/.grafbase/analytics.json` as:
      ```json
      {"optOut":true,"anonymousId":null}
      ```
    - Option 2: Set the `DO_NOT_TRACK` environment variable with any value
  - Opt In (after having opted out) / Reset the anonymous Identifier (may be supported directly later on): delete `$HOME/.grafbase/analytics.json`

## Refactoring

- Splits `Environment` to `Environment` and `Project`, fixing an issue where `login` and `logout` could only be run in projects

# Type of change

- [ ] 💔 Breaking
- [x] 🚀 Feature
- [x] 🐛 Fix
- [ ] 🛠️ Tooling
- [x] 🔨 Refactoring
- [ ] 🧪 Test
- [ ] 📦 Dependency


cli/crates/cli/src/cli_input.rs
```diff
@@ -129,18 +131,78 @@ pub enum SubCommand {
     Unlink,
 }
 
+// TODO see if there's a way to do this automatically (https://github.com/clap-rs/clap/discussions/4921)
+pub trait ArgumentNames {
```
yoav-lavi, line 135: I don't love this solution since it requires manually updating it if we change / add argument names, but I opted out of a partial macro here for now to keep it simple, especially if we end up automating it somehow

cli/crates/common/src/analytics.rs
```diff
@@ -0,0 +1,174 @@
...
+// TODO move this to [`Environment`]
+#[derive(Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+#[allow(clippy::module_name_repetitions)]
+pub struct GrafbaseConfig {
+    pub enable_analytics: bool,
+    pub anonymous_id: Option<Ulid>,
+}
+
+impl ToString for GrafbaseConfig {
```
pimeys, line 38: You should implement `std::fmt::Display` instead, which then gives you automatically `ToString`.

cli/crates/common/src/analytics.rs
```diff
@@ -0,0 +1,174 @@
...
+        Some(_) => None,
+        None => Some(()),
+    }
+}
+
+impl Analytics {
+    pub fn init() -> Result<(), CommonError> {
+        let write_key = option_env!("GRAFBASE_RUDDERSTACK_WRITE_KEY");
+        let dataplane_url = option_env!("GRAFBASE_RUDDERSTACK_DATAPLANE_URL");
+        let do_not_track = std::env::var("DO_NOT_TRACK").ok();
```
pimeys, line 57: 👍🏾

cli/crates/common/src/analytics.rs
```diff
@@ -0,0 +1,174 @@
...
+
+    pub fn track(event_name: &str, properties: Option<Value>) {
+        let event_name = event_name.to_owned();
+        Self::get()
+            .as_ref()
+            .zip(Self::read_data().ok().flatten().and_then(|data| data.anonymous_id))
+            .map(|(analytics, anonymous_id)| {
+                // purposely ignoring errors
+                // TODO possibly change this to a long lived thread once we add more events
+                thread::spawn(move || {
```
pimeys, line 154: Should we store the join handle somewhere and block with it before exiting?

cli/crates/common/src/analytics.rs
yoav-lavi: I'm never sure of this since `Display` feels like a debug / visualization trait while `ToString` is more serialization

cli/crates/common/src/analytics.rs
yoav-lavi: We purposely don't since you won't want to make the CLI run longer due to an analytics request. If it succeeds it succeeds and if it doesn't it doesn't

cli/crates/common/src/analytics.rs
pimeys: Display is to display something. It allows you to use this in every formatting string, and gives you additionally `ToString`. It's always preferrable over `ToString`.

yoav-lavi: Wouldn't `Display` be more of a debugging trait? e.g. the difference between outputting minified JSON and running Prettier, human-centric output

pimeys (approved):

PR #417, opened 2023-05-31
cleanup with invert trait

yoav-lavi:
# Description

## Refactoring

- A small cleanup to `analytics` with an `Invert` trait for `Option`

# Type of change

- [ ] 💔 Breaking
- [ ] 🚀 Feature
- [ ] 🐛 Fix
- [ ] 🛠️ Tooling
- [x] 🔨 Refactoring
- [ ] 🧪 Test
- [ ] 📦 Dependency

### DIFF
--- a/cli/crates/common/src/analytics.rs
+++ b/cli/crates/common/src/analytics.rs
@@ -1,6 +1,6 @@
 #![allow(clippy::let_underscore_untyped)] // derivative
 
-use crate::{environment::Environment, errors::CommonError};
+use crate::{environment::Environment, errors::CommonError, traits::Invert};
 use chrono::{DateTime, Utc};
 use core::panic;
 use derivative::Derivative;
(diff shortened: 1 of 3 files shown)


cli/crates/common/src/analytics.rs
```diff
@@ -64,8 +57,8 @@ impl Analytics {
 
         let analytics_data_and_variables = write_key
```
Finistere, line 58: IMHO it would be more readable to just have:
```
write_key.zip(dataplan_url)
  .filter(|_| do_not_track.is_none() && data.map(|data| data.enable_analytics).unwrap_or_default())
```
Rather than having `invert()` and `zip()`. Especially that `zip(data.filter(|data| !data.enable_analytics).invert());` is really hard to read.

yoav-lavi:
Closing this temporarily

PR #454, opened 2023-06-13
feat(sdk): Add a convenient way to use env vars

pimeys:
# Description

Adds `g.env()` for easy env var support. In addition of the unit tests, also tested manually by generating a GitHub token, and then running a project with the following config:

```ts
import { g, config, connector } from '@grafbase/sdk'

const github = connector.GraphQL({
  url: 'https://api.github.com/graphql',
  headers: (headers) => {
    headers.static('Authorization', `Bearer ${g.env('GITHUB_TOKEN')}`),
    headers.introspection('Authorization', `Bearer ${g.env('GITHUB_TOKEN')}`)
  }
})

g.datasource(github, { namespace: 'GitHub' })

export default config({ schema: g })
```

```fish
❯ env GITHUB_TOKEN="<REDACTED>" npx grafbase dev
```

And it works as it should. If using bash, go with:

```bash
❯ GITHUB_TOKEN="<REDACTED>" npx grafbase dev
```

Also adds support for `.env` file, which should now work.

# Type of change

- [ ] 💔 Breaking
- [x] 🚀 Feature
- [ ] 🐛 Fix
- [ ] 🛠️ Tooling
- [ ] 🔨 Refactoring
- [x] 🧪 Test
- [ ] 📦 Dependency

### DIFF
--- a/packages/grafbase-sdk/src/grafbase-schema.ts
+++ b/packages/grafbase-sdk/src/grafbase-schema.ts
@@ -398,7 +398,7 @@ export class GrafbaseSchema {
   public env(variableName: string): string {
     const value = process.env[variableName]
 
-    if (!value) {
+    if (value === undefined || value === null) {
       throw `Environment variable ${variableName} is not set`
     }
 
--- a/packages/grafbase-sdk/src/grafbase-schema.ts
+++ b/packages/grafbase-sdk/src/grafbase-schema.ts
@@ -390,6 +390,22 @@ export class GrafbaseSchema {
   }
 
   /**
+   * Returns the environment variable with the given variableName.
+   * Throws, if the variable is not set.
+   *
+   * @param variableName - The name of the environment variable.
+   */
+  public env(variableName: string): string {
+    const value = process.env[variableName]
+
+    if (!value) {
+      throw `Environment variable ${variableName} is not set`
+    }
+
+    return value
+  }
+
+  /**
    * Empty the schema.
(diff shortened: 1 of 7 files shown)


packages/grafbase-sdk/src/grafbase-schema.ts
```diff
@@ -389,6 +389,22 @@ export class GrafbaseSchema {
...
+  /**
+   * Returns the environment variable with the given variableName.
+   * Throws, if the variable is not set.
+   *
+   * @param variableName - The name of the environment variable.
+   */
+  public env(variableName: string): string {
+    const value = process.env[variableName]
+
+    if (!value) {
```
yoav-lavi, line 401: This means "" is not set as well, is that intentional?

packages/grafbase-sdk/src/grafbase-schema.ts
```diff
@@ -389,6 +389,22 @@ export class GrafbaseSchema {
...
+  /**
+   * Returns the environment variable with the given variableName.
+   * Throws, if the variable is not set.
+   *
+   * @param variableName - The name of the environment variable.
+   */
+  public env(variableName: string): string {
+    const value = process.env[variableName]
+
+    if (value === undefined || value === null) {
```
yoav-lavi, line 401: You can use == null for this but this is also good

yoav-lavi (approved):

PR #464, opened 2023-06-15
docs(sdk): update readme env/namespace usage

notrab:
This PR updates the README for the Grafbase SDK to reference the use of `g.env()`, optional `namespace` and a missing field type `g.phoneNumber()`.

### DIFF
--- a/packages/grafbase-sdk/README.md
+++ b/packages/grafbase-sdk/README.md
@@ -361,7 +361,7 @@ const stripe = connector.OpenAPI({
 Connectors can be added to the schema using `g.datasource()`, including an optional `namespace`:
 
 ```typescript
-g.datasource(stripe)
+g.datasource(stripe, { namespace: 'Stripe' })
 g.datasource(openai, { namespace: 'OpenAI' })
 ```
 
@@ -386,7 +386,7 @@ const github = connector.GraphQL({
 Connectors can be added to the schema using `g.datasource()`, including an optional `namespace`:
 
 ```typescript
-g.datasource(contentful)
+g.datasource(contentful, { namespace: 'Contentful' })
 g.datasource(github, { namespace: 'GitHub' })
 ```
 
@@ -471,10 +471,12 @@ const derp = auth.JWKS({
 Everywhere where one can define authentication rules, it happens through a lambda with a rules builder.
 
 ```typescript
-;(rules) => {
-  rules.private().read()
-  rules.owner().create()
-  rules.groups(['admin', 'root']).delete()
+{
+  rules: (rules) => {
+    rules.private().read()
+    rules.owner().create()
+    rules.groups(['admin', 'root']).delete()
+  }
 }
 ```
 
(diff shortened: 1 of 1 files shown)


packages/grafbase-sdk/README.md
```diff
@@ -374,10 +383,10 @@ const github = connector.GraphQL({
 })
 ```
 
-Introspecting the connector namespace to the schema happens with the `introspect` method of the schema:
+Connectors can be added to the schema using `g.datasource()`, including an optional `namespace`:
 
 ```typescript
-g.datasource(contentful, { namespace: 'Contentful' })
+g.datasource(contentful)
```
fbjork, line 389: I think we should encourage namespace in all examples.

packages/grafbase-sdk/README.md
```diff
@@ -462,7 +471,7 @@ const derp = auth.JWKS({
 Everywhere where one can define authentication rules, it happens through a lambda with a rules builder.
 
 ```typescript
-(rules) => {
+;(rules) => {
```
fbjork, line 474: typo?

packages/grafbase-sdk/README.md
notrab: Me too. I can revert but the line above explains the change, maybe I can be clearer. One shows without and one with otherwise we don't mention the optional namespace anywhere else.

packages/grafbase-sdk/README.md
notrab: Not really, it's just formatted correctly with Prettier since the example shown is an IIFE.

packages/grafbase-sdk/README.md
fbjork: Got it. Perhaps we add a sentence about when optional namespaces are useful and the pitfalls of not using a namespace.

packages/grafbase-sdk/README.md
notrab: @fbjork I've updated it to be shown differently so it doesn't add the pesky `;`

packages/grafbase-sdk/README.md
notrab: I've added the `namspace` back in to both examples. The real documentation makes a note of the optional namespace and that is best practice to avoid collisions.

packages/grafbase-sdk/README.md
```diff
@@ -2,6 +2,14 @@
 
 A TypeScript library to generate a Grafbase configuration. It replaces the `schema.graphql` file with `grafbase.config.ts`, which should be placed into the project's `grafbase` directory.
 
+## Get Started
+
+```bash
+npx grafbase init --config-format typescript
```
pimeys, line 8: We should say this is for a new project, to add it to the existing one, `(npm|npx yarn|pnpm) install @grafbase/sdk --save-dev`.

pimeys (approved):
Cool, maybe tweak the installation instructions a bit and merge away.

PR #479, opened 2023-06-20
Fix grafbase init -c typescript on folders starting with a dot

tomasol:
# Description
Steps to reproduce:
```
$ mkdir .some
$ cd .some
$ grafbase --trace 2 init -c typescript
2023-06-20T08:04:17.108970Z TRACE grafbase: subcommand: init
Grafbase CLI 0.23.0

Error: could not open the project 'package.json':
Caused by: No such file or directory (os error 2)
```

The problem is that `npm init -y` inside of such folders fails.
This fix replaces the npm call with the package json content taken from a successful call.

# Type of change

- [ ] 💔 Breaking
- [ ] 🚀 Feature
- [x] 🐛 Fix
- [ ] 🛠️ Tooling
- [ ] 🔨 Refactoring
- [ ] 🧪 Test
- [ ] 📦 Dependency

### DIFF
--- a/cli/crates/common/src/environment.rs
+++ b/cli/crates/common/src/environment.rs
@@ -324,11 +324,19 @@ fn find_grafbase_configuration(path: &Path, warnings: &mut Vec<Warning>) -> Opti
 pub fn add_dev_dependency_to_package_json(project_dir: &Path, package: &str, version: &str) -> Result<(), CommonError> {
     let package_json_path = project_dir.join(PACKAGE_JSON_NAME);
 
-    let mut package_json = if !package_json_path.exists() {
+    let mut package_json = if package_json_path.exists() {
+        let file = fs::File::open(&package_json_path).map_err(CommonError::AccessPackageJson)?;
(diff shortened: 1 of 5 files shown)


cli/crates/common/src/environment.rs
```diff
@@ -325,18 +324,38 @@ fn find_grafbase_configuration(path: &Path, warnings: &mut Vec<Warning>) -> Opti
...
-    }
-
-    let file = fs::File::open(&package_json_path).map_err(CommonError::AccessPackageJson)?;
-
-    let Ok(Value::Object(mut package_json)) = serde_json::from_reader(&file) else {
+    let mut package_json = if !package_json_path.exists() {
+        let name = project_dir
+            .file_name()
+            .map(|it| it.to_string_lossy())
+            .unwrap_or(Cow::Borrowed("grafbase-project"));
```
tomasol, line 331: The name is not sanitized in any way. We could slugify it but it does not matter for our workflow. `npm install` succeeds even with the invalid name.

Finistere (approved):

PR #490, opened 2023-06-23
feat(auth): GB-3994: Add authorizer runtime

tomasol:
# Description

Add new UdfKind::Authorizer, running the custom auth provider. Use `requires_udf` flag to install wrangler. Add tests for authorizer.

# Type of change

- [ ] 💔 Breaking
- [x] 🚀 Feature
- [ ] 🐛 Fix
- [ ] 🛠️ Tooling
- [x] 🔨 Refactoring
- [x] 🧪 Test
- [ ] 📦 Dependency

### DIFF
--- a/cli/crates/server/src/errors.rs
+++ b/cli/crates/server/src/errors.rs
@@ -17,8 +17,8 @@ use crate::udf_builder::JavaScriptPackageManager;
 #[derive(Error, Debug)]
 pub enum JavascriptPackageManagerComamndError {
     /// returned if npm/pnpm/yarn cannot be found
-    #[error("could not find {0}: {1}")]
-    NotFound(JavaScriptPackageManager, which::Error),
+    #[error("could not find {0:?}: {1}")]
+    NotFound(Option<JavaScriptPackageManager>, String),
 
     /// returned if any of the npm/pnpm/yarn commands exits unsuccessfully
     #[error("{0} encountered an error: {1}")]
--- a/cli/crates/server/src/errors.rs
+++ b/cli/crates/server/src/errors.rs
@@ -18,9 +18,11 @@ use crate::udf_builder::JavaScriptPackageManager;
 pub enum JavascriptPackageManagerComamndError {
     #[error("working directory '{0}' not found")]
     WorkingDirectoryNotFound(PathBuf),
+    #[error("working directory '{0}' cannot be read.\nCaused by: {1}")]
+    WorkingDirectoryCannotBeRead(PathBuf, std::io::Error),
     /// returned if npm/pnpm/yarn cannot be found
     #[error("could not find {0}: {1}")]
-    NotFound(JavaScriptPackageManager, String),
+    NotFound(JavaScriptPackageManager, which::Error),
 
     /// returned if any of the npm/pnpm/yarn commands exits unsuccessfully
     #[error("{0} encountered an error: {1}")]
--- a/cli/crates/server/src/errors.rs
+++ b/cli/crates/server/src/errors.rs
@@ -16,9 +16,11 @@ use crate::udf_builder::JavaScriptPackageManager;
 
 #[derive(Error, Debug)]
 pub enum JavascriptPackageManagerComamndError {
+    #[error("working directory '{0}' not found")]
+    WorkingDirectoryNotFound(PathBuf),
     /// returned if npm/pnpm/yarn cannot be found
-    #[error("could not find {0:?}: {1}")]
-    NotFound(Option<JavaScriptPackageManager>, String),
+    #[error("could not find {0}: {1}")]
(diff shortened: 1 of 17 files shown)


cli/crates/server/src/errors.rs
```diff
@@ -17,8 +17,8 @@ use crate::udf_builder::JavaScriptPackageManager;
 #[derive(Error, Debug)]
 pub enum JavascriptPackageManagerComamndError {
     /// returned if npm/pnpm/yarn cannot be found
-    #[error("could not find {0}: {1}")]
-    NotFound(JavaScriptPackageManager, which::Error),
+    #[error("could not find {0:?}: {1}")]
+    NotFound(Option<JavaScriptPackageManager>, String),
```
obmarg, line 21: Does this error get shown to users?  `None` or `Some(Npm)` don't seem like very good things to appear in user facing error messages

cli/crates/server/src/udf_builder.rs
```diff
@@ -18,20 +18,30 @@ async fn run_command<P: AsRef<Path>>(
...
 ) -> Result<(), JavascriptPackageManagerComamndError> {
     let command_string = format!("{command_type} {}", arguments.iter().format(" "));
-
-    trace!("running '{command_string}'");
+    let current_directory = current_directory.as_ref();
+    if !current_directory.exists() {
+        return Err(JavascriptPackageManagerComamndError::NotFound(
+            None,
+            format!("Working directory not found: {current_directory:?}"),
+        ));
```
obmarg, line 26: Following on from my previous comment: should this be another error variant, rather than re-using `NotFound`?

`Working directory not found: {}` seems like a better error message than `Could not find None: Working directory not found: {}`

obmarg (approved):

cli/crates/server/src/udf_builder.rs
tomasol: agreed, fixing

cli/crates/server/src/udf_builder.rs
```diff
@@ -18,20 +18,29 @@ async fn run_command<P: AsRef<Path>>(
...
     let command_string = format!("{command_type} {}", arguments.iter().format(" "));
-
-    trace!("running '{command_string}'");
+    let current_directory = current_directory.as_ref();
+    if !current_directory.exists() {
+        return Err(JavascriptPackageManagerComamndError::WorkingDirectoryNotFound(
+            current_directory.to_owned(),
+        ));
+    }
+    debug!("running '{command_string}'");
```
yoav-lavi, line 27: We don't really use anything other than `trace!` at the moment, I'd keep it consistent and then possibly change some of the invocations if we want to add levels

cli/crates/server/src/udf_builder.rs
```diff
@@ -18,20 +18,29 @@ async fn run_command<P: AsRef<Path>>(
     environment: &[(&'static str, &str)],
 ) -> Result<(), JavascriptPackageManagerComamndError> {
     let command_string = format!("{command_type} {}", arguments.iter().format(" "));
-
-    trace!("running '{command_string}'");
+    let current_directory = current_directory.as_ref();
+    if !current_directory.exists() {
```
yoav-lavi, line 22: In most cases we do `try_exists()` to handle cases of missing permissions, I'd do that here as well

cli/crates/server/src/errors.rs
```diff
@@ -16,9 +16,11 @@ use crate::udf_builder::JavaScriptPackageManager;
 
 #[derive(Error, Debug)]
 pub enum JavascriptPackageManagerComamndError {
+    #[error("working directory '{0}' not found")]
+    WorkingDirectoryNotFound(PathBuf),
     /// returned if npm/pnpm/yarn cannot be found
     #[error("could not find {0}: {1}")]
-    NotFound(JavaScriptPackageManager, which::Error),
+    NotFound(JavaScriptPackageManager, String),
```
yoav-lavi, line 23: Why did we change this to a string? It's normally better to keep the original error to be able to display it or display hints by matching it

cli/crates/server/src/udf_builder.rs
tomasol: sure, changing it.

cli/crates/server/src/udf_builder.rs
tomasol: good call

cli/crates/server/src/errors.rs
tomasol: I've changed it back.
The original reason was that I was reusing the variant `NotFound` for case when working directory is not found. 
However if this was a public library I would advise against leaking the dependency for better backwards compatibility.

PR #493, opened 2023-06-23
feat(cli): Init installs dependencies, if using npm

pimeys:
# Description

If initializing a new project with a TypeScript config, and we find out the project uses `npm` and we can find `npm` executable, we run `npm install` to install `@grafbase/sdk` to the project. If the project uses some other package manager, or we can't find `npm` executable, we ask the user to install dependencies.

# Type of change

- [ ] 💔 Breaking
- [x] 🚀 Feature
- [ ] 🐛 Fix
- [ ] 🛠️ Tooling
- [ ] 🔨 Refactoring
- [ ] 🧪 Test
- [ ] 📦 Dependency

### DIFF
--- a/cli/crates/backend/src/errors.rs
+++ b/cli/crates/backend/src/errors.rs
@@ -108,6 +108,9 @@ pub enum BackendError {
     #[error("could not read the repository information for {0}")]
     ReadRepositoryInformation(String),
 
+    #[error("could not find 'npm':\nCaused by: {0}")]
+    NpmNotFound(Box<dyn std::error::Error + Send>),
+
     // wraps a [`CommonError`]
     #[error(transparent)]
     CommonError(#[from] CommonError),
--- a/cli/crates/backend/src/project.rs
+++ b/cli/crates/backend/src/project.rs
@@ -6,16 +6,17 @@ use common::consts::{
     GRAFBASE_DIRECTORY_NAME, GRAFBASE_ENV_FILE_NAME, GRAFBASE_SCHEMA_FILE_NAME, GRAFBASE_SDK_PACKAGE_NAME,
     GRAFBASE_SDK_PACKAGE_VERSION, GRAFBASE_TS_CONFIG_FILE_NAME,
 };
-use common::environment::{self, Project};
+use common::environment::{self, NodePackageManager, Project};
+use common::errors::CommonError;
 use http_cache_reqwest::{CACacheManager, Cache, CacheMode, HttpCache};
 use reqwest::{header, Client};
 use reqwest_middleware::ClientBuilder;
 use serde::Deserialize;
-use std::fs;
 use std::io::{Error as IoError, ErrorKind as IoErrorKind};
 use std::iter::Iterator;
 use std::path::{Path, PathBuf};
 use std::{env, fmt};
+use std::{fs, io};
 use tokio_stream::StreamExt;
 use tokio_util::compat::TokioAsyncReadCompatExt;
 use tokio_util::io::StreamReader;
@@ -135,12 +136,7 @@ pub async fn init(name: Option<&str>, template: Template<'_>) -> Result<(), Back
                         .unwrap_or_default()
                 })
             {
-                environment::add_dev_dependency_to_package_json(
-                    &project_path,
-                    GRAFBASE_SDK_PACKAGE_NAME,
-                    GRAFBASE_SDK_PACKAGE_VERSION,
-                )
-                .map_err(BackendError::CommonError)?;
+                add_dependencies(&project_path)?;
             }
         }
         Template::FromDefault(config_type) => {
@@ -153,17 +149,7 @@ pub async fn init(name: Option<&str>, template: Template<'_>) -> Result<(), Back
             let schema_write_result = match config_type {
                 ConfigType::TypeScript => {
                     let schema_path = grafbase_path.join(GRAFBASE_TS_CONFIG_FILE_NAME);
-
-                    let add_sdk = environment::add_dev_dependency_to_package_json(
-                        &project_path,
-                        GRAFBASE_SDK_PACKAGE_NAME,
-                        GRAFBASE_SDK_PACKAGE_VERSION,
-                    )
-                    .map_err(Into::into);
-
-                    let write_schema = fs::write(schema_path, DEFAULT_SCHEMA_TS).map_err(BackendError::WriteSchema);
-
-                    add_sdk.and(write_schema)
+                    fs::write(schema_path, DEFAULT_SCHEMA_TS).map_err(BackendError::WriteSchema)
                 }
                 ConfigType::GraphQL => {
                     let schema_path = grafbase_path.join(GRAFBASE_SCHEMA_FILE_NAME);
@@ -181,8 +167,37 @@ pub async fn init(name: Option<&str>, template: Template<'_>) -> Result<(), Back
 
             schema_write_result?;
             dot_env_write_result?;
+
+            // This has to be the last thing we do.
+            if let ConfigType::TypeScript = config_type {
+                add_dependencies(&project_path)?;
+            };
         }
     };
+
+    Ok(())
+}
+
+fn add_dependencies(project_path: &Path) -> Result<(), BackendError> {
+    environment::add_dev_dependency_to_package_json(
+        project_path,
+        GRAFBASE_SDK_PACKAGE_NAME,
+        GRAFBASE_SDK_PACKAGE_VERSION,
+    )
+    .map_err(BackendError::CommonError)?;
+
+    let package_manager = environment::guess_project_package_manager(project_path, &project_path.join("package.json"))
+        .unwrap_or_default();
+
+    if let NodePackageManager::Npm = package_manager {
+        environment::install_dependencies(project_path).map_err(|e| match e {
+            CommonError::NpmNotFound(e) => BackendError::NpmNotFound(e),
+            e => BackendError::CommonError(e),
+        })?;
+    } else {
+        let error = io::Error::new(io::ErrorKind::InvalidInput, "Project is not using npm.");
+        return Err(BackendError::NpmNotFound(Box::new(error)));
+    }
 
     Ok(())
 }
--- a/cli/crates/common/src/environment.rs
+++ b/cli/crates/common/src/environment.rs
@@ -398,7 +398,7 @@ pub fn install_dependencies(project_root: &Path) -> Result<(), CommonError> {
 }
 
 pub fn guess_project_package_manager(project_root: &Path, package_json: &Path) -> Option<NodePackageManager> {
-    const LOCK_FILES: &[(&str, NodePackageManager)] = &[
+    const LOCK_FILE_NAMES: &[(&str, NodePackageManager)] = &[
         (PNPM_LOCK_FILE, NodePackageManager::Pnpm),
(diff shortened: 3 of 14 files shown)


cli/crates/backend/src/project.rs
```diff
@@ -181,12 +167,38 @@ pub async fn init(name: Option<&str>, template: Template<'_>) -> Result<(), Back
...
+        project_path,
+        GRAFBASE_SDK_PACKAGE_NAME,
+        GRAFBASE_SDK_PACKAGE_VERSION,
+    )
+    .map_err(BackendError::CommonError)?;
+
+    let package_manager = environment::guess_project_package_manager(project_path, &project_path.join("package.json"))
+        .unwrap_or_default();
+
+    if let JavaScriptPackageManager::Npm = package_manager {
```
yoav-lavi, line 192: I'd perhaps go with `NodePackageManager`, these are specific to Node rather than to JS

yoav-lavi: As in `pnpm` wouldn't work with `deno`

cli/crates/backend/src/project.rs
```diff
@@ -181,12 +167,38 @@ pub async fn init(name: Option<&str>, template: Template<'_>) -> Result<(), Back
...
+        environment::install_dependencies(project_path).map_err(BackendError::CommonError)?;
+    } else {
+        let error = io::Error::new(io::ErrorKind::InvalidInput, "Project is not using npm.");
+        return Err(BackendError::CommonError(CommonError::NpmNotFound(Box::new(error))));
```
yoav-lavi, line 196: I'd duplicate this error rather than emitting a common one that isn't coming from common

cli/crates/common/src/environment.rs
```diff
@@ -330,6 +332,102 @@ fn find_grafbase_configuration(path: &Path, warnings: &mut Vec<Warning>) -> Opti
...
+                );
+
+                Err(CommonError::AccessPackageJson(inner))
+            }
+        }
+    }
+}
+
+pub fn install_dependencies(project_root: &Path) -> Result<(), CommonError> {
+    let command = match which::which("npm") {
```
yoav-lavi, line 386: I'd check the behavior here if you already had it installed but with a different version than latest, it may overwrite

cli/crates/common/src/environment.rs
```diff
@@ -330,6 +332,102 @@ fn find_grafbase_configuration(path: &Path, warnings: &mut Vec<Warning>) -> Opti
...
+        .current_dir(project_root)
+        .output()
+        .map_err(CommonError::NpmInstall)?;
+
+    Ok(())
+}
+
+pub fn guess_project_package_manager(project_root: &Path, package_json: &Path) -> Option<JavaScriptPackageManager> {
+    const LOCK_FILES: &[(&str, JavaScriptPackageManager)] = &[
+        ("pnpm-lock.yaml", JavaScriptPackageManager::Pnpm),
```
yoav-lavi, line 402: Could we use consts for these?

cli/crates/common/src/environment.rs
pimeys: We want to overwrite (discussed in Slack).

cli/crates/backend/src/errors.rs
```diff
@@ -108,6 +108,9 @@ pub enum BackendError {
     #[error("could not read the repository information for {0}")]
     ReadRepositoryInformation(String),
 
+    #[error("could not find 'npm':\nCaused by: {0}")]
+    NpmNotFound(Box<dyn std::error::Error + Send>),
```
yoav-lavi, line 112: Could we narrow the error here? Useful for hints

cli/crates/backend/src/project.rs
```diff
@@ -181,12 +167,41 @@ pub async fn init(name: Option<&str>, template: Template<'_>) -> Result<(), Back
...
 
+fn add_dependencies(project_path: &Path) -> Result<(), BackendError> {
+    environment::add_dev_dependency_to_package_json(
+        project_path,
+        GRAFBASE_SDK_PACKAGE_NAME,
+        GRAFBASE_SDK_PACKAGE_VERSION,
+    )
+    .map_err(BackendError::CommonError)?;
+
+    let package_manager = environment::guess_project_package_manager(project_path, &project_path.join("package.json"))
```
yoav-lavi, line 189: I'd add this as a const

cli/crates/backend/src/project.rs
```diff
@@ -181,12 +167,41 @@ pub async fn init(name: Option<&str>, template: Template<'_>) -> Result<(), Back
...
+        .unwrap_or_default();
+
+    if let NodePackageManager::Npm = package_manager {
+        environment::install_dependencies(project_path).map_err(|e| match e {
+            CommonError::NpmNotFound(e) => BackendError::NpmNotFound(e),
```
yoav-lavi, line 194: Could we expand this to `error`?

yoav-lavi: I'm a bit confused here, wasn't this specifically not a common error before? If it's always a common error we should just wrap it normally

cli/crates/server/src/errors.rs
```diff
@@ -116,7 +117,7 @@ pub enum ServerError {
 
     /// returned if any of the package manager commands ran during resolver build exits unsuccessfully
     #[error("command error: {0}")]
-    WranglerInstallPackageManagerCommandError(#[from] JavascriptPackageManagerComamndError),
+    WranglerInstallPackageManagerCommandError(#[from] JavascriptPackageManagerCommandError),
```
yoav-lavi, line 120: NodePackageManager

cli/crates/backend/src/project.rs
pimeys: Lazy value you mean?

yoav-lavi: Just the `"package.json"` part :)

pimeys: I wrapped the whole thing to a lazy cell, because we call this more than once and it makes sense to cache the result.

cli/crates/common/src/environment.rs
```diff
@@ -330,6 +334,104 @@ fn find_grafbase_configuration(path: &Path, warnings: &mut Vec<Warning>) -> Opti
...
+impl FromStr for NodePackageManager {
+    type Err = CommonError;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        match s {
+            "npm" => Ok(NodePackageManager::Npm),
+            "pnpm" => Ok(NodePackageManager::Pnpm),
+            "yarn" => Ok(NodePackageManager::Yarn),
+            _ => {
+                let inner = io::Error::new(
```
yoav-lavi, line 376: Why does this need an inner IO error rather than a different variant?

cli/crates/common/src/environment.rs
```diff
@@ -330,6 +334,104 @@ fn find_grafbase_configuration(path: &Path, warnings: &mut Vec<Warning>) -> Opti
...
+                Err(CommonError::AccessPackageJson(inner))
+            }
+        }
+    }
+}
+
+pub fn install_dependencies(project_root: &Path) -> Result<(), CommonError> {
+    let command = match which::which("npm") {
+        Ok(command) => command,
+        Err(e) => return Err(CommonError::NpmNotFound(Box::new(e))),
```
yoav-lavi, line 390: Why is this using a dyn error rather than a specific type?

yoav-lavi: That's what we currently do since e.g. hints can use the internal error or display it, we could go over them and box them at some point though to make it easier for third parties

yoav-lavi: But ok, let's leave it as is and discuss the rest later

cli/crates/common/src/environment.rs
```diff
@@ -330,6 +334,104 @@ fn find_grafbase_configuration(path: &Path, warnings: &mut Vec<Warning>) -> Opti
...
+    Command::new(command)
+        .args(["install", "--save-dev"])
+        .current_dir(project_root)
+        .output()
+        .map_err(CommonError::NpmInstall)?;
+
+    Ok(())
+}
+
+pub fn guess_project_package_manager(project_root: &Path, package_json: &Path) -> Option<NodePackageManager> {
```
jakubadamw, line 402: Why the change in the implementation?
For one, this function gets called from the UDF build infrastructure which is async, thus effectively inserting calls blocking on I/O rather than yielding. It's not a tragedy, but a change for the worse.
Any particular reason for the rewrite from into a loop?
Also, we shouldn't use statics for cached computation. The `Project` structure might be a good place to keep the package manager cached if we want it cached.

jakubadamw (changes requested):

cli/crates/common/src/environment.rs
```diff
@@ -330,6 +334,104 @@ fn find_grafbase_configuration(path: &Path, warnings: &mut Vec<Warning>) -> Opti
...
+    const LOCK_FILES: &[(&str, NodePackageManager)] = &[
+        (PNPM_LOCK_FILE, NodePackageManager::Pnpm),
```
jakubadamw, line 404: I saw you did this based on @yoav-lavi's input, but, honestly, I don't see what the improvement here is. The whole slice was a const already.

Also, I'd rename this to `LOCK_FILE_NAMES`.

jakubadamw:
> If the project uses some other package manager
What's the rationale for not doing the same for other package managers? We do that for resolver builds already.

cli/crates/common/src/environment.rs
yoav-lavi: I think this was in regards to a specific string rather than the whole computation, but we could put it in Project

jakubadamw: Except, well, `LOCK_FILE_NAMES` will be a better name

cli/crates/common/src/environment.rs
pimeys: In general, I'd like to avoid async if we don't need async. In CLI's case, there's a tokio executor already with as many threads as there are cores. Having a blocking file system access is much nicer developer experience than playing with the pretty ugly apis from Tokio, especially when their APIs are not really using async IO for file system access, just running it in a separate thread. We're not in a JavaScript event loop, we have quite a many of those loops in Tokio.

For the complexity of async IO, is the CLI code needing that much of throughput? Will loading the package.json be blocking everything in every thread so it will affect throughput? Will we need to handle that many requests in the CLI that this will be an issue? Especially when we load the package manager to a cell, which is kept in cache after the first call?

For `Itertools`, in Prisma we got slapped when we tried to use that. It's a bit of a mess of a crate, especially when some of the iterators actually allocate which is not cool. I always prefer a simple for loop, if there's no need for composition.

You can't cache it in the `Project` structure due to it not existing on init where we need this. I'd need to change the whole project initialization a lot so I could first run init, and after that create a project. Much less friction this way.

pimeys:
> > If the project uses some other package manager
> 
> What's the rationale for not doing the same for other package managers? We do that for resolver builds already.

I did that first, but @yoav-lavi reminded me of edge cases where the project can use multiple package managers. Or the user has `pnpm` in `node_modules` and installed system-wide. These edge cases were enough to decide on installing _only_ if we know none of the special package managers are in use. We can expand on this later, but for now this is the path of least resistance: helping users with new projects to get up to speed, but asking them to install deps manually if we can't decide on the package manager.

PR #502, opened 2023-06-30
Add type conversion between InquireError and CliError

mikaelmello:
# Description

I noticed InquireError was being manually mapped to CliError on every prompt call, a conversion 
that can be made implicit if the resulting type implements From<E> where E is the source type.

So, I thought of creating this PR to simplify the code a little bit in case there isn't a 
reason/context I'm missing.

# Type of change

- [x] 🔨 Refactoring

### DIFF
--- a/cli/crates/cli/src/create.rs
+++ b/cli/crates/cli/src/create.rs
@@ -1,4 +1,4 @@
-use crate::{errors::CliError, output::report, prompts::handle_inquire_error};
+use crate::{errors::CliError, output::report};
 use backend::api::{
     create,
     types::{Account, DatabaseRegion},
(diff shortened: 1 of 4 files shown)


yoav-lavi:
Hey! Apologies for the late response!

Thank you for your contribution! We'd love to see future contributions from you if you'd like to submit others.

The reason this is done manually is since `inquire` exposes CTRL+C as an error, and exiting via `?` was deemed as a less favorable option (as it's even more implicit) than `map_err`. We may refactor this at some point but at the moment we don't want to transition this into a `From` impl due to the reason above. 

Please feel free to comment with any reasons why you feel that a `From` impl is preferable here and we'll reconsider!

PR #522, opened 2023-07-09
CLI: Print out the new operation-related events logged through the bridge to the stdout

jakubadamw:
# Description

This introduces an initial logging of operation-related log events (operation start and completion) that get posted through the bridge.

This also adds a '--debug' argument to the 'dev' command that for now enables logging of introspection queries.

# Type of change

- [ ] 💔 Breaking
- [X] 🚀 Feature
- [ ] 🐛 Fix
- [ ] 🛠️ Tooling
- [ ] 🔨 Refactoring
- [ ] 🧪 Test
- [ ] 📦 Dependency

image

image

### DIFF
--- a/.github/workflows/cli.yml
+++ b/.github/workflows/cli.yml
@@ -17,7 +17,7 @@ permissions:
 
 env:
   GRAFBASE_SKIP_ASSET_VERSION_CHECK: 'true'
-  ASSETS_VERSION: release/e4b391a-2023-07-12
+  ASSETS_VERSION: release/19d76c2-2023-07-12
   PROD_ASSETS: assets.grafbase.com
   CARGO_TERM_COLOR: 'always'
   CARGO_PROFILE_DEV_DEBUG: 0
--- a/cli/crates/cli/src/output/report.rs
+++ b/cli/crates/cli/src/output/report.rs
@@ -157,9 +157,15 @@ pub fn operation_completed(
     r#type: common::types::OperationType,
     duration: std::time::Duration,
 ) {
+    let formatted_duration = if duration < std::time::Duration::from_secs(1) {
+        format!("{}ms", duration.as_millis())
+    } else {
+        format!("{:.1}s", duration.as_secs_f64())
+    };
+
     // FIXME: Add operation type.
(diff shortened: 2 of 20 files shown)


.github/workflows/cli.yml
```diff
@@ -149,7 +149,8 @@ jobs:
         with:
           lint: false
 
-      - name: Build release
+      - if: always()
```
obmarg, line 152: What's with the `always()`s?

cli/crates/cli/src/output/report.rs
```diff
@@ -149,6 +149,32 @@ pub fn udf_message(udf_kind: UdfKind, udf_name: &str, message: &str, level: UdfM
...
+        if !debug {
+            return;
+        }
+    }
+
+    let formatted_duration = if duration < std::time::Duration::from_secs(1) {
+        format!("{:.2}ms", duration.as_secs_f64() / 1000.0)
+    } else {
+        format!("{:.2}s", duration.as_secs_f64())
+    };
```
obmarg, line 171: Noticed that introspection in your screenshots said `0ms` - do any of our actual queries return that fast?  Might be nice to do a `ns` option as well if so?  

Always had a bit of a "wow, so fast" moment when I've seen that elsewhere.

obmarg (approved):

.github/workflows/cli.yml
jakubadamw: So that we get a binary uploaded even if the sanitisation fails. But I only needed it temporarily while the tests were failing, so I'll remove it now

cli/crates/cli/src/output/report.rs
jakubadamw: @obmarg good point, thanks! Yes, I added micro- and nanoseconds now.

obmarg: Sweet - I did of course mean `us` there, but nanos is a good goal if we can acheive it

jakubadamw: @obmarg I added both. Micro (μs) and nano (ns). :-)

PR #527, opened 2023-07-10
feat: pull in header forwarding, add a test.

obmarg:
# Description

Users would like to be able to forward headers from the incoming request to connectors.  This pulls in that functionality and adds a test.

# Type of change

- [ ] 💔 Breaking
- [x] 🚀 Feature
- [ ] 🐛 Fix
- [ ] 🛠️ Tooling
- [ ] 🔨 Refactoring
- [ ] 🧪 Test
- [ ] 📦 Dependency

### DIFF
--- a/cli/crates/cli/tests/graphql-directive/main.rs
+++ b/cli/crates/cli/tests/graphql-directive/main.rs
@@ -8,6 +8,7 @@ use crate::utils::environment::Environment;
 #[path = "../utils/mod.rs"]
 mod utils;
 
+mod headers;
 mod server;
 
 const NAMESPACED_QUERY: &str = "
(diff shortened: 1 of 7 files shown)


fbjork (approved):

cli/crates/cli/tests/openapi/headers.rs
```diff
@@ -0,0 +1,100 @@
...
+        .drain_requests()
+        .into_iter()
+        .map(|request| {
+            // Wow, this is annoying...
+            request
+                .headers
+                .into_iter()
+                .map(|(name, value)| (name.to_string(), value.to_string()))
+                .filter(|(name, _)| {
+                    name != "host" && name != "connection" && name != "accept-encoding" && name != "mf-loop"
```
Finistere, line 69: why not keep them like the grahql connector?

cli/crates/cli/tests/openapi/headers.rs
```diff
@@ -0,0 +1,100 @@
...
+    mount_petstore_spec(&mock_server).await;
+
+    let mut env = Environment::init_async().await;
+    let client = start_grafbase(&mut env, petstore_schema_with_header_forwarding(mock_server.address())).await;
+
+    let http_spy = HttpSpy::new();
+
+    Mock::given(method("GET"))
+        .and(path("/pet/123"))
+        .and(http_spy.clone())
```
Finistere, line 27: Wasn't this enough?
https://docs.rs/wiremock/latest/wiremock/struct.MockServer.html#method.received_requests

cli/crates/cli/tests/graphql-directive/main.rs
```diff
@@ -150,6 +151,43 @@ async fn graphql_test_without_namespace() {
...
 }
 
+#[tokio::test(flavor = "multi_thread")]
+async fn test_header_forwarding() {
+    server::run(54302).await;
+
+    let mut env = Environment::init_async().await;
+    let client = start_grafbase(&mut env, schema(54302, false)).await;
+
+    insta::assert_yaml_snapshot!(
```
Finistere, line 161: What are those tests doing? I see 4 times the same query with different `id`, not sure to see how that's related to the headers. 🤔

obmarg: Copy & paste fail, will delete

cli/crates/cli/tests/openapi/headers.rs
obmarg: So with this one wiremock is on a random port so you need to filter out `host` or the snapshot will never match.  And if I'm filtering out one I figured I might as well filter the rest.

I wanted to filter them in graphql as well - it was just a pain to do so I didn't.

Finistere: I'd just filter out `host` with your comment then IMHO. At least I would add your comment.^^

PR #529, opened 2023-07-11
feat: update tsconfig with support for header forwarding.

obmarg:
# Description

We're adding the ability to forward headers from a users request to openapi & graphql connectors.  This adds support for these to TS config.

As part of this we've deprecated the old `headers.static` method in favour of a new `headers.set` method that supports static headers and forwarded headers.

# Type of change

- [ ] 💔 Breaking
- [x] 🚀 Feature
- [ ] 🐛 Fix
- [ ] 🛠️ Tooling
- [ ] 🔨 Refactoring
- [ ] 🧪 Test
- [ ] 📦 Dependency


pimeys (approved):
nice!

PR #535, opened 2023-07-13
feat(examples): planetscale + resolvers api

notrab:
Guide contents. Needs refactoring. Needs TS config.

# Description

# Type of change

- [ ] 💔 Breaking
- [x] 🚀 Feature
- [ ] 🐛 Fix
- [ ] 🛠️ Tooling
- [ ] 🔨 Refactoring
- [ ] 🧪 Test
- [ ] 📦 Dependency


examples/resolvers-planetscale/grafbase/resolvers/products/update.ts
```diff
@@ -0,0 +1,94 @@
...
+    params.push(id);
+  } else if (slug !== undefined) {
+    updateStatement += " WHERE slug = ?";
+    params.push(slug);
+  } else {
+    // Throw new GraphQLError('ID or Slug must be provided')
+  }
+
+  let selectStatement;
+  let selectParams;
```
Finistere, line 70: Just in case, I'm not familiar with PlanetScale, but with `UPDATE` you can usually add a `RETURNING *` which returns all updated data.

examples/resolvers-planetscale/grafbase/resolvers/products/all.ts
```diff
@@ -0,0 +1,81 @@
...
+        return Boolean(value);
+      }
+      default: {
+        return cast(field, value);
+      }
+    }
+  },
+};
+
+export default async function ProductsAll(_, args) {
```
Finistere, line 27: For a tutorial, as you do queries by hand, I wouldn't bother with relay-style cursor IMHO, just offset + limit

examples/resolvers-planetscale/grafbase/resolvers/products/update.ts
notrab: I didn't think `RETURNING *` was supported by MySQL. Is that a new thing?

examples/resolvers-planetscale/grafbase/resolvers/products/all.ts
notrab: Happy to redo this section of the tutorial if you think it's worth it. I initially wanted to go a bit further than every other tutorial which avoids the difficult task of teaching cursor based.

examples/resolvers-planetscale/grafbase/resolvers/products/update.ts
Finistere: might only be in PostgreSQL 🤔

notrab: Nice find, I originally started with using procedures for all operations in fact but thought it might be a tad over the top for an example 😆

examples/resolvers-planetscale/grafbase/resolvers/products/all.ts
notrab: Maybe we could add 2 sub headings for adding pagination and document cursor/offset versions.

examples/resolvers-planetscale/grafbase/resolvers/products/all.ts
```diff
@@ -0,0 +1,81 @@
...
+  cast(field, value) {
+    switch (field.name) {
+      case "id": {
+        return String(value);
+      }
+      case "onSale": {
+        return Boolean(value);
+      }
+      default: {
+        return cast(field, value);
```
obmarg, line 21: Am I missing something - looks like this is just going to recurse infinitely?

obmarg: I am missing something - `cast` is an import.

examples/resolvers-planetscale/grafbase/resolvers/products/update.ts
```diff
@@ -0,0 +1,94 @@
...
+  if (newOnSale !== undefined) {
+    updateStatement += "onSale = ?, ";
+    params.push(newOnSale);
+  }
+
+  if (params.length === 0) {
+    throw new Error("At least one field to update must be provided.");
+  }
+
+  updateStatement = updateStatement.slice(0, -2);
```
obmarg, line 57: Is it worth putting all the `name = ?` bits into an array and doing `.join(", ")` on them?

Spent 30 seconds wondering "what's this slice doing, does it need a comment" - but could just avoid it altogether?  Although not sure if you think it's clearer this way?

notrab: @obmarg good suggestion. You mean something like this?

```ts
const yo = `UPDATE Products SET ${updateParts.join(", ")} WHERE id = ? OR slug = ?`;
```

obmarg: yeah more or less

examples/resolvers-planetscale/grafbase/resolvers/products/all.ts
```diff
@@ -0,0 +1,81 @@
...
+
+export default async function ProductsAll(_, args) {
+  const { first, last, before, after } = args;
+
+  // Throw GraphQLError when arg is missing when supported
+
+  try {
+    if (first !== undefined && after !== undefined) {
+      const results = await conn.execute(
+        "SELECT * FROM Products WHERE id > ? ORDER BY id ASC LIMIT ?",
```
fbjork, line 35: I would use lower case table names which I think more common.

examples/resolvers-planetscale/grafbase/resolvers/products/create.ts
```diff
@@ -0,0 +1,48 @@
...
+  },
+};
+
+export default async function ProductsCreate(_, { input }) {
+  const { name, slug, price, onSale } = input;
+
+  // Check for unique constraint and error when supported
+
+  try {
+    const { insertId } = await conn.execute(
```
fbjork, line 30: thoughts on using Kelsey or Drizzle since it's very popular these days? That way we SEO rank for that too.

examples/resolvers-planetscale/readme.md
```diff
@@ -0,0 +1,102 @@
+```shell
+CREATE TABLE Products (
+    id INT NOT NULL AUTO_INCREMENT,
+    name VARCHAR(255) NOT NULL,
+    slug VARCHAR(255) NOT NULL UNIQUE,
+    price INT NOT NULL DEFAULT 0,
+    onSale BOOLEAN DEFAULT FALSE,
```
fbjork, line 7:
```suggestion
    on_sale BOOLEAN DEFAULT FALSE,
```

examples/resolvers-planetscale/readme.md
```diff
@@ -0,0 +1,102 @@
+```shell
+CREATE TABLE Products (
```
fbjork, line 2:
```suggestion
CREATE TABLE products (
```

examples/resolvers-planetscale/grafbase/resolvers/products/create.ts
notrab: I've got a mention of this in the guide and will create a separate guide for doing so, that way we have more content to produce :)

examples/resolvers-planetscale/readme.md
notrab: I named it this because of a limitation Grafbase side so it would remove the need to transform and map this field.

examples/resolvers-planetscale/grafbase/resolvers/products/all.ts
notrab: Fine with me, I'll make the changes.

examples/resolvers-planetscale/readme.md
fbjork: What's the limitation?

notrab:
```
[RuleError { locations: [Pos(2:3)], message: "Field \"on_sale\" is not in Camel lowercase, please use \"onSale\" instead." }]
```

notrab: For the sake of the tutorial I wanted to keep a 1:1 mapping, if you think it's worth the explainer I'm happy to adjust though.

fbjork: All good, just trying to understand the limitation. You could perhaps do `SELECT onSale as 'on_sale'`?

PR #537, opened 2023-07-13
feat(sdk): MongoDB connector

pimeys:
# Description

The SDK part of a MongoDB connector definition. Let's merge it quietly now, and add it to the docs and all that when the connector is closer to done.

# Type of change

- [ ] 💔 Breaking
- [x] 🚀 Feature
- [ ] 🐛 Fix
- [ ] 🛠️ Tooling
- [ ] 🔨 Refactoring
- [ ] 🧪 Test
- [ ] 📦 Dependency

### DIFF
--- a/packages/grafbase-sdk/src/typedefs/auth.ts
+++ b/packages/grafbase-sdk/src/typedefs/auth.ts
@@ -63,7 +63,7 @@ export class AuthDefinition {
   /**
    * Sets the name of the field in the database, if different than the name of the field.
    */
-  public map(name: string): MapDefinition {
+  public mapped(name: string): MapDefinition {
     return new MapDefinition(this, name)
   }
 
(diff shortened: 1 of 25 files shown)


packages/grafbase-sdk/src/typedefs/auth.ts
```diff
@@ -59,6 +60,13 @@ export class AuthDefinition {
     return new CacheDefinition(this, new FieldLevelCache(params))
   }
 
+  /**
+   * Sets the name of the field in the database, if different than the name of the field.
+   */
+  public map(name: string): MapDefinition {
```
obmarg, line 66: A random thought: `map` might be an odd name for this in a TS context - quite different from the usual `.map` function.

obmarg (approved):

PR #546, opened 2023-07-18
feat(cli): GB-4358: introduce a more thorough set of arguments to the dev command

jakubadamw:
# Description

These will control the level of logs for GraphQL operations and functions (resolvers, authorizers) etc.

```
Run your Grafbase project locally
Usage: grafbase dev [OPTIONS]
Options:
  -p, --port <PORT>                                                  Use a specific port [default: 4000]
  -s, --search                                                       If a given port is unavailable, search for another
      --disable-watch                                                Do not listen for schema changes and reload
      --log-level-functions <LOG_LEVEL_FUNCTIONS>                    Log level to print from function invocations, defaults to or overrides ‘log-level’ [possible values: none, error, warn, info, debug]
      --log-level-graphql-operations <LOG_LEVEL_GRAPHQL_OPERATIONS>  Log level to print for GraphQL operations, defaults to or overrides ‘log-level’ [possible values: none, error, warn, info, debug]
      --log-level <LOG_LEVEL>                                        Default log level to print [default: info] [possible values: none, error, warn, info, debug]
  -h, --help                                                         Print help
```

# Type of change

- [ ] 💔 Breaking
- [X] 🚀 Feature
- [ ] 🐛 Fix
- [ ] 🛠️ Tooling
- [ ] 🔨 Refactoring
- [ ] 🧪 Test
- [ ] 📦 Dependency

### DIFF
--- a/cli/crates/cli/src/cli_input.rs
+++ b/cli/crates/cli/src/cli_input.rs
@@ -1,10 +1,19 @@
 use crate::create::CreateArguments;
 use clap::{arg, command, CommandFactory, Parser, ValueEnum};
 use clap_complete::{shells, Generator};
-use common::consts::{DEFAULT_LOG_FILTER, TRACE_LOG_FILTER};
+use common::{
+    consts::{DEFAULT_LOG_FILTER, TRACE_LOG_FILTER},
+    types::LogLevel,
+};
 use std::{fmt, path::PathBuf};
 
 const DEFAULT_PORT: u16 = 4000;
+
+#[derive(Clone, Copy)]
+pub struct LogLevelFilters {
+    pub functions: LogLevel,
+    pub graphql_operations: LogLevel,
+}
 
 #[derive(Debug, Parser)]
 pub struct DevCommand {
@@ -17,9 +26,33 @@ pub struct DevCommand {
     /// Do not listen for schema changes and reload
     #[arg(long)]
     pub disable_watch: bool,
-    /// Log more events taking place such as introspection queries
-    #[arg(short, long)]
-    pub debug: bool,
+    /// Level of logs coming from function (resolver, authorizer) invocations to print, overrides 'log-level'
+    #[arg(long)]
+    pub log_level_functions: Option<LogLevel>,
+    /// Level of logs related to incoming GraphQL operations to print, overrides 'log-level'
+    #[arg(long)]
+    pub log_level_graphql_operations: Option<LogLevel>,
+    /// Default level of logs to print
+    #[arg(long)]
+    pub log_level: Option<LogLevel>,
+}
+
+impl DevCommand {
+    pub fn log_levels(&self) -> LogLevelFilters {
+        const DEFAULT_LOG_LEVEL_FUNCTIONS: LogLevel = LogLevel::Info;
+        const DEFAULT_LOG_LEVEL_GRAPHQL_OPERATIONS: LogLevel = LogLevel::Info;
+
+        LogLevelFilters {
+            functions: self
+                .log_level_functions
+                .or(self.log_level)
+                .unwrap_or(DEFAULT_LOG_LEVEL_FUNCTIONS),
+            graphql_operations: self
+                .log_level_functions
+                .or(self.log_level)
+                .unwrap_or(DEFAULT_LOG_LEVEL_GRAPHQL_OPERATIONS),
+        }
+    }
 }
 
 #[derive(Debug, Parser, Clone, Copy)]
@@ -61,11 +94,10 @@ pub struct CompletionsCommand {
 }
 
 #[derive(Debug, Clone, Copy, ValueEnum)]
+#[clap(rename_all = "lowercase")]
 pub enum ConfigFormat {
     /// Adds a TypeScript configuration file
-    #[value(name = "typescript")]
     TypeScript,
-    #[value(name = "graphql")]
     /// Adds a GraphQL configuration file
     GraphQL,
 }
--- a/cli/crates/cli/src/cli_input.rs
+++ b/cli/crates/cli/src/cli_input.rs
@@ -9,10 +9,32 @@ use std::{fmt, path::PathBuf};
 
 const DEFAULT_PORT: u16 = 4000;
 
+#[derive(Clone, Copy, Debug, PartialEq, PartialOrd, serde::Deserialize, clap::ValueEnum)]
+#[clap(rename_all = "snake_case")]
+pub enum LogLevelFilter {
+    None,
+    Error,
+    Warn,
+    Info,
+    Debug,
+}
+
+impl From<LogLevelFilter> for Option<LogLevel> {
+    fn from(value: LogLevelFilter) -> Self {
+        match value {
+            LogLevelFilter::None => None,
+            LogLevelFilter::Error => Some(LogLevel::Error),
+            LogLevelFilter::Warn => Some(LogLevel::Warn),
+            LogLevelFilter::Info => Some(LogLevel::Info),
+            LogLevelFilter::Debug => Some(LogLevel::Debug),
+        }
+    }
+}
+
 #[derive(Clone, Copy)]
 pub struct LogLevelFilters {
-    pub functions: LogLevel,
-    pub graphql_operations: LogLevel,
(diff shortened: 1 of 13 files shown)


cli/crates/cli/src/cli_input.rs
```diff
@@ -17,9 +48,24 @@ pub struct DevCommand {
     /// Do not listen for schema changes and reload
     #[arg(long)]
     pub disable_watch: bool,
-    /// Log more events taking place such as introspection queries
-    #[arg(short, long)]
-    pub debug: bool,
+    /// Log level to print from function invocations, defaults to 'log-level'
+    #[arg(long)]
```
yoav-lavi, line 52: I'd maybe name this since the parameter would be "FUNCTION_LOG_LEVEL"

yoav-lavi: The others as well

cli/crates/cli/src/output/report.rs
```diff
@@ -140,16 +143,68 @@ pub fn complete_udf_build(udf_kind: UdfKind, udf_name: &str, duration: std::time
...
+        LogEventType::OperationStarted { .. } => return,
+        LogEventType::OperationCompleted { name, duration, r#type } => {
+            let colour = match r#type {
+                common::types::OperationType::Query { is_introspection } => {
+                    if is_introspection && log_level_filter < LogLevel::Debug {
+                        return;
+                    }
+                    watercolor::colored::Color::Green
+                }
+                // Pink.
```
yoav-lavi, line 187: WDYM pink?

cli/crates/cli/src/cli_input.rs
jakubadamw: Sorry, not following. What parameter would be 'FUNCTION_LOG_LEVEL'?

cli/crates/cli/src/output/report.rs
jakubadamw: It used to be pink, now I changed it to green, but didn't update the comment… The danger of comments.

cli/crates/cli/src/dev.rs
```diff
@@ -16,7 +17,13 @@ static READY: Once = Once::new();
 ///
 /// returns [`CliError::ServerPanic`] if the development server panics
 #[allow(clippy::fn_params_excessive_bools)]
-pub fn dev(search: bool, watch: bool, external_port: u16, debug: bool, tracing: bool) -> Result<(), CliError> {
+pub fn dev(
```
pimeys, line 20: Not part of this PR, but I'd combine all the boolean flags into named bitflags.

cli/crates/cli/src/cli_input.rs
pimeys: @yoav-lavi can you elaborate?

cli/crates/cli/src/output/report.rs
```diff
@@ -140,16 +143,65 @@ pub fn complete_udf_build(udf_kind: UdfKind, udf_name: &str, duration: std::time
...
+        LogLevel::Debug => watercolor::colored::Color::BrightBlack,
+        LogLevel::Error => watercolor::colored::Color::Red,
+        LogLevel::Info => watercolor::colored::Color::Cyan,
+        LogLevel::Warn => watercolor::colored::Color::Yellow,
+    };
+    println!("{}", format!("[{udf_kind} '{udf_name}'] {message}").color(colour));
 }
 
-pub fn operation_started(_request_id: &str, _name: &Option<String>) {}
+pub fn operation_log(_request_id: &str, log_event_type: LogEventType, log_level_filter: Option<LogLevel>) {
```
pimeys, line 169: Do we need to have the request id parameter here if we don't use it?

pimeys (commented):
Just a few nitpicks. In general I wonder could this be handled with tracing and writing a custom subscriber? I know it sucks a bit writing subscribers for tracing, and also it might not be feasible with our wasm setup. I think the dev command starts to have a bit too many parameters already.

@yoav-lavi should give his final blessing, he still has an open question.

cli/crates/server/src/types.rs
```diff
@@ -23,18 +42,12 @@ pub enum ServerMessage {
...
-        request_id: String,
-        name: Option<String>,
-    },
-    OperationCompleted {
+    OperationLogMessage {
         request_id: String,
-        name: Option<String>,
-        duration: std::time::Duration,
-        r#type: common::types::OperationType,
+        event_type: LogEventType,
```
Finistere, line 50: Maybe wrap `LogEvent` here? or is there a difference?

cli/crates/cli/src/output/report.rs
```diff
@@ -140,16 +143,65 @@ pub fn complete_udf_build(udf_kind: UdfKind, udf_name: &str, duration: std::time
...
+    let Some(log_level_filter) = log_level_filter else {
+        return;
+    };
+    if log_level_filter < LogLevel::Info {
```
Finistere, line 173: if log level is debug you don't log anything? 🤔

cli/crates/cli/src/cli_input.rs
```diff
@@ -1,11 +1,42 @@
...
+pub enum LogLevelFilter {
+    None,
+    Error,
+    Warn,
+    Info,
+    Debug,
+}
+
+impl From<LogLevelFilter> for Option<LogLevel> {
+    fn from(value: LogLevelFilter) -> Self {
```
Finistere, line 23: Not sure how that would work out, but I wonder if this could just be done with something like `impl PartialOrd<LogLevel> for LogLevelFilter` or something like that

Finistere (approved):

cli/crates/cli/src/output/report.rs
jakubadamw: Right now we don't, but in the future we most likely will

jakubadamw: I can remove it, though. Sure!

cli/crates/cli/src/output/report.rs
jakubadamw:
```
#[derive(Clone, Copy, Debug, PartialEq, PartialOrd, serde::Deserialize)]
pub enum LogLevel {
    Error,
    Warn,
    Info,
    Debug,
}
```

🙂

cli/crates/cli/src/cli_input.rs
yoav-lavi: Specifically referring to ` --log-level-functions <LOG_LEVEL_FUNCTIONS>`, `LOG_LEVEL_FUNCTIONS` as a parameter name seems a bit weirdly put

jakubadamw:
> Just a few nitpicks. In general I wonder could this be handled with tracing and writing a custom subscriber? I know it sucks a bit writing subscribers for tracing, and also it might not be feasible with our wasm setup. I think the dev command starts to have a bit too many parameters already.
> 
> @yoav-lavi should give his final blessing, he still has an open question.

Well, but it's a UX/UI design choice to have those parameters rather than a single 'RUST_LOG'-like parameter/environment variable. As for using tracing for this – it may be plausible, yes. I can look into refactoring this when I am back.

cli/crates/cli/src/cli_input.rs
jakubadamw: I considered that but to me operations like `PartialOrd` etc. should be defined between types that are equivalent. Semantically, these aren't, actually. `LogLevel` is the level of a single message, whereas filter belongs to another category (includes `None` that a level can't be). Doesn't feel like `PartialOrd` would be appropriate…

Finistere: Well don't have a strong opinion, I think it probably would have made code simpler rather than dealing with the optional special case. 🤷

cli/crates/server/src/types.rs
jakubadamw: One is the type for the bridge protocol. I guess there isn't a difference, but we have separate types for the other kinds of things happening (UDF invocation) to stress the separation of these protocols. I can take a look at merging these, but we should do it for all in that case.

Finistere: fair enough!

PR #560, opened 2023-07-25
Host Pathfinder in the CLI

yoav-lavi:
# Description

# Type of change

- [ ] 💔 Breaking
- [ ] 🚀 Feature
- [ ] 🐛 Fix
- [ ] 🛠️ Tooling
- [ ] 🔨 Refactoring
- [ ] 🧪 Test
- [ ] 📦 Dependency


cli/crates/backend/src/api/login.rs
```diff
@@ -86,8 +81,13 @@ pub async fn login(message_sender: MspcSender<LoginMessage>) -> Result<(), ApiEr
         Err(error) => return Err(ApiError::ReadUserDotGrafbaseFolder(error)),
     }
 
-    let port = find_available_port_in_range(EPHEMERAL_PORT_RANGE, LocalAddressType::Localhost)
-        .ok_or(ApiError::FindAvailablePort)?;
+    let listener = TcpListener::bind((Ipv4Addr::LOCALHOST, 0))
+        .await
+        .map_err(|_| ApiError::FindAvailablePort)?;
```
Finistere, line 86: OPT: could also probably have been done with something like `.ok().and_then(...).ok_or_else(|| ApiError::FindAvailablePort)?` chain I think

cli/crates/server/src/errors.rs
```diff
@@ -142,8 +142,8 @@ pub enum ServerError {
     #[error("could not read the project database directory\nCaused by: {0}")]
     ReadDatabaseDir(IoError),
 
-    /// returned if an available port cannot be found for the bridge server
-    #[error("could not find an available port for the bridge server")]
+    /// returned if an available port cannot be found for the bridge server or playground
+    #[error("could not find an available port for an internal server")]
```
Finistere, line 146: OPT: IMHO "could not find an available port" would have been enough.

Finistere (commented):
The code itself looks good to me, but it's unclear to me why dynaql proxies to pathfinder hosted by the cli. 🤔 
Intuitively I would have a the cli host a webserver and forward the calls to dynaql if relevant. Why isn't it the case?

cli/crates/server/src/errors.rs
yoav-lavi: That could be confused with the main port for the worker

cli/crates/backend/src/api/login.rs
yoav-lavi: Will take a look into this for a future version

yoav-lavi:
>The code itself looks good to me, but it's unclear to me why dynaql proxies to pathfinder hosted by the cli. 🤔
Intuitively I would have a the cli host a webserver and forward the calls to dynaql if relevant. Why isn't it the case?
The idea was to have them on the same port, at the moment proxying just the playground rather than all calls is the smaller change but we can consider the opposite later on

Finistere (approved):

cli/crates/server/src/errors.rs
Finistere: is there a difference though for the user? Or do we need that for ourselves for investigation?

yoav-lavi: It signals that it's not related to the port search you passed, e.g. if you chose a really large port

PR #561, opened 2023-07-25
fix: console.log not working

obmarg:
We had a case mismatch on log levels between our JS and Rust.

Also added a test to make sure we don't regress.

Fixes GB-4406


Finistere (approved):

Finistere (commented):
Good catch!

fbjork (approved):

PR #572, opened 2023-08-03
fix: update search protocol

Finistere:
# Description

# Type of change

- [ ] 💔 Breaking
- [ ] 🚀 Feature
- [ ] 🐛 Fix
- [ ] 🛠️ Tooling
- [ ] 🔨 Refactoring
- [ ] 🧪 Test
- [ ] 📦 Dependency

### DIFF
--- a/cli/crates/server/src/bridge/search.rs
+++ b/cli/crates/server/src/bridge/search.rs
@@ -8,6 +8,7 @@ use sqlx::SqlitePool;
 
 use tantivy::schema::Field;
 use tantivy::Document;
+use ulid::Ulid;
 
 use std::fs::File;
 use std::io::BufReader;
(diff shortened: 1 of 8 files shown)


cli/crates/server/src/bridge/search.rs
```diff
@@ -56,20 +57,24 @@ impl<'a> Index<'a> {
                 searcher.search_backward_before(last as usize, &before.try_into()?)?
             }
         };
-        Ok(hits.map_id(|id| String::from_utf8(id).unwrap()))
+        Ok(hits.map_id(|id| {
+            let mut id = String::from_utf8(id).unwrap();
+            let _ = id.drain(..(id.len() - 26));
```
obmarg, line 62: A comment here might be good?  Looks quite weird, cannot immediately tell why we're doing this

obmarg (approved):

cli/crates/server/src/bridge/search.rs
Finistere: just removing the prefix: "post#<ulid>". Didn't want to make too many assumptions, but expecting an ulid at the end is one already. Adding a comment

PR #588, opened 2023-08-07
feat(cache): add cache access scopes to the sdk

hfgbarrigas:
# Description

Adds cache access scopes support in the sdk.

# Type of change

- [ ] 💔 Breaking
- [x] 🚀 Feature
- [ ] 🐛 Fix
- [ ] 🛠️ Tooling
- [ ] 🔨 Refactoring
- [ ] 🧪 Test
- [ ] 📦 Dependency

### DIFF
--- a/packages/grafbase-sdk/src/cache.ts
+++ b/packages/grafbase-sdk/src/cache.ts
@@ -28,6 +28,19 @@ export type MutationInvalidation =
   | { field: string }
 
 /**
+ * Defines the access scope for the cache.
+ * - 'apikey' will use the request's api_key details as part of the cache key.
+ * - `public` will allow any authenticated request access to the cache key.
+ * - `{ claim: string }` will use the `claim` value from the request's jwt as part of the cache key.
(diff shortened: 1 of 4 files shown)


packages/grafbase-sdk/src/cache.ts
```diff
@@ -27,6 +27,19 @@ export type MutationInvalidation =
   | 'list'
   | { field: string }
 
+/**
+ * Defines the access scope for the cache.
```
fbjork, line 31: Why not just call it Cache Scopes?
`Defines the cache scope.`

packages/grafbase-sdk/tests/unit/types.test.ts
```diff
@@ -103,6 +103,20 @@ describe('Type generator', () => {
     `)
   })
 
+  it('generates one with cache using access scopes', () => {
```
pimeys, line 106: misses a test for the `public` scope?

pimeys (approved):
maybe one extra test, then ready to go.

PR #596, opened 2023-08-11
feat(report): GB-4418: Add support for the new log event types (`fetch()` calls from connectors and resolvers

jakubadamw:
# Description

This adds logging for `fetch()` requests coming from connectors and custom resolvers.

image
image

Adds support for the `--verbose` flag while we are it.

# Type of change

- [ ] 💔 Breaking
- [X] 🚀 Feature
- [ ] 🐛 Fix
- [ ] 🛠️ Tooling
- [ ] 🔨 Refactoring
- [ ] 🧪 Test
- [ ] 📦 Dependency


cli/crates/cli/src/cli_input.rs
```diff
@@ -55,15 +55,23 @@ pub struct DevCommand {
     #[arg(long, value_name = "GRAPHQL_OPERATION_LOG_LEVEL")]
     pub log_level_graphql_operations: Option<LogLevelFilter>,
     /// Default log level to print
-    #[arg(long, default_value = "info")]
-    pub log_level: LogLevelFilter,
+    #[arg(long)]
+    pub log_level: Option<LogLevelFilter>,
+    /// A shortcut to enable fairly detailed logging
+    #[arg(short, long)]
```
yoav-lavi, line 61: Why isn't verbose a specific log level? Seems a bit redundant

jakubadamw: Because it's not a log level. It's a shortcut for a certain combination of log levels for the various scopes (right now `debug` for all). You can use the scope-specific log levels for granularity or the verbose flag to set them all. Right now it defaults to debug to all, but we may change it to another combination. Hence it doesn't map to a single log level.

fbjork (approved):

PR #611, opened 2023-08-16
chore(ci): tweak/format workflows/actions

JeanMertz:
This PR does some basic cleaning/formatting of our workflows and actions. There should be no breaking changes, but please do review carefully.

One non-stylistic change I did make was swapping `dtolnay/rust-toolchain@master` out for `actions-rust-lang/setup-rust-toolchain@v1` which is built upon the former, but also gives optimized caching settings for Rust projects, and adds "problem matchers" to have GitHub add notifications on the relevant lines when clippy/rustfmt fails.

closes GB-4527


.github/actions/init_rust_job/action.yml
```diff
@@ -33,24 +33,24 @@ runs:
...
 
     - name: rustup
       shell: bash
-      run: |
-        rustup target list
+      run: rustup target list
 
     - name: Install cargo-nextest & libs linux
-      if: ${{ startsWith(inputs.platform, 'linux') }}
+      if: ${{ inputs.platform == 'linux' }}
```
mettke, line 53: Does not work for linux-arm

.github/workflows/cli.yml
```diff
@@ -8,13 +8,9 @@ on:
       - 'cli-*'
 
 permissions:
```
mettke, line 10: I think its helpful to have a comment saying why these permissions are given

JeanMertz: Well, they aren't added in other workflows either, and also, it appears the comments are not updated with the actual settings? e.g. it says "states check reads", but it says `checks: write`.

Also, "allow repo checkout", I don't believe that needs write permission, so that too seems to be outdated? Other than that, the last two are just stating what the code already states, e.g. "Allow PRs write" vs `pull_requests: write`.

.github/actions/init_rust_job/action.yml
JeanMertz: The `platform` input field is a "choice" input, and it can only be one of `windows`, `macos` and `linux`.

JeanMertz: Turns out, we are both correct. The choices are limited to those three, but that doesn't apparently mean you can't also use other values for the input. I'll update the code, and add `linux-arm` as another valid choice 👍

JeanMertz: Resolved in [63e3f7c](https://github.com/grafbase/grafbase/pull/611/commits/63e3f7c48186a96422b2a61f7f7ba00ed8d653b6).

.github/actions/cdn_assets/action.yml
```diff
@@ -2,10 +2,6 @@ name: Download CLI assets
 description: Download CLI assets from Grafbase CDN
 
 inputs:
-  working-directory:
-    description: Where to run
-    required: false
-    default: cli
```
yoav-lavi, line 8: Why was this moved to each individual step?

.github/workflows/cli.yml
yoav-lavi: I agree, we should update these but not remove them if they're out of date

.github/workflows/cli.yml
```diff
@@ -33,6 +29,11 @@ env:
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
```
yoav-lavi, line 32: Why was this added?

.github/actions/cdn_assets/action.yml
JeanMertz: It was removed as an input value, since we're not actually passing this input from anywhere.

.github/workflows/cli.yml
JeanMertz: I can't update them because I don't know why these specific permissions are required. In my opinion, keeping outdated comments is worse than removing them (they convey information that is false).

If you have any idea what the comments should say, I'd be happy to add it back.

.github/workflows/cli.yml
JeanMertz: Good question. I suspect I did so to avoid older builds taking up resources as new ones are started. It looks like the other workflows in this repository also have this enabled, which might be another contributing factor to me thinking it would be good to do the same here.

Any reason we don't want to do this?

yoav-lavi:
Closing this for now as it's not relevant for the first milestone

PR #616, opened 2023-08-18
GB-3698, GB-4557 - Proxy the worker from the CLI rather than the other way around

yoav-lavi:
# Description

## Features

- Proxies the worker with a server running in the CLI
  - The error server now only runs on `/graphql` queries as Pathfinder will still run in the root
  - A future PR will explore exposing IPv6 via this interface
  - A followup PR will remove the proxy from the worker
  -  In the future we may let miniflare select the port to avoid (theoretical) edge cases with that specific port 

## Fixes

- Does not restart Pathfinder when reloading changes to the schema

## Tooling

- Ignores .jj folders
- adds a .ignore hardlink to .gitignore
- Update the Nix flake

# Type of change

- [ ] 💔 Breaking
- [x] 🚀 Feature
- [x] 🐛 Fix
- [x] 🛠️ Tooling
- [ ] 🔨 Refactoring
- [ ] 🧪 Test
- [ ] 📦 Dependency


cli/crates/backend/src/server_api.rs
```diff
@@ -15,12 +12,6 @@ type ServerInfo = (thread::JoinHandle<Result<(), ServerError>>, Receiver<ServerM
...
 /// returns [`BackendError::PortInUse`] if search is off and the supplied port is in use
-pub fn start_server(start_port: u16, search: bool, watch: bool, tracing: bool) -> Result<ServerInfo, BackendError> {
-    let port = find_available_port(search, start_port, LocalAddressType::Localhost).ok_or(if search {
-        BackendError::AvailablePort
-    } else {
-        BackendError::PortInUse(start_port)
-    })?;
-
-    Ok(server::start(port, watch, tracing))
+pub fn start_server(start_port: u16, search: bool, watch: bool, tracing: bool) -> ServerInfo {
```
yoav-lavi, line 15:
```suggestion
#[must_use]
pub fn start_server(start_port: u16, search: bool, watch: bool, tracing: bool) -> ServerInfo {
```

JeanMertz (approved):

PR #629, opened 2023-08-23
chore: open-source internal crates

JeanMertz:
This PR open-sources a set of crates previously maintained internally by us. This is the first step towards creating a broader ecosystem around Grafbase and our tooling.

More to come. Stay tuned...


yoav-lavi:
As discussed in Slack, I think it'd be a better idea to have the crates in a project directory rather than the root

Finistere:
Haven't seen any discussion on it. A good chunk if not all is irrelevant to the CLI and only designed with EFS & dynamodb in mind. The relevant parts are already in this repo as it creates indices on the fly.

obmarg:
Is there a way we could do this that doesn't get rid of all of the git history for the associated files?  Maybe [this](https://stackoverflow.com/a/17864475) to produce a clean repo with just the `oss` history followed by a subtree merge into `grafbase`?

yoav-lavi:
@obmarg we discussed this, apparently it doesn't work since we moved the files to an oss dir so there's no history before that if you use filtering by dir. I still think we should look into it more though

obmarg:
> @obmarg we discussed this, apparently it doesn't work since we moved the files to an oss dir so there's no history before that if you use filtering by dir. I still think we should look into it more though

[This answer certainly suggests we could be able to do _some_ kind of filtering to just a set of files](https://stackoverflow.com/a/32946929)

PR #651, opened 2023-09-05
feat(resolvers): optionally create symlink to grafbase-wasm-sdk

hfgbarrigas:
# Description

This PR adds a symlink to grafbase-wasm-sdk and updates the wrangler.toml used on resolvers to use a rule that makes this wasm available when uploading the worker.

The symlink is required because wrangler build rules use relative paths.

# Type of change

- [ ] 💔 Breaking
- [x] 🚀 Feature
- [ ] 🐛 Fix
- [ ] 🛠️ Tooling
- [ ] 🔨 Refactoring
- [ ] 🧪 Test
- [ ] 📦 Dependency

### DIFF
--- a/cli/crates/server/src/errors.rs
+++ b/cli/crates/server/src/errors.rs
@@ -216,6 +216,10 @@ pub enum UdfBuildError {
     #[error("could not write to a temporary file '{0}': {1}")]
     CreateNotWriteToTemporaryFile(PathBuf, IoError),
 
+    /// returned if creating symlink to a file fails.
+    #[error("could not link to file: {0}")]
+    SymlinkFailure(IoError),
+
     #[error("could not find a {0} referenced in the schema under the path {1}.{{js,ts}}")]
     UdfDoesNotExist(UdfKind, PathBuf),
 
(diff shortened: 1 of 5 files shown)


cli/crates/server/src/errors.rs
```diff
@@ -216,6 +216,10 @@ pub enum UdfBuildError {
     #[error("could not write to a temporary file '{0}': {1}")]
     CreateNotWriteToTemporaryFile(PathBuf, IoError),
 
+    /// returned if creating symlink to a file fails.
+    #[error("could not link to file: {0}")]
+    SymlinkToFile(IoError),
```
Finistere, line 221:
```suggestion
    SymlinkFailure(IoError),
```
or `SymlinkCouldNotBeCreated`. The `SymlinkToFile` name doesn't say what failed, hence the need of the comment

cli/crates/server/src/udf_builder.rs
```diff
@@ -160,6 +161,11 @@ pub async fn build(
     let new_package_json_contents = serde_json::to_string_pretty(&package_json).expect("must be valid JSON");
     trace!("new package.json contents:\n{new_package_json_contents}");
 
+    // symlink to grafbase-wasm-sdk
```
Finistere, line 164: would have been worth saying why instead, the function is explicit enough for what it does IMHO. Basically what you say in the PR. But since git blame can be used, good enough.^^

PR #665, opened 2023-09-06
fix: error propagation & nested list resolution

obmarg:
While testing `@defer` I've discovered a few bugs: 

1. We were propagating errors up to the top level of a response rather than the nearest nullable field.
2. We weren't reporting the path correctly on errors.
3. Nested lists would get flattened into a single level of list.
4. Nullability validation at the various levels of the list (and of items in lists) was also pretty sketchy.

This PR fixes those:

1. We were being a bit overzealous with the question mark operator.  I've replaced one or two points with proper error->null conversion to fix this.
2. Was because we called `set_error_path` all over the place which overwrote the path with a less specific path.  I've updated `set_error_path` to only set the path if there's not one already, which fixes that.
3. Our list resolver code just didn't take nested lists into account properly. I've rewritten it with that in mind.
4. Fixed as part of 3 - specifically had to track the nullability of various parts of the list to do this.

I'm not certain this is perfect but it's a lot better than it was - added a bunch of tests to verify various cases I could think of.

Fixes GB-4761
Fixes GB-4769

### DIFF
--- a/engine/crates/grafbase-engine/src/resolver_utils/field.rs
+++ b/engine/crates/grafbase-engine/src/resolver_utils/field.rs
@@ -50,13 +50,24 @@ pub async fn resolve_field(
         return Ok(None);
     };
 
-    match CurrentResolverType::new(&field, ctx) {
+    let result = match CurrentResolverType::new(&field, ctx) {
         CurrentResolverType::PRIMITIVE => resolve_primitive_field(ctx, parent_type, field, parent_resolver_value).await,
         CurrentResolverType::CONTAINER => resolve_container_field(ctx, field, parent_resolver_value).await,
         CurrentResolverType::ARRAY => resolve_array_field(ctx, field, parent_resolver_value).await,
     }
-    .map(Some)
-    .map_err(|error| ctx.set_error_path(error))
+    .map_err(|error| ctx.set_error_path(error));
+
+    match result {
+        Ok(result) => Ok(Some(result)),
+        Err(e) if !field.ty.is_non_null() => {
+            ctx.add_error(e);
+            Ok(Some(ctx.response_graph.write().await.insert_node(CompactValue::Null)))
+        }
+        Err(error) => {
+            // Propagate the error to parents who can add it to the response and null things out
+            Err(error)
+        }
+    }
 }
 
 async fn resolve_primitive_field(
@@ -85,18 +96,17 @@ async fn resolve_primitive_field(
                     .unwrap(),
                 );
                 Err(ServerError::new(
-                    format!("An error happened while fetching {:?}", ctx.item.node.name),
+                    format!(
+                        "An error happened while fetching `{}`, expected a non null value but found a null",
+                        field.name
+                    ),
                     Some(ctx.item.pos),
                 ))
             } else {
                 Ok(result.take())
             }
         }
-        Err(err) if field.ty.is_non_null() => Err(err),
-        Err(err) => {
-            ctx.add_error(err);
-            Ok(serde_json::Value::Null)
-        }
+        Err(err) => return Err(err),
     }?;
 
     let field_type = ctx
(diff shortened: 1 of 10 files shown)


engine/crates/grafbase-engine/src/resolver_utils/field.rs
```diff
@@ -50,13 +50,24 @@ pub async fn resolve_field(
...
         CurrentResolverType::CONTAINER => resolve_container_field(ctx, field, parent_resolver_value).await,
         CurrentResolverType::ARRAY => resolve_array_field(ctx, field, parent_resolver_value).await,
     }
-    .map(Some)
-    .map_err(|error| ctx.set_error_path(error))
+    .map_err(|error| ctx.set_error_path(error));
+
+    match result {
+        Ok(result) => Ok(Some(result)),
+        Err(e) if !field.ty.is_non_null() => {
```
Finistere, line 62: opt: `!field.ty.is_non_null()` is a bit hard to read, `field.ty.is_nullable()` would be simpler.

engine/crates/grafbase-engine/src/resolver_utils/list.rs
```diff
@@ -1,129 +1,201 @@
...
+            (ListKind::NonNullList(_), serde_json::Value::Null) => {
+                return Err(ctx.set_error_path(ServerError::new(
+                    format!(
+                        "An error occurred while fetching `{}`, a non-nullable value was expected but no value was found.",
+                        field.node.name.node
+                    ),
+                    Some(ctx.item.pos),
+                )));
+            }
+            (_, serde_json::Value::Array(_)) => value.item_iter().expect("we checked its an array").collect::<Vec<_>>(),
```
Finistere, line 54: might be worth a comment that you're using `item_ter` to avoid a copy of the data.

engine/crates/grafbase-engine/src/resolver_utils/list.rs
```diff
@@ -1,129 +1,201 @@
...
-                            .insert_node(ResponsePrimitive::new(result.into())))
+                    if found_null {
+                        ctx.add_error(ServerError::new(
+                            format!(
+                                "An error occurred while fetching `{}`, a non-nullable value was expected but nm value was found.",
+                                field.node.name.node
+                            ),
+                            Some(ctx.item.pos),
+                        ));
+                        children.push(ctx.response_graph.write().await.insert_node(CompactValue::Null));
```
Finistere, line 93: shouldn't you rather replace the node instead of adding a new one? The goal is to have a compacter representation with `CompactValue::Null` no?

engine/crates/grafbase-engine/src/resolver_utils/list.rs
```diff
@@ -226,3 +298,122 @@ pub async fn resolve_list_native<'a, T: LegacyOutputType + 'a>(
...
+
+/// The nullability of a list _and_ its contents
+#[derive(Debug, PartialEq, Clone, Copy)]
+enum ListKind {
+    NullableList(ListInner),
+    NonNullList(ListInner),
+}
+
+impl ListKind {
+    pub fn inner_nullablity(self) -> ListInner {
```
Finistere, line 319: Not sure, but seems like having a boolean struct field `has_nullable_item` with a method `has_nullable_item()` would make it simpler than having to deal with an enum.

Finistere (approved):

engine/crates/grafbase-engine/src/resolver_utils/list.rs
Finistere: Not at all, might just be worth a comment. A bit surprising to see you adding a null when you already have a null value. So to be honest without taking a deeper look, I can't explain why.^^

obmarg: Actually never mind I'll just do it

obmarg: Also ignore anything I said above: I didn't look at the context of this line, thought we were talking about another bit of code

PR #666, opened 2023-09-06
feat: error handling for @defer

obmarg:
I introduced basic `@defer` support to grafbase-engine in #2696.  One of the things I skipped was error handling.  

This was trickier than you might expect because error handling requires the path to be set correctly in the `Context`, but `Context.path_node` made heavy use of references in a way that made it really hard to defer.  I've ended up re-working it to a `Vector` of `ArcIntern<String>` instead of a reverse linked list.  It's much easier to make use of.  I'm using `im::Vector` instead of just a `Vec` to theoretically make cloning slightly cheaper, though I've been careful to hide that detail so we can swap it out if it proves to be a waste of time.

### DIFF
--- a/engine/crates/grafbase-engine/src/query_path.rs
+++ b/engine/crates/grafbase-engine/src/query_path.rs
@@ -23,7 +23,7 @@ impl QueryPath {
         self.0.push_back(segment.into());
     }
 
-    pub fn last_segment(&self) -> Option<&QueryPathSegment> {
+    pub fn last(&self) -> Option<&QueryPathSegment> {
         self.0.last()
     }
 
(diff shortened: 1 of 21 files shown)


engine/crates/grafbase-engine/src/query_path.rs
```diff
@@ -0,0 +1,83 @@
...
+impl QueryPath {
+    pub fn empty() -> Self {
+        QueryPath::default()
+    }
+
+    pub fn push(&mut self, segment: impl Into<QueryPathSegment>) {
+        self.0.push_back(segment.into());
+    }
+
+    pub fn last_segment(&self) -> Option<&QueryPathSegment> {
```
Finistere, line 26: I would just use `last` and `iter`. Specifying `segment` is unnecessary IMHO, and it's not consistent with `push` either.

Finistere (approved):

PR #667, opened 2023-09-06
feat: set hasNext correctly in `@defer`

obmarg:
Incremental payloads as used by `@defer` have a `hasNext` property that tells you whether to expect any more payloads after the current one.  I skipped over setting this correctly when I first implemented defer - just setting it to true all the time.  This PR fixes that.

### DIFF
--- a/engine/crates/grafbase-engine/src/schema.rs
+++ b/engine/crates/grafbase-engine/src/schema.rs
@@ -653,8 +653,12 @@ impl Schema {
 
                     // For now we're taking the simple approach and running all the deferred
                     // workloads serially. We can look into doing something smarter later.
-                    while let Some(workload) = receiver.receive() {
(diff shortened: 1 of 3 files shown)


engine/crates/grafbase-engine/src/schema.rs
```diff
@@ -723,7 +727,7 @@ async fn process_deferred_workload(
         label: workload.label,
         data,
         path: workload.path,
-        has_next: true, // TODO: Need to handle has_next properly...
+        has_next: true, // We hardcode this to true here, the function calling us should override
```
Finistere, line 730: IMHO should be `false` by default then, feels like supposing there is another workload when it's not even aware of it. but super tiny nitpick obviously.^^

Finistere (approved):

PR #670, opened 2023-09-06
feat(resolvers): add local kv

hfgbarrigas:
# Description

This PR configures KV for local usage

# Type of change

- [ ] 💔 Breaking
- [x] 🚀 Feature
- [ ] 🐛 Fix
- [ ] 🛠️ Tooling
- [ ] 🔨 Refactoring
- [ ] 🧪 Test
- [ ] 📦 Dependency

### DIFF
--- a/cli/crates/server/src/errors.rs
+++ b/cli/crates/server/src/errors.rs
@@ -220,6 +220,10 @@ pub enum UdfBuildError {
     #[error("could not link to file: {0}")]
     SymlinkFailure(IoError),
 
+    /// returned if kv data path is invalid. E.g: has non unicode characters
+    #[error("invalid KV data path: {0}")]
+    InvalidKvDataPath(String),
+
     #[error("could not find a {0} referenced in the schema under the path {1}.{{js,ts}}")]
     UdfDoesNotExist(UdfKind, PathBuf),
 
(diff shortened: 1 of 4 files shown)


cli/crates/server/src/udf_builder.rs
```diff
@@ -242,19 +242,38 @@ pub async fn build(
         .await
         .map_err(|err| UdfBuildError::CreateUdfArtifactFile(entrypoint_js_path.clone(), udf_kind, err))?;
 
+    let grafbase_kv_data_path = environment
+        .user_dot_grafbase_path
+        .join(crate::consts::GRAFBASE_KV_DATA_PATH)
+        .to_string_lossy()
```
yoav-lavi, line 248: We may want a proper error here if the path has non unicode chars

yoav-lavi (changes requested):
See the one comment, otherwise looks good

cli/crates/server/src/errors.rs
```diff
@@ -220,6 +220,10 @@ pub enum UdfBuildError {
     #[error("could not link to file: {0}")]
     SymlinkFailure(IoError),
 
+    /// returned if kv data path is invalid. E.g: has non unicode characters
+    #[error("invalid KV data path: {0}")]
```
yoav-lavi, line 224: We could add a hint for this that we don't support non-unicode paths, but not urgent if you don't want to add it here

yoav-lavi (approved):

PR #671, opened 2023-09-06
feat(connector): Neon findOne + relations + all that

pimeys:
# Implementation of PostgreSQL/Neon findOne

This (largeish) PR implements all facilities to find a unique row from a PostgreSQL database. This includes joins for nested selection, either unique which returns an object or null, and non-unique giving a collection of rows.

## Field mapping

The trick we do for PostgreSQL is we first introspect the database schema into an object called `DatabaseDefinition`. This object is a cyclic tree, allowing one to iterate over the tables, table columns, enums, relations and so forth. The object lets you to traverse from table to column, and back to table from the column.

A relation is a special construct that maps a foreign key of the database. For a given SQL schema:

```sql
CREATE TABLE "User" (
  id SERIAL PRIMARY KEY
);

CREATE TABLE "Profile" (
  user_id INT UNIQUE,
  CONSTRAINT "Profile_User" FOREIGN KEY (user_id) REFERENCES "User" (id)
)
```

we define a `ForwardRelationWalker` to the "Profile", where referenced table is "User" and referencing table is "Profile". And, of course, referenced columns are `[id]` and referencing columns `[user_id]`. For this to work from the "User" table, we create a `BackRelationWalker`, which has referenced table as "Profile" and referencing table as "User" (and the columns accordingly).

Now, we can define a combined `RelationWalker` we can iterate over from any table. The `relation` will give the correct tables and columns to construct a join.

Every type and field we generate in GraphQL, we map those to the corresponding table, column or relation. These mappings are found in the `DatabaseDefinition`.

When entering the `PostgresResolver`, we first create a `PostgresContext` that serves as an interface to iterate over selection, filters and so on. This context is aware of the database definition, and can provide types from the database definition for easier query building.

## Querying

We utilize HTTP transport, which prevents us from using SQL transactions. This means every query we perform must be atomic, and must be doable in one request. For this to work, we utilize PostgreSQL JSON aggregation functionality, and render the whole result set as JSON already in the database.

The query utilizes lateral joins. If these are not familiar to you, a lateral join allows using data from other selects in the same query, so we can find a "User" first and use the `id` to find the "Profile" in the lateral select.

I wrote an [annotated example](https://gist.github.com/pimeys/a7535acb0922fa432562539f5d8123c3) of how the query looks like when rendered.

For generating the SQL, we have a new crate [grafbase-sql-ast](https://github.com/grafbase/sql-ast/). This crate implements an AST structure for building complex SQL queries programmatically, and a renderer which outputs a parameterized query string together with the parameters as `serde_json::Value`. See the [documentation](https://docs.rs/grafbase-sql-ast/) for more info.

## Schemas

We introspect all the schemas for the user. A schema is a PostgreSQL namespace, meaning schema `public` can have table `User` and schema `private` can also have a table with the same name. We do not respect these namespaces in GraphQL, so if we have a name collision, the first user is called `PublicUser` and the second `PrivateUser`. All queries, joins et.al. should _just work_ between the schemas.

## New types

- `Time` for time values (`time` and `timetz` in PostgreSQL)
- `UnsignedBigInt` for unsigned large integer values (`oid` in PostgreSQL)
- `NaiveDateTime` for datetime values with no timezone (`timestamp` in PostgreSQL)
- `Uuid` for universally unique identifiers (`uuid` in PostgreSQL)

## Known issues

- Defining `before` or `after` argument to a nested collection (including all of the cursor/pageinfo stuff) select will panic. You can [follow the issue](https://linear.app/grafbase/issue/GB-4677/query-support-pagination) to see when this is working.

- A special case, where the user has a compound unique key (secondary) on the table, and one of the columns is nullable, defining a query `user(by: { nameEmail: { name: "foo" } })` which skips the email leads to wrong results. For now, the best way to query this is `user(by: { nameEmail: { name: "foo", email: null } })`, but we need to implicitly set the missing column as null in the query. [There's an issue for that](https://linear.app/grafbase/issue/GB-4909/implicit-nulls-on-unique-findone).

## Testing

Over half of the PR is tests, which are in the `integration-tests` directory.

## How to read the PR

Check the generated types and queries first. Understand how the `DatabaseDefinition`, `Names`, `Relations` and `Walker` structures work. See the resolver, the `PostgresContext` and related iterators what they do, and read the query generation. I tried to comment things where needed, but please ask if unsure what something does.

Closes: GB-4675
Closes: GB-4685


engine/crates/integration-tests/tests/execution/mod.rs
```diff
@@ -65,197 +63,6 @@ fn aliases() {
     });
 }
 
-#[test]
```
pimeys, line 68: merge thingy, fix tomorrow

engine/crates/integration-tests/tests/postgresql/introspection.rs
```diff
@@ -0,0 +1,388 @@
...
+        type User {
+          name: String!
+          email: String!
+        }
+
+        input UserByInput {
+          nameEmail: UserNameEmailInput
+        }
+
+        input UserNameEmailInput {
```
pimeys, line 183: These all fields need to be `!`.

engine/crates/parser-postgresql/src/registry/types/input/oneof.rs
```diff
@@ -0,0 +1 @@
+
```
pimeys, line 1: remove

engine/crates/parser-sdl/src/connector_parsers.rs
```diff
@@ -82,13 +82,10 @@ pub(crate) fn merge_registry(ctx: &mut VisitorContext<'_>, mut src_registry: Reg
 
     let mut main_registry = ctx.registry.borrow_mut();
 
-    // I am sort of making the assumption that connectors won't generate names that
```
pimeys, line 85: fixed a while ago. they are checked if they're unique

obmarg: where is this done?

pimeys: Would be nicer to have this a bit more implicit. Now we have to add the iteration to that function for every new connector...

obmarg: That code doesn't do what this comment was referring to.  That code makes sure the names of connectors are unique - this comment is referring to the names of types generated by the connectors - I think this comment still stands.

engine/crates/engine/src/registry/scalars/decimal.rs
```diff
@@ -43,7 +43,8 @@ impl DynamicParse for DecimalScalar {
 
     fn to_value(value: serde_json::Value) -> Result<crate::Value, crate::Error> {
         match value {
-            serde_json::Value::String(phone) => Ok(ConstValue::String(phone)),
+            serde_json::Value::String(number) => Ok(ConstValue::String(number)),
+            serde_json::Value::Number(number) => Ok(ConstValue::String(number.to_string())),
```
pimeys, line 47: https://grafbase.slack.com/archives/C04V5QCF8MP/p1694716505620239

engine/crates/integration-tests/src/postgresql.rs
```diff
@@ -0,0 +1,207 @@
...
+            .connection
+            .execute(query)
+            .await
+            .expect("error in query execute")
+    }
+
+    pub async fn execute(&self, operation: impl AsRef<str>) -> Response {
+        self.inner
+            .engine
+            .get_or_init(async { Engine::new(self.inner.schema.clone()).await })
```
pimeys, line 195: This prevents a race. If we initialize the `Engine` together with `TestApi`, it runs the introspection right away, and after that gives the control to the test that can create tables and so on. The engine doesn't know anything about these tables, so we postpone the initialization until the first GraphQL query/mutation arrives.

engine/crates/parser-sdl/src/connector_parsers.rs
pimeys: Good point. I'll return this comment.

obmarg (approved):

engine/crates/integration-tests/src/postgresql.rs
```diff
@@ -0,0 +1,207 @@
...
+
+        let builder = EngineBuilder::new(&api.inner.schema);
+
+        let result = parser_sdl::parse(&api.inner.schema, &HashMap::new(), &builder)
+            .await
+            .expect("error in parsing the schema")
+            .registry
+            .export_sdl(false);
+
+        parse_schema::<String>(&result).unwrap().to_string()
```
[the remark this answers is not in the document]
pimeys: To get a bit cleaner output. We should prettify already in our own code, but it misses nice indentation and all that...

engine/crates/engine/src/registry/scalars/time.rs
```diff
@@ -0,0 +1,62 @@
+use engine_value::ConstValue;
+use time::{format_description::well_known::Rfc3339, Time};
```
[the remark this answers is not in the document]
pimeys: Oh, now I remember why I chose time: it can parse time together with timezone, we need that for the `timetz` type in PostgreSQL.

engine/crates/parser-postgresql/src/registry/types.rs
```diff
@@ -66,8 +66,15 @@ fn generate_types(
                 key: column.database_name().to_string(),
             });
 
-            if column.database_type().is_enum() {
-                field.resolver = field.resolver.and_then(Transformer::RemoteEnum);
+            let extra_transformer = match column.database_type() {
+                DatabaseType::Scalar(ScalarType::Bytea) => Some(Transformer::BytesToBase64),
+                DatabaseType::Scalar(ScalarType::ByteaArray) => Some(Transformer::ByteArrayToBase64Array),
```
obmarg, line 71: Is the extra a on the end of `Bytea` & `ByteaArray` a typo?

pimeys: No, the type is `bytea`.

engine/crates/engine/src/registry/resolvers/postgresql/context.rs
```diff
@@ -0,0 +1,87 @@
...
+
+    /// Access to the schema registry.
+    pub fn registry(&self) -> &Registry {
+        self.context.registry()
+    }
+
+    /// A simple `user(by: { id: 1 })` filter, that has exactly one equals operation.
+    pub fn by_filter(&self) -> ServerResult<SimpleFilterIterator<'a>> {
+        let filter_map: Map<String, Value> = self.context.input_by_name("by")?;
+        let input_type = self.context.find_argument_type("by")?;
```
Finistere, line 77: do you really need to map input types to the table when you already have the `resolver_context.ty` providing the actual `meta_type`? You never change it inside `SImplerFilterIterator` as far as I see. You could also just provide the `TableWalker` from `self.table()` if lifetime allows for that instead of retrieving the same one each time. Or I suppose you intend to re-use that iterator for `findMany` filters?

engine/crates/engine/src/registry/resolvers/postgresql/context/selection/collection_args.rs
```diff
@@ -0,0 +1,161 @@
...
+        let before = value.field.get_argument("before").and_then(|value| value.as_string());
+        let after = value.field.get_argument("after").and_then(|value| value.as_string());
+
+        let mut order_by_argument = value
+            .field
+            .get_argument("orderBy")
+            .and_then(|value| value.as_slice())
+            .map(Cow::from);
+
+        // For `last` to work, we need some known order for the inner query.
```
Finistere, line 32: why is `first` any different? you have the same problem in both directions with cursor-based pagination. with `first: 100, after: X` we need to be sure `X` is unique.

engine/crates/engine/src/registry/resolvers/postgresql/request/query.rs
```diff
@@ -0,0 +1,154 @@
...
+                join_data.lateral();
+                collecting_select.left_join(join_data);
+            }
+        }
+    }
+
+    let mut json_select = Select::from_table(Table::from(collecting_select).alias(builder.table().database_name()));
+    json_select.value(row_to_json(builder.table().database_name(), false).alias(builder.field_name().to_string()));
+
+    if let Some(args) = builder.collection_args() {
```
Finistere, line 111: any reason to do that before the `match`?

engine/crates/integration-tests/src/postgresql.rs
Finistere: worth mentioning in a comment then if you bothered making a comment in GitHub.

engine/crates/engine/src/registry/resolvers/postgresql/context/selection/collection_args.rs
pimeys: This is more of a notion that we do `LIMIT` here too, but the order of the inner query is different.

Finistere: not sure what you mean, you're only specifying an ordering with `last`, not with `first`. Even though `first` still needs some ordering due to cursor navigation. So I don't get that part.

engine/crates/engine/src/registry/resolvers/postgresql/context.rs
pimeys: Probably not. This is me not really understanding still how the engine works.

engine/crates/engine/src/registry/resolvers/postgresql/context/selection/collection_args.rs
pimeys: Hmm, maybe I should enforce ordering then. I was thinking that it comes implicitly from the clustered primary or if missing, secondary key. Maybe I should not rely on that...

Finistere: no idea, would need to check postgres doc for several versions at least. It's probably the case. My point is mostly than `first` and `last` aren't different in that regard, so if you need to add ordering for one, the other one isn't any different.

PR #672, opened 2023-09-07
chore(parser): Parse used directives in a schema

Finistere:
# Description

# Type of change

- [ ] 💔 Breaking
- [ ] 🚀 Feature
- [ ] 🐛 Fix
- [ ] 🛠️ Tooling
- [ ] 🔨 Refactoring
- [ ] 🧪 Test
- [ ] 📦 Dependency

### DIFF
--- a/engine/crates/sdl-parser/src/lib.rs
+++ b/engine/crates/sdl-parser/src/lib.rs
@@ -11,7 +11,7 @@ use grafbase_engine::{
     },
     Pos,
 };
-use grafbase_engine_parser::{parse_schema, types::ServiceDocument, Error as ParserError};
+use grafbase_engine_parser::{types::ServiceDocument, Error as ParserError};
 use grafbase_types::UdfKind;
 use rules::{
     auth_directive::AuthDirective,
@@ -65,6 +65,7 @@ use crate::rules::{
 };
 
 pub mod connector_parsers;
+pub mod usage;
 
 mod directive_de;
 mod dynamic_string;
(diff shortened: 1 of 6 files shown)


engine/crates/sdl-parser/src/lib.rs
```diff
@@ -144,8 +165,79 @@ pub async fn parse<'a>(
...
+                                .extend(value.node.directives.into_iter().map(directive_name));
+                        }
+                    }
+                    TypeKind::InputObject(input_object) => {
+                        for field in input_object.fields {
+                            used.input_value
+                                .extend(field.node.directives.into_iter().map(directive_name));
+                        }
+                    }
+                    TypeKind::Union(_) | TypeKind::Scalar => (),
```
obmarg, line 225: Should we also gather directives from the enum definition, input object definition and the unions & scalars?

Not super important since we don't have any directives on those atm, but would save us having to remember to update this in the future.

engine/crates/sdl-parser/src/lib.rs
```diff
@@ -144,8 +165,79 @@ pub async fn parse<'a>(
...
 
-    let schema = parse_schema(schema)?;
+    grafbase_engine::parser::parse_schema(schema)
+}
 
+pub fn parse_used_directives(schema: &str) -> grafbase_engine::parser::Result<UsedDirectives> {
+    let doc = parse_schema(schema)?;
+    let mut used = UsedDirectives::default();
+    fn directive_name(directive: Positioned<ConstDirective>) -> String {
+        directive.node.name.node.to_string()
```
obmarg, line 175: Up to you, but what do you think about putting this stuff in it's own module & just `pub use`ing it from the root?

obmarg (approved):

engine/crates/sdl-parser/src/lib.rs
obmarg: Oh wait, never mind - I see those are being gathered outside of this case.

Finistere: damn, you're right missed it. Those aren't defined by the type kind but by the type definition. Thanks!

engine/crates/sdl-parser/src/lib.rs
Finistere: fair point

engine/crates/sdl-parser/src/lib.rs
Finistere: oh yes, right xD

engine/crates/sdl-parser/src/lib.rs
Finistere: done

PR #673, opened 2023-09-07
chore: simplify ExecutionEngine trait

obmarg:
I'm likely going to have to add another method or two to the ExecutionEngine trait soon, as well as updating a few places that it's used.  But it seemed quite difficult to work with, so I've tried to simplify it a bit first:

1. It had a ton of associated types that really only seem to exist because it doesn't have access to gateway-protocol types.  I've just moved it into `gateway-protocol` to fix this.  Not sure if it really needed to be in it's own crate.  Now that I'm typing this I'm wondering if I could have just added a dep between on `gateway-protocol` instead.  Not sure if that's any better though, let me know in the comments.
2. Updated it to take `self: Arc<Self>` so we don't need to pass a bunch of context to each function that could just live on self.
3. Tried to simplify the construction of ExecutionUpstream a bit, deduplicating the `local` & `not(local)` branches a bit.

This work has already happened in the `api` repo, just moving the relevant bits over here now.

### DIFF
--- a/engine/Cargo.lock
+++ b/engine/Cargo.lock
@@ -1251,7 +1251,6 @@ dependencies = [
  "cfg-if",
  "dynamodb",
  "gateway-protocol",
- "gateway-types",
  "grafbase-engine",
  "grafbase-local",
  "grafbase-runtime",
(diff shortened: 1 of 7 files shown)


hfgbarrigas (approved):
Nice!

PR #674, opened 2023-09-07
Revert "chore: simplify ExecutionEngine trait"

yoav-lavi:
Reverts grafbase/grafbase#673


obmarg (changes requested):
I don't think this needs reverted (based on a slack chat)

PR #689, opened 2023-09-13
feat: add SSE & incremental delivery to local gateway

obmarg:
The engine currently has support for `@defer` but it's useless without some sort of streaming transport.  This adds support for the [GraphQLOverSSE][1] & [Incremental Delivery][2] transports into the `gateway-local` crate.  Though I'll still need to make another PR in `api` followed by a PR here before this is functional.

While testing this I found that doing a `@defer` at the top level of the query was causing a panic because of some mismatched exepctations around `resolver_chain_node`.  So I've also fixed that in this PR

[1]: https://github.com/graphql/graphql-over-http/blob/main/rfcs/GraphQLOverSSE.md
[2]: https://github.com/graphql/graphql-over-http/blob/main/rfcs/IncrementalDelivery.md

### DIFF
--- a/engine/crates/integration-tests/tests/defer/mod.rs
+++ b/engine/crates/integration-tests/tests/defer/mod.rs
@@ -384,6 +384,71 @@ fn test_defer_with_errors() {
     });
 }
 
+#[test]
+fn test_defer_at_root() {
+    runtime().block_on(async {
+        let mock_server = wiremock::MockServer::start().await;
+        let engine = build_engine(petstore_schema(mock_server.address())).await;
+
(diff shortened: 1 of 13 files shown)


engine/crates/integration-tests/tests/defer/mod.rs
```diff
@@ -384,6 +384,71 @@ fn test_defer_with_errors() {
...
+            engine
+                .execute_stream(
+                r#"
+                    query {
+                        petstore {
+                            pet(petId: 123) {
+                                id
+                                name
+                            }
+						}
```
Finistere, line 406: weird formatting

Finistere (approved):
Impressive work!

PR #693, opened 2023-09-14
fix(udf): GB-4829: Avoid installing wrangler when it's already installed with the right version

jakubadamw:
# Description

This PR introduces a check ensuring that we don't reinstall wrangler excessively. If we find a valid wrangler already installed in ~/.grafbase/wrangler – by trying to invoke it through `npm exec` – and it's of the right version, no attempt to install it again will be carried out. Even if `npm install` itself for the existing package would already have been fairly quick.

# Type of change

- [ ] 💔 Breaking
- [ ] 🚀 Feature
- [X] 🐛 Fix
- [ ] 🛠️ Tooling
- [ ] 🔨 Refactoring
- [ ] 🧪 Test
- [ ] 📦 Dependency

### DIFF
--- a/cli/crates/server/src/udf_builder.rs
+++ b/cli/crates/server/src/udf_builder.rs
@@ -317,7 +317,7 @@ async fn symlink_grafbase_wasm_sdk(
     Ok(())
 }
 
-async fn installed_wrangler_version(wrangler_installation_path: impl AsRef<Path>, tracing: bool) -> Option<String> {
+async fn installed_wrangler_version(wrangler_installation_path: impl AsRef<Path>) -> Option<String> {
     let wrangler_installation_path = wrangler_installation_path.as_ref();
     let wrangler_arguments = &[
         "exec",
@@ -331,7 +331,7 @@ async fn installed_wrangler_version(wrangler_installation_path: impl AsRef<Path>
         JavaScriptPackageManager::Npm,
         wrangler_arguments,
         wrangler_installation_path,
-        tracing,
+        false,
         &[],
     )
     .await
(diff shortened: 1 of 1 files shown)


cli/crates/server/src/udf_builder.rs
```diff
@@ -56,7 +56,7 @@ async fn run_command<P: AsRef<Path>>(
 
     if output.status.success() {
         trace!("'{command_string}' succeeded");
-        Ok(())
+        Ok(output.stdout)
```
obmarg, line 59: Correct me if I'm wrong, but does this being populated not depend on a call to `.stdout(Stdio::piped())` which only happens if the user has passed in trace on the CLI?

cli/crates/server/src/udf_builder.rs
```diff
@@ -325,23 +351,41 @@ pub async fn install_wrangler(environment: &Environment, tracing: bool) -> Resul
...
         tracing,
         &[],
     )
     .await?;
     run_command(
         JavaScriptPackageManager::Npm,
-        &["install"],
-        environment.wrangler_installation_path.to_str().expect("must be valid"),
+        &["install", "--prefix", wrangler_installation_path_str],
+        wrangler_installation_path_str,
```
obmarg, line 388: I was under the impression `npm add` was an alias for `npm install` - is this second call actually doing anything?

jakubadamw: @obmarg silly me, of course.

cli/crates/server/src/udf_builder.rs
jakubadamw: Yes, good point, I disabled tracing for this particular invocation then.

obmarg (approved):

PR #694, opened 2023-09-14
fix: send valid json in the complete body

obmarg:
I noticed when writing tests of SSE that a naive client implementation might assume `body` is always JSON, but I was sending the word `complete` in the body of the complete event.  I don't think it's wrong, but might as well make it valid JSON.


yoav-lavi (approved):
This could have been split into 3-4 PRs ideally but approving for now

PR #700, opened 2023-09-15
chore(ci): Run the engine integration tests

pimeys:
# Description

These are not run anywhere right now.

### DIFF
--- a/.github/workflows/engine.yml
+++ b/.github/workflows/engine.yml
@@ -1,0 +1,84 @@
+name: engine
+
+on:
+  workflow_dispatch:
+  pull_request:
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  detect-change-type:
+    runs-on: buildjet-8vcpu-ubuntu-2204
+    outputs:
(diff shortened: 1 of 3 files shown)


.github/workflows/engine.yml
```diff
@@ -0,0 +1,92 @@
...
+  checks: write
+  # Allow repo checkout
+  contents: write
+  # Allow PRs write
+  pull-requests: write
+  # Allow issues write
+  issues: write
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
```
obmarg, line 18: Probably worth having this cancel old builds

```suggestion
concurrency:
  group: ${{ github.workflow }}-${{ github.ref }}
  cancel-in-progress: true
```

.github/workflows/engine.yml
```diff
@@ -0,0 +1,92 @@
...
+
+permissions:
+  # Allow checks read
+  checks: write
+  # Allow repo checkout
+  contents: write
+  # Allow PRs write
+  pull-requests: write
+  # Allow issues write
+  issues: write
```
obmarg, line 15: Do we actually need all of these?

obmarg (approved):
Sweet

PR #712, opened 2023-09-18
fix: use SendWrapper in less places

obmarg:
I've was adding some GraphQL connector tests into the integration-tests today, when I ran into some panics from our use of `SendWrapper`.  It was safe to use `SendWrapper` in wasm because it's inherently single threaded - less so in native code.

Thankfully we generally only need SendWrapper when we're building for wasm.  So I've updated all the places we use it to only do so on WASM.


pimeys (approved):
nice

PR #714, opened 2023-09-18
fix: dont forward `@defer` to connected gql apis

obmarg:
Now that we support `@defer` we need to be careful to not send it to connected GraphQL APIs.  They might well not support the directive, and even if they do we don't have any support for streaming responses from connected APIs.

This filters out the directive to handle this.  We can maybe do something better in the future/

### DIFF
--- a/engine/crates/engine/src/registry/resolvers/graphql/serializer.rs
+++ b/engine/crates/engine/src/registry/resolvers/graphql/serializer.rs
@@ -555,9 +555,9 @@ mod tests {
     #[rstest]
     #[case::one_bare("query { foo @include }")]
     #[case::one_arguments("query { foo @include(if: true) }")]
-    #[case::many_bare("query { foo @include @deprecated }")]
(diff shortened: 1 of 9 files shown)


pimeys (approved):

engine/crates/engine/src/registry/resolvers/graphql/serializer.rs
```diff
@@ -503,6 +507,13 @@ impl From<crate::Error> for Error {
     }
 }
 
+fn should_forward_directive(name: &str) -> bool {
+    // For now we only support forwarding the skip & defer directives.
```
Finistere, line 511:
```suggestion
    // For now we only support forwarding the skip & include directives.
```

Finistere (approved):

engine/crates/engine/src/registry/resolvers/graphql/serializer.rs
obmarg: 😓 apparently I have defer on the brain

PR #719, opened 2023-09-19
fix: handle variables inside inputs in graphql connector

obmarg:
The GraphQL connector needs to track which variables are used inside a query in order to know which variables to send to the downstream API.  It was doing this, but it was only pulling top-level variables rather than any nested inside a `Value`.  This PR fixes that and adds a test that catches the failure.

### DIFF
--- a/engine/crates/engine/value/src/lib.rs
+++ b/engine/crates/engine/value/src/lib.rs
@@ -542,6 +542,11 @@ impl Value {
             _ => None,
         }
     }
+
+    /// Returns an iterator over the variables that are used in this Value
+    pub fn variables_used(&self) -> VariableIterator<'_> {
+        VariableIterator::new(self)
+    }
 }
 
 impl Default for Value {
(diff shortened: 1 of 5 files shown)


engine/crates/engine/value/src/lib.rs
```diff
@@ -643,3 +648,34 @@ fn write_object<K: Display, V: Display>(
     }
     f.write_char('}')
 }
+
+/// Iterator over the variables that are used inside a Value
+pub struct VariableIterator<'a> {
+    values: Vec<&'a Value>,
```
pimeys, line 654: Why not `&'a [&'a Value]` together with `position: usize`. Your iterator returns references anyhow, so using a slice here would allow the whole struct to be `Copy`.

obmarg: How would I recurse through nested objects & lists if I had a slice?

pimeys: Maybe add `nested: Option<VariableIterator<'a>>` here and then first do `nested.and_then(|iter| iter.next())` and if it returns something, return that immediately.

obmarg: I can't use a slice either, since I'm sometimes pulling from an `IndexMap<String, Value>`.

I'm sure I could do it with a

```
struct VariableIterator<'a> {
   values: Box<dyn Iterator<Item=Value>>,
   nested: Option<Box<VariableIterator<'a>>
}
```

or something, but I kind of like the simplicity of the current and I'm not too fussed about allocations

PR #721, opened 2023-09-19
chore: Move additional code from API

Finistere:
Moving missing parts from API to have a fully functional engine.


obmarg (approved):
There's going to be some confusing af commits in that history I bet, but never mind 🔀

PR #725, opened 2023-09-19
feat(connector): Neon pagination

pimeys:
Implements cursors, page infos and before/after filters for Neon/PostgreSQL. Again, about 2k lines of this PR are tests, so the actual feature should be about 1k lines to read.

Some weird things: we load all the data, even nested data, in one query. This kind of puts us implementing page infos in resolvers, and additionally the cursors are resolved values, so they are also handled in special resolvers. This required me to add some query-specific data to the resolved value, which allows cursor calculation if needed. This whole thing has a bit too much copying we should maybe address later.

Another one that will bring questions is our ordering preference. MongoDB has nulls first _always_, but PostgreSQL has this as a server setting, or nulls last when ascending. For the first iteration, we take a stance to have nulls _always first_ no matter what is the order. This simplifies the pagination code, and we can extend it later to let user to choose `ASC`, `DESC`, `ASC_NULLS_LAST`, `ASC_NULLS_FIRST`, `DESC_NULLS_LAST` or `DESC_NULLS_FIRST`.

Closes: GB-4677

### DIFF
--- a/engine/crates/engine/src/registry/resolvers/transformer.rs
+++ b/engine/crates/engine/src/registry/resolvers/transformer.rs
@@ -1,7 +1,5 @@
 #![allow(deprecated)]
 
-use std::hash::Hash;
-
 use base64::{engine::general_purpose::STANDARD_NO_PAD, Engine};
 use dynamodb::attribute_to_value;
 use dynomite::AttributeValue;
@@ -9,6 +7,8 @@ use grafbase_sql_ast::ast::Order;
 use indexmap::IndexMap;
 use postgresql_types::{cursor::SQLCursor, database_definition::TableId};
 use runtime::search::GraphqlCursor;
+use serde_json::Value;
+use std::hash::Hash;
 
 use super::{
     dynamo_querying::{DynamoResolver, IdCursor},
@@ -167,7 +167,7 @@ impl Transformer {
                 Ok(new_value)
             }
             Transformer::PaginationData => {
-                let pagination = last_resolver_value
+                let pagination = dbg!(last_resolver_value)
                     .and_then(|x| x.pagination.as_ref())
                     .map(ResolvedPaginationInfo::output);
                 Ok(ResolvedValue::new(serde_json::to_value(pagination)?))
@@ -178,7 +178,7 @@ impl Transformer {
             // TODO: look into optimizing nested single edges
             Transformer::SingleEdge { key, relation_name } => {
                 let old_val = match last_resolver_value.and_then(|x| x.data_resolved().get(key)) {
-                    Some(serde_json::Value::Array(arr)) => {
+                    Some(Value::Array(arr)) => {
                         // Check than the old_val is an array with only 1 element.
                         if arr.len() > 1 {
                             ctx.add_error(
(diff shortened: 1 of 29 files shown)


engine/crates/parser-postgresql/src/registry/types/page_info.rs
```diff
@@ -0,0 +1,41 @@
...
+use engine::registry::{resolvers::transformer::Transformer, MetaField, ObjectType};
+
+pub(super) fn register(input_ctx: &InputContext<'_>, output_ctx: &mut OutputContext) {
+    let object = ObjectType::new(
+        input_ctx.type_name("PageInfo"),
+        [
+            MetaField {
+                name: "hasPreviousPage".to_string(),
+                ty: "Boolean!".into(),
+                resolver: Transformer::PaginationData.and_then(Transformer::select("has_previous_page")),
```
obmarg, line 12: Probably not that important, but seems odd we don't put this `PaginationData` resolver onto the `pageInfo` field and save having to do it for every single field.

Though maybe there's a reason that won't work...

engine/crates/engine/src/registry/resolvers/transformer.rs
```diff
@@ -275,66 +291,202 @@ impl Transformer {
...
+                Ok(new_value)
+            }
+            Transformer::PostgresPageInfo => {
+                let resolved_value =
+                    last_resolver_value.ok_or_else(|| Error::new("Internal error resolving postgres page info"))?;
+
+                let mut rows: Vec<Value> = match resolved_value.data_resolved() {
+                    Value::Array(rows) => rows.clone(),
+                    _ => return Ok(resolved_value.clone()),
+                };
```
obmarg, line 368: I think you could make this clone less expensive in certain circumstances if you changed this function to take an owned `Option<ResolvedValue>` instead of an `Option<&ResolvedValue>` (which I have been intending to do at some point anyway) and then using take instead of data_resolved:

```suggestion
                if !resolved_value.data_resolved().is_array() {
                    return Ok(resolved_value)
                }
                let Value::Array(mut rows) = resolved_value.take() else {
                    unreachable!();
                }
```

`ResolvedValue::take` will _sometimes_ clone the underlying data, but it'll avoid it if it can.  At the very least the case where you're doing `FindMany.and_then(PostgresPageInfo)` should avoid the clone I think

pimeys: How much work would it be to do this ownership change in this PR? I've been thinking about doing so, but didn't want to _change all the things_ immediately.

engine/crates/parser-postgresql/src/registry/types/page_info.rs
pimeys: Yeah, I tried that. So we then don't have the pagination data in the last resolved value, just the data from the select is available (if I just keep the select here, move the pagination data to the `pageInfo`).

I can also just try to select the whole object at a time for `pageInfo`, but the default resolver for these fields is then `Parent` and it tries to convert the `PageInfo` object into a boolean etc. later on which breaks.

obmarg (approved):

engine/crates/engine/src/registry/resolvers/transformer.rs
obmarg: I don't think this is a large change, but it can definitely wait till you're not on holiday.

PR #732, opened 2023-09-21
test: that custom resolvers can be deferred

obmarg:
#731 fixed the problem with deferring custom resolvers - this adds a test for it.

Fixes GB-4762


engine/crates/integration-tests/tests/defer/mod.rs
```diff
@@ -689,6 +690,170 @@ fn test_invalid_defer_parameters() {
...
+            "data": null,
+            "errors": [
+              {
+                "locations": [
+                  {
+                    "column": 29,
+                    "line": 1
+                  }
+                ],
+                "message": "I'm afraid I can't do that Dave",
```
pimeys, line 813: > It’s called “Daisy”. Daisy, Daisy, give me your answer do. I’m half crazy all for the love of you. It won’t be a stylish marriage, I can’t afford a carriage. But you’ll look sweet upon the seat of a bicycle built for two.

pimeys (approved):

PR #760, opened 2023-10-04
feat: first pass of federation subgraph support

obmarg:
The first chunk of work  for supporting apollo federation subgraphs:

- Parsing a directive to enable it.
- Recording entities in the Registry (we had a way of doing this inherited from async_graphql, but I've changed it up)
- Resolving the `_entities` field
- Some tests

Fixes GB-4995
Fixes GB-4996
Fixes GB-4997


engine/crates/engine/src/registry/federation.rs
```diff
@@ -0,0 +1,145 @@
...
+/// There should be one instance of this for each MetaType that represents
+/// a federation entity.
+#[derive(Clone, Debug, serde::Serialize, serde::Deserialize, Default)]
+pub struct FederationEntity {
+    keys: Vec<FederationKey>,
+}
+
+#[derive(Clone, Debug, serde::Serialize, serde::Deserialize)]
+pub enum FederationResolver {
+    DynamoUnique,
```
pimeys, line 18: Do I understand this correctly, but we can only resolve data from dynamo with the subgraph in the MVP?

engine/crates/engine/src/registry/resolvers/federation.rs
```diff
@@ -0,0 +1,90 @@
...
+    // the resolvers out by passing them a more accurate ResolverContext.
+    // Not sure this is the best idea, but lets find out.
+    let actual_type = ctx.registry().lookup_expecting(&representation.ty).map_err(|error|
+        // This probably indicates a malformed registry, lets return an error
+        Error::new(format!("Internal error: {} could not be looked up in registry: {}", representation.ty, error.message))
+    )?;
+    let resolver_context = ResolverContext::new(&ctx).with_ty(actual_type);
+
+    let data = match key_being_resolved.resolver() {
+        FederationResolver::DynamoUnique => {
```
pimeys, line 68: Do we need to extend this every time we add new resolvers for new databases?

engine/crates/integration-tests/setup-local-dynamo.sh
```diff
@@ -0,0 +1,7 @@
+dy --region local admin create table database --keys __pk __sk
```
pimeys, line 1: Could this happen automatically when we start the docker image?

engine/crates/engine/src/registry/federation.rs
obmarg: That's all we can do in this PR - but I'm not done with the MVP yet.  Dynamo was just the easiest thing to add so I could test everything else that's in here.

engine/crates/engine/src/registry/resolvers/federation.rs
obmarg: If we want those to be supported by federation - yes.  I am hoping this glue will be fairly trivial like it is here, but will see how true that is when we add another connector or two.  

We could probably look into some kind of trait/standard resolver/whatever at some point - but it's not yet clear what that'd need to look like (and I didn't want to go refactoring dynamo in particular since we'll be ditching it at some point).

engine/crates/integration-tests/setup-local-dynamo.sh
pimeys: Can we still just run:

```console
> docker compose up
> cargo nextest run
```

and be done with it, or do we always need to run this script before the tests?

engine/crates/engine/src/registry/resolvers/federation.rs
pimeys: Some standard behavior would be nice. Let's figure that out later, I'm just worried that we'll forget this part of the codebase when we implement new connectors...

engine/crates/integration-tests/setup-local-dynamo.sh
obmarg: yeah, you just need to start docker & run the tests.

you know what, i'm getting rid of this script - it's presence is probably more confusing than it's worth

PR #761, opened 2023-10-04
fix(connector): Allow emitting input field for nullable composite key

pimeys:
This fixes an edge case that might seem small, but would lead to catastrophic results in the worst case. Let's imagine a table:

```sql
CREATE TABLE "user" (
  name VARCHAR(255) NOT NULL,
  email VARCHAR(255) NULL,
  CONSTRAINT foo_key UNIQUE (name, email)
);
```

The input type would be something like this:

```graphql
query {
  user(by: { nameEmail: { name: "Naukio", email: null } }) {
    name
    email
  }
}
```

This would trigger a query with `WHERE name = 'Naukio' AND email IS NULL`. Now, the column being nullable, the generated graphql type allows one to emit the value completely, so this is also allowed:

```graphql
query {
  user(by: { nameEmail: { name: "Naukio" } }) {
    name
    email
  }
}
```

Before the fix, it renders a query `WHERE name = 'Naukio'`, which then could return more than one rows, leading to  undefined behavior. And what's worse, when we implement the deletion or updates of one item, this would delete/update more than one item in the database.

The fix maps all fields _not defined_ in the query that are part of the constraint, and adds `AND field IS NULL` for the fields missing in the input.

Closes: GB-4909

### DIFF
--- a/engine/crates/parser-postgresql/src/registry/context/output/builders.rs
+++ b/engine/crates/parser-postgresql/src/registry/context/output/builders.rs
@@ -1,5 +1,5 @@
 use engine::registry::{Constraint, EnumType, InputObjectType, MetaEnumValue, MetaField, MetaInputValue, ObjectType};
-use postgresql_types::database_definition::{RelationId, TableColumnId, TableId};
+use postgresql_types::database_definition::{RelationId, TableColumnId, TableId, UniqueConstraintId};
 
 #[derive(Debug)]
 pub struct ObjectTypeBuilder {
@@ -45,6 +45,7 @@ pub struct InputTypeBuilder {
     pub(super) input_object_type: InputObjectType,
     pub(super) type_mapping: Vec<(String, TableId)>,
     pub(super) field_mapping: Vec<(String, TableColumnId)>,
+    pub(super) unique_constraint_mapping: Vec<(String, UniqueConstraintId)>,
     pub(super) relation_mapping: Vec<(String, RelationId)>,
     pub(super) nested: Vec<InputObjectType>,
 }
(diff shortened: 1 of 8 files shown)


engine/crates/parser-postgresql/src/registry/context/output/builders.rs
```diff
@@ -55,19 +56,20 @@ impl InputTypeBuilder {
...
+            unique_constraint_mapping: Vec::new(),
             relation_mapping: Vec::new(),
             nested: Vec::new(),
         }
     }
 
-    pub(crate) fn push_input_column(&mut self, value: MetaInputValue, column_id: TableColumnId) {
-        self.field_mapping.push((value.name.to_string(), column_id));
-        self.push_non_mapped_input_column(value);
+    pub(crate) fn map_unique_constraint(&mut self, field: &str, constraint_id: UniqueConstraintId) {
```
Finistere, line 67: why not `push_unique_constraint` like `push_input_column`? both are "mapping".

engine/crates/parser-postgresql/src/registry/queries/input/oneof.rs
```diff
@@ -30,6 +30,7 @@ pub(crate) fn register(input_ctx: &InputContext<'_>, table: TableWalker<'_>, out
 
                 let query_name = type_prefix.to_camel_case();
 
+                builder.map_unique_constraint(&query_name, constraint.id());
```
Finistere, line 33: Won't this break with two tables `A` and `B` having both a constraint on columns with the same name?

Finistere (approved):

engine/crates/parser-postgresql/src/registry/queries/input/oneof.rs
pimeys: This builder is per table, meaning there will be no other tables managing this collection. In the `Names` struct, the mapping is per `(TableId, StringId)`, so it's not a problem.

engine/crates/parser-postgresql/src/registry/context/output/builders.rs
pimeys: Yeah, kind of. I'm not creating anything with this, I'm just mapping this field to this constraint.

Finistere: up to you, just struck me reading the PR, but I only have a limited view through it.^^

PR #769, opened 2023-10-05
feat: OpenAPI federation inference

obmarg:
I added basic (and definitely broken) federation support in #760.  This PR expands that to partially cover the OpenAPI connector.

To do this I'm using the resource -> operation relationships that we're already inferring from the OpenAPI specs.  This inference definitely doesn't work on all OpenAPI specs, but it works reasonably for some so it's a good starting point.  A future bit of work will provide a way for users to specify their own federation keys & resolvers on top of the OpenAPI connector.

I've added some tests with the stripe schema in here - it's not inferring anywhere near as many entities as I'd expect so I'm almost definitely going to need to follow up on this with some fixes.


pimeys (approved):
lgtm

PR #770, opened 2023-10-05
feat: cli start

Finistere:
Not really perfect as an instance of Miniflare is spawned for each
resolver... But except that performance issues, it should behave like
expected.

# Description

# Type of change

- [ ] 💔 Breaking
- [ ] 🚀 Feature
- [ ] 🐛 Fix
- [ ] 🛠️ Tooling
- [ ] 🔨 Refactoring
- [ ] 🧪 Test
- [ ] 📦 Dependency


cli/crates/cli/src/cli_input.rs
```diff
@@ -213,6 +216,50 @@ pub struct LogsCommand {
...
+    #[arg(long, value_name = "GRAPHQL_OPERATION_LOG_LEVEL")]
+    pub log_level_graphql_operations: Option<LogLevelFilter>,
+    /// Log level to print for fetch requests, defaults to 'log-level'
+    #[arg(long, value_name = "FETCH_REQUEST_LOG_LEVEL")]
+    pub log_level_fetch_requests: Option<LogLevelFilter>,
+    /// Default log level to print
+    #[arg(long)]
+    pub log_level: Option<LogLevelFilter>,
+    #[arg(long)]
+    pub listen: Option<IpAddr>,
```
yoav-lavi, line 237: I'd possibly name this host or something more indicative

Finistere: well it's not really a host, in PostgreSQL they're naming it `listen_addresses`. We could name it `address` or `ip` eventually. wdyt?

yoav-lavi: `address` is probably better IMO but you could poll it

yoav-lavi: Do other CLIs use `listen`?

Finistere: let's go with address then :)

yoav-lavi: I'd poll it

cli/crates/server/src/servers.rs
```diff
@@ -13,31 +13,74 @@ use common::consts::MAX_PORT;
...
+    listen: IpAddr,
+    port: u16,
+    tracing: bool,
+    message_sender: tokio::sync::mpsc::UnboundedSender<ServerMessage>,
+) -> Result<(), ServerError> {
+    use bridge::BridgeState;
+
+    export_embedded_files()?;
+    create_project_dot_grafbase_directory()?;
+    validate_dependencies().await?;
```
yoav-lavi, line 46: Do we still need this if you don't use UDFs?

cli/crates/server/src/bridge/udf.rs
```diff
@@ -396,7 +398,7 @@ pub async fn invoke(
 
     messages.sort_by_key(|(logged_time, _)| *logged_time);
     for (_, message) in messages {
-        bridge_sender.send(message).await.unwrap();
+        bridge_sender.send(message).unwrap();
```
yoav-lavi, line 401: unwrap here

cli/crates/server/src/servers.rs
Finistere: we use UDFs, It's not great as I'm creating an instance of miniflare per resolver right now, but we do use it. Have started a PR to have a single miniflare, but didn't finish.

yoav-lavi: I mean, could we do this only if UDFs are in use?

cli/crates/server/src/bridge/udf.rs
Finistere: well yeah like before 🤷

yoav-lavi: I'd use an expect still if we're modifying it anyhow

Finistere: ok

cli/crates/server/src/servers.rs
Finistere: but yeah fair point

Finistere: yeah definitely, didn't properly check what it was doing, good catch :)

yoav-lavi: Do we need that for the SDK though?
We might also be able to do the export given that condition

Finistere: ah agreed again. Keeping `create_project_dot_grafbase_directory()?` though, as it's not necessarily tied to resolvers, no?

yoav-lavi: Yeah that's used for the DB as well. I'd just check that we also include a check for the SDK in the condition since you'll need node for it and the extract likely

Finistere: what do you mean with SDK? the sdk inside resolvers will only be used if there are resolvers. So we're good with the UDF check, no?

yoav-lavi: I mean if you use the TS config, that uses node and IIRC an asset

Finistere: ahhh you're right, doesn't use assets but does use node obviously.

yoav-lavi: But I'm assuming we can check for that

Finistere: Well moved the check behind udf check and doing it in the schema parsing if we're using node. Double-check, but it's simple. 🤷

cli/crates/server/src/servers.rs
```diff
@@ -52,54 +96,40 @@ const EVENT_BUS_BOUND: usize = 5;
...
-) -> (JoinHandle<Result<(), ServerError>>, Receiver<ServerMessage>) {
-    let (sender, receiver): (Sender<ServerMessage>, Receiver<ServerMessage>) = mpsc::channel();
-
+    message_sender: tokio::sync::mpsc::UnboundedSender<ServerMessage>,
+) -> Result<(), ServerError> {
     let project = Project::get();
 
-    let handle = thread::spawn(move || {
-        export_embedded_files()?;
+    export_embedded_files()?;
```
yoav-lavi, line 108: @Finistere I don't think we need this then

yoav-lavi (approved):
Great work!

PR #776, opened 2023-10-06
chore(ci): Downgrade Windows runner to 2019 version (should be faster)

pimeys:
# Description

See: https://github.com/actions/checkout/issues/1186


fbjork (approved):

pimeys:
never seems to catch a runner, so seems like this will not work.

PR #782, opened 2023-10-09
feat(connector): Postgres create mutations

pimeys:
# PostgreSQL create mutations

Adds the create and createMany mutations for PostgreSQL. Again, most of this PR is tests and the actual feature is not that much to read.

## Create one

Creates one row to the database.

```graphql
query {
  userCreate(input: { id: 1, name: "Musti" }) {
    id
    name
  }
}
```

The return value is a user:

```json
{
  "data": {
    "userCreate": {
      "id": 1,
      "name": "Musti"
    }
  }
}
```

So far only scalar fields can be used in the creation. We'll be coming back to the nested creates in the following issue: https://linear.app/grafbase/issue/GB-5047/nested-creates

## Create many

Creates multiple rows to the database

```graphql
query {
  userCreateMany(input: [{ id: 1, name: "Musti" }, { id: 2, name: "Naukio" }]) {
    id
    name
  }
}
```

The return value is two users:

```json
{
  "data": {
    "userCreateMany": [
      {
        "id": 1,
        "name": "Musti"
      },
      {
        "id": 2,
        "name": "Naukio"
      }
    ]
  }
}
```

## Insert query

The query is not just a simple insert, but we're preparing it for the nested inserts that are coming later:

```sql
WITH "user" AS (INSERT INTO "User" (id) VALUES (1) RETURNING id)
SELECT json_build_object('id', "user"."id")
```

You can read some of the reasoning from the [gist](https://gist.github.com/pimeys/fe141cc61133f91d15a35a3a0f97266b) how this enables nested inserts.

Closes: GB-4678


obmarg (approved):

engine/crates/engine/src/registry/scalars/hex_bytes.rs
```diff
@@ -0,0 +1,53 @@
...
+pub struct HexBytesScalar;
+
+impl<'a> SDLDefinitionScalar<'a> for HexBytesScalar {
+    fn name() -> Option<&'a str> {
+        Some("HexBytes")
+    }
+
+    fn description() -> Option<&'a str> {
+        Some("Bytes stored as hex. The value is returned as base64-encoded string.")
+    }
```
tomhoule, line 17: Is it relevant that the bytes are stored as hex for the consumer? The input is also expected to be base64, if I read the lines below correctly.

tomhoule: It looks like we already have Bytes. I'm not clear on why we need a separate type for values where the input/output format between Grafbase and the database is always a hex string. It's more connector-dependent than value dependent, postgres takes hex strings.

pimeys: This is because all translations to and from this type are common to all connectors. And all transformers are when you load a value, not when you store one. After consulting @Finistere, the decided best practice for now is to create another type, if the value is stored differently in the database. In the future, we'd like to have connector-specific transformers.

PR #783, opened 2023-10-10
Refactor a little bit how the operation type gets propagated from the executor to the gateway

jakubadamw:
# Description

Necessary for some of the changes on the platform.

# Type of change

- [ ] 💔 Breaking
- [ ] 🚀 Feature
- [ ] 🐛 Fix
- [ ] 🛠️ Tooling
- [X] 🔨 Refactoring
- [ ] 🧪 Test
- [ ] 📦 Dependency

### DIFF
--- a/engine/crates/common-types/src/lib.rs
+++ b/engine/crates/common-types/src/lib.rs
@@ -45,7 +45,6 @@ pub enum LogLevel {
     Debug,
 }
 
-#[serde_with::serde_as]
 #[derive(serde::Deserialize, serde::Serialize, Debug)]
 pub struct Operation<'a> {
     pub name: Option<std::borrow::Cow<'a, str>>,
--- a/engine/crates/common-types/src/lib.rs
+++ b/engine/crates/common-types/src/lib.rs
@@ -20,11 +20,20 @@ pub enum UdfKind {
     Authorizer,
 }
 
-#[derive(serde::Deserialize, serde::Serialize, Clone, Copy, Debug)]
+#[derive(serde::Deserialize, serde::Serialize, Clone, Copy, Debug, strum::Display, PartialEq)]
+#[strum(serialize_all = "lowercase")]
 pub enum OperationType {
     Query { is_introspection: bool },
     Mutation,
     Subscription,
+}
+
+impl Default for OperationType {
+    fn default() -> Self {
+        Self::Query {
+            is_introspection: false,
+        }
+    }
 }
 
 #[derive(Clone, Copy, Debug, PartialEq, PartialOrd, serde::Deserialize, serde::Serialize)]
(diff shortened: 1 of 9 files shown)


engine/crates/common-types/src/lib.rs
```diff
@@ -36,6 +45,13 @@ pub enum LogLevel {
     Debug,
 }
 
+#[serde_with::serde_as]
```
obmarg, line 48: nitpicking, but doesn't look like we need serde_as here?

engine/crates/graphql-extensions/src/runtime_log.rs
```diff
@@ -97,23 +97,19 @@ impl Extension for RuntimeLogExtension {
...
-                        ParserOperationType::Subscription => OperationType::Subscription,
+        if let Some(operation) = response.graphql_operation.as_ref() {
+            self.log_event_receiver
+                .invoke(
+                    runtime_ctx.ray_id(),
+                    runtime_ctx.log.request_log_event_id,
+                    LogEventType::OperationCompleted {
+                        name: operation_name.map(From::from),
+                        duration,
+                        r#type: operation.r#type,
```
tomhoule, line 108: Am I right to think this simplification is the motivation for the change?

engine/crates/engine/src/schema.rs
```diff
@@ -566,9 +577,7 @@ impl Schema {
                             .execute(env.operation_name.as_deref(), &env.operation, &mut fut)
                             .await
                     }
-                    // here we don't know the type of the operation because it failed preparing the request
-                    // defaulting but this might not be the best option
-                    Err(errors) => Response::from_errors(errors, Default::default()),
+                    Err(errors) => Response::bad_request(errors),
```
tomhoule, line 580: Was the Default impl on OperationType for this call site? Is it still necessary?

tomhoule (approved):
Superficial review, but lgtm :+1:

engine/crates/engine/src/response/mod.rs
```diff
@@ -38,8 +45,33 @@ pub struct Response {
     #[serde(skip)]
     pub http_headers: HeaderMap,
 
-    /// GraphQL operation type derived from incoming request
-    pub operation_type: OperationType,
+    /// GraphQL operation.
+    pub graphql_operation: Option<ResponseOperation>,
```
obmarg, line 49: Not really a comment on your change since we were doing it already, but feels kind of bad passing data back in the response like this....

obmarg (approved):

engine/crates/engine/src/response/mod.rs
jakubadamw: We actually `take()` it in the gateway in my API PR to hide it from the client, but you're right – I should add `#[serde(skip_serializing_if = "Option::is_none")]`

engine/crates/common-types/src/lib.rs
jakubadamw: Good catch!

engine/crates/engine/src/schema.rs
jakubadamw: Good catch!

engine/crates/graphql-extensions/src/runtime_log.rs
jakubadamw: This, plus getting rid of [this API-side processing](https://github.com/grafbase/api/pull/2852/files#diff-666c2f5e9ac8daf19cde946d15b8f52b36192dbdef530ef7fab2eed76571a8f4L823-L870) here.

PR #787, opened 2023-10-10
feat(connector): Postgres update mutations

pimeys:
# PostgreSQL update mutations

Adds the update and updateMany mutations for PostgreSQL.

Tests: 2425 lines
Implementation: 469 lines

## Update one

Updates one row in the database.

```graphql
query {
  userUpdate(by: { id: 1 }, input: { name: "Musti" }) {
    id
    name
  }
}
```

The return value is a user:

```json
{
  "data": {
    "userUpdate": {
      "id": 1,
      "name": "Musti"
    }
  }
}
```

So far only scalar fields can be used in the updates. We'll be coming back to the nested updates in the following issue: https://linear.app/grafbase/issue/GB-5089/nested-updates

## Update many

Updates multiple rows in the database

```graphql
query {
  userUpdateMany(filter: { id: { eq: 1 } }, input: { name: "Musti" }) {
    id
    name
  }
}
```

The return value is a list of users:

```json
{
  "data": {
    "userUpdateMany": [
      {
        "id": 1,
        "name": "Musti"
      }
    ]
  }
}
```

## Update query

The query is not just a simple update, but we're preparing it for the nested updates that are coming later:

```sql
WITH "user" AS (UPDATE "User" SET "name" = 'Musti' WHERE id = 1 RETURNING id, name)
SELECT json_build_object('id', "user"."id", 'name', "user"."name")
```

Closes: GB-4680

### DIFF
--- a/cli/Cargo.lock
+++ b/cli/Cargo.lock
@@ -2477,9 +2477,9 @@ dependencies = [
 
 [[package]]
 name = "grafbase-sql-ast"
-version = "0.1.8"
+version = "0.1.9"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "80d39c76dd44c3cec2c81e527770394ac6371da5883454addcedafae5470bd98"
+checksum = "c3aac7444d8a2748651d0ad4b7f27dc08b616b2babfe0ebe2656b2ee82463396"
 dependencies = [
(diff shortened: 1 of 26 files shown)


yoav-lavi (approved):
Nice work!

PR #790, opened 2023-10-11
feat: unresolvable federation entities

obmarg:
A key part of federation is supporting "unresolvable" entities - entities that are owned by another subgraph but referenced by entities in the current subgraph.

For example with if we had two subgraphs, one for `Todo`s and one for `TodoList`s, the `Todo` subgraph could use this schema:

```
type Todo @model @key(fields: "id") {
   title: String
   listId: ID!
   list: TodoList!
}

type TodoList @key(fields: "id", resolvable: false) {
   id: ID!
}
```

This (combined with a custom resolver) allows the `Todo` subgraph to return a "representation" of a `TodoList`: `{"__typename": "TodoList", "id": "123"}` - which gives the federation router the data it needs to request properties of the `TodoList` from the other subgraph.

I've taken a bunch of shortcuts to get this out quick, but will resolve those later.

Part of GB-5058
Part of GB-5084


engine/crates/parser-sdl/src/rules/basic_type.rs
```diff
@@ -29,15 +33,24 @@ impl<'a> Visitor<'a> for BasicType {
         type_definition: &'a engine::Positioned<engine_parser::types::TypeDefinition>,
     ) {
         let directives = &type_definition.node.directives;
-        if_chain! {
```
obmarg, line 32: Got rid of this `if_chain` stuff because it's quite annoying - breaks rust analyzer and we have `let else` as an alternative now anyway

engine/crates/parser-sdl/src/rules/basic_type.rs
```diff
@@ -54,20 +67,56 @@ impl<'a> Visitor<'a> for BasicType {
...
+            &type_name,
+        );
+
+        // If the type is a non primitive and also not modelized, it means we need to
+        // create the Input version of it.
+        // If the input is non used by other queries/mutation, it'll be removed from the
+        // final schema.
+        add_input_type_non_primitive(ctx, object, &type_name);
+
+        // We also need to parse any unresolvable keys
```
obmarg, line 86: The actual changes to this function start here

engine/crates/parser-sdl/src/rules/federation/directive.rs
```diff
@@ -56,17 +56,6 @@ mod tests {
...
 
-    #[cfg(comebacktothisone)]
-    #[test]
-    fn test_federation_with_no_models() {
-        let schema = r#"extend schema @federation(version: "2.3")"#;
-        let registry = crate::to_parse_result_with_variables(schema, &HashMap::new())
-            .unwrap()
-            .registry;
-        // TODO: This fails.  Not sure if it should?
-        assert_eq!(registry.export_sdl(false), registry.export_sdl(true));
```
obmarg, line 67: Decided this test wasn't worth the bother at the moment

engine/crates/parser-sdl/src/rules/basic_type.rs
tomhoule: 100% agree

engine/crates/parser-sdl/src/rules/basic_type.rs
```diff
@@ -54,20 +67,56 @@ impl<'a> Visitor<'a> for BasicType {
...
+                })
+                .partition_result::<Vec<_>, Vec<_>, _, _>();
+
+            ctx.append_errors(errors);
+
+            for (pos, key) in oks {
+                if key.resolvable {
+                    ctx.report_error(
+                        vec![pos],
+                        "Found a resolvable key on a basic type, which is currently unsupported",
```
tomhoule, line 109: Am I reading this file correctly, that BasicType is a GraphQL output type, in contrast to a model? (`type Foo { ... }`?)

I first read about resolvable and non-resolvable keys this afternoon so I'm likely missing important bits, but wouldn't we need to support both resolvable and unresolvable keys to define subgraphs? The basic thing I don't understand about this error is what is missing to make them work, in basic scenarios it wouldn't influence how the object is returned. (does it have to do with `__Entities`?)

tomhoule: Note: I haven't had time to read the internal design docs on federation yet, I'm doing that next and the answer is probably there.

engine/crates/parser-sdl/src/rules/federation/key.rs
```diff
@@ -0,0 +1,116 @@
...
+            E: serde::de::Error,
+        {
+            // This whole implementation is extremely naive and doesn't support a lot of stuff.
+            // Will fix in https://linear.app/grafbase/issue/GB-5086
+            if value.contains('{') {
+                return Err(E::custom("nested fields in keys aren't supported at the moment"));
+            }
+
+            Ok(value
+                .split(' ')
```
tomhoule, line 50: Naive question: I haven't seen this in the unit tests but I guess the expected format is `@key(fields: "a b c")`?

tomhoule (approved):
Sorry for the delay reviewing. Reading up on federation beforehand helped make me understand the PR though, and this is a good way to see more of the code. No important feedback in my comments. LGTM :+1:

engine/crates/parser-sdl/src/rules/federation/key.rs
obmarg: To be more accurate: it's a GraphQL query without arguments or directives.  So `@key(fields: "a b { c { d }}")` would be valid.  Though this implementation only supports `a b`.

tomhoule: Oooh, I see. Technically commas would be allowed too, then? Probably not relevant to implement at this stage.

obmarg: Yeah, I would assume so.  Going to write/use an actual parser when I revisi the mentioned issue.

PR #791, opened 2023-10-11
GB-5006 - Add federation to SDK

yoav-lavi:
# Description

## Features

- Allows setting the federation version in the SDK

# Type of change

- [ ] 💔 Breaking
- [x] 🚀 Feature
- [ ] 🐛 Fix
- [ ] 🛠️ Tooling
- [ ] 🔨 Refactoring
- [ ] 🧪 Test
- [ ] 📦 Dependency

### DIFF
--- a/packages/grafbase-sdk/src/federation.ts
+++ b/packages/grafbase-sdk/src/federation.ts
@@ -1,5 +1,5 @@
 export interface FederationParams {
-  version: string
+  version: '2.3'
 }
 
 export class Federation {
--- a/packages/grafbase-sdk/src/federation.ts
+++ b/packages/grafbase-sdk/src/federation.ts
@@ -1,0 +1,15 @@
+export interface FederationParams {
+  version: string
+}
+
+export class Federation {
+  private version: string
+
+  constructor(params: FederationParams) {
+    this.version = params.version
+  }
+
+  public toString(): string {
+    return `\nextend schema @federation(version: "${this.version}")\n`
+  }
+}
--- a/packages/grafbase-sdk/tests/unit/federation.test.ts
+++ b/packages/grafbase-sdk/tests/unit/federation.test.ts
@@ -1,0 +1,42 @@
+import { config, g } from '../../src/index'
+import { describe, expect, it, beforeEach } from '@jest/globals'
+import { renderGraphQL } from '../utils'
+
+describe('Federation generator', () => {
+  beforeEach(() => g.clear())
+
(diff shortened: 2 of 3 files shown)


packages/grafbase-sdk/src/federation.ts
```diff
@@ -0,0 +1,15 @@
+export interface FederationParams {
+  version: string
+}
+
+export class Federation {
+  private version: string
```
obmarg, line 6: Wonder if it's worth typing this more specifically - we only accept the string `2.3` currently, so I _think_ we could give users faster feedback with the type system?

obmarg (approved):
Good stuff, thanks for helping with this

packages/grafbase-sdk/src/federation.ts
yoav-lavi: Will add that as a specific value then

yoav-lavi: Done

packages/grafbase-sdk/tests/unit/federation.test.ts
```diff
@@ -0,0 +1,42 @@
...
+      federation: {
+        version: '2.3'
+      }
+    })
+
+    expect(renderGraphQL(cfg)).toMatchInlineSnapshot(`
+      "type Post {
+        id: ID!
+      }
+      extend schema @federation(version: "2.3")
```
pimeys: The extensions should maybe happen before type declarations, as we do with other extensions such as the cache.

packages/grafbase-sdk/tests/unit/federation.test.ts
```diff
@@ -0,0 +1,42 @@
...
+      "
+    `)
+  })
+
+  it('does not render federation when disabled', async () => {
```
pimeys, line 28: Is this necessary? We have plenty of tests where we don't have federation set and they should show breakage.

pimeys (approved):

packages/grafbase-sdk/tests/unit/federation.test.ts
yoav-lavi: Fair enough

yoav-lavi: Removed

packages/grafbase-sdk/tests/unit/federation.test.ts
yoav-lavi: Done

PR #792, opened 2023-10-11
feat(sdk): add types for resolver  `context` and `info` args

tomhoule:
# Description

This commit adds two exports to `@grafbase/sdk':

- `Context`: the type of the `context` arg in resolvers.
- `Info`: the type of the `info` arg in resolvers.

The source is, at much as possible, annotated with the source of truth
for each part of the types in the comments.

This commit implements that section of the RFC: https://www.notion.so/grafbase/Typed-Resolvers-b4a0b47f4b5b4d2b9815e0aa82c6a8a8?pvs=4#7e9386db2ac648a28d5fd8db11f9b4b1

Questions:

- In several places, I used `any`, but we may want to use `unknown`
  instead (different DX for sure). Are there opinions? Graphql.JS for
  example uses `Record<string, unknown>` for variables.
- My current plan for testing that these types match reality is to
  release `@grafbase/sdk` once this is merged, then manually check in
  a deployment. Is there a relevant test suite that would also work?
  Or would a local project with the CLI in dev mode give enough
  confidence?

Part of GB-5097

# Type of change

- [ ] 💔 Breaking
- [x] 🚀 Feature
- [ ] 🐛 Fix
- [ ] 🛠️ Tooling
- [ ] 🔨 Refactoring
- [ ] 🧪 Test
- [ ] 📦 Dependency

### DIFF
--- a/packages/grafbase-sdk/src/resolver/context.ts
+++ b/packages/grafbase-sdk/src/resolver/context.ts
@@ -1,0 +1,71 @@
+import { KVListResult, KVMetadata, KVGetOptions, KVSetOptions, KVListOptions } from './context/kv'
+import { Classification } from './context/ai'
+
+/**
+ * The type of the `context` argument in a Grafbase edge resolver.
+ *
+ * Reference: https://grafbase.com/docs/edge-gateway/resolvers#context
+ *
+ * @example
+ *
+ * import { Context, Info } from '@grafbase/sdk'
+ *
+ * export default async function(_parent, _args, context: Context) {
+ *   // ...
+ * }
+ */
+export type ResolverContext = {
+  /** Context about the HTTP request being handled. */
+  request: {
+    headers: Record<string, any>,
+  }
+  /**
+   * Grafbase KV
+   *
+   * If you want to use this, please make sure to [enable KV in your Grafbase configuration](https://grafbase.com/docs/edge-gateway/resolvers#enable-kv).
+   *
+   * See the reference documentation: https://grafbase.com/docs/edge-gateway/resolvers#kv
+   */
+  kv: {
+    /** Retrieve the value and metadata for a key. See [the docs](https://grafbase.com/docs/edge-gateway/resolvers#get) for examples. */
+    get: (key: string, options?: KVGetOptions) => Promise<{ metadata: KVMetadata, value: any }>
+    /** Create a new key-value pair or update an existing one. See [the docs](https://grafbase.com/docs/edge-gateway/resolvers#set) for examples. */
+    set: (key: string, value: any, options?: KVSetOptions) => Promise<void>
+    /** Delete a key and its value. */
+    delete: (key: string) => Promise<void>
+    /** Fetch the list of all keys. See [the docs](https://grafbase.com/docs/edge-gateway/resolvers#list) for examples. */
+    list: (options?: KVListOptions) => Promise<KVListResult>
+  },
+  ai: {
+    textLlm: (args: {
+      model?: string // closed set of possibilities? see common/grafbase-sdk/src/api/ai/models.rs in grafbase/api
+      prompt?: string
+      messages?: { role: 'system' | 'user', content: string }[]
+    }) => Promise<{ response: string }>
+    textClassification: (args: {
+      model?: string
+      text: string
+    }) => Promise<{ response: [Classification, Classification] }>
+    textTranslation: (args: {
+      model?: string
+      text: string
+      /** Defaults to English. */
+      from?: string
+      to: string
+    }) => Promise<{ translated_text: string }>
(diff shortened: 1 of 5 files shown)


packages/grafbase-sdk/src/resolver/context/kv.ts
```diff
@@ -0,0 +1,14 @@
+// No type is enforced by KV, users will get back what they put in.
+export type KVMetadata = any
+
+export type KVListResult = {
+	keys: KVListKey[]
+	list_complete: boolean
+  cursor?: string
+}
```
tomhoule, line 8: TODO formatting here, it's tabs, we can't have that

packages/grafbase-sdk/src/resolver/info.ts
```diff
@@ -0,0 +1,34 @@
...
+    /**  The fields traversed prior to the called resolver. */
+    path: ResolverInfoPath
+    /** The variables defined for the query. */
+    variableValues: { [variableName: string]: any }
+}
+
+/** A field traversed prior to calling a resolver. Also see ResolverInfo. */
+export type ResolverInfoPath = {
+    // definition: engine/crates/engine/src/registry/resolvers/custom.rs and engine/crates/engine/src/query_path.rs
+    /** The name of the field or */
```
tomhoule, line 28: TODO finish that sentence

pimeys:
Maybe you can create a project where you point to this library directly from the filesystem. I hope npm is smart enough to allow that.

pimeys (approved):

tomhoule:
> Maybe you can create a project where you point to this library directly from the filesystem. I hope npm is smart enough to allow that.

I did that to test that the exports and completion work as expected. I'll check if I can do that with a running dev server, but you're right, it should be possible.

packages/grafbase-sdk/src/resolver/context.ts
```diff
@@ -0,0 +1,68 @@
...
+    headers: Record<string, any>,
+  }
+  /**
+   * Grafbase KV
+   *
+   * If you want to use this, please make sure to [enable KV in your Grafbase configuration](https://grafbase.com/docs/edge-gateway/resolvers#enable-kv).
+   *
+   * See the reference documentation: https://grafbase.com/docs/edge-gateway/resolvers#kv
+   */
+  kv: {
```
notrab, line 29: There's also [`delete`](https://grafbase.com/docs/edge-gateway/resolvers#delete)

packages/grafbase-sdk/src/resolver/context.ts
```diff
@@ -0,0 +1,68 @@
...
+    /** Retrieve the value and metadata for a key. See [the docs](https://grafbase.com/docs/edge-gateway/resolvers#get) for examples. */
+    get: (key: string) => Promise<{ metadata: KVMetadata, value: any }>
+    /** Create a new key-value pair or update an existing one. See [the docs](https://grafbase.com/docs/edge-gateway/resolvers#set) for examples. */
+    set: (key: string, value: any, metadata?: KVMetadata) => Promise<void>
```
notrab, line 33: The `metadata` is nested inside this object, this arg also has `ttl` property.

packages/grafbase-sdk/src/resolver/context/kv.ts
```diff
@@ -0,0 +1,18 @@
+/** The metadata for a key in KV. Data of any shape can be stored. */
+export type KVMetadata = any
+
+// https://developers.cloudflare.com/kv/api/list-keys/
+/** The result of a [`list()`](https://grafbase.com/docs/edge-gateway/resolvers#list) call.  */
+export type KVListResult = {
+  keys: KVListKey[]
+  list_complete: boolean
+  /** If `list_complete` is false then you will receive a `cursor` value you can pass to `list()` to obtain the next set of results. */
```
notrab, line 9: Should this comment not be on the line above?

packages/grafbase-sdk/src/resolver/context.ts
```diff
@@ -0,0 +1,68 @@
...
+    /** Fetch the list of all keys. See [the docs](https://grafbase.com/docs/edge-gateway/resolvers#list) for examples. */
+    list: (cursor?: string) => Promise<KVListResult>
```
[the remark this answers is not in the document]
tomhoule: Should be fixed in latest push, thanks!

packages/grafbase-sdk/src/resolver/context.ts
tomhoule: Should be fixed now.

packages/grafbase-sdk/src/resolver/context.ts
tomhoule: Indeed, found the definition for these and fixed that. Thanks!

hfgbarrigas (approved):

PR #794, opened 2023-10-12
feat: federation resolver for resolvable basic types

obmarg:
Sometimes one subgraph will want to add fields to a type that is owned by another subgraph.  This partially enables suppport for that in Grafbase.  e.g. to add a name to a `TodoList` owned by another subgraph you could use this schema:

```graphql
extend schema @federation(version: "2.3")

type TodoList @key(fields: "id") {
    id: ID!
    name: String! @resolver(name: "todoListName")
}
```

### DIFF
--- a/engine/crates/parser-sdl/src/tests.rs
+++ b/engine/crates/parser-sdl/src/tests.rs
@@ -5,21 +5,23 @@ use engine::{registry::Registry, Schema};
 use function_name::named;
 use serde_json as _;
 
-use crate::rules::visitor::RuleError;
-
 macro_rules! assert_validation_error {
     ($schema:literal, $expected_message:literal) => {
         assert_matches!(
-            super::parse_registry($schema)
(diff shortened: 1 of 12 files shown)


engine/crates/parser-sdl/src/tests.rs
```diff
@@ -5,22 +5,24 @@ use engine::{registry::Registry, Schema};
...
                 // It only matters that we find the expected error.
                 .and_then(|errors| errors.into_iter().next()),
-            Some(RuleError { message, .. }) if message == $expected_message
-        );
+            Some(crate::RuleError { message, .. }) => {
+                assert_eq!(message, $expected_message);
+            }
+        )
     };
 }
```
tomhoule, line 22: :100:

tomhoule (approved):
This reads nicely, I think I get the moving parts better since yesterday's review. LGTM!

PR #798, opened 2023-10-12
feat(connector): Connect to Postgres through TCP

pimeys:
# PostgreSQL TCP connectivity

Changes our connection method from Neon HTTP bridge to tokio-postgres and TCP. This has some caveats one should know.

## Forked tokio-postgres

We needed to [fork](https://github.com/grafbase/rust-postgres) tokio-postgres to get it working _correctly_ and _fast_ in a serverless context together with a connection pool such as pgbouncer or Hyperdrive.

A connection pool for serverless Postgres is _by definition_ running in a so called _transaction mode_. This means, that a connection is reserved for you only as long as the current transaction is alive. A single query is also a transaction, and for us it means that every query can and will run in a different connection.

Now, a typical Postgres query is two roundtrips (a bit simplified):

```sql
PREPARE s1 (int) AS SELECT $1;
```

and then

```sql
EXECUTE s1(1);
```

Now, this is two queries, right? And what if the `EXECUTE` runs in a different connection than `PREPARE`? We can of course force them to run in the same connection by spawning a transaction first, but this would mean two more roundtrips for `BEGIN` and `COMMIT` or `ROLLBACK`. Which is... not great.

The other problem is how the naming of the statements is done by the client (tokio-postgres). It thinks it's always having the same connection underneath the surface, so for the first query it names the statement as `s1`, the next one is `s2` and so forth. Now if the pool gives us a different connection for every query, naturally we'll be having collisions with names.

The first patch for tokio-postgres enables us to use a Postgres feature of anonymous (or unnamed) statements. This means that we combine the two queries, prepare first without a name and execute immediately. We will lose a bit of performance for having to analyze and create a statement every time even for the same query, but the statement is not kept in the connection and deleted immediately. This solves the name clashes.

The second set of fixes is to simplify Postgres type handling. Now we have this first preparing roundtrip, which returns us what types it expects us to use when we call the query with parameters. There is another mode of querying with text-only mode. In this case Postgres takes every single parameter as strings, and returns us the values as strings. What we need to do then is to stringify the user input, and parse the result set what comes back for us.

If we do this, we can _completely skip the preparing roundtrip_. So, we use a bit more CPU and RAM, but we save 5-100 milliseconds from a roundtrip, which is valuable on a serverless context where we can't store statements into memory and reuse them, as we can in on-prem installations.

These two patches are combined from Cloudflare and Neon into our own branch. It is quite possible they will never be merged to upstream.

## Pooling

This connector should _never be used without a connection pool_. If running it against a Postgres instance that's not backed by Hyperdrive or pgbouncer, it will quite fast reserve every single connection for itself and make the database admin quite unhappy.

When we install this to a customer that's running it on-prem, we have to design our resolver to keep the connections to databases alive, and we must add an internal connection pool for Postgres to enable better parallel execution. For our current edge use-cases this pull request is enough and should do fine.

## Testing

This has _not yet_ been tested on production. There are two initialization functions for the new `TcpTransport` type: one for native client such as the CLI, using tokio, tokio-postgres and rustls, and another for WASM that uses the Cloudflare socket, compiles tokio-postgres without tokio and initializes it in a quite different way. The second branch compiles, but I don't know yet how well it works.

Next week we'll be testing it together with @hfgbarrigas by deploying it to production, together with Neon database and Hyperdrive pool. And maybe writing one e2e test too.

Closes: GB-4683


engine/crates/postgres-types/src/transport/tcp.rs
```diff
@@ -0,0 +1,147 @@
...
+        config.dbname(dbname);
+
+        let hostname = url.host_str().ok_or_else(|| {
+            crate::error::Error::InvalidConnectionString(String::from(
+                "the connection string does not define a valid hostname",
+            ))
+        })?;
+
+        let socket = worker::Socket::builder()
+            .secure_transport(worker::SecureTransport::On)
```
pimeys, line 77: https://docs.rs/worker/0.0.18/worker/enum.SecureTransport.html

Finistere:
> The other problem is how the naming of the statements is done by the client (tokio-postgres). It thinks it's always having the same connection underneath the surface, so for the first query it names the statement as s1, the next one is s2 and so forth. Now if the pool gives us a different connection for every query, naturally we'll be having collisions with names.

Stupid question, couldn't we use the sha256 hash of the query for the statement name? (solving the collision problem)

> The second set of fixes is to simplify Postgres type handling. Now we have this first preparing roundtrip, which returns us what types it expects us to use when we call the query with parameters. There is another mode of querying with text-only mode.

Just to understand it correctly. With anonymous statements, do you still have a preparing roundtrip?

Finistere (approved):

PR #807, opened 2023-10-13
examples: add a KV + typed resolver and authorizer example

tomhoule:
I haven't found existing examples (in `/examples`) using KV, so this seemed like an opportunity. The original purpose of this PR is to have a working example of typed resolvers and authorizers with the recently added types from SDK, so we can link to it in the changelog.

It is very very basic and not much of a functioning app, feedback welcome.

### DIFF
--- a/packages/grafbase-sdk/src/authorizer/context.ts
+++ b/packages/grafbase-sdk/src/authorizer/context.ts
@@ -7,7 +7,7 @@
  *
  * import { AuthorizerContext, VerifiedIdentity } from '@grafbase/sdk'
  *
- * export default async ({ request }: AuthorizerContext): VerifiedIdentity? {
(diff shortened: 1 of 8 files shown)


packages/grafbase-sdk/src/authorizer/context.ts
```diff
@@ -7,7 +7,7 @@
  *
  * import { AuthorizerContext, VerifiedIdentity } from '@grafbase/sdk'
  *
- * export default async ({ request }: AuthorizerContext): VerifiedIdentity? {
+ * export default async function({ request }: AuthorizerContext): VerifiedIdentity? {
```
tomhoule, line 10: Wrong syntax.

Finistere (approved):

PR #810, opened 2023-10-13
feat: pull the current SDK version from its package.json

obmarg:
I released 0.7.0 of the SDK earlier, but forgot to update the `GRAFBASE_SDK_PACKAGE_VERSION` constant that the CLI uses to create package.json files.

This PR automates that process - instead of having a hardcoded constant we now read the package.json and populate the constant with whatever is in there.


pimeys (approved):
nice.

fbjork (approved):

tomhoule (commented):
Nice!

PR #813, opened 2023-10-14
Dockerfile

fbjork:
# Description

Adds Dockerfile to enable running Grafbase in a container.

Currently the Dockerfile lives in the root folder because the cli and engine live at the same level. The engine folder should move to the cli folder along with the Dockerfile eventually.

To run the container locally with a grafbase folder containing the schema:
```
docker run --name grafbase -v "./grafbase:/grafbase" grafbase
```

TODO: ensure resolvers work.

# Type of change

- [ ] 💔 Breaking
- [ ] 🚀 Feature
- [ ] 🐛 Fix
- [ ] 🛠️ Tooling
- [ ] 🔨 Refactoring
- [ ] 🧪 Test
- [ ] 📦 Dependency

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -19,6 +19,9 @@ WORKDIR /grafbase
 
 RUN apk add --no-cache nodejs npm
 
+RUN adduser -g wheel -D grafbase -h "/data" && mkdir -p /data && chown grafbase: /data
+USER grafbase
+
 COPY --from=build /grafbase/cli/target/release/grafbase /bin/grafbase
 
 ENTRYPOINT ["/bin/grafbase"]
@@ -26,3 +29,6 @@ ENTRYPOINT ["/bin/grafbase"]
 CMD ["start"]
 
 EXPOSE 4000
+
+VOLUME ["/data"]
+WORKDIR "/data"
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,23 @@
+# Build
+FROM rust:1.73 AS build
+
+WORKDIR /grafbase
+
+COPY ./cli ./cli
+COPY ./engine ./engine
+
+COPY ./cli/Cargo.lock ./cli/Cargo.lock
+COPY ./cli/Cargo.toml ./cli/Cargo.toml
+COPY ./engine/Cargo.lock ./engine/Cargo.lock
+COPY ./engine/Cargo.toml ./engine/Cargo.toml
+
+WORKDIR /grafbase/cli
+
+RUN cargo build --release
+
+# Run
+FROM debian:bullseye AS run
+
+COPY --from=build /grafbase/cli/target/release/grafbase .
+
+CMD ["./grafbase", "start"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -16,7 +16,9 @@ WORKDIR /grafbase/cli
 RUN cargo build --release
 
 # Run
-FROM debian:bullseye AS run
+FROM debian:bookworm AS run
+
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -0,0 +1,29 @@
...
+COPY ./cli/Cargo.toml ./cli/Cargo.toml
+COPY ./engine/Cargo.lock ./engine/Cargo.lock
+COPY ./engine/Cargo.toml ./engine/Cargo.toml
+
+WORKDIR /grafbase/cli
+
+RUN cargo build --release
+
+# Run
+FROM debian:bookworm AS run
```
hfgbarrigas, line 19: try using one of the following and see how it goes:

- distroless images
- alpine images
- if you're having fun, try scratch images

should definitely decrease the size

Dockerfile
```diff
@@ -0,0 +1,29 @@
+# Build
+FROM rust:1.73 AS build
+
+WORKDIR /grafbase
+
+COPY ./cli ./cli
+COPY ./engine ./engine
+
+COPY ./cli/Cargo.lock ./cli/Cargo.lock
```
Finistere, line 9: Unless I'm missing something, copying the cargo files after having copied the whole engine and cli directories has no use. Any change to the cli cargo files will make docker start from the first `COPY ./cli ./cli` anyway.

fbjork: Yeah, removing for now. The idea is to cache the dependencies layer, but we can come back to that later.

Dockerfile
obmarg: We're not statically linking (unless I've missed something) so probably limits to how far away from debian we can go before we start running into problems

Dockerfile
```diff
@@ -0,0 +1,28 @@
+# Build
+FROM rust:1.73-alpine3.18 AS build
```
mettke, line 2: I recommend using image hashes instead of versions. Otherwise a future push to the image may break the build. Renovate should take care of updates and we should have a build pipeline to validate that building and running works

Dockerfile
```diff
@@ -0,0 +1,28 @@
...
+# Run
+FROM alpine:3.18
+
+WORKDIR /grafbase
+
+RUN apk add --no-cache nodejs npm
+
+COPY --from=build /grafbase/cli/target/release/grafbase /bin/grafbase
+
+ENTRYPOINT ["/bin/grafbase"]
```
mettke, line 24: We should not run as root. Instead we should create a user under which the cli runs.

Dockerfile
```diff
@@ -0,0 +1,28 @@
...
+
+CMD ["start"]
+
```
mettke, line 27: We also should define a volume and use it as workdir. This way its possible to run the container with read only root filesystem while still being able to access and change files inside the volume

Dockerfile
fbjork: Can you suggest the change?

Dockerfile
fbjork: Can you suggest the change?

Dockerfile
fbjork: Example?

Dockerfile
mettke:
```
VOLUME [ "/data" ]
WORKDIR "/data"
```

Dockerfile
fbjork: it seems a bit extreme give we use a widely used image, so I will hold off on this for now.

obmarg (approved):

PR #818, opened 2023-10-16
fix: Fix grafbase dev with TS config

Finistere:
Somehow a change wasn't added to the last PR and the test for the TS config wasn't really testing that it worked.

### DIFF
--- a/cli/crates/server/src/servers.rs
+++ b/cli/crates/server/src/servers.rs
@@ -105,6 +105,8 @@ pub async fn start(
 ) -> Result<(), ServerError> {
     let project = Project::get();
 
+    // Exporting Pathfinder, TS parser & miniflare for resolvers.
+    export_embedded_files()?;
     create_project_dot_grafbase_directory()?;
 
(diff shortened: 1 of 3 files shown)


tomhoule (commented):
Question for me to understand the CLI better, rather than feedback: how/where are the embedded files exported when the config is graphql? I think e.g. pathfinder is in there.

cli/crates/server/src/servers.rs
```diff
@@ -105,6 +105,8 @@ pub async fn start(
 ) -> Result<(), ServerError> {
     let project = Project::get();
 
+    // Exporting Pathfinder, TS parser & miniflare for resolvers.
+    export_embedded_files()?;
```
yoav-lavi, line 109: @Finistere IIRC we said we'd check if TS config is in use first no?

Finistere: we need Pathfinder in all cases

yoav-lavi (approved):

PR #819, opened 2023-10-16
feat(cli): include Pathfinder here instead of through CDN assets

tomhoule:
# Description

The CLI includes [Pathfinder](https://github.com/grafbase/pathfinder) as part of the `grafbase dev` workflow. A bundled build of Pathfinder is included in the assets file (`assets.tar.gz`) that is built locally when we iterate in this repo, and downloaded from a canonical source for CLI releases. The assets are included in the compiled CLI binaries, then decompressed to `~/.grafbase` when we start a server.

Before this commit, a bundled build of Pathfinder is included in the assets archive that is built and downloaded separately from this repo.

What this commit changes:

- the Pathfinder wrapper app (the `cli-app` package) moves from the website repo to this repo, under `packages/`.
- the Pathfinder build included in the CLI binary is built from source in this repo. **This makes updating Pathfinder much easier: we just bump the version in packages/cli-app/packages.json**.

This commit does not change how the assets archive that is included in the CLI is handled. It only adds the local build of pathfinder to that archive when the `server` crate is built, via a cargo build script.

The build script expects to find a built version of cli-app, either in the default location or at a path specified via `GRAFBASE_CLI_PATHFINDER_BUNDLE_PATH`. It will not try to build it itself — that can be done, but that felt like too much magic. On failure, it points to a relevant section in the README that explains why the error happens and how to build cli-app to remediate.

This approach lets us progressively move the contents and construction logic for `assets.tar.gz` from other repos to this repo, without changing how the assets are consumed.

Upsides:

- Incremental migration away from existing assets bundling
- Avoids introducing new include_bytes!() with their own unpacking logic. Keeps a single entry point for these files.
- Easy to iterate on locally (just iterate on and build cli-app)
- Easy to upgrade pathfinder (upgrade through cli-app package.json and package lock).

Downsides:

- The build scripts unpacks `assets.tar.gz`, adds pathfinder, then rebuilds the archive. We don't have the final built assets in an immutable location anymore. 
- Requires a JS project build to build the CLI
  - In development, you only need to build once.
  - The build process is documented and the error carefully points you to how to build.
  - If we want to move away from downloaded, pre-bundled assets that are built elsewhere, we will need to build the assets in this repo.
- Does more work at build time, potentially making builds slower.
  - The build script will only re-run if the bundle to include changed, it won't make iteration slower.

closes GB-4984

# Type of change

- [ ] 💔 Breaking
- [ ] 🚀 Feature
- [ ] 🐛 Fix
- [ ] 🛠️ Tooling
- [ ] 🔨 Refactoring
- [ ] 🧪 Test
- [ ] 📦 Dependency

### DIFF
--- a/.github/actions/cdn_assets/action.yml
+++ b/.github/actions/cdn_assets/action.yml
@@ -21,15 +21,15 @@ runs:
       working-directory: ${{ inputs.working-directory }}
       run: |
         curl https://${{ inputs.cdn }}/cli/${{ inputs.assets_version }}.tar.gz --output assets.tar.gz --fail
-        curl https://${{ inputs.cdn }}/cli/pathfinder/static.tgz --output static.tgz --fail
-        tar -zxvf static.tgz
-        rm static.tgz
-        gunzip assets.tar.gz
-        tar -rf assets.tar static
-        rm -r static
-        gzip assets.tar
         cp assets.tar.gz crates/server/assets
         mkdir ~/.grafbase
         cp assets.tar.gz ~/.grafbase
         (cd ~/.grafbase && tar -xf assets.tar.gz)
         touch ~/.grafbase/version.txt
+
+    - name: Build cli-app
+      shell: bash
+      working-directory: './packages/cli-app'
+      run: |
+        npx pnpm i
+        npx pnpm run cli-app:build
--- a/.github/actions/cli_assets/action.yml
+++ b/.github/actions/cli_assets/action.yml
@@ -1,5 +1,5 @@
-name: Download CLI assets
-description: Download CLI assets from Grafbase CDN
+name: Download and build CLI assets
+description: Download and build CLI assets from Grafbase CDN and from source
 
 inputs:
   working-directory:
--- a/.github/workflows/cli-partial-release.yml
+++ b/.github/workflows/cli-partial-release.yml
@@ -54,8 +54,8 @@ jobs:
         with:
           cmd: yq '.env.ASSETS_VERSION' './.github/workflows/cli.yml'
 
-      - name: Fetch CDN assets
-        uses: ./.github/actions/cdn_assets
+      - name: Build CLI assets
+        uses: ./.github/actions/cli_assets
         with:
           cdn: ${{ steps.lookup-prod-assets.outputs.result }}
           assets_version: ${{ steps.lookup-assets-version.outputs.result }}
--- a/.github/workflows/cli.yml
+++ b/.github/workflows/cli.yml
@@ -84,7 +84,7 @@ jobs:
           restore-key: ${{ runner.os }}-${{ runner.arch }}-cargo-dev
 
       - name: Prepare CLI assets
-        uses: ./.github/actions/cdn_assets
+        uses: ./.github/actions/cli_assets
         with:
           cdn: ${{ env.PROD_ASSETS }}
           assets_version: ${{ env.ASSETS_VERSION }}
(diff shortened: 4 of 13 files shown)


packages/cli-app/tsconfig.json
```diff
@@ -0,0 +1,8 @@
+{
+  "extends": "../tsconfig/base.json",
```
jonathanawesome, line 2: This file doesn't exist in this repo. Here are the original contents:

```js
{
  "compilerOptions": {
    "target": "es2020",
    "lib": ["dom", "dom.iterable", "esnext"],
    "noImplicitAny": false,
    "allowJs": true,
    "skipLibCheck": true,
    "strict": true,
    "forceConsistentCasingInFileNames": false,
    "noEmit": true,
    "esModuleInterop": true,
    "module": "esnext",
    "moduleResolution": "node",
    "resolveJsonModule": true,
    "isolatedModules": true,
    "jsx": "preserve",
    "incremental": true,
    "baseUrl": "./src"
  },
  "exclude": ["node_modules"]
}
```

tomhoule: Thanks! I haven't really put work into cli-app in this branch, I was thinking of importing (and rename it if necessary) it in a first PR to keep the commit history (https://github.com/grafbase/grafbase/pull/820).

.github/actions/cdn_assets/action.yml
```diff
@@ -21,15 +21,15 @@ runs:
...
         cp assets.tar.gz ~/.grafbase
         (cd ~/.grafbase && tar -xf assets.tar.gz)
         touch ~/.grafbase/version.txt
+
+    - name: Build cli-app
+      shell: bash
+      working-directory: './packages/cli-app'
+      run: |
+        npx pnpm i
+        npx pnpm run cli-app:build
```
tomhoule, line 35: Should we rename the action?

obmarg: Yes (or do the CLI build outside of this)

tomhoule: done in [3048de1](https://github.com/grafbase/grafbase/pull/819/commits/3048de138b1dba9e1ed9d8c40c82e7a1098afe78)

.github/actions/cli_assets/action.yml
```diff
@@ -21,15 +21,15 @@ runs:
...
-        gzip assets.tar
         cp assets.tar.gz crates/server/assets
         mkdir ~/.grafbase
         cp assets.tar.gz ~/.grafbase
         (cd ~/.grafbase && tar -xf assets.tar.gz)
         touch ~/.grafbase/version.txt
+
+    - name: Build cli-app
+      shell: bash
+      working-directory: './packages/cli-app'
```
obmarg, line 32: Not really a comment on this PR but  `cli-app` is quite a misleading package name now the code doesn't live in the pathfinder repo

fbjork: perhaps cli-pathfinder?

obmarg (approved):

.github/actions/cli_assets/action.yml
jonathanawesome: but, it _is_ the cli app. does `cli-ui` make folks more comfortable?

obmarg: > but, it is the cli app

We have at least two things in this repo that could be described as "the CLI app"

PR #828, opened 2023-10-17
fix: allow openapi connector to have un-prefixed types.

obmarg:
Prior to this fix it was awkward to get un-prefixed types out of the OpenAPI connector.  You could specify `name: "", namespace: false` to get it to work, but #827 is about to make empty names impossible so there'd be no way to do this.

Since we've recently added federation subgraph support to the connector this is more important than it might otherwise be.  In a federated graph we'd really want to refer to a `User` type and not a `UsersApiUser` type - the connector name prefix gets in the way of making a nice subgraph.

I've fixed this by:

- Tying the default prefix to the `namespace: bool` setting.  If this is false, types will not be prefixed.  This is what the GraphQL connector was doing already so it seems sensible.
- Added a `prefixTypes: string` setting to the OpenAPI & GraphQL connectors that users can use to override the default prefix.  This gives users a way of getting the old behaviour back should they want to (and giving users a way to disable the prefix and keep namespacing if for some reason that's their want).

Fixes GB-5117
Fixes GB-5136


engine/crates/parser-graphql/src/metadata.rs
```diff
@@ -0,0 +1,11 @@
+use engine::registry::ConnectorHeaders;
+use url::Url;
+
+pub struct GraphQlConnectorMetadata<'a> {
+    pub(crate) name: &'a str,
+    pub(crate) namespace: bool,
```
tomhoule, line 6: The explanation from your PR description would be useful on this field. This struct is about the connector arguments if I understand this right, so maybe this is already documented elsewhere, then ignore the suggestion.

tomhoule (approved):
Nice!

engine/crates/parser-graphql/src/metadata.rs
tomhoule: Oh ok, I didn't notice it wasn't used.

PR #832, opened 2023-10-17
feat: A bunch of improvements for grafbase start and build command

Finistere:
- Only a single instance of miniflare is used, shared across workers.
- UDFs are built in parallel
- Adding a very simple cache mechanism, avoiding recompiling UDFs
  if nothing changed.

I'm also adding a `build` command which does all the UDF compilation in a separate step. It's only helpful for the `start` command currently, allowing one to `grafbase build` during a docker image build.

Resolves: GB-5109,GB-5110

# Description

# Type of change

- [ ] 💔 Breaking
- [ ] 🚀 Feature
- [ ] 🐛 Fix
- [ ] 🛠️ Tooling
- [ ] 🔨 Refactoring
- [ ] 🧪 Test
- [ ] 📦 Dependency

### DIFF
--- a/cli/crates/cli/src/cli_input.rs
+++ b/cli/crates/cli/src/cli_input.rs
@@ -7,18 +7,20 @@ use common::{
 };
 use std::{
     net::{IpAddr, Ipv4Addr},
+    num::NonZeroUsize,
     path::PathBuf,
 };
 use ulid::Ulid;
 
 const DEFAULT_PORT: u16 = 4000;
 
-#[derive(Clone, Copy, Debug, PartialEq, PartialOrd, serde::Deserialize, clap::ValueEnum)]
+#[derive(Default, Clone, Copy, Debug, PartialEq, PartialOrd, serde::Deserialize, clap::ValueEnum)]
 #[clap(rename_all = "snake_case")]
 pub enum LogLevelFilter {
     None,
     Error,
     Warn,
+    #[default]
     Info,
     Debug,
 }
@@ -36,7 +38,7 @@ impl LogLevelFilter {
     }
 }
 
-#[derive(Clone, Copy)]
+#[derive(Default, Clone, Copy)]
 pub struct LogLevelFilters {
     pub functions: LogLevelFilter,
     pub graphql_operations: LogLevelFilter,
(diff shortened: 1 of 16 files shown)


cli/crates/cli/src/cli_input.rs
```diff
@@ -218,6 +220,24 @@ pub struct LogsCommand {
...
+#[derive(Debug, clap::Args)]
+pub struct BuildCommand {
+    /// Number of resolver builds running in parallel
+    #[arg(long)]
+    pub parallelism: Option<u16>,
+}
+
+impl BuildCommand {
+    pub fn parallelism(&self) -> NonZeroUsize {
+        let p = self.parallelism.unwrap_or(0);
```
yoav-lavi, line 232: I'd avoid single letter variables

cli/crates/cli/src/cli_input.rs
```diff
@@ -218,6 +220,24 @@ pub struct LogsCommand {
...
+        if p == 0 {
+            std::thread::available_parallelism().unwrap_or(NonZeroUsize::new(1).unwrap())
```
yoav-lavi, line 234: Could we use expect instead of unwrap?

cli/crates/cli/src/cli_input.rs
Finistere: fixed

cli/crates/server/src/servers.rs
```diff
@@ -28,58 +31,99 @@ use std::{fs, process::Stdio};
...
+
+            let mut hasher = sha2::Sha256::new();
+
+            for entry in walkdir::WalkDir::new(&project.grafbase_directory_path)
+                .sort_by_file_name()
+                .follow_links(true)
+                .into_iter()
+                .filter_map(Result::ok)
+                // Only path we can somewhat safely ignore is the schema one
+                .filter(|e| !e.file_type().is_dir() && e.path() != project.schema_path.path())
```
yoav-lavi, line 76: Same here

Finistere: fixed

pimeys (approved):

cli/crates/cli/src/start.rs
```diff
@@ -21,79 +23,43 @@ impl MessageGroup {
...
+pub(crate) fn run<F>(
+    log_level_filters: LogLevelFilters,
+    build: impl FnOnce(UnboundedSender<ServerMessage>) -> F,
+) -> Result<(), CliError>
+where
+    F: Future<Output = Result<(), ServerError>> + Send + 'static,
+{
+    let (message_sender, message_receiver) = tokio::sync::mpsc::unbounded_channel::<ServerMessage>();
+    let action = build(message_sender);
+    let log_repoter = log_reporter(message_receiver, log_level_filters);
```
tomhoule, line 53: Tiny typo: let log_reporter

tomhoule (approved):
No meaningful feedback from me, other than I read the PR, and it all makes sense :100:

cli/crates/cli/src/start.rs
Finistere: thanks, fixed!

PR #835, opened 2023-10-18
feat(cli): implement typescript code generation for resolvers

tomhoule:
# Description

This PR is large, but mostly because of tests.

First iteration of the implementation of the GraphQL SDL -> TypeScript
resolver types part of Typed Resolvers.

This is a new standalone crate in `cli`. It does not reuse existing
infrastructure because the amount of analysis required on the schema is
very limited: we do not want to do extra validation because we want to
tolerate fairly broken schemas and still generate code,and code
generation has to be fast for integration in the `gb dev` loop.

Unlike graphql-code-generator, we do not generate types for the resolver
functions. By design, the generated module has one TypeScript type per
GraphQL type. They have the same name (except TS reserved keywords), and
they are in the same order in the generated file as in the source file.

We do not want types for the resolver functions because we are going to
try using the resolvers as the source of truth (RFC section pending).
Resolver function types can easily be added later if that approach does
not pan out.

Not implemented yet:

- Custom scalar types. They are typed as `any`. We probably want to have
  a configurable location for the generated code to import them from.
- More subtle interface typing. The current approach is very naive, it
  only includes the interface fields.

closes GB-5146

# Type of change

- [ ] 💔 Breaking
- [x] 🚀 Feature
- [ ] 🐛 Fix
- [ ] 🛠️ Tooling
- [ ] 🔨 Refactoring
- [ ] 🧪 Test
- [ ] 📦 Dependency

### DIFF
--- a/cli/crates/typed-resolvers/src/analyze.rs
+++ b/cli/crates/typed-resolvers/src/analyze.rs
@@ -1,0 +1,444 @@
+use graphql_parser::schema as ast;
+use std::{collections::HashMap, ops, str::FromStr};
+
+pub(crate) fn analyze<'doc>(graphql_document: &'doc ast::Document<'doc, &'doc str>) -> AnalyzedSchema<'doc> {
+    let mut schema = AnalyzedSchema::default();
+
+    analyze_top_level(graphql_document, &mut schema);
+    analyze_fields(graphql_document, &mut schema);
+
+    schema
+}
+
+/// First pass. Resolve what definitions exist in the schema (objects, unions, etc.).
+fn analyze_top_level<'doc>(graphql_document: &'doc ast::Document<'doc, &'doc str>, schema: &mut AnalyzedSchema<'doc>) {
+    for definition in &graphql_document.definitions {
+        match definition {
+            ast::Definition::DirectiveDefinition(_) | ast::Definition::SchemaDefinition(_) => (), // not interested
+            ast::Definition::TypeDefinition(ast::TypeDefinition::Object(output_type)) => {
+                schema.push_output_type(Object {
+                    name: output_type.name,
+                    docs: output_type.description.as_deref(),
+                    kind: ObjectKind::Object,
+                });
+            }
+            ast::Definition::TypeDefinition(ast::TypeDefinition::InputObject(object_definition)) => {
+                schema.push_output_type(Object {
+                    name: object_definition.name,
+                    docs: object_definition.description.as_deref(),
+                    kind: ObjectKind::InputObject,
+                });
+            }
+            ast::Definition::TypeDefinition(ast::TypeDefinition::Interface(object_definition)) => {
+                schema.push_output_type(Object {
+                    name: object_definition.name,
+                    docs: object_definition.description.as_deref(),
+                    kind: ObjectKind::Interface,
+                });
+            }
+            ast::Definition::TypeDefinition(ast::TypeDefinition::Scalar(scalar)) => {
+                schema.push_custom_scalar(CustomScalar {
+                    name: scalar.name,
+                    docs: scalar.description.as_deref(),
+                });
+            }
+            ast::Definition::TypeDefinition(ast::TypeDefinition::Union(union_definition)) => {
+                schema.push_union(Union {
+                    name: union_definition.name,
+                });
+            }
+            ast::Definition::TypeDefinition(ast::TypeDefinition::Enum(enum_definition)) => {
+                let id = schema.push_enum(Enum {
+                    name: enum_definition.name,
+                    docs: enum_definition.description.as_deref(),
+                });
+
+                for variant in &enum_definition.values {
+                    schema.push_enum_variant(id, variant.name);
+                }
+            }
+            ast::Definition::TypeExtension(_) => {
+                // We ignore type extensions in the first pass because we can't know if they extend
+                // types defined in the document yet.
+            }
+        }
+    }
+}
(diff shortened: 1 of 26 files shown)


cli/crates/typed-resolvers/src/analyze.rs
tomhoule, line 1: This module may be a tad too large, split it?

cli/crates/typed-resolver-tests/build.rs
```diff
@@ -0,0 +1,55 @@
...
+    ffi::OsStr,
+    fs,
+    io::{self, Write as _},
+    path, sync,
+};
+
+const PARSE_TESTS_ROOT_DIR: &str = "tests/schema_types";
+const CARGO_MANIFEST_DIR: &str = env!("CARGO_MANIFEST_DIR");
+
+fn main() -> io::Result<()> {
```
tomhoule, line 12: Document what the build script does at a very high level in README.

cli/crates/typed-resolver-tests/tests/schema_types/interfaces.expected.ts
```diff
@@ -0,0 +1,22 @@
...
+};
+
+export type Dog = {
+  __typename?: 'Dog'
+  id: string
+  name: string
+};
+
+export type LivingThing = {
+  metabolicRate: number
```
tomhoule, line 15: Is this right? Should we have `__typename?: 'Dog' | 'Human'` here? It would be easy to implement. graphql-code-generator doesn't, because I assume they assume `__resolveType` as the mechanism for interfaces / unions typename.

cli/crates/typed-resolvers/src/codegen.rs
```diff
@@ -0,0 +1,149 @@
...
+                let object_type = &schema[*id];
+                let graphql_type_name = object_type.name;
+                let ts_type_name = safe_ts_type_name(graphql_type_name);
+                let is_input_object = matches!(object_type.kind, ObjectKind::InputObject);
+
+                maybe_docs(out, object_type.docs, "")?;
+                writeln!(out, "export type {ts_type_name} = {{")?;
+
+                if let ObjectKind::Object = object_type.kind {
+                    writeln!(out, "{INDENT}__typename?: '{graphql_type_name}';")?;
```
obmarg, line 22: By the looks of things you might not need it I've got an indenting `Write` impl here you could steal: https://github.com/obmarg/cynic/blob/main/cynic/src/queries/indent.rs#L3-L39

cli/crates/typed-resolver-tests/build.rs
```diff
@@ -0,0 +1,58 @@
...
+        let test_name = case.file_stem().unwrap().to_str().unwrap();
+        let case_path = case.to_str().unwrap();
+        writeln!(
+            out_file,
+            r###"#[test] fn {test_name}() {{ run_test(r##"{case_path}"##) }}"###
+        )?;
+    }
+
+    Ok(())
+}
```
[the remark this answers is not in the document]
tomhoule: I'm not familiar with that crate, looking into it. If we can avoid implementing this ourselves, it's a win.

cli/crates/typed-resolver-tests/Cargo.toml
```diff
@@ -0,0 +1,12 @@
+[package]
+name = "typed-resolver-tests"
+version.workspace = true
+edition.workspace = true
+license.workspace = true
+homepage.workspace = true
+keywords.workspace = true
+repository.workspace = true
```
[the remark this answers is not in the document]
tomhoule: Definitely! Thanks

cli/crates/typed-resolvers/src/codegen.rs
tomhoule: So `indented(format!("__typename?: '{graphql_type_name}')`? I don't have a strong opinion. In this file there's only ever one level of indentation, so I felt fine without a Display wrapper, but whatever is most readable is good.

cli/crates/typed-resolver-tests/build.rs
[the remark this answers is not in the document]
tomhoule: I've written a bunch of these build scripts so I'm a bit numb to the pain by now.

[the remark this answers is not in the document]
obmarg: Yeah, same - but again, just a thought.  If it makes no sense then ignore me

cli/crates/typed-resolvers/src/codegen.rs
obmarg: (or in some cases:

```
let out = indented(out);
writeln!(out, "whatever");
writeln!(out, "whatever");
```

But anyway, feel free to ignore this comment, may not be worthwhile

pimeys (approved):
looks good, go ahead and continue the work.

cli/crates/typed-resolver-tests/build.rs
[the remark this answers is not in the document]
tomhoule: I wasn't aware of this library so I'm taking the occasion to try it now, will come back to this thread with my impressions.

cli/crates/typed-resolvers/src/codegen.rs
tomhoule: (discussed on slack: tried it, adopted it, it's a nice win in terms of code organization and size, thanks!)

PR #838, opened 2023-10-19
fix: proper FieldSet parsing

obmarg:
Federation makes quite a lot of use of `FieldSet`s - a simplified GraphQL string without arguments, fragments, spreads etc.

For the sake of speed I added a naive parser of these in a previous PR.  But I'm in the middle of implementing `@requires` which will likely run into the limitations of my previous approach.

So, this PR adds some more thorough parsing.

### DIFF
--- a/engine/crates/parser-openapi/src/tests/snapshots/parser_openapi__tests__petstore_output-2.snap
+++ b/engine/crates/parser-openapi/src/tests/snapshots/parser_openapi__tests__petstore_output-2.snap
@@ -2166,12 +2166,14 @@ Registry {
         "PetstoreUser": FederationEntity {
             keys: [
                 FederationKey {
-                    selections: [
-                        Selection {
-                            field: "username",
-                            selections: [],
-                        },
-                    ],
+                    selections: FieldSet(
+                        [
+                            Selection {
+                                field: "username",
+                                selections: [],
+                            },
+                        ],
+                    ),
                     resolver: Some(
                         Http(
                             HttpResolver {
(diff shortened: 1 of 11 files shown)


engine/crates/parser-openapi/src/tests/snapshots/parser_openapi__tests__petstore_output-2.snap
```diff
@@ -2166,12 +2166,14 @@ Registry {
...
-                        },
-                    ],
+                    selections: FieldSet(
+                        [
+                            Selection {
+                                field: "username",
+                                selections: [],
+                            },
+                        ],
+                    ),
```
tomhoule, line 2176: Is the registry the data structure we serialize and retrieve from KV that we have to keep backwards compatible? Federation support is very new so breaking changes would be ok here, just checking my understanding.

engine/crates/parser-sdl/src/rules/federation/field_set.rs
```diff
@@ -0,0 +1,143 @@
+use std::str::FromStr;
+
+use engine::registry::{self, field_set::Selection};
+use nom::{
```
tomhoule, line 4: nice to know that we have nom already in there to reach out when we need it

tomhoule (approved):
Very nice!

engine/crates/parser-sdl/src/rules/federation/field_set.rs
obmarg: Yeah.  I think we have `nom`, `combine` _and_ `pest` used in various bits of our codebase.  Possibly excessive having all three

engine/crates/parser-openapi/src/tests/snapshots/parser_openapi__tests__petstore_output-2.snap
tomhoule: Ah yes, I misread that, I thought something became an array that was not, but it's just indentation that looks like that. Yep the newtype should be transparent.

PR #839, opened 2023-10-19
feat(pg): adjust PostgresContext to be generic over transport

hfgbarrigas:
# Description

Adjusts the pg connector to be a bit more flexible on the transport to be used at runtime

# Type of change

- [ ] 💔 Breaking
- [x] 🚀 Feature
- [ ] 🐛 Fix
- [ ] 🛠️ Tooling
- [ ] 🔨 Refactoring
- [ ] 🧪 Test
- [ ] 📦 Dependency


engine/crates/engine/src/registry/resolvers/postgres/request/create_many.rs
```diff
@@ -1,14 +1,20 @@
...
     let (sql, params) = renderer::Postgres::build(query::insert::build(&ctx, input)?);
 
     if ctx.mutation_is_returning() {
-        let operation = ctx.transport().parameterized_query::<RowData>(&sql, params);
+        let operation = ctx.transport().parameterized_query(&sql, params).map_ok(|values| {
+            values
+                .into_iter()
+                .map(|value| serde_json::from_value::<RowData>(value).expect("should deserialize to RowData"))
+                .collect()
+        });
```
obmarg, line 17: Could you use that map_result function here?

engine/crates/integration-tests/src/engine/builder.rs
```diff
@@ -96,6 +103,10 @@ impl EngineBuilder {
             schema_builder = enable_local_dynamo(schema_builder).await;
         }
 
+        if self.local_postgres {
+            schema_builder = schema_builder.data(runtime_local::LocalPgTransportFactory::runtime_factory());
+        }
```
obmarg, line 108: Do you think it's worth just having local_postgres on all the time?

Dynamo has a specific `with_local_dynamo` function because we need to setup dynamo tables for the tests where we're using dynamo and I didn't want to do that on every single test.  But looks like adding this to the context doesn't have much of a cost so we could just do it all the time?

Dunno, up to you

engine/crates/postgres-types/src/transport.rs
```diff
@@ -31,23 +31,25 @@ impl Column {
...
 #[cfg_attr(target_arch = "wasm32", async_trait(?Send))]
 #[cfg_attr(not(target_arch = "wasm32"), async_trait)]
-pub trait Transport {
+pub trait Transport: Send + Sync {
     async fn parameterized_execute(&self, query: &str, params: Vec<Value>) -> crate::Result<i64>;
 
-    async fn parameterized_query<T>(&self, query: &str, params: Vec<Value>) -> crate::Result<Vec<T>>
-    where
-        T: DeserializeOwned + Send;
+    async fn parameterized_query(&self, query: &str, params: Vec<Value>) -> crate::Result<Vec<Value>>;
```
obmarg, line 37: I think the map_result function works fine - but you could also have split this trait in two like I did with this `Context` trait here: https://github.com/grafbase/grafbase/blob/main/engine/crates/engine/src/context/ext.rs#L35

You put the `parameterized_execute` function in a `Transport` trait.  That trait will be dyn safe, then you put all the non dyn safe functions on a `TransportExt` trait and [implement that for the dyn safe trait like I did here](https://github.com/grafbase/grafbase/blob/main/engine/crates/engine/src/context/ext.rs#L303C66-L303C66)

obmarg (approved):
Nice, just a few comments

engine/crates/engine/src/registry/resolvers/postgres/request/create_many.rs
hfgbarrigas: oh definitely, left overs

engine/crates/integration-tests/src/engine/builder.rs
hfgbarrigas: aye sounds good

engine/crates/postgres-types/src/transport.rs
hfgbarrigas: Me and Julius had a chat and opted to instead return a `Stream<Result<Value, Error>>` from the `Transport` . Felt more adequate to bubble up to the caller any further processing.

pimeys (approved):

PR #841, opened 2023-10-20
chore: better debug logging in parser-openapi

obmarg:
The types `parser-openapi` works with are often just wrappers around a `NodeIndex`.  When working with these it can be quite awkward to debug - if you try to log details of the types you're working with you just get `OutputType(123)` or whatever.

This PR adds a `DebugNode` trait and implements it for some of these wrapper types - basically a clone of `Debug` that takes an `OpenApiGraph` (which is where the actual data these IDs refer to lives).

Also added some better tracing setup for the `parser-openapi` tests.

I've had this sitting on my local machine for a week or two, figured i should get it pushed up before I forget

### DIFF
--- a/engine/crates/parser-openapi/src/graph/debug.rs
+++ b/engine/crates/parser-openapi/src/graph/debug.rs
@@ -1,0 +1,39 @@
+use std::fmt::Debug;
+
+use super::OpenApiGraph;
+
+/// std::fmt::Debug for a node in our graph
+///
+/// Our nodes are usually just wrappers around indices so
+/// the default Debug impl is pretty useless.  This trait adds an
+/// extra graph parameter so we can get at the actual data.
+pub trait DebugNode: Sized {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>, graph: &OpenApiGraph) -> std::fmt::Result;
+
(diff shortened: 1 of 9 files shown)


engine/crates/parser-openapi/src/graph/debug.rs
```diff
@@ -0,0 +1,39 @@
+use std::fmt::Debug;
+
+use super::OpenApiGraph;
+
+/// std::fmt::Debug for a node in our graph
```
yoav-lavi, line 5: Why not implement the normal debug trait? is it already implemented?

yoav-lavi (approved):

engine/crates/parser-openapi/src/graph/debug.rs
obmarg: The normal debug trait only has access to `&self`.  I'm implementing this trait for types that look like `struct OutputType(u32)` - so all the normal debug trait can print is `OutputType(123)`.  Which is not very useful for debugging things.

This trait has access to both `&self` and `&OpenApiGraph` - so it can look up the actual data that the `u32` inside `OutputType` refers to and print useful information.

PR #842, opened 2023-10-20
cli: Subgraph introspection

pimeys:
Introduces a new subcommand `subgraph` and one subcommad to that called `introspect`. This command first runs a query to the given endpoint:

```graphql
query {
  _service {
    sdl
  }
}
```

If this query succeeds, that's the given output. If not, we run the standard GraphQL introspection and output _that_ schema.

Closes: GB-5141

### DIFF
--- a/cli/crates/cli/src/cli_input.rs
+++ b/cli/crates/cli/src/cli_input.rs
@@ -283,6 +283,40 @@ impl StartCommand {
     }
 }
 
+#[derive(Debug, clap::Args)]
+pub struct IntrospectCommand {
+    /// GraphQL URL to introspect
+    url: String,
+    /// Add a header to the introspection request
+    #[clap(short = 'H', long, value_parser, num_args = 0..)]
+    header: Vec<String>,
+}
+
+impl IntrospectCommand {
+    pub fn url(&self) -> &str {
+        &self.url
+    }
+
+    pub fn headers(&self) -> impl Iterator<Item = (&str, &str)> {
+        self.header.iter().map(|header| {
+            let mut splitted = header.split(':');
+            (splitted.next().unwrap_or(""), splitted.next().unwrap_or("").trim())
+        })
+    }
+}
+
+#[derive(Debug, Parser)]
+pub enum SubgraphCommandKind {
+    /// Introspect a subgraph endpoint and print its schema
+    Introspect(IntrospectCommand),
+}
+
+#[derive(Debug, Parser)]
+pub struct SubgraphCommand {
+    #[command(subcommand)]
+    pub kind: SubgraphCommandKind,
+}
+
 #[derive(Debug, Parser, strum::AsRefStr, strum::Display)]
 #[strum(serialize_all = "lowercase")]
 pub enum SubCommand {
@@ -314,6 +348,8 @@ pub enum SubCommand {
     /// Build the Grafbase project in advance to avoid the resolver build step in the start
     /// command.
     Build(BuildCommand),
+    /// Operations related to GraphQL subgraphs
+    Subgraph(SubgraphCommand),
 }
 
 // TODO see if there's a way to do this automatically (https://github.com/clap-rs/clap/discussions/4921)
@@ -381,6 +417,7 @@ impl ArgumentNames for SubCommand {
             | SubCommand::Start(_)
             | SubCommand::Build(_)
             | SubCommand::Completions(_)
+            | SubCommand::Subgraph(_)
             | SubCommand::Logs(_) => None,
         }
     }
--- a/cli/crates/cli/tests/introspection.rs
+++ b/cli/crates/cli/tests/introspection.rs
@@ -1,0 +1,257 @@
+#![allow(unused_crate_dependencies)]
+#![recursion_limit = "256"]
+
+#[path = "graphql-directive/server.rs"]
+mod server;
+
+mod utils;
+
+use serde_json::json;
+use utils::environment::Environment;
+use wiremock::{
+    matchers::{body_json, header, method, path},
+    Mock, MockServer, ResponseTemplate,
+};
+
+#[tokio::test(flavor = "multi_thread")]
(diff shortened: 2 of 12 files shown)


cli/crates/cli/src/cli_input.rs
```diff
@@ -283,6 +283,38 @@ impl StartCommand {
     }
 }
 
+#[derive(Debug, clap::Args)]
+pub struct IntrospectCommand {
+    /// GraphQL URL to introspect. Defaults to "http://localhost:4000"
```
fbjork, line 288: url should be required without a default value IMO

pimeys: This is quite useful if you're in a local context. We default to `http://localhost:4000`, and you just want to display the current schema from the running service.

pimeys: So you have a local subgraph running, you want to display and output the schema to somewhere, all you need to do is:

```console
grafbase subgraph introspect
```

pimeys: Agreed. Will remove that.

cli/crates/graphql-introspection/src/standard.rs
```diff
@@ -0,0 +1,30 @@
...
+        false,
+        url,
+        ConnectorHeaders::default(),
+        headers.iter().map(|(key, value)| (*key, *value)),
+        None,
+    )
+    .await;
+
+    match result {
+        Ok(registry) => Ok(registry.export_sdl(true)),
```
obmarg, line 19: Is there a reason to use `parser_graphql` & round-trip via a `Registry` with this?

I wouldn't necessarily trust it to not add GraphQL connector specific bits to the schema.

cli/crates/cli/tests/introspection.rs
```diff
@@ -0,0 +1,2361 @@
...
+    }));
+
+    Mock::given(method("POST"))
+        .and(path("/graphql"))
+        .and(body_json(request))
+        .respond_with(response)
+        .mount(&server)
+        .await;
+
+    let request = json!({
```
obmarg, line 237: Would semi-recommend just starting an actual GraphQL server and introspecting that.

Introspection isn't guaranteed to match this extract string and it'd be a lot less verbose.

cli/crates/graphql-introspection/src/subgraph.rs
```diff
@@ -0,0 +1,72 @@
...
+
+#[derive(Debug, serde::Deserialize)]
+struct Data {
+    #[serde(rename = "_service")]
+    service: Service,
+}
+
+#[derive(Debug, serde::Deserialize)]
+struct Response {
+    data: Option<Data>,
```
tomhoule, line 21: Would we be interested in `errors` here too? The status isn't always the indicator for GraphQL APIs.

tomhoule (approved):

cli/crates/cli/src/cli_input.rs
```diff
@@ -283,6 +283,39 @@ impl StartCommand {
...
+    url: Option<String>,
+    #[clap(short = 'H', long, value_parser, num_args = 0.., value_delimiter = ',')]
```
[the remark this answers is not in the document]
pimeys:
```console
❯ cargo run -- subgraph introspect http://localhost:4000 -H "foo: lol,bar"
    Blocking waiting for file lock on build directory
   Compiling grafbase v0.40.2 (/home/pimeys/code/grafbase/grafbase/cli/crates/cli)
    Finished dev [unoptimized + debuginfo] target(s) in 48.34s
     Running `/home/pimeys/code/grafbase/grafbase/cli/target/debug/grafbase subgraph introspect 'http://localhost:4000' -H 'foo: lol,bar'`
Grafbase CLI 0.40.2

[crates/cli/src/cli_input.rs:301] &self.header = [
    "foo: lol",
    "bar",
]
```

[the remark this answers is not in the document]
pimeys: oh my god...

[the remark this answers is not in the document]
pimeys: Not specifying the `value_delimiter` it works correctly.

cli/crates/graphql-introspection/src/standard.rs
pimeys: Good point. Should we use some standard mechanism, such as cynic here?

cli/crates/graphql-introspection/src/subgraph.rs
pimeys: What we figure out here is if data is `None`, we have errors and that's an indicator to do the standard introspection. I'm not super interested about what the error data contains...

cli/crates/graphql-introspection/src/standard.rs
obmarg: Yeah, might be a safer bet to just use it directly instead of going via `parser_graphql`.  [It's relatively easy to do introspection->SDL](https://github.com/obmarg/cynic/blob/main/cynic-cli/src/introspect.rs#L18-L32)

cli/crates/graphql-introspection/src/lib.rs
```diff
@@ -0,0 +1,38 @@
...
+pub use error::Error;
+use url::Url;
+
+pub async fn introspect(url: &str, headers: &[(&str, &str)]) -> Result<String, String> {
+    let url = match Url::parse(url) {
+        Ok(url) if url.path() == "/graphql" => url,
+        Ok(mut url) => {
+            url.set_path("/graphql");
+            url
+        }
```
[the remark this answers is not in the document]
obmarg: Yeah, the standard doesn't care about URLs.  Think avoiding the assumption might be better - this comment was brought to you by me trying the thing out:

CleanShot 2023-10-23 at 16 54 09@2x

obmarg (approved):

PR #843, opened 2023-10-22
feat(cli): typed resolvers with code generation

tomhoule:
# Description

This commit implements a first iteration of completely type safe
resolvers through code generation and resolver module analysis.

At a high level: we generate typescript types automatically in the
`grafbase dev` loop. They are written inside the `generated`
subdirectory of the project `grafbase` directory.

They are used as in the following example:

```
import { Resolver } from '@grafbase/generated'

const resolver: Resolver['post.comments'] = async (parent, args, ctx, info) => {
        // ...
}

export default resolver
```

Implementation-wise, this commit brings the following changes to the
existing code generation logic:

- replace graphql-parser with async-graphql-parser. graphql-parser isn't
  up to date on the spec and panics when it encounters schema
  extensions.
- implement resolver argument type generation

In the CLI,

- Integrate codegen in the `grafbase dev` loop.
- Integrate resolver checks in the `grafbase dev` loop.

Now, resolver checks. They use [the `swc`
parser](https://crates.io/crates/swc_ecma_parser) directly to parse and
analyze the TS resolver modules. That analysis is the building block for
auto-wiring of resolver functions (automatically assigning the right
types to the right resolvers for the right fields in the schema). For
now, it module analysis only serves better ergonomics and DX. The
following checks are implemented:

- The resolver is valid TypeScript
- The resolver has a default export:

```
  × The module is missing a default export. Grafbase expects a resolver function as default export.
   ╭─[./tests/resolver_discovery/missing_default_export/resolvers/missingDefaultExport.ts:3:1]
 3 │
 4 │ export const meowResolver = () => true
   ·  ───────────────────┬──────────────────
   ·                     ╰── Maybe this should be the default export?
   ╰────
  help: Export your resolver as default: `export default resolver`
```

- The schema field the resolver claims to resolve (via its type
  annotation) declares the resolver in the schema:

```
  × Orphan resolver at ./tests/resolver_discovery/missing_resolver_directive/resolvers/missingDirective.ts. The `Test.shouldHaveAResolver` field in your schema does not define a resolver.
   ╭─[./tests/resolver_discovery/missing_resolver_directive/resolvers/missingDirective.ts:2:1]
 2 │
 3 │ const resolver: Resolvers["Test.shouldHaveAResolver"] = (stuff) => {}
   ·                            ─────────────┬────────────
   ·                                         ╰── Inferred from this.
 4 │
   ╰────
  help: Try declaring the resolver: `@resolver(name: "missingDirective")` in GraphQL or `.resolver("missingDirective")` in TypeScript.
```

- The schema field the resolver claims to resolve (via its type
  annotation) declares a resolver file at a different location:

```
  × The resolver for `User.creditScore` isn't at the right location. In your schema, the resolver for `User.creditScore` is declared at `nested/somewhere/else`.
   ╭─[./tests/resolver_discovery/wrong_location/resolvers/nested/wrongLoc.ts:2:1]
 2 │
 3 │ const resolver: Resolver["User.creditScore"] = true
   ·                           ─────────┬────────
   ·                                    ╰── Inferred from this.
 4 │
   ╰────
  help: You can move the file or change the resolver name in your schema to "nested/wrongLoc".
```

I measured the CLI binary size increase in release mode from the new dependencies (no binary stripping or anything clever) to 2.5MB.

Not in this PR:

- ResolverFn in SDK. We should add the `ResolverFn` type to
  `@grafbase/sdk` to streamline the generated code.
- Auto-wiring, i.e. adding the `@resolver` directives based on resolver
  module analysis.. The challenge here is that we would need to extend
  or modify schemas sent to the bridge. It's an invasive change in the
  CLI, so I decided to leave it out of scope to get typed resolvers out
  as directly as possible. We can iterate on it in follow up PRs when
  there is bandwidth.
- Custom scalars. They are typed as `any`. There are solutions in other
  codegen based tools, but it will likely require configuration (where
  to import the types for custom scalars from), so out of scope for now.

closes GB-5147
closes GB-5068

# Type of change

- [ ] 💔 Breaking
- [x] 🚀 Feature
- [ ] 🐛 Fix
- [ ] 🛠️ Tooling
- [ ] 🔨 Refactoring
- [ ] 🧪 Test
- [ ] 📦 Dependency


cli/crates/server/src/codegen_worker.rs
```diff
@@ -0,0 +1,96 @@
+use crate::{event::Event, types::ServerMessage};
```
tomhoule, line 1: Should this file be named `codegen_server` instead for consistency with the rest of the modules in the crate?

tomhoule: Also: should it do more error reporting? It is mostly silent now.

cli/crates/server/src/codegen_worker.rs
```diff
@@ -0,0 +1,96 @@
...
+                        generate_ts_resolver_types(sdl, &generated_ts_resolver_types_path);
+                    }
+                    _ if path.extension() == Some(ffi::OsStr::new("ts"))
+                        && path.ancestors().any(|ancestor| ancestor == resolvers_path) =>
+                    {
+                        // A resolver changed, check it.
+                        let Some(sdl) = last_seen_sdl.as_deref() else {
+                            continue;
+                        };
+                        if let Err(err) = typed_resolvers::check_resolver(sdl, &path) {
```
tomhoule, line 64: In this file in general, I stayed conservative about checking resolvers, we don't check them all eagerly on startup. It's only for convenience, so I limited the feedback to the resolvers the user is currently working on to avoid spam and maybe even false positives. The checks do not produce data we use yet, and they are not blocking for testing, deploying.

cli/crates/server/src/codegen_worker.rs
```diff
@@ -0,0 +1,93 @@
...
+                            event_sender
+                                .send(ServerMessage::CompilationError(format!("{err:?}")))
```
tomhoule, line 63: Should there be a separate ServerMessage variant for this? The output looks good enough, I tested manually.

tomhoule:
It looks like the linux arm builds may be running out of memory on `cargo test`. Will look into it.

tomhoule:
WIP changelog: https://website-git-changelog-typed-resolvers-part-2.grafbase-vercel.dev/changelog/fully-typed-resolvers

.github/actions/sanitize/action.yml
```diff
@@ -49,4 +49,4 @@ runs:
       shell: bash
       run: |
         cd ${{ inputs.working-directory }}
-        RUST_BACKTRACE=1 cargo nextest run --profile ci
+        CARGO_BUILD_JOBS=2 RUST_BACKTRACE=1 cargo nextest run --profile ci
```
tomhoule, line 52: This solves the problem of builds running out of memory on linux-arm, at the cost of reduced parallelism. I see 8vcpu in the names. Let's try 4? or different approach?

cli/crates/server/src/event.rs
```diff
@@ -1,23 +1,26 @@
...
+    /// emitted when a schema or UDF change is detected
     /// and the server should be reloaded
     Reload(PathBuf),
     /// emitted when the bridge is ready to receive requests
     BridgeReady,
     /// emitted when the proxy server has a startup error
     ProxyError,
+    /// emitted when an SDL schema produced from the TS config has been written. Contains the file
+    /// path where it was written.
+    NewSdlFromTsConfig(Box<Path>),
```
pimeys, line 16: Why `Box<Path>` and not `PathBuf`? Just curious.

cli/crates/typed-resolvers/Cargo.toml
```diff
@@ -9,7 +9,15 @@ repository.workspace = true
...
 [dependencies]
-graphql-parser = "0.4"
+async-graphql-parser = "6"
+async-graphql-value = "6"
+walkdir = "2"
+
+swc_ecma_parser = "0.141.2"
+swc_ecma_ast = "0.110.0"
+swc_common = "0.33"
+miette = { version = "5.10.0", features = ["fancy"] }
```
pimeys, line 19: :3

cli/crates/typed-resolvers/src/analyze.rs
```diff
@@ -11,209 +13,247 @@ pub(crate) fn analyze<'doc>(graphql_document: &'doc ast::Document<'doc, &'doc st
...
-fn analyze_ast_field<'doc>(
-    field: &'doc ast::Field<'doc, &'doc str>,
-    schema: &AnalyzedSchema<'doc>,
-) -> Option<Field<'doc>> {
-    let (name, inner_is_nullable, list_wrappers) = type_from_nested(&field.field_type);
+fn analyze_ast_field<'doc>(field: &'doc ast::FieldDefinition, schema: &AnalyzedSchema<'doc>) -> Option<Field<'doc>> {
+    let resolver_name = field
+        .directives
+        .iter()
+        .find(|directive| directive.node.name.node == "resolver")
```
pimeys, line 202: This stuff from the graphql parser is super annoying. We don't need it here, but for the schema registry we should hide this `directive.node.name.node` dance...

cli/crates/typed-resolvers/src/check_resolver.rs
```diff
@@ -0,0 +1,254 @@
...
+    let (src, module) = parse_file(path)?;
+    let Some((resolver_path_in_schema, resolver_field_span)) = find_resolver_signature(&module).map_err(|err| {
+        err.with_source_code(miette::NamedSource::new(
+            path.display().to_string(),
+            src.as_str().to_owned(),
+        ))
+    })?
+    else {
+        return Ok(());
+    };
```
pimeys, line 21:
```suggestion
    let (src, module) = parse_file(path)?;

    let found_resolver_signature = find_resolver_signature(&module).map_err(|error| {
        error.with_source_code(miette::NamedSource::new(
            path.display().to_string(),
            src.as_str().to_owned(),
        ))
    })?;

    let Some((resolver_path_in_schema, resolver_field_span)) = found_resolver_signature else {
        return Ok(());
    };
```

cli/crates/typed-resolvers/tests/resolver_discovery/missing_default_export/expected.out
```diff
@@ -0,0 +1,11 @@
+=== ERRORS ===
```
pimeys, line 1: This whole error is super :3

cli/crates/typed-resolvers/tests/schema_types/basic.expected.ts
```diff
@@ -1,3 +1,16 @@
+// This is a generated file. It should not be edited manually.
+//
+// You can decide to commit this file or add it to your `.gitignore`.
+//
+// By convention, this module is imported as `@grafbase/generated`. To make this syntax possible,
+// add a `paths` entry to your `tsconfig.json`.
```
pimeys, line 6: Should we add this automatically?

examples/resolvers-kv/grafbase/grafbase.config.ts
```diff
@@ -13,6 +13,7 @@ const authProvider = auth.Authorizer({
 // https://grafbase.com/docs/database
 
 g.model('Question', {
+    id: g.id(),
```
pimeys, line 16: In the Grafbase database, the `id` is implicit and shouldn't be defined in the model.

We're sunsetting the database though so we can soon delete the whole file.

cli/crates/cli/tests/dev_watch.rs
```diff
@@ -45,4 +45,10 @@ fn dev_watch() {
     assert!(user_id.starts_with("user_"));
     assert!(user_birthday.ends_with('Z'));
     assert_eq!(user_verified, "VERIFIED");
+
+    {
+        // Check that the TS resolver types are being generated.
+        let generated_types_path = env.directory.join("grafbase/generated/index.ts");
```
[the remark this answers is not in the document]
tomhoule: I think they're different because the schemas are in the "dot grafbase" directory, `~/.grafbase`, whereas this is inside the grafbase directory in the project. I don't think we would want it to be in `~/.grafbase` because that would make it trickier to import. It's also part of the project and some people will want to commit it.

cli/crates/server/src/codegen_worker.rs
```diff
@@ -0,0 +1,93 @@
...
+
+fn read_sdl<'a>(path: &Path, last_seen_sdl: &'a mut Option<String>) -> Option<&'a str> {
+    let sdl = fs::read_to_string(path).ok()?;
+    *last_seen_sdl = Some(sdl);
+    last_seen_sdl.as_deref()
+}
+
+fn generate_ts_resolver_types(graphql_sdl: &str, target_file_path: &Path) {
+    let mut out = String::new();
+    if typed_resolvers::generate_ts_resolver_types(graphql_sdl, &mut out).is_ok() {
```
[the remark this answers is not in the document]
tomhoule: I'll write that in a comment, but if we fail to generate the types (e.g. the GraphQL string is invalid), then you end up with an empty file, or worse, you had generated types, and now they're replaced with an empty file, so I decided to go to a string first. An alternative solution would be explicitly deleting the file if generation fails. Wdyt?

cli/crates/server/src/event.rs
tomhoule: Like `Box<str>` vs `String`:

```
[src/main.rs:3] std::mem::size_of::<Box<Path>>() = 16
[src/main.rs:4] std::mem::size_of::<PathBuf>() = 24
```

`Box<Path>` doesn't have the `capacity` field. It's more efficient if you're not planning to mutate it.

cli/crates/typed-resolvers/Cargo.toml
tomhoule: I learned this library for this PR, and I think I'm going to keep using it, the APIs make sense and the errors are very nice.

cli/crates/typed-resolvers/src/analyze.rs
tomhoule: Yes we won't want to work with ASTs directly, we'll have something like `ParserDatabase` for things like checks.

examples/resolvers-kv/grafbase/grafbase.config.ts
tomhoule: :+1:

cli/crates/cli/tests/dev_watch.rs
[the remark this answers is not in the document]
pimeys: Do you think this place where we now store the resolvers could also be used for storing the client types? Could these files collide?

cli/crates/server/src/event.rs
pimeys: That is the best reason for me. Nice.

cli/crates/cli/tests/dev_watch.rs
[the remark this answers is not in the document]
tomhoule: Hmm, that's a good question. Maybe `./grafbase/resolver/types.ts` kind of path? We would only have to be careful with the gitignores but maybe it makes sense.

cli/crates/typed-resolvers/Cargo.toml
```diff
@@ -9,7 +9,15 @@ repository.workspace = true
 publish = false
 
 [dependencies]
-graphql-parser = "0.4"
+async-graphql-parser = "6"
+async-graphql-value = "6"
```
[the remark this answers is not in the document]
tomhoule: I wasn't aware of that, definitely makes more sense, will do this.

cli/crates/typed-resolvers/tests/schema_types/basic.expected.ts
tomhoule: Yeah we definitely could, but I'm a bit worried about messing with people's tsconfig.

cli/crates/typed-resolvers/Cargo.toml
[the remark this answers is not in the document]
tomhoule: Done

.github/actions/sanitize/action.yml
obmarg: Our runners have 12GB of RAM - kinda worrying if we're managing to exhaust that

cli/crates/typed-resolvers/tests/schema_types/basic.expected.ts
pimeys: We do that for `package.json` already...

cli/crates/typed-resolvers/Cargo.toml
pimeys: we should consider it for graphql parsing too...

.github/actions/sanitize/action.yml
tomhoule: Yes I think it's probably one very specific thing (test, dependency, ?) that has pathological behaviour, I've worked on projects with more tests that didn't have that problem.

tomhoule: I've seen `cargo test` in CLI take a very long time locally as well, for anecdotal evidence.

cli/crates/typed-resolvers/Cargo.toml
tomhoule: Absolutely

PR #849, opened 2023-10-23
feat: `@requires` directive

obmarg:
Federation defines a `@requires` directive that is used to declare that the resolver for a field requires certain other fields to function correctly.  The federation router can then make sure it passes those fields to the relevant subgraph in addition to any of the fields in the entities key.

Pretty simple so far.  

But just exposing the directive to federation would leave us with a pretty poor DX when a resolver requires fields that aren't provided by a subgraph - in those cases the user would still be stuck unable to write their resolver.  This was an existing problem (documented in GB-4014) but is more of a problem now that we're doing federation

So for those cases I've extended our execution to pay attention to the requires field and run some extra resolution to make sure we've populated whatever fields the resolver requires.  This isn't done in a particularly efficient way: if several fields require the same expensive-resolver to run then we'll run that expensive resolver several times.  But we can always try to optimise later if that turns out to be a problem.

This change also hasn't fully solved this problem for all our connectors - connectors like mongo & postgres only fetch the requested subset of their data from the database and that doesn't currently include any `@requires` fields. But OpenAPI is currently the only thing in scope for federation so we can revisit that later.  

I also skipped adding support for this to our normal `@model` parsing code because it was a pain to do and we're sunsetting it anyway.


yoav-lavi (approved):
Nice work!

PR #850, opened 2023-10-23
feat: add `@requires` support to SDK

obmarg:
I forgot to do this in #849

### DIFF
--- a/packages/grafbase-sdk/src/typedefs/resolver.ts
+++ b/packages/grafbase-sdk/src/typedefs/resolver.ts
@@ -19,10 +19,12 @@ export type Resolvable =
 export class ResolverDefinition {
   private field: Resolvable
   private resolver: string
+  private requiresFields: string | null
 
   constructor(field: Resolvable, resolver: string) {
     this.field = field
     this.resolver = resolver
+    this.requiresFields = null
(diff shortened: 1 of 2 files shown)


packages/grafbase-sdk/src/typedefs/resolver.ts
```diff
@@ -43,7 +45,22 @@ export class ResolverDefinition {
     return new CacheDefinition(this, new FieldLevelCache(params))
   }
 
+  /**
+   * Declares that this resolver requires certain fields to function correctly
+   *
+   * @param fields - The fields this resolver requires
```
pimeys, line 51: Why is this plural? It's a string param, so should it be `field: string`?

pimeys (approved):

packages/grafbase-sdk/src/typedefs/resolver.ts
obmarg: Delimited by whitespace, with nesting.  It's basically a subset of GraphQL syntax (missing arguments & directives specifically)

tomhoule:
Will incorporate this in https://github.com/grafbase/grafbase/pull/851 once merged.

PR #854, opened 2023-10-25
cli: future proof generated TS type names, improve DX

tomhoule:
This commit is prompted by a fear that has been visiting me at night since 9aba0fc513e67afb2a877c187f71f782f10c299e. The name of that fear is name collisions in generated code.

Before this commit, we have `safe_ts_type_name()`, a function that takes care of making a GraphQL type name TypeScript-safe, turning `undefined` into `_undefined` for example. We then generate one `export type` or moral equivalent for each GraphQL definition in the GraphQL schema.

There are several problems with that approach:

- Name collisions with other types we generate (`Resolver`) or types from the SDK. Of course we could add extra exceptions and rename `Resolver` to `_Resolver` for example, but that doesn't seem fool proof.
- Discoverability. If we rename types, they are not as discoverable.

The solution implemented here is nesting: the generated TypeScript code exports exactly two types: `Schema` and `Resolver`. The TypeScript type definitions for each GraphQL schema definition (objects, unions, etc.) and resolver live inside these two types. That makes name collisions impossible and completely removes the need to escape/change type names.

That comes with an extra improvement in DX. If you hover on the `parent` argument when writing a resolver, before you would see a type like `User` (or `_undefined`). Now you will see the fields inside.

Closes GB-5170

Before:

After:

# Type of change

- [ ] 💔 Breaking
- [ ] 🚀 Feature
- [ ] 🐛 Fix
- [ ] 🛠️ Tooling
- [x] 🔨 Refactoring
- [ ] 🧪 Test
- [ ] 📦 Dependency

### DIFF
--- a/cli/crates/typed-resolvers/src/codegen.rs
+++ b/cli/crates/typed-resolvers/src/codegen.rs
@@ -2,26 +2,27 @@ use crate::analyze::{AnalyzedSchema, BuiltinScalar, Definition, GraphqlType, Lis
 use std::fmt;
 
 const INDENT: &str = "  ";
+const DOUBLE_INDENT: &str = "    ";
 
 pub(crate) fn generate_module<O>(schema: &AnalyzedSchema<'_>, out: &mut O) -> fmt::Result
 where
     O: fmt::Write,
 {
     out.write_str(HEADER)?;
+    out.write_str("export type Schema = {\n")?;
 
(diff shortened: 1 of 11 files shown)


pimeys (approved):
Haha, good catch, superb fix. And TypeScript is _so weird_.

PR #860, opened 2023-10-26
composition: first bits

tomhoule:
# Description

This commit is a proposal for a general crate structure for supergraph
composition. Everything is a suggestion and up for discussion. It is
definitely very far from complete, and some areas (error reporting) are
just sketched.

What this commit _does_ contains: a proposed public API with a compose
function:

```rust
pub fn compose(subgraphs: &Subgraphs) -> CompositionResult
```

Where `CompositionResult` is SDL for a supergraph and diagnostics.

As well as the general flow, a test suite and the two following
composition rules as a proofs of concept:

- Declaration of different kinds cannot be merged. For example if you
  have `type User { ... }` and `interface User { ... }` in two different
  subgraphs, this is a composition error.
- Objects can be merged as long as their fields do not overlap (no name
  collision). This is the simplest rule of object merging. Handling any
  directive at all is out of scope for this PR.

Here is a non-exhaustive list of design choices that may be
controversial and that we may want to discuss:

1. Subgraphs are considered as a whole set, and the supergraph is an
  entirely separate data structure. There is a possibly viable
  alternative path: merging subgraphs pair-wise, with the resulting
  merged schema being the supergraph. I'm not sure I see all the
  tradeoffs clearly, but a few thoughts:

  - We can probably produce better diagnostics (errors) with global
    knowledge of all the subgraphs.
  - Supergraph SDL looks very different from the union of the subgraph
    schemas with additional rules. We would need to filter out, add
    attributes in non obvious ways to post-process the last remaining
    schema at the end of pair-wise merges.

2. Composition is separate from validation of individual subgraph
  schemas, but it assumes valid schemas as input.

3. Composition produces a SDL string for a supergraph, not a structured
  representation. This is because the requirements of the consumers are
  not known, and a data structure designed for rendering to SDL will
  look different from a data structure for the same supergraph but aimed
  at efficiently resolving queries. For example, `@key` directives will
  matter a great deal for composition, but as far as I can tell, they're
  not in the supergraph SDL.

  This API proposed in this commit picks the easy solution and only
  designs for rendering. The `Supergraph` type is opaque and write only
  however, so we should be able to change how it is organized internally
  relatively easily if we decide we want to use it for something other
  than rendering as GraphQL SDL in the future.

  So I see two questions to resolve:

  - Does it make sense to have a common data structure to represent a
    GraphQL schema, or do we opt for specialized data structures like
    `Subgraphs` in this commit? I think specializing is worth the
    effort, for separation of concerns and simplicity.
  - If we do want specialized data structures, should composition build
    the supergraph data structure or just SDL, and leave the supergraph
    schema data structure implementation to the router?

4. String interning from the start. The main argument against doing this
  now is that it would be premature optimization. On the other hand, we
  know that when merging many subgraphs, we will store, manipulate and
  compare many strings.

  - Storage is cheaper with interning: we have only one copy of each
    string in memory.
  - Manipulation is easier with interning: since `StringId`s are `Copy`,
    we don't need to worry about lifetimes (unlike `&str`), and we don't
    need to worry about ownership and allocation (unlike `String`)
  - Comparison is cheaper: comparing small integers is cheaper than
    comparing strings.

5. Efficient data access patterns from the start. More specifically,
  avoid doing anything `O(n²)` or worse on the number of types in the
  supergraph, or the number of fields in a type (think about `Query`).
  Composing large supergraphs out of many subgraphs in a reasonable
  amount of time is a requirement. It does mean that we have to maintain
  extra data structures as "secondary indexes".

# Type of change

- [ ] 💔 Breaking
- [x] 🚀 Feature
- [ ] 🐛 Fix
- [ ] 🛠️ Tooling
- [ ] 🔨 Refactoring
- [ ] 🧪 Test
- [ ] 📦 Dependency

### DIFF
--- a/crates/composition/src/compose_supergraph.rs
+++ b/crates/composition/src/compose_supergraph.rs
@@ -1,0 +1,52 @@
+use crate::{
+    subgraphs::{DefinitionKind, DefinitionWalker},
+    Context,
+};
+
+pub(crate) fn build_supergraph(ctx: &mut Context<'_>) {
+    ctx.subgraphs
+        .iter_definition_groups(|first, rest| match first.kind() {
+            DefinitionKind::Object => merge_object_definitions(ctx, first, rest),
+            _ => todo!(),
+        });
+
+    ctx.subgraphs.iter_field_groups(|first, rest| {
+        if let Some(next) = rest.next() {
+            ctx.diagnostics.push_fatal(format!(
+                "The field `{}` on `{}` is defined in two subgraphs (`{}` and `{}`).",
+                first.name_str(),
+                first.parent_definition().name_str(),
+                first.parent_definition().subgraph().name_str(),
+                next.parent_definition().subgraph().name_str(),
+            ));
+        }
+
+        ctx.supergraph.insert_field(
+            first.parent_definition().name(),
+            first.name(),
+            first.type_name(),
+        )
+    });
+}
+
+fn merge_object_definitions<'a>(
+    ctx: &mut Context<'_>,
+    first: DefinitionWalker<'a>,
+    mut rest: impl Iterator<Item = DefinitionWalker<'a>>,
+) {
+    let kind = first.kind();
+
+    if let Some(incompatible) = rest.find(|definition| definition.kind() != kind) {
+        let first_kind = first.kind();
+        let second_kind = incompatible.kind();
+        let name = first.name_str();
+        let first_subgraph = first.subgraph().name_str();
+        let second_subgraph = incompatible.subgraph().name_str();
+        ctx.diagnostics.push_fatal(format!(
+            "Cannot merge {first_kind:?} with {second_kind:?} (`{name}` in `{first_subgraph}` and `{second_subgraph}`)",
+        ));
+    }
+
+    ctx.supergraph
+        .insert_definition(first.name(), DefinitionKind::Object);
+}
--- a/crates/composition/src/diagnostics.rs
+++ b/crates/composition/src/diagnostics.rs
@@ -1,0 +1,28 @@
+/// Warnings and errors produced by composition.
+#[derive(Default)]
+pub struct Diagnostics(Vec<Diagnostic>);
+
+impl Diagnostics {
+    pub(crate) fn any_fatal(&self) -> bool {
+        self.0.iter().any(|diagnostic| diagnostic.is_fatal)
+    }
+
+    /// Iterate over all diagnostic messages.
+    pub fn iter_messages(&self) -> impl Iterator<Item = &str> {
+        self.0.iter().map(|diagnostic| diagnostic.message.as_str())
+    }
+
+    pub(crate) fn push_fatal(&mut self, message: String) {
+        self.0.push(Diagnostic {
+            message,
+            is_fatal: true,
+        });
+    }
+}
+
+/// A composition diagnostic.
+pub(crate) struct Diagnostic {
+    message: String,
+    /// Should this diagnostic be interpreted as a composition failure?
+    is_fatal: bool,
+}
--- a/crates/composition/src/strings.rs
+++ b/crates/composition/src/strings.rs
@@ -1,0 +1,42 @@
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord)]
+pub(crate) struct StringId(usize);
+
+impl StringId {
+    pub(crate) const MIN: StringId = StringId(usize::MIN);
+    pub(crate) const MAX: StringId = StringId(usize::MAX);
+}
+
+/// Very simple implementation of string interning.
+#[derive(Default)]
+pub(crate) struct Strings(indexmap::IndexSet<Box<str>>);
+
+impl Strings {
+    /// Interns a string-like value, avoiding allocation if the same string has already been
+    /// interned.
+    pub(crate) fn intern<T>(&mut self, string: T) -> StringId
+    where
+        T: AsRef<str> + Into<Box<str>>,
+    {
+        if let Some(idx) = self.0.get_index_of(string.as_ref()) {
+            return StringId(idx);
+        }
+
+        StringId(self.0.insert_full(string.into()).0)
+    }
+
+    /// Resolve a [StringId] into an `&str`. The unwrap is safe because we only issue [StringId]s
+    /// here. The only plausible error cases is if you pass in a [StringId] produced by another
+    /// [Strings] instance.
+    pub(crate) fn resolve(&self, id: StringId) -> &str {
+        self.0.get_index(id.0).unwrap().as_ref()
+    }
+}
+
+// Sugar for [Strings::resolve()].
+impl std::ops::Index<StringId> for Strings {
+    type Output = str;
+
+    fn index(&self, index: StringId) -> &Self::Output {
+        self.resolve(index)
+    }
+}
--- a/crates/composition/src/subgraphs.rs
+++ b/crates/composition/src/subgraphs.rs
@@ -1,0 +1,180 @@
+mod walkers;
+
+pub(crate) use walkers::*;
+
+use crate::strings::{StringId, Strings};
+use itertools::Itertools;
+use std::collections::BTreeSet;
+
+/// A set of subgraphs to be composed.
+#[derive(Default)]
+pub struct Subgraphs {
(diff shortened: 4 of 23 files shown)


crates/composition/src/diagnostics.rs
```diff
@@ -0,0 +1,28 @@
+/// Warnings and errors produced by composition.
+#[derive(Default)]
+pub struct Diagnostics(Vec<Diagnostic>);
```
tomhoule, line 3: This module is a very very rough sketch, we'll want more info in our diagnostics.

crates/composition/src/strings.rs
```diff
@@ -0,0 +1,42 @@
...
+    /// Resolve a [StringId] into an `&str`. The unwrap is safe because we only issue [StringId]s
+    /// here. The only plausible error cases is if you pass in a [StringId] produced by another
+    /// [Strings] instance.
+    pub(crate) fn resolve(&self, id: StringId) -> &str {
+        self.0.get_index(id.0).unwrap().as_ref()
+    }
+}
+
+// Sugar for [Strings::resolve()].
+impl std::ops::Index<StringId> for Strings {
```
tomhoule, line 36: Not sure if this is too much sugar, I'm leaning towards just calling `resolve()` everywhere.

crates/composition/tests/composition/object_fields_clash/supergraph.graphql
```diff
@@ -0,0 +1 @@
+The field `name` on `Band` is defined in two subgraphs (`label` and `critics`).
```
tomhoule, line 1: We probably want a separate file for errors.

crates/composition/src/subgraphs/walkers.rs
```diff
@@ -0,0 +1,96 @@
...
+
+    pub fn kind(self) -> DefinitionKind {
+        self.definition().kind
+    }
+
+    pub fn subgraph(self) -> SubgraphWalker<'a> {
+        self.walk(self.definition().subgraph_id)
+    }
+
+    // pub fn object_fields(self) -> impl Iterator<Item = ObjectFieldWalker<'a>> {
```
tomhoule, line 52: Left in as an example of traversing the schema.

crates/composition/src/subgraphs.rs
```diff
@@ -0,0 +1,168 @@
...
+use itertools::Itertools;
+use std::collections::BTreeSet;
+
+/// A set of subgraphs to be composed.
+#[derive(Default)]
+pub struct Subgraphs {
+    pub(crate) strings: Strings,
+    subgraphs: Vec<Subgraph>,
+    definitions: Vec<Definition>,
+    fields: Vec<Field>,
```
tomhoule, line 15: Needs a disclaimer about what vecs are sorted and why, and binary search.

crates/composition/src/strings.rs
tomhoule: (see supergraph::render for examples)

crates/composition/src/subgraphs.rs
```diff
@@ -0,0 +1,170 @@
...
+    // InputObject,
+    // Union,
+    // CustomScalar,
+}
+
+/// A field in an object, interface or input object type.
+pub(crate) struct Field {
+    parent_id: DefinitionId,
+    name: StringId,
+    type_name: StringId,
```
tomhoule, line 153: obviously too simplistic (we lose list and `!` wrapper types), just for demonstration

crates/composition/src/compose_supergraph.rs
tomhoule, line 1: This module is the one where we try to compose subgraphs into a supergraph.

crates/composition/src/diagnostics.rs
tomhoule: Also in general error handling: there are few errors to handle, but still we shouldn't panic as much as in this sketch.

crates/composition/src/compose_supergraph.rs
```diff
@@ -0,0 +1,52 @@
...
+    ctx.subgraphs
+        .iter_definition_groups(|first, rest| match first.kind() {
+            DefinitionKind::Object => merge_object_definitions(ctx, first, rest),
+            _ => todo!(),
+        });
+
+    ctx.subgraphs.iter_field_groups(|first, rest| {
+        if let Some(next) = rest.next() {
+            ctx.diagnostics.push_fatal(format!(
+                "The field `{}` on `{}` is defined in two subgraphs (`{}` and `{}`).",
```
pimeys, line 16: I don't fully understand this yet, but could there be three subgraphs defining the same field, would this error then be wrong?

crates/composition/src/compose_supergraph.rs
```diff
@@ -0,0 +1,52 @@
...
+    let kind = first.kind();
+
+    if let Some(incompatible) = rest.find(|definition| definition.kind() != kind) {
+        let first_kind = first.kind();
+        let second_kind = incompatible.kind();
+        let name = first.name_str();
+        let first_subgraph = first.subgraph().name_str();
+        let second_subgraph = incompatible.subgraph().name_str();
+        ctx.diagnostics.push_fatal(format!(
+            "Cannot merge {first_kind:?} with {second_kind:?} (`{name}` in `{first_subgraph}` and `{second_subgraph}`)",
```
pimeys, line 46: Same here, what about three subgraphs with name collisions?

crates/composition/tests/composition/object_fields_clash/supergraph.graphql
pimeys: This is not graphql, so yes.

crates/composition/tests/composition/objects_basic/supergraph.graphql
```diff
@@ -0,0 +1,6 @@
+type User {
```
pimeys, line 1: Do we expect some concrete order in the final type?

pimeys (approved):
nice. nitpicks only, but approved.

crates/composition/src/subgraphs.rs
```diff
@@ -0,0 +1,170 @@
...
+        crate::ingest_subgraph::ingest_subgraph(subgraph_schema, name, self)
+    }
+
+    /// Iterate over groups of definitions to compose. The definitions are grouped by name. The
+    /// argument is a closure that receives, for each group, the first definition of the group and
+    /// an iterator over all the other definitions in the group. The order of iteration is
+    /// deterministic but unspecified.
+    pub(crate) fn iter_definition_groups<'a>(
+        &'a self,
+        mut compose_fn: impl FnMut(DefinitionWalker<'a>, &mut dyn Iterator<Item = DefinitionWalker<'a>>),
```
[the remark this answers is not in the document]
tomhoule: You can't type alias `impl Trait` parameters, but we could have `iter_definition_groups<'a, ComposeFn>(...) where ComposeFn: ...`. That?

crates/composition/src/compose_supergraph.rs
tomhoule: You would get the error for the first two, so it's not strictly wrong, but it's a confusing error, I agree. We can make the error better by mentioning all the subgraphs where the object type defines the same field. And we should add "without `@shareable`" but we're not at that level of validation yet.

crates/composition/tests/composition/objects_basic/supergraph.graphql
tomhoule: As far as I can tell it's not specified by the spec and apollo produces stable but random order (probably the order it iterates the fields in different subgraphs). Not lexicographic order.

crates/composition/src/subgraphs.rs
tomhoule: done

crates/composition/src/subgraphs.rs
```diff
@@ -0,0 +1,170 @@
...
+pub struct Subgraphs {
+    pub(crate) strings: Strings,
+    subgraphs: Vec<Subgraph>,
+    definitions: Vec<Definition>,
+    fields: Vec<Field>,
+
+    // Secondary indexes.
+
+    // (definition name, definition id)
+    definition_names: BTreeSet<(StringId, DefinitionId)>,
```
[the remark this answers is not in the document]
tomhoule: done

PR #861, opened 2023-10-26
typed-resolvers: make all fields that could have resolvers attached optional

tomhoule:
For object and interface types, in the generated TypeScript code. This is a correctness problem:

- In the `parent` argument, because that data may not be present (e.g. sibling resolver's responses)
- In the return types, because the resolver is not necessarily responsible for returning that data.

Thanks @obmarg  for noticing the problems.

closes GB-5177

# Type of change

- [ ] 💔 Breaking
- [ ] 🚀 Feature
- [x] 🐛 Fix
- [ ] 🛠️ Tooling
- [ ] 🔨 Refactoring
- [ ] 🧪 Test
- [ ] 📦 Dependency

### DIFF
--- a/cli/crates/typed-resolvers/src/codegen.rs
+++ b/cli/crates/typed-resolvers/src/codegen.rs
@@ -33,10 +33,11 @@ where
 
                 for field in fields {
                     let field_optional = !is_input_object
-                        && matches!(
+                        && (matches!(
                             field.r#type.kind,
                             TypeKind::Definition(Definition::Object(_) | Definition::Union(_))
-                        );
+                        ) || field.resolver_name.is_some()
+                            || field.has_arguments);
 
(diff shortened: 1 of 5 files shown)


cli/crates/typed-resolvers/src/codegen.rs
```diff
@@ -33,10 +33,11 @@ where
 
                 for field in fields {
                     let field_optional = !is_input_object
-                        && matches!(
+                        && (matches!(
                             field.r#type.kind,
                             TypeKind::Definition(Definition::Object(_) | Definition::Union(_))
-                        );
+                        ) || field.resolver_name.is_some()
+                            || field.has_arguments);
```
tomhoule, line 40: Not the prettiest formatting, happy to push something nicer if that seems better.

obmarg (approved):

PR #865, opened 2023-10-27
composition: implement @shareable and @link basics

tomhoule:
Implemented in this PR:

- `@link` directive imports and aliasing (see docs in diff)
- Fields with `@shareable` are allowed to be defined in multiple
  subgraphs

Also sets up very basic CI for crates/composition.

Part of GB-5176

# Type of change

- [ ] 💔 Breaking
- [x] 🚀 Feature
- [ ] 🐛 Fix
- [ ] 🛠️ Tooling
- [ ] 🔨 Refactoring
- [ ] 🧪 Test
- [ ] 📦 Dependency

### DIFF
--- a/.github/workflows/composition.yml
+++ b/.github/workflows/composition.yml
@@ -1,0 +1,27 @@
+on:
+  pull_request:
+    paths:
+      - 'crates/composition/**/*'
+
+jobs:
+  composition:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout Repository
+        uses: actions/checkout@v3
+
+      - name: Rust job init
+        uses: ./.github/actions/init_rust_job
+        with:
+          platform: linux
+          cache-key: ${{ runner.os }}-${{ runner.arch }}-cargo-dev-${{ hashFiles('crates/composition/Cargo.toml') }}
+          restore-key: ${{ runner.os }}-${{ runner.arch }}-cargo-dev
+
+      - run: cargo clippy
+        working-directory: 'crates/composition'
+
+      - run: cargo test
+        working-directory: 'crates/composition'
+      
+      - run: cargo fmt --check
+         working-directory: 'crates/composition'
--- a/crates/composition/src/ingest_subgraph/schema_definitions.rs
+++ b/crates/composition/src/ingest_subgraph/schema_definitions.rs
@@ -1,0 +1,232 @@
+use async_graphql_parser::types as ast;
+use async_graphql_value::ConstValue;
+use std::borrow::Cow;
+
+pub(super) fn ingest_schema_definitions(
+    document: &ast::ServiceDocument,
+) -> FederationDirectivesMatcher<'_> {
+    document
+        .definitions
+        .iter()
+        .filter_map(|def| {
+            if let ast::TypeSystemDefinition::Schema(schema) = def {
+                Some(schema)
+            } else {
+                None
+            }
+        })
+        .flat_map(|schema_definition| schema_definition.node.directives.iter())
+        .map(|d| &d.node)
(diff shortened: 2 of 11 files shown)


crates/composition/src/ingest_subgraph/schema_definitions.rs
```diff
@@ -0,0 +1,228 @@
...
+                    out.push((
+                        name.trim_start_matches('@'),
+                        alias.unwrap_or(name).trim_start_matches('@'),
+                    ));
+                }
+            }
+            _ => (),
+        }
+    }
+}
```
obmarg, line 134: I don't have particularly strong opinions but `ConstValue` implements `Deserializer` so we _could_ use serde to derive this

crates/composition/src/subgraphs.rs
```diff
@@ -66,22 +67,20 @@ impl Subgraphs {
     /// and field name. The argument is a closure that receives, for each group, the first field of
     /// the group and an iterator over all the other fields in the group. The order of iteration is
     /// deterministic but unspecified.
```
obmarg, line 69: This comment doesn't look correct anymore

.github/workflows/composition.yml
```diff
@@ -0,0 +1,24 @@
...
+        with:
+          platform: linux
+          cache-key: ${{ runner.os }}-${{ runner.arch }}-cargo-dev-${{ hashFiles('crates/composition/Cargo.toml') }}
+          restore-key: ${{ runner.os }}-${{ runner.arch }}-cargo-dev
+
+      - run: cargo clippy
+        working-directory: 'crates/composition'
+
+      - run: cargo test
+        working-directory: 'crates/composition'
```
obmarg, line 24: Maybe worth checking format as well:

```suggestion
      - run: cargo test
        working-directory: 'crates/composition'
      
      - run: cargo fmt --check
         working-directory: 'crates/composition'
```

obmarg (approved):

crates/composition/src/ingest_subgraph/schema_definitions.rs
```diff
@@ -0,0 +1,228 @@
...
+
+impl<'a> FederationDirectivesMatcher<'a> {
+    pub(crate) fn is_federation_directive(directive: &ast::ConstDirective) -> bool {
+        directive
+            .get_argument("url")
+            .map(|url| match &url.node {
+                ConstValue::String(s) => s.contains("dev/federation/v2"),
+                _ => false,
+            })
+            .unwrap_or_default()
```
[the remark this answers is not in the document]
tomhoule: I moved code around and the check for the directive name got lost in the process. Good catch, fixing now.

crates/composition/src/ingest_subgraph/schema_definitions.rs
tomhoule: Oh I didn't know that. I'll give it a try and push if that works too and that saves us work.

crates/composition/src/subgraphs.rs
tomhoule: True, good catch, fixing it.

crates/composition/src/subgraphs.rs
```diff
@@ -178,3 +180,17 @@ fn push_and_return_id<T, Id>(elems: &mut Vec<T>, new_elem: T, make_id: fn(usize)
     elems.push(new_elem);
     id
 }
+
+impl ops::Index<SubgraphId> for Subgraphs {
```
tomhoule, line 184: Can't remember why this was added, will check if we still need it.

crates/composition/src/subgraphs/object.rs
tomhoule, line 1: Dead code, should be gone.

PR #870, opened 2023-10-27
composition: initial `@key` support

tomhoule:
This is basic and mostly about ingestion. Composition will probably take a different shape as more rules are implemented (composite keys, nested keys, multiple keys).

Two validation rules are implemented in this PR:

- If an object is an entity is one subgraph, it has to be an entity in all subgraphs.
- If a field is the key in one subgraph, it has to be part of the key in all the others.

Part of GB-5176


crates/composition/src/subgraphs/selection_sets.rs
```diff
@@ -0,0 +1,120 @@
...
+    pub(crate) fn selection_set_from_str(&mut self, fields: &str) -> Result<SelectionId, String> {
+        // Cheating for now, we should import the parser from engines instead.
+        let fields = format!("{{ {fields} }}");
+        let parsed = async_graphql_parser::parse_query(fields).map_err(|err| err.to_string())?;
+
+        let async_graphql_parser::types::ExecutableDocument {
+            operations: async_graphql_parser::types::DocumentOperations::Single(operation),
+            ..
+        } = parsed
+        else {
```
tomhoule, line 52: Not great formatting, but it's what rustfmt produces.

tomhoule: This function is a bit long, maybe it needs to be split in two and get its own module.

crates/composition/src/compose_supergraph.rs
```diff
@@ -32,14 +39,37 @@ fn merge_object_definitions<'a>(
...
+            entity_subgraphs
+                .into_iter()
+                .map(|d| d.subgraph().name_str())
+                .join(", "),
+            non_entity_subgraphs
+                .into_iter()
+                .map(|d| d.subgraph().name_str())
+                .join(", "),
+        ));
+    }
```
tomhoule, line 63: This is obviously not how we want the composition functions to look like in the end. I think I am starting to get a feel for how to organize them, but I'll keep implementing the basics before refactoring for the more complicated rules.

crates/composition/src/subgraphs/selection_sets.rs
```diff
@@ -0,0 +1,120 @@
...
+        self.subgraphs.strings.resolve(self.id.field)
+    }
+
+    pub(crate) fn children(self) -> impl Iterator<Item = SelectionWalker<'a>> {
+        self.subgraphs
+            .selection_sets
+            .children(self.id.id)
+            .map(move |selection| self.walk(selection))
+    }
+}
```
tomhoule, line 120: This file should probably have unit tests, it's not completely trivial.

crates/composition/src/ingest_subgraph/object.rs
```diff
@@ -0,0 +1,42 @@
...
+            };
+            let is_resolvable = directive
+                .node
+                .get_argument("resolvable")
+                .and_then(|v| match v.node {
+                    ConstValue::Boolean(b) => Some(b),
+                    _ => None,
+                })
+                .unwrap_or(true); // defaults to true
+            subgraphs.push_key(definition_id, selection_id, is_resolvable);
```
Finistere, line 37: opt: usually not a fan of having both an output and an output argument, but I guess that'll be reworked X times anyway.

crates/composition/src/subgraphs.rs
```diff
@@ -19,6 +21,13 @@ pub struct Subgraphs {
     // Invariant: `fields` is sorted by `Field::object_id`. We rely on it for binary search.
     fields: Vec<Field>,
 
+    /// All the keys (`@key(...)`) in all the subgraphs in one container.
+    keys: Keys,
+
+    /// Storage for selection sets. Technically only a restricted version of selection sets allowed
+    /// inside `@key(fields: "...")`.
+    selection_sets: SelectionSets,
```
Finistere, line 29: why not include it inside `Keys`? The selection is specific to the key in my mind with a list of subgraphs supporting it (+ resolvable).
The structure of `Selection` could be simpler if embedded inside `Key` with something like:
```
struct Selection {
  fieldId: FieldId
  children: Vec<Selection>
}
```
Is there any benefit of merging all selections inside a BTreeSet?

crates/composition/src/ingest_subgraph/object.rs
tomhoule: Yes that felt dirty as well as I was writing it. I was writing down the reasoning for this particular case in this comment box and it's not strong. I'll change this function to do everything with the output argument.

crates/composition/src/subgraphs.rs
tomhoule: > why not include it inside Keys? The selection is specific to the key in my mind with a list of subgraphs supporting it (+ resolvable).

It may be the wrong decision because I haven't implemented these other directives yet, but I think `@provides` and `@requires` also use selection sets (https://www.apollographql.com/docs/federation/federated-types/federated-directives/#provides).

> Is there any benefit of merging all selections inside a BTreeSet?

Apart from expected smaller memory footprint and less fragmentation no. Usually there's ways to query the data structures (like range queries) that open themselves up with that kind of approaches, but it won't buy us anything in this case It's probably slower than nested vecs too, since the normal case is going to be no nesting. I'll revert to the simpler approach, I'm not convinced the shared data structure for collections is a good rabbit hole to go down into anymore.

crates/composition/src/ingest_subgraph/object.rs
tomhoule: Ah no, it's because we don't store `is_shareable` inside the object, but rather inside the fields, to have a single source of truth, so it isn't in `Subgraphs`. I'll think about a clean way to incorporate that.

crates/composition/src/subgraphs.rs
Finistere: Yes both `@provides` and `@requires` have selection sets, but the way it's constructed currently nothing will be shared unless I miss something. 🤷 Or am I missing something? In terms of memory whether having `parentId` usize vs `children` vec probably is quite close in practice for this

tomhoule: I haven't thought in details about how these two would be implemented, but I imagined something like:

```rust
struct Provides {
  selection: SelectionId
}
```

sharing the same data structure for selection sets inside Subgraphs (not that it wouldn't create new entries even for the same selection sets, indeed no sharing would happen at that level, the BTreeSet would grow large). It's orthogonal to nested vecs vs. btreeset too, we could still have a selection_sets store inside Subgraphs the same way. For now let's nest.

For the memory usage, in principle (SelectionId, SelectionId, StringId) is three words, whereas a Field struct even without parent reference is (StringId, Vec), so four words. And the BTreeSet groups allocations, the hope was that you would end with fewer allocations than selection sets in practice. But I agree at the scale we're talking about it doesn't matter until we notice it being a problem, and the reasoning isn't airtight.

I've been bitten by too many small vecs nested inside struct fields in the past so I instinctively avoid them, but in this case it's the better solution for now, it's simpler, I'm changing that module to nesting vecs inside `Key` like you suggested.

tomhoule: Done

Finistere (approved):

PR #876, opened 2023-10-30
composition: implement input type and object field arguments

tomhoule:
Part of GB-5176

### DIFF
--- a/crates/composition/src/compose_supergraph.rs
+++ b/crates/composition/src/compose_supergraph.rs
@@ -1,3 +1,7 @@
+mod input_object;
+mod object;
+
+use self::input_object::*;
 use crate::{
     subgraphs::{DefinitionKind, DefinitionWalker, FieldWalker},
     Context,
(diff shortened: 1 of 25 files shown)


crates/composition/src/compose_supergraph.rs
```diff
@@ -70,6 +75,10 @@ fn merge_object_definitions<'a>(
 fn merge_field_definitions(ctx: &mut Context<'_>, fields: &[FieldWalker<'_>]) {
     let Some(first) = fields.get(0) else { return };
 
+    if first.parent_definition().kind() != DefinitionKind::Object {
+        return;
+    }
```
tomhoule, line 80: Not done in this PR, but I think the way this is going to evolve is matching on the definition kind here too. Then we can specialize on merging interface fields, object fields or input object fields. In separate functions of course for readability.

crates/composition/src/supergraph/render.rs
```diff
@@ -11,13 +11,23 @@ impl Supergraph {
...
+                        let args = if args.is_empty() {
+                            String::new()
+                        } else {
+                            let inner = args
+                                .iter()
+                                .map(|(name, ty)| (strings.resolve(*name), strings.resolve(*ty)))
+                                .map(|(name, ty)| format!("{name}: {ty}"))
+                                .join(", ");
+                            format!("({inner})")
+                        };
```
tomhoule, line 27: dirty but this whole module is going away soon

PR #881, opened 2023-10-31
composition: interfaces, custom scalars & enums

tomhoule:
- Interfaces. Non-entity interfaces only for now. Like unions,
  interfaces are merged by composition.
- Custom scalars. No particular rule, they are all picked up and merged
  into the resulting final schema.
- Enums. The rules are more complicated. If the enum is used in output
  position only, the enums from different subgraphs are merged. If the
  enum is used in input position, only the values they have in common
  are retained. And if the enum is used in both argument and output
  position, the enums must match exactly in all schemas.

Part of GB-5176
Closes GB-5228
Closes GB-5230
Closes GB-5231
# Type of change

- [ ] 💔 Breaking
- [x] 🚀 Feature
- [ ] 🐛 Fix
- [ ] 🛠️ Tooling
- [ ] 🔨 Refactoring
- [ ] 🧪 Test
- [ ] 📦 Dependency

### DIFF
--- a/crates/composition/src/compose_supergraph/enums.rs
+++ b/crates/composition/src/compose_supergraph/enums.rs
@@ -1,0 +1,134 @@
+use super::*;
+use crate::{strings::StringId, Subgraphs};
+use std::collections::HashSet;
+
+pub(super) fn merge_enum_definitions(
+    first: &DefinitionWalker<'_>,
+    definitions: &[DefinitionWalker<'_>],
+    ctx: &mut Context<'_>,
+) {
+    let enum_name = first.name();
+
+    match (
+        enum_is_used_in_input(enum_name, ctx.subgraphs),
+        enum_is_used_in_return_position(enum_name, ctx.subgraphs),
+    ) {
+        (true, false) => merge_intersection(first, definitions, ctx),
+        (false, true) => merge_union(first, definitions, ctx),
+        (true, true) => merge_exactly_matching(first, definitions, ctx),
+        (false, false) => {
+            // The enum isn't used at all, omit it from the supergraph.
+        }
+    }
+}
+
+/// Returns whether the enum is used anywhere in a field argument or an input type field.
+fn enum_is_used_in_input(enum_name: StringId, subgraphs: &Subgraphs) -> bool {
+    let in_field_arguments = || {
+        subgraphs
+            .iter_fields()
+            .flat_map(|field| field.arguments())
+            .any(|arg| arg.argument_type().type_name() == enum_name)
+    };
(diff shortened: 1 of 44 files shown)


crates/composition/src/compose_supergraph/enums.rs
```diff
@@ -0,0 +1,134 @@
...
+    for item in iterator {
+        if slice[idx] != item {
+            return false;
+        }
+
+        idx += 1;
+    }
+
+    idx == slice.len()
+}
```
tomhoule, line 134: If someone has something shorter / easier to read with std or itertools (haven't tried zip_longest(), afraid of the item enum), I'll take it gladly.

crates/composition/src/supergraph.rs
```diff
@@ -38,7 +40,13 @@ impl Supergraph {
     ) {
         if self
             .fields
-            .insert((parent_type_name, field_name), (arguments, field_type))
+            .insert(
+                (parent_type_name, field_name),
+                Field {
+                    arguments,
+                    field_type,
+                },
```
pimeys, line 48: maybe put this in a variable, doesn't look that weird

crates/composition/tests/composition/custom_scalars_basic/supergraph.graphql
```diff
@@ -0,0 +1,8 @@
+scalar Yan
+scalar Tan
+scalar Texture
+type Sheep {
+    texture: Texture
+    sheepIdx: Tethera
+}
+scalar Tethera
```
pimeys, line 8: we should maybe at some point think about the ordering of these

crates/composition/src/compose_supergraph/enums.rs
```diff
@@ -0,0 +1,134 @@
...
+
+    ctx.supergraph
+        .insert_definition(first.name(), DefinitionKind::Enum);
+
+    for value in expected {
+        ctx.supergraph.insert_enum_value(first.name(), value)
+    }
+}
+
+fn is_slice_match<T: PartialEq>(slice: &[T], iterator: impl Iterator<Item = T>) -> bool {
```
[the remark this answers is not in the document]
tomhoule: In this case the iterators we have aren't exactsize, unfortunately.

crates/composition/tests/composition/custom_scalars_basic/supergraph.graphql
tomhoule: This is way, way down the list, and the rendering to graphql is going to be rewritten today

PR #882, opened 2023-10-31
Fix: workers not loading in cli-app

jonathanawesome:
Updates to the worker bundling strategy in Pathfinder have resulted in our `cli-app` not being able to load the 3 monaco-editor workers.

This PR updates the `base` option in the `vite.config` of `cli-app` to "/static", which results in a correct path for each of the workers.


tomhoule (approved):
I'm not an expert in vite, but that said, looks good to me :+1:

tomhoule:
It took me a while to package the assets in a way that works and where they got updated (we should really do something about this), but now I can confirm that locally, the errors are gone :tada:

fbjork (approved):

tomhoule:
I'm clicking the update branch button to trigger a rebuild, I'm trying to fix the yellow hanging jobs problem once and for all in https://github.com/grafbase/grafbase/pull/892.

fbjork:
@jonathanawesome @tomhoule ready to go?

tomhoule:
Yes, I think the CI problems are sorted out for good after https://github.com/grafbase/grafbase/pull/892

PR #884, opened 2023-10-31
feat(connector): Transaction support for Postgres

pimeys:
This allows us to use transactions with the transport, and run the same query/execute methods as we do with the normal functions.

### DIFF
--- a/engine/crates/postgres-types/src/transport/tcp/transaction.rs
+++ b/engine/crates/postgres-types/src/transport/tcp/transaction.rs
@@ -1,0 +1,24 @@
+use crate::{error::Error, transport::Transport};
+use async_trait::async_trait;
+use futures::stream::BoxStream;
+use serde_json::Value;
+use tokio_postgres::Transaction;
+
+use super::executor;
+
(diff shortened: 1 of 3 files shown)


engine/crates/postgres-types/src/transport/tcp/transaction.rs
```diff
@@ -0,0 +1,24 @@
...
+        executor::query(self, query, params)
+    }
+
+    async fn parameterized_execute(&self, query: &str, params: Vec<Value>) -> crate::Result<i64> {
+        executor::execute(self, query, params).await
+    }
+
+    // we don't care about this in a tx
+    fn connection_string(&self) -> &str {
+        ""
```
tomhoule, line 22: maybe we could panic instead? That should never be reached.

tomhoule (approved):

engine/crates/postgres-types/src/transport/tcp/transaction.rs
pimeys: This is a stupid thing I use for testing only. I'm probably removing it in the future when I have a bit of time.

PR #889, opened 2023-11-03
composition: return structured federated graphs

tomhoule:
# Description

Until this commit, the output of composition is a GraphQL SDL string without any federation-specific metadata attached.

This commit implements a much more structured output format that is designed to be serialized. The format is defined in a new crate: `grafbase-federated-graph`. It is a lightly normalized representation of the schema, with all strings and all field types deduplicated.

The composition crate is reorganized for clarity and to make that structured output possible. See the comments in the diff for more details.

The new `grafbase-federated-graph` crate also defines a function to render a structured federated graph to a GraphQL SDL string.

# Type of change

- [ ] 💔 Breaking
- [x] 🚀 Feature
- [ ] 🐛 Fix
- [ ] 🛠️ Tooling
- [ ] 🔨 Refactoring
- [ ] 🧪 Test
- [ ] 📦 Dependency


crates/composition/src/composition_ir.rs
```diff
@@ -0,0 +1,222 @@
...
+            },
+        )
+    }
+
+    pub(crate) fn insert_object(&mut self, object_name: StringWalker<'_>) {
+        let name = self.insert_string(object_name);
+        let object = federated::Object {
+            name,
+            implements_interfaces: Vec::new(),
+            resolvable_keys: Vec::new(),
```
tomhoule, line 112: As you can see there, we're not filling in some of the important data. I didn't want to do it in this PR to avoid making it huger.

crates/composition/src/emit_federated_graph.rs
```diff
@@ -0,0 +1,132 @@
...
+
+        let push_field = |out| {
+            let field = federated::Field {
+                name: field_name,
+                field_type_id,
+                arguments,
+
+                provides: Vec::new(),
+                requires: Vec::new(),
+                resolvable_in: Vec::new(),
```
tomhoule, line 77: Here as well, not filling in important info yet. That should come next.

.github/workflows/composition.yml
```diff
@@ -20,10 +23,10 @@ jobs:
...
-        working-directory: 'crates/composition'
+        working-directory: crates/${{ matrix.package }}
 
       - run: cargo test
-        working-directory: 'crates/composition'
+        working-directory: crates/${{ matrix.package }}
       
       - run: cargo fmt --check
-        working-directory: 'crates/composition'
+        working-directory: crates/${{ matrix.package }}
```
tomhoule, line 32: Now thinking maybe we should have some duplication and run these sequentially, it seems wasteful to spin up two runners for that.

crates/composition/src/composition_ir.rs
```diff
@@ -0,0 +1,222 @@
...
+
+    pub(crate) strings: StringsIr,
+    pub(crate) fields: Vec<FieldIr>,
+    pub(crate) union_members: BTreeSet<(subgraphs::StringId, subgraphs::StringId)>,
+}
+
+impl CompositionIr {
+    pub(crate) fn insert_enum(&mut self, enum_name: StringWalker<'_>) -> federated::EnumId {
+        let name = self.insert_string(enum_name);
+        push_definition_and_return_id(
```
tomhoule, line 35: Not too convinced by this util function, maybe we can make these calls more readable.

crates/composition/src/composition_ir.rs
```diff
@@ -0,0 +1,222 @@
...
+/// structure is distinct from [FederatedGraph](grafbase_federated_graph::FederatedGraph) because
+/// it is accumulated out of order during composition. Only after all composed top-level
+/// definitions (objects, interfaces, scalars, enums, unions) are known can we construct
+/// [federated::FieldTypeId]s and [federated::ObjectId]s for union members for the federated graph.
+///
+/// This is a **write only** data structure during composition. The source of truth for the
+/// contents of the federated graph is the subgraphs.
+#[derive(Default)]
+pub(crate) struct CompositionIr {
+    pub(crate) field_type_kinds_map: HashMap<subgraphs::StringId, federated::FieldTypeKind>,
```
tomhoule, line 18: Todo: rename this, the name is bad. Also maybe `FieldTypeKind` should be renamed to `DefinitionId`?

crates/composition/src/composition_ir.rs
obmarg: Yeah.  Maybe just because I'm unfamiliar with this code, but it's not really clear to me what this code is doing.

`push_definition_and_return_id` takes 4 arguments but is only 3 lines long - feels like it might be better to just write out the code (though I could be wrong)

crates/composition/src/lib.rs
```diff
@@ -2,33 +2,46 @@
...
+        federated_graph,
+        diagnostics,
     }
 }
+
+fn push_and_return_id<T, Id>(elems: &mut Vec<T>, new_elem: T, make_id: fn(usize) -> Id) -> Id {
+    let id = make_id(elems.len());
+    elems.push(new_elem);
+    id
+}
```
obmarg, line 47: Kinda wondering about this function as well.  Maybe just me, but feels like the make_id parameter is obscuring what's going on without providing much benefit?

You _could_ just return a `usize` and do `federated::FieldTypeId(push_and_return_id(&mut whatever, Whatever))`?

Dunno, don't have strong feelngs, just a thought.

obmarg: I might personally have been tempted to stick this on a `Vec` extension trait as well, to save the  `&mut`.

`FieldTypeId(self.field_types.push_and_return_id(FieldType { .. }))` seems like it might be alright?  Dunno, definitely not an important comment

crates/composition/src/subgraphs.rs
```diff
@@ -107,6 +110,14 @@ impl Subgraphs {
...
         }
     }
+
+    /// Iterates all builtin scalars _that are in use in at least one subgraph_.
+    pub(crate) fn iter_builtin_scalars(&self) -> impl Iterator<Item = StringWalker<'_>> + '_ {
+        ["ID", "String", "Boolean", "Int", "Float"]
+            .into_iter()
+            .filter_map(|name| self.strings.lookup(name))
+            .map(|string| self.walk(string))
+    }
```
obmarg, line 120: Are we putting built in scalars into the composition then?  Or is this used for something else.

I'm not even sure what we should be doing around these, just wondering.

crates/composition/tests/composition/shareable_basic/federated.graphql
```diff
@@ -1,11 +1,12 @@
 type Customer {
-    id: ID
+    id: ID!
     name: String
     other: Int
     newsletterSubscribed: Boolean
-    subscriptionPlan: Plan
+    subscriptionPlan: Plan!
```
obmarg, line 6: I assume all these non-nullable changes are expected?

crates/federated-graph/src/federated_graph.rs
```diff
@@ -0,0 +1,276 @@
...
+    FieldId + fields + Field,
+    FieldTypeId + field_types + FieldType,
+    InputObjectId + input_objects + InputObject,
+    InterfaceId + interfaces + Interface,
+    ObjectId + objects + Object,
+    ScalarId + scalars + Scalar,
+    StringId + strings + String,
+    SubgraphId + subgraphs + Subgraph,
+    UnionId + unions + Union,
+}
```
obmarg, line 276: Nice

obmarg (approved):

crates/composition/src/composition_ir.rs
obmarg: Agreed, I'm not really clear what this field is from it's name

obmarg: Particularly the `field` prefix - kind of looks like it's just a map from types to kinds?  

Though maybe I'm misunderstanding - as I said, finding it quite difficult to follow all the functions in here with the nested helper functions

crates/composition/src/subgraphs.rs
tomhoule: The reason I had to do this is that we have the `FieldTypeKind` enum. We can either special case the builtin scalars in there with an additional enum variant or add them to the data structure. I don't have strong opinions either way. I also had serialization to protobuf in mind at the time I think, where enums with data aren't that easy to model, so it would have been a DefinitionKind enum and an int. But with serve, we could have a BuiltinScalar variant that contains an enum, easily.

crates/composition/tests/composition/shareable_basic/federated.graphql
tomhoule: Yes, we weren't writing the bangs and list type wrappers at all before, so all types were written as nullable non-list. Now they have the wrappers.

crates/composition/src/lib.rs
tomhoule: Yes both good ideas. I'll experiment and see what look and feels better.

crates/composition/src/subgraphs.rs
tomhoule: One advantage of this is that it's fewer special cases, fewer types, all scalars are scalars (builtin or custom). But there's definitely many cases where we want special treatment for the builtin scalars. So it can make sense to have them separate. I'm a bit undecided.

crates/composition/src/composition_ir.rs
tomhoule: It's a map from a name to a field type (but field type was already taken, so it became field type kind, which isn't a good name). I'll at least add a comment here and maybe more in other places, and rename the field and the enum.

tomhoule: Where field type is actually a definition, so a scalar, or an enum, or an interface, etc.

crates/composition/src/lib.rs
tomhoule: I went with the extension trait, I think it does read better.

crates/composition/src/composition_ir.rs
tomhoule: Went on and did that, I think it does read better.

crates/composition/src/subgraphs.rs
obmarg: Yeah, we're going to have to special case them _somewhere_ either way.  Not sure where it would be better to do that, but I'm fine with this for now - can always change it if it turns out to not make sense.

crates/composition/src/composition_ir.rs
```diff
@@ -0,0 +1,179 @@
...
+/// structure is distinct from [FederatedGraph](grafbase_federated_graph::FederatedGraph) because
+/// it is accumulated out of order during composition. Only after all composed top-level
+/// definitions (objects, interfaces, scalars, enums, unions) are known can we construct
+/// [federated::FieldTypeId]s and [federated::ObjectId]s for union members for the federated graph.
+///
+/// This is a **write only** data structure during composition. The source of truth for the
+/// contents of the federated graph is the subgraphs.
+#[derive(Default)]
+pub(crate) struct CompositionIr {
+    pub(crate) definitions_by_name: HashMap<subgraphs::StringId, federated::Definition>,
```
obmarg, line 18: This seems much better

crates/composition/src/subgraphs.rs
tomhoule: Sounds good :+1:

PR #902, opened 2023-11-06
composition: populate entity provenance

tomhoule:
Populate the grafbase_federated_graph::Object::resolvable_keys list for each object.

closes GB-5266

# Type of change

- [ ] 💔 Breaking
- [x] 🚀 Feature
- [ ] 🐛 Fix
- [ ] 🛠️ Tooling
- [ ] 🔨 Refactoring
- [ ] 🧪 Test
- [ ] 📦 Dependency

### DIFF
--- a/crates/composition/src/emit_federated_graph.rs
+++ b/crates/composition/src/emit_federated_graph.rs
@@ -2,12 +2,19 @@ mod field_types_map;
 
 use self::field_types_map::FieldTypesMap;
 use crate::{
-    composition_ir::{CompositionIr, FieldIr, StringsIr},
+    composition_ir::{CompositionIr, FieldIr, KeyIr, StringsIr},
     subgraphs, Subgraphs, VecExt,
 };
 use grafbase_federated_graph as federated;
 use itertools::Itertools;
 use std::collections::BTreeSet;
+
+struct Context<'a> {
+    strings_ir: &'a mut StringsIr,
+    field_types_map: &'a mut FieldTypesMap,
+    out: &'a mut federated::FederatedGraph,
+    subgraphs: &'a Subgraphs,
+}
 
 /// This can't fail. All the relevant, correct information should already be in the CompositionIr.
 pub(crate) fn emit_federated_graph(
@@ -26,15 +33,18 @@ pub(crate) fn emit_federated_graph(
         ..Default::default()
     };
 
-    emit_fields(
-        &ir.fields,
+    let mut ctx = Context {
+        strings_ir: &mut strings_ir,
+        field_types_map: &mut field_types_map,
+        out: &mut out,
         subgraphs,
-        &mut strings_ir,
-        &mut field_types_map,
-        &mut out,
-    );
+    };
 
-    emit_union_members(&ir.union_members, &mut field_types_map, &mut out);
+    emit_subgraphs(&mut ctx);
+    emit_fields(&ir.fields, &mut ctx);
+
+    emit_union_members(&ir.union_members, &mut ctx);
+    emit_keys(&ir.resolvable_keys, &mut ctx);
 
     out.field_types = field_types_map.field_types;
     out.strings = strings_ir.strings;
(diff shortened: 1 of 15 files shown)


crates/composition/src/emit_federated_graph.rs
```diff
@@ -129,3 +144,70 @@ fn emit_union_members(
...
+                    .object_fields
+                    .iter()
+                    .find(|object_field| {
+                        let field_name = ctx.out[object_field.field_id].name;
+                        object_field.object_id == object_id
+                            && ctx.strings_ir[field_name]
+                                == ctx.subgraphs.walk(selection.field).as_str()
+                    })
+                    .map(|of| of.field_id)
+                    .unwrap(),
```
tomhoule, line 182: Ugly and inefficient. But we're trying to iterate fast, so I decided not to improve. If you feel it would be better, I'm glad to do it.

crates/composition/src/emit_federated_graph.rs
```diff
@@ -129,3 +144,70 @@ fn emit_union_members(
...
+        .map(|selection| {
+            let field = match parent_id {
+                federated::Definition::Object(object_id) => ctx
+                    .out
+                    .object_fields
+                    .iter()
+                    .find(|object_field| {
+                        let field_name = ctx.out[object_field.field_id].name;
+                        object_field.object_id == object_id
+                            && ctx.strings_ir[field_name]
```
tomhoule, line 178: I don't like that you can index by both ctx.strings_ir and ctx.out here, and only the first one is correct. I'll fix that now or later depending on feedback. There's an identical situation with FieldTypesMap.

tomhoule: (because ctx.out is the federated graph and we haven't put the strings in there yet, in these functions, they're still in strings_ir at that point in time)

Finistere (approved):

crates/composition/src/emit_federated_graph.rs
tomhoule: Yep, that's the case since `emit_keys` comes after `emit_fields` (and the objects are already inserted earlier during composition). The Vecs in FederatedGraph aren't sorted, the only guarantee is that the ids are correct. Everything is produced already valid, there's no binary search involved.

Finistere: if ids are correct, should the `object_id` be enough to find the right fields?

tomhoule: So in the scope of this function we have:

1. The federated graph with all fields and objects
2. The object (or interface) id to which the selection set applies to
3. A `StringId` for the selection (the name of the selected field)

What we do here is find find an `ObjectField` (or `InterfaceField`) in (1.) where (2.) is the ObjectField::object_id and (3.) is the name of field referenced by ObjectField::field_id.

Where we could be more efficient is in finding the right `ObjectField`s. Maybe a binary search on the ObjectField::object_id could work to narrow down, but I'm not sure we can guarantee the that the vec is sorted here — I can check. The other solution would be building some kind of index (`HashMap<(ObjectId, StringId), FieldId>` or similar).

Finistere: mixed the structure, thought for a moment that it was `ObjectFields { object_id, fields: Vec<FieldId> }`. I would sort the vec to generate the keys. 🤷 
It doesn't need to be part of the contract.

crates/composition/src/emit_federated_graph.rs
tomhoule: Fixed in 083e6f3dcd370e8d01cde4f120f26ef83e34b0c1 — it does feel much better now. Strings and field types are in a single place.

crates/composition/src/emit_federated_graph.rs
tomhoule: Fixed in 058cff7d5048b9d3a989610349227a538411bf28

PR #904, opened 2023-11-07
composition: implement provenance for object fields

tomhoule:
That is to say, which field in the federated graph is resolvable by
which subgraph.

This PR also moves the `strings` module from the top level inside `subgraphs` for clarity, no change other than that, it's just moved.

closes GB-5267

# Type of change

- [ ] 💔 Breaking
- [x] 🚀 Feature
- [ ] 🐛 Fix
- [ ] 🛠️ Tooling
- [ ] 🔨 Refactoring
- [ ] 🧪 Test
- [ ] 📦 Dependency


.github/workflows/cli.yml
```diff
@@ -408,7 +408,7 @@ jobs:
         if: ${{ contains(needs.*.result, 'failure') || contains(needs.*.result, 'cancelled') }}
       - name: Comment PR
         uses: thollander/actions-comment-pull-request@v2
-        if: ${{ github.event_name == 'pull_request' }}
+        if: ${{ github.event_name == 'pull_request' && contains(needs.*.result, 'success') }}
```
tomhoule, line 411: Without this extra condition, the action posts a comment about CLI binaries even when the CLI build jobs are skipped.

crates/composition/src/compose.rs
```diff
@@ -131,12 +131,19 @@ fn merge_field_definitions(ctx: &mut Context<'_>, fields: &[FieldWalker<'_>]) {
     }
 
     let arguments = object::merge_field_arguments(*first, fields);
+    let resolvable_in = fields
+        .iter()
+        .map(|field| {
+            grafbase_federated_graph::SubgraphId(field.parent_definition().subgraph().id.idx())
+        })
+        .collect();
```
obmarg, line 139: How are we going to deal with fields that are present in multiple subgraphs?

obmarg: Forget I spoke, apparently I'm still half asleep and can't read

obmarg (approved):

PR #915, opened 2023-11-08
fix: don't assume templates are SDL based in init

obmarg:
`grafbase init` can set up a project from a template.  The output assumed that all the templates were written using SDL though, and always pointed the user at a `schema.graphql` file.  This makes it a little smarter: it'll now do the right thing depending on the template.

Unfortunately had to get rid of a `tokio::main` attr because it didn't like  return values, but 🤷‍♂️.

### DIFF
--- a/cli/Cargo.lock
+++ b/cli/Cargo.lock
@@ -4480,7 +4480,7 @@ dependencies = [
 [[package]]
 name = "postgres-protocol"
 version = "0.6.6"
-source = "git+https://github.com/grafbase/rust-postgres/?branch=grafbase#69df3c14a26e604396ef46caa7ded791686f34ce"
+source = "git+https://github.com/grafbase/rust-postgres/?branch=grafbase#9913ef3e5e61205d232773ac3caabbc95d310944"
 dependencies = [
  "base64 0.21.2",
  "byteorder",
@@ -4498,7 +4498,7 @@ dependencies = [
 [[package]]
 name = "postgres-types"
 version = "0.2.6"
-source = "git+https://github.com/grafbase/rust-postgres/?branch=grafbase#69df3c14a26e604396ef46caa7ded791686f34ce"
+source = "git+https://github.com/grafbase/rust-postgres/?branch=grafbase#9913ef3e5e61205d232773ac3caabbc95d310944"
 dependencies = [
  "bytes",
  "fallible-iterator",
@@ -6695,7 +6695,7 @@ dependencies = [
 [[package]]
 name = "tokio-postgres"
 version = "0.7.10"
-source = "git+https://github.com/grafbase/rust-postgres/?branch=grafbase#69df3c14a26e604396ef46caa7ded791686f34ce"
+source = "git+https://github.com/grafbase/rust-postgres/?branch=grafbase#9913ef3e5e61205d232773ac3caabbc95d310944"
 dependencies = [
  "async-trait",
  "byteorder",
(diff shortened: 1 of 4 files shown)


cli/crates/backend/src/project.rs
```diff
@@ -96,95 +96,105 @@ pub enum Template<'a> {
 /// - returns [`BackendError::GetRepositoryInformation`] if the request to get the information for a repository returned a non 200-299 status
 ///
 /// - returns [`BackendError::ReadRepositoryInformation`] if the request to get the information for a repository returned a response that could not be parsed
-#[tokio::main]
-pub async fn init(name: Option<&str>, template: Template<'_>) -> Result<(), BackendError> {
+// #[tokio::main]
```
obmarg, line 99:
```suggestion
```

cli/crates/backend/src/project.rs
```diff
@@ -96,95 +96,104 @@ pub enum Template<'a> {
...
                         &project_path,
                         GRAFBASE_SDK_PACKAGE_NAME,
                         GRAFBASE_SDK_PACKAGE_VERSION,
                     )
-                    .map_err(Into::into);
+                    .map_err(BackendError::CommonError)?;
+                    return Ok(ConfigType::TypeScript);
+                }
+
+                Ok(ConfigType::GraphQL)
```
obmarg, line 152: This diff looks huge, but these return values are the only non-indentation change

cli/crates/backend/src/project.rs
```diff
@@ -96,95 +96,104 @@ pub enum Template<'a> {
...
 
-            let dot_env_write_result = fs::write(dot_env_path, DEFAULT_DOT_ENV).map_err(BackendError::WriteSchema);
+                schema_write_result?;
+                dot_env_write_result?;
 
-            if schema_write_result.is_err() || dot_env_write_result.is_err() {
-                tokio::fs::remove_dir_all(&grafbase_path)
-                    .await
-                    .map_err(BackendError::DeleteGrafbaseDirectory)?;
+                Ok(config_type)
```
obmarg, line 193: Also this one

cli/crates/cli/src/output/report.rs
```diff
@@ -51,12 +51,12 @@ pub fn start_dev_server(resolvers_reported: bool, port: u16, start_port: u16) {
     );
 }
 
-pub fn project_created(name: Option<&str>, template: Template<'_>) {
+pub fn project_created(name: Option<&str>, config_type: ConfigType /*template: Template<'_>*/) {
```
obmarg, line 54:
```suggestion
pub fn project_created(name: Option<&str>, config_type: ConfigType) {
```

cli/Cargo.lock
```diff
@@ -4480,7 +4480,7 @@ dependencies = [
 [[package]]
 name = "postgres-protocol"
 version = "0.6.6"
-source = "git+https://github.com/grafbase/rust-postgres/?branch=grafbase#69df3c14a26e604396ef46caa7ded791686f34ce"
+source = "git+https://github.com/grafbase/rust-postgres/?branch=grafbase#9913ef3e5e61205d232773ac3caabbc95d310944"
```
tomhoule, line 4483: Is this expected?

tomhoule (approved):
Nice!

cli/Cargo.lock
obmarg: No, no it is not - well spotted.

PR #918, opened 2023-11-08
fix: `@deprecated` directive

obmarg:
While testing the federation subgraph stuff, I noticed we don't support the `@deprecated` directive properly.  This is core GraphQL stuff so figured I should probably fix.

Part of GB-5001

### DIFF
--- a/engine/crates/parser-sdl/src/rules/enum_type.rs
+++ b/engine/crates/parser-sdl/src/rules/enum_type.rs
@@ -5,9 +5,11 @@
 //! TODO: Manage deprecation
 use engine::{registry, registry::MetaEnumValue};
 use engine_parser::types::TypeKind;
-use if_chain::if_chain;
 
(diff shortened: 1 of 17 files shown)


engine/crates/parser-sdl/src/rules/enum_type.rs
```diff
@@ -17,23 +19,31 @@ impl<'a> Visitor<'a> for EnumType {
...
+
+        ctx.registry.get_mut().create_type(
+            |_| {
+                registry::EnumType::new(type_name.clone(), values)
+                    .with_description(type_definition.node.description.clone().map(|x| x.node))
+                    .into()
+            },
+            &type_name,
+            &type_name,
+        );
```
tomhoule, line 47: nice

tomhoule (commented):
:100:

PR #920, opened 2023-11-08
feat(cli): Local federation development mode

pimeys:
Adds local development mode for federated graphs.

To start the server:

```
gb dev --federated
```

This doesn't need to be in a project (yet), but that will change this week.

Then go into a project which is a subgraph, start the dev in there

```
gb dev
```

And finally publish the subgraph to the federated graph:

```
gb publish --dev --name one --url "http://localhost:4001/graphql" -H "x-api-key: kekw"
```

The URL should be the one which answers to GraphQL queries in the subgraph. It'll introspect the schema and compose it to the subgraph.

Closes: GB-5343


cli/crates/federated-dev/src/dev/composer/messages.rs
```diff
@@ -0,0 +1 @@
+
```
obmarg, line 1: Is this file meant to be here?

cli/crates/federated-dev/src/dev/admin/types.rs
```diff
@@ -0,0 +1,38 @@
...
+#[derive(InputObject)]
+pub(crate) struct PublishSubgraphInput {
+    pub(crate) name: String,
+    pub(crate) url: Url,
+    pub(crate) headers: Vec<Header>,
+}
+
+#[derive(SimpleObject)]
+pub(crate) struct PublishSubgraphSuccess {
+    __typename: &'static str,
```
obmarg, line 29: Quite odd that you're defining this - `__typename` implicitly exists on all GraphQL composite outputs.

Could you just do `struct PublishSubgraphSuccess` and save yourself some effort?

cli/crates/federated-dev/src/subgraph.rs
```diff
@@ -0,0 +1,61 @@
...
+            url: "{url}",
+            headers: [{headers}]
+          }}) {{ typename }}
+        }}
+    "#};
+
+    let request = json!({
+        "query": mutation,
+        "variables": {}
+    });
```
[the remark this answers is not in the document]
obmarg: Oh, also - where's the `typename` field coming from?  It's defined as `__typename` in your struct.  (Imagine async-graphql might be stripping off the `__` so your object has both `typename` (the field you defined) _and_ `__typename` (the implicit field).

Might be worth cleaning up?

obmarg (approved):
Good stuff, just a few small comments

cli/crates/federated-dev/src/dev/composer.rs
```diff
@@ -0,0 +1,170 @@
...
+
+    fn ingest_subgraphs(&self, add_new: Option<(&str, &Subgraph)>) -> Subgraphs {
+        let mut subgraphs = Subgraphs::default();
+
+        for (name, subgraph) in &self.graphs {
+            if add_new.is_some_and(|(added_name, _)| added_name == name) {
+                continue;
+            }
+
+            subgraphs.ingest(subgraph.schema(), name);
```
obmarg, line 94: I just changed the signature of (I think) this function btw - sorry for the breakage.  Probably want to pull in `main` and add `subgraph.url()` (or whatever it is) on the end

cli/crates/federated-dev/src/dev/admin/types.rs
pimeys: async-graphql expects at least one field in the object. I don't really want to return anything if successful, so I wonder what we could have here?

obmarg: Could just return a `boolean` or similar rather than a whole object?

pimeys: Ah, it was late yesterday, of course...

cli/crates/cli/src/cli_input.rs
```diff
@@ -340,24 +354,60 @@ pub struct SchemaCommand {
...
+    /// Add a header to the introspection request
+    #[clap(short = 'H', long, value_parser, num_args = 0..)]
+    header: Vec<String>,
+}
+
+impl PublishCommand {
+    pub(crate) fn headers(&self) -> impl Iterator<Item = (&str, &str)> {
+        self.header.iter().map(|header| {
+            let mut splitted = header.split(':');
+            (splitted.next().unwrap_or(""), splitted.next().unwrap_or("").trim())
```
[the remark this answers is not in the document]
pimeys: Yeah I have an idea how to do this correctly. It kind of requires a proper parser, but luckily a proper parser here is _very_ simple.

PR #925, opened 2023-11-09
fix: Correctly rename input variables for GraphQL connector

Finistere:
# Description

# Type of change

- [ ] 💔 Breaking
- [ ] 🚀 Feature
- [ ] 🐛 Fix
- [ ] 🛠️ Tooling
- [ ] 🔨 Refactoring
- [ ] 🧪 Test
- [ ] 📦 Dependency

### DIFF
--- a/engine/crates/engine/src/registry/resolvers/graphql/serializer.rs
+++ b/engine/crates/engine/src/registry/resolvers/graphql/serializer.rs
@@ -12,9 +12,10 @@ use engine_parser::{
     Positioned,
 };
 use engine_value::{Name, Value};
+use type_names::WrappingType;
 
 use super::Target;
-use crate::registry::{type_kinds::SelectionSetTarget, type_names, MetaField, Registry};
+use crate::registry::{type_kinds::SelectionSetTarget, type_names, MetaField, MetaTypeName, Registry};
 
 /// Serialize a list of [`Selection`]s into a GraphQL query string.
 ///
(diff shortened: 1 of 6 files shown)


engine/crates/engine/src/registry/resolvers/graphql/serializer.rs
```diff
@@ -469,10 +470,22 @@ impl<'a: 'b, 'b: 'a, 'c: 'a> Serializer<'a, 'b> {
...
+            .strip_prefix(self.prefix.unwrap_or_default())
+            .unwrap_or(ty)
+            .to_string();
+        for wrapper in wrappers {
+            ty = match wrapper {
+                WrappingType::List => format!("[{}]", ty),
+                WrappingType::NonNull => format!("{}!", ty),
+            };
+        }
+        ty
```
obmarg, line 488: Imagine we're not doing this enough times for it to matter, but you could do this with less repeated string allocations:

```suggestion
        let wrappers = WrappingType::all_for(ty).collect::<Vec<_>>();
        let mut ty = String::with_capacity(ty.len());
        for wrapper in &wrappers {
            if let WrappingType::List = wrapper {
                write!(&mut ty, "[").ok();
            }
        }
        
        let stripped_type = MetaTypeName::concrete_typename(ty)
             .strip_prefix(self.prefix.unwrap_or_default())
            .unwrap_or(ty)
            .to_string();
        write!(&mut ty, "{stripped_type}", ).ok();
        
        for wrapper in wrappers.iter().rev() {
            match wrapper {
                WrappingType::List => write!(&mut ty, "]").ok(),
                WrappingType::NonNull => write!(&mu ty, "!").ok(),
            };
        }
        ty
```

Finistere: oh nice! Wasn't my proudest piece of code to be honest xD

obmarg (approved):

PR #927, opened 2023-11-10
Initial query execution flow

Finistere:
# Description

# Type of change

- [ ] 💔 Breaking
- [ ] 🚀 Feature
- [ ] 🐛 Fix
- [ ] 🛠️ Tooling
- [ ] 🔨 Refactoring
- [ ] 🧪 Test
- [ ] 📦 Dependency


engine/crates/engine-v2/src/response_graph/mod.rs
```diff
@@ -0,0 +1,137 @@
...
+    Object(ObjectNodeId),
+}
+
+// Maybe an enum? Need to support `__typename`
+pub struct FieldEdge {
+    // probably needs a better name. it's the position for requested fields. For added fields,
+    // it's the position of the query field that needed it.
+    pub pos: Pos,
+    // resolving fragments eagerly, it makes manipulating SelectionSet easier during planning.
+    pub type_condition: Option<TypeCondition>,
```
obmarg, line 120: Have you given any thought to the other uses of fragments?  I can see this approach working alright for type conditions, but not so sure about `@defer`, `@skip`, `@include`?  Not so sure

engine/crates/engine-v2/src/plan/logical/graph.rs
```diff
@@ -0,0 +1,93 @@
+use super::LogicalPlan;
+
+#[derive(PartialEq, Eq, PartialOrd, Ord, Clone, Copy, Hash)]
+pub struct PlanId(usize); // should be a u16, more than enough for any sane query
```
obmarg, line 4: I'd vote u32 but sure

engine/crates/engine-v2/src/response_graph/mod.rs
Finistere: excellent point. for `@skip` & `@include` it shouldn't be hard as the condition can be detected quite easily. For `@defer`, it's definitely more tricky. I think in this case we would probably have a `deferred` field in the `NodeSelectionSet` containing the deferred fragment fields. A data source might only be responsible of only some parts of fragment fields, so I'm a bit skeptical of keeping the fragment notion which does add more complexity. The only benefit I see is more readable queries generated against subgraphs. A valid one, but not critical imho.

engine/crates/engine-v2/src/plan/logical/graph.rs
Finistere: removed the comment, we'll see it another day. Given how the plans are currently generated I think 65K is more than enough. But it's a decision that can be easily changed later. 🤷

engine/crates/engine-v2/src/executor/mod.rs
```diff
@@ -0,0 +1,51 @@
...
+impl Executor {
+    fn build(engine: &Engine, resolver: &schema::Resolver, request: ExecutorRequest<'_>) -> Self {
+        match resolver {
+            Resolver::Subgraph(resolver) => graphql::GraphqlExecutor::build(engine, resolver, request),
+        }
+    }
+
+    async fn execute(self, proxy: ResponseGraphProxy) {
+        match self {
+            Executor::GraphQL(plan) => plan.execute(proxy).await,
```
obmarg, line 36: Nitpicking , but the thing inside the `Executor::GraphQL` is a `GraphQlExecutor` - so `executor` seems like a less confusing name.

engine/crates/engine-v2/schema/src/conversion.rs
```diff
@@ -281,3 +299,9 @@ from_id_newtypes! {
     SubgraphId => SubgraphId,
     UnionId => UnionId,
 }
+
+impl DataSourceId {
+    fn from_subgraph_id(id: federated_graph::SubgraphId) -> Self {
+        Self(NonZeroU32::new(id.0 as u32 + 1).unwrap())
```
obmarg, line 305: Not sure I'm _against_ these `NonZeroU32`s - but they do seem like they're increasing our risk of off by one errors significantly...

obmarg (commented):
Not finished reviewing yet, but half my requests are for docstrings so going to submit my review and come back to it later

engine/crates/engine-v2/schema/src/lib.rs
```diff
@@ -76,8 +87,88 @@ pub struct Key {
     pub fields: SelectionSet,
 }
 
-pub type SelectionSet = Vec<Selection>;
+#[derive(Default, Clone)]
+pub struct SelectionSet {
```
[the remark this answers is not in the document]
Finistere: The schema contains selection sets for `@requires` & `@provides`

engine/crates/engine-v2/schema/src/lib.rs
```diff
@@ -76,8 +87,88 @@ pub struct Key {
...
+    // sorted by field id
+    items: Vec<Selection>,
+}
+
+impl FromIterator<Selection> for SelectionSet {
+    fn from_iter<T: IntoIterator<Item = Selection>>(iter: T) -> Self {
+        let mut items = iter.into_iter().collect::<Vec<_>>();
+        items.sort_unstable_by_key(|selection| selection.field);
```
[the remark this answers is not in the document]
Finistere: Mostly to make `merge` easier as I'm manipulating those selection sets directly during planning. It's true that the spec doesn't explicitly specify those as unordered. But I don't think it really matters, no? In this case, it's the selection set used to represent `@provides` and `@requires`, but I'm ignoring any user order everywhere else for the response.

engine/crates/engine-v2/src/executor/mod.rs
Finistere: yeah leftover

engine/crates/engine-v2/schema/src/conversion.rs
Finistere: Very fair point, moved all the IDs into a sub-module so the only way to create them is through `From<usize>`. As I'm using `Option<ObjectId>` in response objects it felt like a waste of space to allocate additional bytes.

engine/crates/engine-v2/schema/src/translator.rs
```diff
@@ -0,0 +1,38 @@
+use crate::{FieldId, InterfaceId, ObjectId, Schema, UnionId};
+
+const TYPENAME: &str = "__typename";
+
+pub trait Translator {
+    fn field(&self, field_id: FieldId) -> &str;
+    fn typename(&self, object_id: ObjectId) -> &str;
+    fn union_typename_field(&self, union_id: UnionId) -> &str;
+    fn interface_typename_field(&self, interface_id: InterfaceId) -> &str;
+}
```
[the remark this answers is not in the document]
Finistere: removed it. The idea was to allow resolvers to define how to rename/translate fields. It was simple when reading from the response but a lot more tricky when writing data into it.

engine/crates/engine-v2/src/request.rs
```diff
@@ -0,0 +1,149 @@
...
+                            schema::Definition::Object(object_id) => self.schema.object_fields(object_id).collect(),
+                            schema::Definition::Interface(interface_id) => {
+                                self.schema.interface_fields(interface_id).collect()
+                            }
+                            _ => {
+                                return Err(ServerError::new(
+                                    format!("Field {name} does not have any fields."),
+                                    Some(pos),
+                                ));
+                            }
```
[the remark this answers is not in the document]
Finistere: yeah indeed, good catch

engine/crates/engine-v2/src/response_graph/selection_set/output.rs
```diff
@@ -0,0 +1,61 @@
+use super::{NodeSelection, NodeSelectionSet};
+use crate::response_graph::FieldEdgeId;
+
+#[derive(Default, Debug, Clone, PartialEq, Eq)]
+pub struct OutputNodeSelectionSet {
+    items: Vec<OutputNodeSelection>,
+}
```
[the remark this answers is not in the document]
Finistere: renamed those to `ReadSelectionSet` and `WriteSelectionSet`. They're respectively used to read from the response and write into it. So plan inputs and outputs. Compared to the bare `SelectionSet` they had more attributes to rename fields on the fly. Currently, they're the same after removing code, but it makes sense to me to keep them nonetheless imho.

engine/crates/engine-v2/src/response_graph/input.rs
```diff
@@ -0,0 +1,114 @@
+use serde::ser::{SerializeMap, SerializeSeq};
+use serde_json::Value;
+
+use super::{InputNodeSelectionSet, Node, ObjectNode, ObjectNodeId, ResponseGraph};
+
+pub struct Input<'a> {
```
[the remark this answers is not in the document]
Finistere: renamed to `ResponseObjectsView` it's used as the input for a plan avoiding any copies

engine/crates/engine-v2/schema/src/lib.rs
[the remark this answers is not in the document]
obmarg: The spec explicitly specifies that the order of output should match the order of the query: http://spec.graphql.org/October2021/#sec-Serialized-Map-Ordering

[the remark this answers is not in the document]
obmarg: Mutation fields also need to be executed serially in the order that they were specified in the operation: http://spec.graphql.org/October2021/#sec-Mutation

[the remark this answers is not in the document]
Finistere: Excellent catch.

engine/crates/engine-v2/src/executor/coordinator.rs
```diff
@@ -0,0 +1,93 @@
...
+                        // We should data back through a stream to later support additional data
+                        // being sent like @defer or streaming
+                        break;
+                    }
+                }
+            }
+        }
+
+        // Response will be sent through a byte stream back. Client shouldn't be aware of wether we
+        // defer / stream / etc.
```
obmarg, line 85: > Client shouldn't be aware of whether we // defer / stream / etc.

Just a note for the future: I don't think this layer should deal with the byte-stream aspect of it.  This should probably return a stream of some `Response: Serialize` type, and leave the converting-to-byte-stream bit up to a higher layer.

Finistere: indeed, would definitely be better

obmarg: There's far too many different transports (and http libraries) for the core engine to be concerning itself with all those details

engine/crates/engine-v2/src/request.rs
```diff
@@ -0,0 +1,152 @@
...
+            OperationType::Mutation => self
+                .schema
+                .root_operation_types
+                .mutation
+                .expect("Mutation operation type not supported by schema."),
+            OperationType::Subscription => self
+                .schema
+                .root_operation_types
+                .subscription
+                .expect("Subscription operation type not supported by schema."),
```
obmarg, line 42: This function returns a Result, so these should probably be `ok_or_else(...)?`

obmarg (approved):
Good work.  Still a bunch to do, but lets get this merged sometime and iterate

engine/crates/engine-v2/src/request.rs
Finistere: oh yeah definitely. I'll need to make a first pass over the code to add some error management which is pretty much nonexistent currently. ^^

PR #928, opened 2023-11-10
updated default configuration files GB-4618

fbjork:
# Description

* Updates the default SDL/TS config to use connectors instead of database models
* Updates the copy for newly initialized projects

# Type of change

- [ ] 💔 Breaking
- [ ] 🚀 Feature
- [ ] 🐛 Fix
- [ ] 🛠️ Tooling
- [ ] 🔨 Refactoring
- [ ] 🧪 Test
- [ ] 📦 Dependency

### DIFF
--- a/cli/.prettierrc.json
+++ b/cli/.prettierrc.json
@@ -1,5 +1,5 @@
 {
-  "semi": true,
+  "semi": false,
   "trailingComma": "all",
   "singleQuote": true,
   "printWidth": 120,
--- a/cli/crates/backend/assets/grafbase.default.config.ts
+++ b/cli/crates/backend/assets/grafbase.default.config.ts
@@ -1,54 +1,40 @@
-import { g, auth, config } from '@grafbase/sdk'
+import { g, config } from '@grafbase/sdk'
 
 // Welcome to Grafbase!
-// Define your data models, integrate auth, permission rules, custom resolvers, search, and more with Grafbase.
-// Integrate Auth
-// https://grafbase.com/docs/auth
 //
-// const authProvider = auth.OpenIDConnect({
-//   issuer: process.env.ISSUER_URL ?? ''
-// })
+// Configure authentication, data sources, resolvers and caching for your GraphQL API.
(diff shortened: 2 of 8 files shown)


cli/.prettierrc.json
```diff
@@ -1,5 +1,5 @@
 {
-  "semi": true,
+  "semi": false,
```
yoav-lavi, line 2: Why was this changed?
It'll likely case a lot of formatting changes if we have TS/JS files

fbjork: This should have been the default, this is what we have in the root config already and in other repos. I'll run formatting to ensure it's consistent.

yoav-lavi: There's other differences with this one (which is used for the CLI project), we could discuss using the same one but I'd do it separately personally

fbjork: Yes, will do it separately to ensure consistency across the mono repo.

cli/crates/backend/assets/grafbase.default.config.ts
```diff
@@ -1,54 +1,40 @@
...
+// g.datasource(pg)
 
-  // Extend models with resolvers
-  // https://grafbase.com/docs/edge-gateway/resolvers
-  // gravatar: g.url().resolver('user/gravatar')
+// Resolvers - https://grafbase.com/docs/resolvers
+//
+g.query('helloWorld', {
+  returns: g.string(),
+  resolver: 'hello-world',
```
obmarg, line 16: Kind of wonder if we should have a `hello-world` resolver in the default output so that this works...

obmarg (approved):

cli/crates/backend/assets/grafbase.default.config.ts
fbjork: Yeah I thought about this too, but didn't want to include any resolvers. by default. I could either comment this one out, but then the schema is empty. The error will only happen when you query the field.

PR #929, opened 2023-11-10
fix(cli): Change the project root semantics such as `grafbase/` becomes only a fallback for the project root

jakubadamw:
# Description

Introduce a fallback mechanism in searching the paths such as the current `grafbase` subdirectory remains supported. It will be phased out in the future. At the same time, the main search path for 'schema.graphql' and 'grafbase.config.ts' will now be the current directory.

# Type of change

- [ ] 💔 Breaking
- [X] 🚀 Feature
- [ ] 🐛 Fix
- [ ] 🛠️ Tooling
- [ ] 🔨 Refactoring
- [ ] 🧪 Test
- [ ] 📦 Dependency

### DIFF
--- a/cli/crates/common/src/environment.rs
+++ b/cli/crates/common/src/environment.rs
@@ -1,10 +1,10 @@
 #![allow(dead_code)]
 
-use crate::consts::AUTHORIZERS_DIRECTORY_NAME;
+use crate::consts::{AUTHORIZERS_DIRECTORY_NAME, GRAFBASE_DIRECTORY_NAME};
 use crate::types::UdfKind;
 use crate::{
     consts::{
-        DATABASE_DIRECTORY, DOT_GRAFBASE_DIRECTORY, GRAFBASE_DIRECTORY_NAME, GRAFBASE_HOME, GRAFBASE_SCHEMA_FILE_NAME,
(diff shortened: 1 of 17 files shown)


cli/crates/common/src/environment.rs
```diff
@@ -195,20 +192,15 @@ impl Project {
...
 
-        let grafbase_directory_path = schema_path
-            .parent()
-            .expect("the schema directory must have a parent by definiton")
-            .to_path_buf();
-
-        let path = grafbase_directory_path
+        let path = schema_path
             .parent()
             .expect("the grafbase directory must have a parent directory by definition")
```
tomhoule, line 197: Do we always expect a parent here? It seems unlikely that users would have their project at the filesystem root, but it's not unthinkable.

tomhoule (approved):
I'm not the most familiar with the CLI internals but this looks good :+1:

cli/crates/common/src/environment.rs
jakubadamw: `schema_path` would then be `/schema.graphql` or `/grafbase.config.ts`, so `parent()` for those is '/'.

tomhoule: Ah yes, good point, all good then :+1:

PR #930, opened 2023-11-10
cli: make codegen opt-in with an experimental feature

tomhoule:
Unfortunately we can't use the same mechanism as the other experimental features, because the typed resolvers logic is outside of the engine and doesn't have access to the registry.

Confirmed that this works locally, ideally there would be automated tests, but this should go in today's release.

closes GB-5332

### DIFF
--- a/cli/crates/typed-resolvers/src/error.rs
+++ b/cli/crates/typed-resolvers/src/error.rs
@@ -1,33 +1,12 @@
 use engine_parser::Error as ParseError;
 use std::fmt;
 
-#[derive(Debug)]
+#[derive(thiserror::Error, Debug)]
 pub enum CodegenError {
+    #[error("The `codegen` experimental feature is not enabled")]
     ExperimentalFeatureNotEnabled,
-    ParseError(ParseError),
-    FmtError(fmt::Error),
-}
-
-impl fmt::Display for CodegenError {
-    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        match self {
-            CodegenError::ExperimentalFeatureNotEnabled => {
-                f.write_str("The `codegen` experimental feature is not enabled")
-            }
-            CodegenError::ParseError(err) => fmt::Display::fmt(err, f),
-            CodegenError::FmtError(err) => fmt::Display::fmt(err, f),
-        }
-    }
-}
-
-impl From<fmt::Error> for CodegenError {
-    fn from(value: fmt::Error) -> Self {
-        CodegenError::FmtError(value)
-    }
-}
-
-impl From<ParseError> for CodegenError {
-    fn from(value: ParseError) -> Self {
-        CodegenError::ParseError(value)
-    }
+    #[error(transparent)]
+    ParseError(#[from] ParseError),
+    #[error(transparent)]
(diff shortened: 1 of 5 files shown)


obmarg (approved):

cli/crates/typed-resolvers/src/lib.rs
```diff
@@ -56,3 +61,20 @@ pub fn check_resolvers(resolvers_root: &Path, schema: &analyze::AnalyzedSchema<'
 
     AnalyzedResolvers { errs }
 }
+
+fn experimental_codegen_is_true(parsed_schema: &engine_parser::types::ServiceDocument) -> bool {
```
yoav-lavi, line 65: is_true feels a bit weird on a boolean function, "enabled" perhaps?

cli/crates/typed-resolvers/src/lib.rs
```diff
@@ -56,3 +61,20 @@ pub fn check_resolvers(resolvers_root: &Path, schema: &analyze::AnalyzedSchema<'
...
+    parsed_schema
+        .definitions
+        .iter()
+        .filter_map(|def| match def {
+            engine_parser::types::TypeSystemDefinition::Schema(schema_definition) => Some(schema_definition),
+            _ => None,
+        })
+        .flat_map(|def| def.node.directives.iter())
+        .any(|directive| {
+            directive.node.name.as_str() == "experimental"
```
yoav-lavi, line 75: Could we add consts for these?

cli/crates/typed-resolvers/src/error.rs
```diff
@@ -3,13 +3,17 @@ use std::fmt;
 
 #[derive(Debug)]
 pub enum CodegenError {
```
yoav-lavi, line 5: This isn't necessarily specific to this PR, but why not use this error for these?

cli/crates/typed-resolvers/src/lib.rs
tomhoule: Good point, doing this

cli/crates/typed-resolvers/src/error.rs
tomhoule: To avoid dependencies where they're not strictly necessary, but I can definitely change to thiserror for uniformity in this PR or later if that sounds better.

cli/crates/typed-resolvers/src/lib.rs
tomhoule: Done and pushed

cli/crates/typed-resolvers/src/lib.rs
tomhoule: Done and pushed

cli/crates/typed-resolvers/src/error.rs
yoav-lavi: Well the CLI already has thiserror so it should theoretically be fine

yoav-lavi: Whatever you prefer in terms of where to do it

tomhoule: done and pushed

PR #933, opened 2023-11-10
feat: GraphQL schema validation checks

tomhoule:
Based on the 2021 version of the GraphQL spec. This should cover the spec pretty much exhaustively. The error messages are using formulations users would be familiar with.

TODO

- README


engine/crates/validation/src/validate/directive_definitions.rs
```diff
@@ -0,0 +1,45 @@
...
+            .get(directive_name)
+            .map(|directive| directive.node.is_repeatable)
+            .unwrap_or(true)
+        {
+            return;
+        }
+
+        ctx.push_error(miette::miette!("{directive_name} is not repeatable"));
+    });
+}
```
obmarg, line 45: Should we also be checking that the directive is valid in its current position, or are we doing that somewhere else?

engine/crates/validation/src/validate/interface_implementers.rs
```diff
@@ -0,0 +1,256 @@
...
+// http://spec.graphql.org/draft/#IsSubType()
+fn validate_field_type_implements_interface_field(
+    interface_name: &str,
+    implementer_name: &str,
+    candidate_field: &Positioned<ast::FieldDefinition>,
+    interface_field: &Positioned<ast::FieldDefinition>,
+    ctx: &mut Context<'_>,
+) {
+    // Check that the wrapper types are compatible
+    {}
```
obmarg, line 155: This meant to be here?

engine/crates/validation/src/validate/schema_definition.rs
```diff
@@ -0,0 +1,104 @@
...
+        (def.subscription, "Subscription"),
+    ];
+
+    for idx in 0..(names.len()) {
+        let name = &names[idx];
+        let name = name.0.unwrap_or(name.1);
+        for other_name in names[idx..].iter().skip(1) {
+            let other_name = other_name.0.unwrap_or(other_name.1);
+            if name == other_name {
+                ctx.push_error(miette::miette!("Typed used twice in schema definition: `{name}`"));
```
obmarg, line 52: `Typed` looks like a typo?

```suggestion
                ctx.push_error(miette::miette!("Type used twice in schema definition: `{name}`"));
```

engine/crates/validation/src/validate/schema_definition.rs
```diff
@@ -0,0 +1,104 @@
...
+                    default.to_lowercase()
+                ));
+            }
+            // http://spec.graphql.org/draft/#sec-Root-Operation-Types
+            Some(referenced) => match referenced.node.kind {
+                ast::TypeKind::Object(_) => (),
+                _ => {
+                    let type_name = referenced.node.name.node.as_str();
+                    ctx.push_error(miette::miette!(
+                        "Query root type must be an Object type, it cannot be set to {type_name}"
```
obmarg, line 82: Should this error message say "Query/Mutation/Subscription root type" rather than just `Query root type`?

engine/crates/validation/tests/validation_errors/input_object_cycles.graphql
```diff
@@ -0,0 +1,21 @@
+input A {
+    otherA: A!
+}
+
+input B {
+  other: [B]
```
obmarg, line 6: Worth making this `[B!]!` so you get a proper test of lists that can't hit the field-is-nullable case?

obmarg (approved):
Nice, just a few small comments

engine/crates/validation/src/validate/arguments.rs
```diff
@@ -0,0 +1,51 @@
...
+            ValidateInputTypeResult::UnknownType => diagnostics::unknown_type(type_name, &location(), ctx),
+            ValidateInputTypeResult::NotAnInputType => {
+                diagnostics::output_type_in_input_position(type_name, &location(), ctx);
+            }
+        }
+
+        let is_non_null_without_default = !arg.node.ty.node.nullable && arg.node.default_value.is_none();
+        if is_non_null_without_default && arg.node.directives.iter().any(|dir| dir.node.name.node == "deprecated") {
+            ctx.push_error(miette::miette!(
+                "Required argument {}.{}({}:) cannot be deprecated.",
```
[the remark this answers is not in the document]
tomhoule: I started working with the draft spec, and it looks like that bit me — http://spec.graphql.org/draft/#sec--deprecated and a mention that required arguments can't be deprecated. Will remove, good catch!

engine/crates/validation/src/validate/directive_definitions.rs
tomhoule: Looks like an omission.

engine/crates/validation/src/validate/schema_definition.rs
tomhoule: good catch

engine/crates/validation/src/validate/schema_definition.rs
tomhoule: It's one of the not-so-great errors I copied from the ecosystem. I agree we should do better here, I'll adjust.

engine/crates/validation/src/validate/arguments.rs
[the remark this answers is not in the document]
obmarg: Ah, did wonder if you were working off a future spec.  Could maybe feature flag this under `draft` or something so we can easily revive it whenever the next release is out?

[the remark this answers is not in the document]
tomhoule: Yes we'll probably want config in some form so I might as well add it now. Another thing I noticed is that other implementations do not enforce that a type/enum/whatever that you `extend` exists in the same schema, but the spec is pretty clear. Another case we might want configuration for.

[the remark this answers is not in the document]
obmarg: Some of the federation subgraph stuff I was working with looked like it was using `extend` to indicate an extension of another subgraphs types - so definitely seems like might want to support both cases.

engine/crates/validation/src/validate/interface_implementers.rs
tomhoule: Done

engine/crates/validation/src/validate/schema_definition.rs
tomhoule: Done and pushed

engine/crates/validation/tests/validation_errors/input_object_cycles.graphql
tomhoule: Good point — done and pushed

obmarg (commented):
From the latest commit, looks like we might not be validating directives in a bunch of places?  At least fields, enum values, input fields, argument definitions, schema definition from what I can't see

engine/crates/validation/src/validate/arguments.rs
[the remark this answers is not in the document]
tomhoule: Should be addressed by https://github.com/grafbase/grafbase/pull/933/commits/0570025713303243db2aa8bc34d25c12c7f51440

tomhoule:
>  From the latest commit, looks like we might not be validating directives in a bunch of places? At least fields, enum values, input fields, argument definitions, schema definition from what I can't see

Likely forgotten, double checking based on the locations list

tomhoule:
Pushed the missing directive validations in cef922ecab33e7d380ffbc3b14c64acf9f22a5dc — I think your list was exhaustive.

PR #941, opened 2023-11-13
feat: implement rendering of join spec SDL from FederatedGraph

tomhoule:
Spec: https://specs.apollo.dev/join/v0.3/
Closes GB-5331
Prerequisite for GB-5331 as we'll need this for testing.

# Type of change

- [ ] 💔 Breaking
- [x] 🚀 Feature
- [ ] 🐛 Fix
- [ ] 🛠️ Tooling
- [ ] 🔨 Refactoring
- [ ] 🧪 Test
- [ ] 📦 Dependency


tomhoule:
One detail: every object/interface where an `@join__field(graph: G)` appears should have an `@join__type(graph: G)`, with or without selection set, according to the spec. It isn't implemented in this PR, still thinking about the requirements and how best to model that in FederatedGraph. I think we can do without for now.

engine/crates/federated-graph/src/render_sdl.rs
```diff
@@ -184,3 +195,44 @@ fn render_field_arguments(args: &[FieldArgument], graph: &FederatedGraph) -> Str
         out
     }
 }
+
+struct SelectionSetDisplay<'a>(&'a FieldSet, &'a FederatedGraph);
```
obmarg, line 199: Not important, but since this is printing a FieldSet should it be named`FieldSetDisplay`?

obmarg (approved):

PR #946, opened 2023-11-14
composition: release to crates.io

tomhoule:
This is the bare minimum change to get the crates to crates.io. We should do this now in the spirit of release early, release often, and improve from there. That will already enable depending on the crates.io version of the crates instead of the git repo in other places.


engine/crates/composition/Cargo.toml
```diff
@@ -2,9 +2,13 @@
 name = "grafbase-composition"
 version = "0.1.0"
 edition = "2021"
+license = "Apache-2.0"
+description = "An implementation of Grafbase federated schema composition"
+repository = "https://github.com/grafbase/grafbase"
```
fbjork, line 7: should this be https://github.com/grafbase/grafbase/tree/main/engine/crates/composition ?

tomhoule: done

engine/crates/federated-graph/Cargo.toml
```diff
@@ -1,7 +1,11 @@
 [package]
-name = "grafbase-federated-graph"
+name = "graphql-federated-graph"
 version = "0.1.0"
 edition = "2021"
+license = "Apache-2.0"
```
yoav-lavi, line 5: Not MPL2?

engine/crates/composition/Cargo.toml
```diff
@@ -1,10 +1,14 @@
 [package]
-name = "grafbase-composition"
-version = "0.1.0"
+name = "graphql-composition"
```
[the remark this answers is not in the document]
tomhoule: Yes, this is implementing the spec so the idea with the crates.io release is that it will end up making sense as open source that other projects can use.

engine/crates/federated-graph/Cargo.toml
tomhoule: I don't know, I aligned with the CLI license but I don't have an opinion on that.

yoav-lavi: (if this is implementing a spec as you said before maybe that's not relevant)

yoav-lavi (approved):
Nice!

tomhoule:
Linear issue link https://linear.app/grafbase/issue/GB-5226/publish-composition-crate-to-cratesio

PR #948, opened 2023-11-14
chore: make `MockGraphQlServer` generic over any schema

obmarg:
For testing federation we're going to need a variety of different GraphQL servers in our integration tests.  This commit is prep work for that: it makes the MockGraphQlServer able to support any `async_graphql` schema.

A follow up PR will add some that we can use for federation testing

### DIFF
--- a/engine/crates/integration-tests/src/mocks/graphql.rs
+++ b/engine/crates/integration-tests/src/mocks/graphql.rs
@@ -1,12 +1,13 @@
 //! A mock GraphQL server for testing the GraphQL connector
 
-use std::{net::TcpListener, time::Duration};
+use std::{net::TcpListener, sync::Arc, time::Duration};
 
-use async_graphql::{
-    EmptyMutation, EmptySubscription, InputObject, Interface, Object, Schema, SimpleObject, Union, ID,
-};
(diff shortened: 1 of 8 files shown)


engine/crates/integration-tests/src/mocks/graphql.rs
```diff
@@ -65,159 +73,36 @@ async fn graphql_handler(headers: HeaderMap, req: GraphQLRequest) -> GraphQLResp
...
-    value: String,
-}
-
-#[derive(SimpleObject)]
-struct PullRequest {
-    title: String,
-    checks: Vec<String>,
-    author: UserOrBot,
-}
+    // let schema = Schema::build(Query { headers }, EmptyMutation, EmptySubscription).finish();
```
obmarg, line 76:
```suggestion
```

Finistere: leftover?

Finistere (approved):

engine/crates/integration-tests/src/mocks/graphql.rs
obmarg: Yeah, spotted that in self-review, gone now

PR #950, opened 2023-11-14
test: federation tests of variables & fragments

obmarg:
- Added some helpers to reduce boilerplate in these tests
- Added some simple tests of fragments for whenever we support those
- Added some simple tests of variables for whenever we support those

Part of GB-5337


yoav-lavi (approved):
Nice work!

PR #953, opened 2023-11-15
feat: base engine v2

Finistere:
Refactored some parts to make more sense. Several parts have been moved
to a request module as they're referencing the operation
fields/path/etc. For example, I replaced the term `ResponsePath` with
`OperationPath` which is a better representation of what it actually is.
The real `ResponsePath` will be needed for errors as it'll take into
account array indices. I prefixed the structs with `Operation` felt more
appropriate than `Request`, but not entirely sure either.

The response is now also better organized with read & write sub-modules.

The integration tests still don't pass because I'm not generating a
proper GraphQL query nor do we have the subgraph url. But the rough flow seems ok. Basic errors and inserting data into the response seemed fine. So should probably be fine for introspection.

# Description

# Type of change

- [ ] 💔 Breaking
- [ ] 🚀 Feature
- [ ] 🐛 Fix
- [ ] 🛠️ Tooling
- [ ] 🔨 Refactoring
- [ ] 🧪 Test
- [ ] 📦 Dependency


engine/crates/engine-v2/src/executor/coordinator.rs
```diff
@@ -1,76 +1,106 @@
...
+            selection_set: planned_operation.selection_set,
+            operation_fields: planned_operation.fields,
+            response: Arc::new(Mutex::new(Response::new(planned_operation.strings))),
+            plans: planned_operation.plans,
         }
     }
 
     pub async fn execute(&mut self) -> serde_json::Value {
-        let (sender, mut receiver) = futures_channel::mpsc::unbounded::<CoordinationTask>();
+        let mut futures = FuturesUnordered::<Pin<Box<dyn Future<Output = CoordinationTask> + Send + 'static>>>::new();
```
obmarg, line 37: Is there a reason this can't be `FuturesUnordered::<BoxFuture<'static, CoordinationTask>>`?  Bit easier to read

obmarg (approved):

engine/crates/engine-v2/src/executor/coordinator.rs
Finistere: Nop it's exactly the same actually, just didn't think of it

Finistere: will change it

PR #954, opened 2023-11-15
fix: remove hardcoded JWTs from owner tests

obmarg:
These were hardcoded with an expiration date of 1700000000 which was yesterday, causing all the tests to suddenly fail. I've updated the tests to just generate their own JWTs, which should prevent this becoming an issue again.

Hardcoded JWTs in tests is kind of a bad idea anyway - harder to debug, harder to write, requires more documentation and you run into silly things like this.


yoav-lavi (approved):

jakubadamw (approved):
Lovely!

PR #957, opened 2023-11-15
composition: implement @provides, @requires, @external

tomhoule:
# Description

closes GB-5270

# Type of change

- [ ] 💔 Breaking
- [x] 🚀 Feature
- [ ] 🐛 Fix
- [ ] 🛠️ Tooling
- [ ] 🔨 Refactoring
- [ ] 🧪 Test
- [ ] 📦 Dependency

### DIFF
--- a/engine/crates/composition/src/compose/object.rs
+++ b/engine/crates/composition/src/compose/object.rs
@@ -1,7 +1,6 @@
-use std::collections::HashSet;
-
 use super::*;
 use crate::subgraphs::{FieldTypeId, StringId};
+use std::collections::HashSet;
 
 /// The arguments of a federated graph's fields are the interseciton of the subgraph's arguments for
 /// that field.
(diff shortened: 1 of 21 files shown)


tomhoule:
Parts of the PR are pretty verbose, maybe it needs some more love to make these sections more readable.

obmarg (approved):

engine/crates/composition/src/compose/object.rs
```diff
@@ -24,3 +26,67 @@ pub(super) fn merge_field_arguments<'a>(
...
+            key_subgraphs
+                .into_iter()
+                .map(|f| f.parent_definition().subgraph().name().as_str())
+                .join(", "),
+            non_key_subgraphs
+                .into_iter()
+                .map(|f| f.parent_definition().subgraph().name().as_str())
+                .join(", "),
+        ));
+    }
```
obmarg, line 70: I kind of thought subgraphs could have different sets of keys, so not sure about this.  Though I'm not sure if I've actually read that somewhere or if it was just an assumption on my part

Finistere: They can: https://www.apollographql.com/docs/federation/entities-advanced#differing-keys-across-subgraphs

PR #959, opened 2023-11-15
feat: graphql-composiiton now accepts a url

obmarg:
The output from `graphql-composition` has a `url` field in it, which was being populated from the `name`.  This adds another parameter for the `url` and populates from that instead.

Tried not to overthink this: it's just a plain &str argument rather than a `Url` or anything more strongly typed.  Though happy to change if anyone has preferences.


tomhoule (approved):
:100:

PR #960, opened 2023-11-16
fix(deploy): Fix the root path handling for newly created projects.

jakubadamw:
# Description

The traversal when constructing the archive should use a minimum depth of 1.
This also changes how the relative project root is determined (it isn't hardcoced to '.'). This coincidentally allows people to keep using UDFs outside the relative project root path, which should be a common use case.

# Type of change

- [ ] 💔 Breaking
- [ ] 🚀 Feature
- [X] 🐛 Fix
- [ ] 🛠️ Tooling
- [ ] 🔨 Refactoring
- [ ] 🧪 Test
- [ ] 📦 Dependency

### DIFF
--- a/cli/crates/backend/src/api/deploy.rs
+++ b/cli/crates/backend/src/api/deploy.rs
@@ -12,7 +12,7 @@ use cynic::http::ReqwestExt;
 use cynic::{Id, MutationBuilder};
 use reqwest::{header, Body, Client};
 use std::ffi::OsStr;
-use std::path::PathBuf;
+use std::path::Path;
 use tokio::fs::read_to_string;
 use tokio_util::codec::{BytesCodec, FramedRead};
 use tokio_util::compat::TokioAsyncReadCompatExt;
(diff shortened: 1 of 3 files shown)


tomhoule (approved):

cli/crates/backend/src/api/deploy.rs
```diff
@@ -57,8 +57,10 @@ pub async fn deploy() -> Result<(), ApiError> {
             .map_err(ApiError::AppendToArchive)?;
     }
 
-    let walker = WalkDir::new(&project.path).into_iter();
-    for entry in walker.filter_entry(|entry| entry_not_in_blacklist(entry, &project.path)) {
+    println!("Project path: {}", project.path.to_string_lossy());
```
yoav-lavi, line 60: @jakubadamw This should probably be removed

jakubadamw: Yes, sorry…

PR #962, opened 2023-11-16
feat: First passing integration-tests for engine v2

Finistere:
Added a simple query builder for the graphql executor and simplified a
few things. Converting the response to `serde_json::Value` for now as
I had forgotten it's nested...

The rest is just moving stuff around.

# Description

# Type of change

- [ ] 💔 Breaking
- [ ] 🚀 Feature
- [ ] 🐛 Fix
- [ ] 🛠️ Tooling
- [ ] 🔨 Refactoring
- [ ] 🧪 Test
- [ ] 📦 Dependency


engine/crates/engine-v2/src/executor/graphql/query.rs
```diff
@@ -0,0 +1,152 @@
...
+        Ok(())
+    }
+
+    fn write_arguments(&self, buffer: &mut Buffer, arguments: &Vec<OperationArgument>) -> Result<(), Error> {
+        if !arguments.is_empty() {
+            buffer.write("(")?;
+
+            let mut arguments = arguments.iter().peekable();
+            while let Some(argument) = arguments.next() {
+                buffer.add_variable_references(argument.value.variables_used().map(|name| name.to_string()));
```
jakubadamw, line 110: https://docs.rs/itertools/latest/itertools/trait.Itertools.html#method.format_with could make this a bit nicer (and reallocation-efficient!), but not a huge issue

jakubadamw: Same in the other functions.

jakubadamw (approved):
Nice!

tomhoule (approved):
:100:

engine/crates/engine-v2/schema/src/names.rs
```diff
@@ -0,0 +1,29 @@
...
+/// renames later.
+/// Argument names aren't supported given current schema design.
+pub trait Names {
+    fn field(&self, field_id: FieldId) -> &str;
+    fn object(&self, object_id: ObjectId) -> &str;
+    fn union(&self, union_id: UnionId) -> &str;
+    fn interface(&self, interface_id: InterfaceId) -> &str;
+}
+
+impl Names for &Schema {
```
obmarg, line 13: Does this need to be implemented for `&Schema`?

```suggestion
impl Names for Schema {
```

Finistere: ah good catch, misused `dyn`.

engine/crates/engine-v2/src/executor/graphql/query.rs
Finistere: Oh interesting, will take a look at it when extend this with directives & fragments. thanks!

engine/crates/engine-v2/src/plan/planner.rs
```diff
@@ -127,18 +129,17 @@ impl<'a> Planner<'a> {
         let mut plan_ids = vec![];
         let mut children = vec![];
         // will need to reworked with inputs adding fields.
```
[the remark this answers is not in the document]
Finistere: ops, indeed. good catch :)

PR #964, opened 2023-11-16
Adding introspetion fields to engine v2 schema

Finistere:
Adding the introspection fields into the schema for engine-v2. They're not used currently and they're missing a resolver, but they're there. :)

### DIFF
--- a/engine/crates/engine-v2/schema/src/introspection.rs
+++ b/engine/crates/engine-v2/schema/src/introspection.rs
@@ -1,0 +1,416 @@
+use std::{
+    collections::HashMap,
+    ops::{Deref, DerefMut},
+};
+
+use crate::{
+    Definition, EnumId, EnumValue, Field, FieldArgument, FieldId, FieldType, FieldTypeId, ObjectField, ObjectId,
+    ScalarId, Schema, StringId, Value, Wrapping,
+};
+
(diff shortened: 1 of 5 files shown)


engine/crates/engine-v2/schema/src/introspection.rs
```diff
@@ -0,0 +1,416 @@
...
+}
+
+impl<'a> DerefMut for IntrospectionFields<'a> {
+    fn deref_mut(&mut self) -> &mut Self::Target {
+        self.schema
+    }
+}
+
+impl<'a> IntrospectionFields<'a> {
+    // /!\ You MUST call schema.ensure_proper_ordering() after calling this function. /!\
```
obmarg, line 30: Seems like you could enforce this by taking ownership of `Schema` and having a `.finish()` function that calls `ensure_proper_ordering()` then returns ownership of the schema.

Probably safer than assuming people will read this comment

obmarg (approved):

engine/crates/engine-v2/schema/src/introspection.rs
Finistere: hum indeed, good point

Finistere: fixed in next pr

PR #967, opened 2023-11-16
composition: implement composite key support

tomhoule:
# Description

Concretely, entities defining keys like `@key(fields: "id name")` and
even nested selections like `@key(fields: "patient { id } recordId")`.

A lot of the complexity in this PR comes from the machinery to detect
that:

1. An object is part of an entity in a subgraph because one of its field
   is part of a nested key on another entity.
2. A field is part of a key, not on its parent object but another object
   that includes it in its key through a nested selection.

closes GB-5264

# Type of change

- [ ] 💔 Breaking
- [x] 🚀 Feature
- [ ] 🐛 Fix
- [ ] 🛠️ Tooling
- [ ] 🔨 Refactoring
- [ ] 🧪 Test
- [ ] 📦 Dependency

### DIFF
--- a/engine/crates/composition/src/compose/object.rs
+++ b/engine/crates/composition/src/compose/object.rs
@@ -30,7 +30,7 @@ pub(super) fn merge_field_arguments<'a>(
 pub(super) fn compose_object_fields<'a>(first: FieldWalker<'a>, fields: &[FieldWalker<'a>], ctx: &mut Context<'a>) {
     if fields
         .iter()
-        .filter(|f| !(f.is_shareable() || f.is_external() || f.is_key()))
+        .filter(|f| !(f.is_shareable() || f.is_external() || f.is_part_of_key()))
         .count()
         > 1
     {
@@ -45,8 +45,11 @@ pub(super) fn compose_object_fields<'a>(first: FieldWalker<'a>, fields: &[FieldW
         ));
     }
 
-    let first_is_key = first.is_key();
-    if fields.iter().any(|field| field.is_key() != first_is_key) {
+    let first_is_part_of_key = first.is_part_of_key();
+    if fields
+        .iter()
+        .any(|field| field.is_part_of_key() != first_is_part_of_key)
+    {
         let name = format!(
             "{}.{}",
             first.parent_definition().name().as_str(),
@@ -54,7 +57,7 @@ pub(super) fn compose_object_fields<'a>(first: FieldWalker<'a>, fields: &[FieldW
         );
         let (key_subgraphs, non_key_subgraphs) = fields
             .iter()
-            .partition::<Vec<FieldWalker<'_>>, _>(|field| field.is_key());
+            .partition::<Vec<FieldWalker<'_>>, _>(|field| field.is_part_of_key());
 
         ctx.diagnostics.push_fatal(format!(
             "The field `{name}` is part of `@key` in {} but not in {}",
--- a/engine/crates/composition/src/ingest_subgraph/nested_key_fields.rs
+++ b/engine/crates/composition/src/ingest_subgraph/nested_key_fields.rs
@@ -1,0 +1,47 @@
+use crate::subgraphs::*;
+
+/// See [crate::subgraphs::Keys::nested_key_fields].
+pub(super) fn ingest_nested_key_fields(subgraph_id: SubgraphId, subgraphs: &mut Subgraphs) {
+    subgraphs.with_nested_key_fields(|subgraphs, nested_key_fields| {
+        for definition in subgraphs.walk(subgraph_id).definitions() {
+            for key in definition.entity_keys() {
+                for field in key.fields() {
+                    let Some(field_type) = definition
+                        .find_field(field.field)
+                        .and_then(|field| field.r#type().definition(subgraph_id))
+                    else {
+                        continue;
+                    };
+
+                    for subselection_field in &field.subselection {
+                        ingest_nested_key_fields_rec(field_type, subselection_field, nested_key_fields);
+                    }
+                }
+            }
+        }
+    });
+}
+
+fn ingest_nested_key_fields_rec(
+    parent_definition: DefinitionWalker<'_>,
+    selection: &Selection,
+    nested_key_fields: &mut NestedKeyFields,
+) {
+    let Some(field) = parent_definition.find_field(selection.field) else {
+        return;
+    };
+
+    nested_key_fields.insert(field);
+
+    if selection.subselection.is_empty() {
+        return;
+    }
+
+    let Some(selection_field_type) = field.r#type().definition(parent_definition.subgraph().id) else {
+        return;
+    };
+
+    for subselection in &selection.subselection {
+        ingest_nested_key_fields_rec(selection_field_type, subselection, nested_key_fields);
+    }
+}
--- a/engine/crates/composition/src/subgraphs/definitions.rs
+++ b/engine/crates/composition/src/subgraphs/definitions.rs
@@ -1,6 +1,6 @@
 use super::*;
 
-#[derive(Clone, Copy, Debug, PartialEq, Eq, PartialOrd, Ord)]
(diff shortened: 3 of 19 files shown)


engine/crates/composition/src/compose/object.rs
```diff
@@ -45,16 +45,16 @@ pub(super) fn compose_object_fields<'a>(first: FieldWalker<'a>, fields: &[FieldW
         ));
     }
 
-    let first_is_key = first.is_key();
-    if fields.iter().any(|field| field.is_key() != first_is_key) {
+    let first_is_key = first.is_part_of_key();
```
tomhoule, line 48: should be renamed to `first_is_part_of_key`

engine/crates/composition/src/subgraphs/fields.rs
```diff
@@ -100,23 +100,14 @@ impl<'a> FieldWalker<'a> {
         self.field().arguments.iter().map(move |id| self.walk(*id))
     }
 
-    /// Returns true iff there is an `@key` directive containing exactly this field (no composite
-    /// key).
-    pub fn is_key(self) -> bool {
+    /// Returns true iff there is an `@key` directive containing this field, possibly with others
+    /// as part of a composite key.
+    pub(crate) fn is_part_of_key(self) -> bool {
```
tomhoule, line 105: Note that there's nothing about nesting in the implementation. What does it mean for a nested field to be part of the key (in another value type or entity)? For example:

```graphql
# First subgraph

type A @key("b { id } c") {
  b: B!
  c: String!
}

type B {
  id: ID!
  foo: String
}

# Other subgraph

type B @key("id") {
  id: ID!
  bar: String
}
```

Is that valid?

tomhoule: Apparently it should be.

tomhoule: Implemented.

tomhoule:
The PR could use more tests, there's a lot of edge cases when you start to consider nested keys.

engine/crates/composition/src/ingest_subgraph/nested_key_fields.rs
```diff
@@ -0,0 +1,47 @@
...
+                    };
+
+                    for subselection_field in &field.subselection {
+                        ingest_nested_key_fields_rec(field_type, subselection_field, nested_key_fields);
+                    }
+                }
+            }
+        }
+    });
+}
```
pimeys, line 23: What about:

```suggestion
pub(super) fn ingest_nested_key_fields(subgraph_id: SubgraphId, subgraphs: &mut Subgraphs) {
    subgraphs.with_nested_key_fields(|subgraphs, nested_key_fields| {
        let keys = subgraphs
            .walk(subgraph_id)
            .definitions()
            .flat_map(|definition| definition.entity_keys());

        for key in keys {
            for field in key.fields() {
                let Some(field_type) = key
                    .parent_definition()
                    .find_field(field.field)
                    .and_then(|field| field.r#type().definition(subgraph_id))
                else {
                    continue;
                };

                for subselection_field in &field.subselection {
                    ingest_nested_key_fields_rec(field_type, subselection_field, nested_key_fields);
                }
            }
        }
    });
}
```

You could make this even better if `KeyWalker::fields` would return an iterator of walkers, and `Selection` could know in which key it's in.

engine/crates/composition/src/subgraphs/definitions.rs
```diff
@@ -90,3 +90,17 @@ impl<'a> DefinitionWalker<'a> {
         self.walk(self.definition().subgraph_id)
     }
 }
+
+impl<'a> SubgraphWalker<'a> {
```
pimeys, line 94: Btw, why we want to split these implementations to different modules? Now it's a bit tricky to have a full vision what is implemented in `SubgraphWalker`, if it's in multiple places.

engine/crates/composition/src/subgraphs/definitions.rs
```diff
@@ -90,3 +90,17 @@ impl<'a> DefinitionWalker<'a> {
...
+    pub(crate) fn definitions(self) -> impl Iterator<Item = DefinitionWalker<'a>> {
+        let subgraph_id = self.id;
+        let definitions = &self.subgraphs.definitions.0;
+        let start = definitions.partition_point(|def| def.subgraph_id < self.id);
+        let subgraph_definitions = definitions[start..]
```
pimeys, line 99: This stuff smells like something we could have in a more generic way available somewhere.

engine/crates/composition/src/subgraphs/keys.rs
```diff
@@ -61,6 +65,44 @@ impl Subgraphs {
...
+    /// type Nested {
+    ///   identifier: ID!
+    /// }
+    ///
+    /// ```
+    ///
+    /// `Nested.identifier` is a nested key field.
+    fields: HashSet<FieldId>,
+
+    /// Objects that are part of keys that are not defined on the object itself.
```
pimeys, line 97: Can an object be defined on the object itself? Just asking, this reads a bit weird.

engine/crates/composition/src/subgraphs/keys.rs
```diff
@@ -97,12 +139,39 @@ impl<'a> KeyWalker<'a> {
...
 
 impl<'a> DefinitionWalker<'a> {
+    pub fn is_entity(self) -> bool {
+        self.entity_keys().next().is_some()
+            || self
+                .subgraphs
+                .keys
+                .nested_key_fields
+                .objects_with_nested_keys
+                .contains(&self.id)
```
pimeys, line 149: This stuff is ripe for some apis on a bit higher level, so you don't need to traverse the tree so deep to ask that question.

engine/crates/composition/src/subgraphs/keys.rs
```diff
@@ -97,12 +139,39 @@ impl<'a> KeyWalker<'a> {
...
+
     pub fn entity_keys(self) -> impl Iterator<Item = KeyWalker<'a>> {
-        let start = self.subgraphs.keys.0.partition_point(|(parent, _)| *parent < self.id);
-        self.subgraphs.keys.0[start..]
+        let start = self
+            .subgraphs
+            .keys
+            .keys
+            .partition_point(|(parent, _)| *parent < self.id);
+        self.subgraphs.keys.keys[start..]
```
pimeys, line 158: Again, there's a possibility for some generic function that could handle the partitioning.

pimeys (approved):
Nitpicks, but good stuff.

engine/crates/composition/src/subgraphs/definitions.rs
tomhoule: I'm experimenting with this because it lets me keep the data in the `subgraphs::keys` module private (for example), so everything interacting with that piece of data is encapsulated in the module. Jury is still out on whether that's a better approach, but the crate is structured like that at the moment.

engine/crates/composition/src/subgraphs/definitions.rs
tomhoule: See `RangeWalker` in @Finistere 's introspection PR.

engine/crates/composition/src/subgraphs/keys.rs
tomhoule: I can add a code example.

engine/crates/composition/src/subgraphs/keys.rs
tomhoule: It's nested because we're defining methods on `Subgraphs`, but we're in the module that defines NestedKeyFields.

engine/crates/composition/src/ingest_subgraph/nested_key_fields.rs
tomhoule: `definition` would be out of scope, unless you add a map inside the flat map. Selection is not doable because it's not normalized so you can't have unique identifiers, it was discussed in a previous review with Benjamin.

PR #968, opened 2023-11-16
feat: hook federation-dev up to engine-v2

obmarg:


cli/crates/federated-dev/src/dev/router.rs
```diff
@@ -1,28 +1,68 @@
...
+        let mut stream = streams.merge();
+
+        while let Some(message) = stream.next().await {
+            match (message, self.engine.as_ref()) {
+                (RouterMessage::Graph(graph), _) => self.engine = Some(new_engine(graph)),
+                (RouterMessage::Request(request, response_sender), Some(engine)) => {
+                    tokio::spawn(run_request(request, response_sender, Arc::clone(engine)));
+                }
+                (_, None) => {
+                    todo!("error")
```
obmarg, line 40: oops, missed this

engine/crates/integration-tests/src/engine/mod.rs
```diff
@@ -124,6 +120,19 @@ pub struct GraphQlRequest {
...
+        let mut request = Request::new(self.query);
+        if let Some(name) = self.operation_name {
+            request = request.operation_name(name);
+        }
+        if let Some(variables) = self.variables {
+            request = request.variables(variables);
+        }
+        request
+    }
+}
```
tomhoule, line 134: nice

cli/crates/federated-dev/src/dev/router.rs
```diff
@@ -1,28 +1,84 @@
...
+
+                    self.engine = Some(new_engine(graph));
+                }
+                (RouterMessage::Request(request, response_sender), Some(engine)) => {
+                    log::trace!("router got a new request with an existing engine");
+
+                    tokio::spawn(run_request(request, response_sender, Arc::clone(engine)));
+                }
+                (RouterMessage::Request(_, response_sender), None) => {
+                    log::trace!("router got a new request with a missingengine");
```
tomhoule, line 49: nitpick: missing engine

cli/crates/federated-dev/src/dev/router.rs
```diff
@@ -1,28 +1,84 @@
...
+
+                    response_sender
+                        .send(Err(ServerError::new(
+                            "there are no subgraphs registered currently",
+                            None,
+                        )))
+                        .ok();
+                }
+            }
```
tomhoule, line 58: This whole function reads very nicely. Just by curiosity because I've never worked with merged streams for multiple channels, do you use it over `tokio::select!()` over the channels in a loop for specific reasons or do you just find streams nicer?

tomhoule (approved):
Grain of salt because I haven't worked on engine-v2 or the federated dev server yet so I may be missing context, but this looks great!

cli/crates/federated-dev/src/dev/router.rs
obmarg: I've never been a massive fan of the select! macros in practice.  Like the concept but dislike them in practice.

Syntactically they're a bit messy imo.  Not too bad, but not the whole `pattern = async_expression => { thing to do }` is a bit ick.

And conceptually I've found them a tiny bit confusing.  The built in futures one has a requirement that the futures are fused which is annoying to have to think about. The tokio one cancels all but the selected future which is probably fine in practice but is something I'd rather not have to even think about.

Whereas merging like this is conceptually incredibly simple, and syntactically just normal rust.

tomhoule: That makes a lot of sense, I never liked the select macro syntax very much either. I'll try merged streams next time.

cli/crates/federated-dev/src/dev/router.rs
obmarg: don't want to wait on ci again so i'll fix this in next pr

PR #975, opened 2023-11-17
fix(udf): Fix an issue with path resolution for resolvers/authorizers with the new root scheme

jakubadamw:
# Description

We have failed to account for the new default project root path in how we're resolving the paths to UDFs.
Ensure the tests are parametrised by both back-compat and present scheme.

# Type of change

- [ ] 💔 Breaking
- [ ] 🚀 Feature
- [X] 🐛 Fix
- [ ] 🛠️ Tooling
- [ ] 🔨 Refactoring
- [ ] 🧪 Test
- [ ] 📦 Dependency


obmarg (approved):
Nice.

PR #978, opened 2023-11-17
chore: don't return result from some actor handler functions

obmarg:
Some of the handler functions in federation dev return `Result`, which makes it really easy to use `?`.  Normally this would be good, but in this case `?` is going to break the loop inside these actors, causing them to shutdown their channels, and the system to subtly stop working properly.

This updates two of the more critical actors to handle errors a bit more explicitly.  I've left Refresher & Ticket alone for now because their only instances of `?` are probably fine to early return.  Though maybe we should just be consistent, any thoughts?

### DIFF
--- a/cli/crates/federated-dev/src/dev/composer.rs
+++ b/cli/crates/federated-dev/src/dev/composer.rs
@@ -25,36 +25,38 @@ impl Composer {
         }
     }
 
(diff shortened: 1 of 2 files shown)


tomhoule (approved):
:100:

PR #986, opened 2023-11-20
fix(CI): upload no built artifacts to rust actions cache

tomhoule:
# Description

The working hypothesis is that we are observing the darwin jobs sometimes hanging forever and getting cancelled when uploading the cached cargo artifacts at the end of the job (they're 1-2GB). The solution proposed here branch is to cache the dependency sources but none of the compiled artifacts from `target/`.

# Type of change

- [ ] 💔 Breaking
- [ ] 🚀 Feature
- [x] 🐛 Fix
- [ ] 🛠️ Tooling
- [ ] 🔨 Refactoring
- [ ] 🧪 Test
- [ ] 📦 Dependency


.github/workflows/cli.yml
```diff
@@ -331,7 +331,7 @@ jobs:
         with:
           platform: macos
           cache-key: ${{ runner.os }}-${{ runner.arch }}-${{ matrix.target }}-cargo-release-${{ hashFiles('cli/Cargo.lock') }}
-          restore-key: ${{ runner.os }}-${{ runner.arch }}-${{ matrix.target }}
+          restore-key: ${{ runner.os }}-${{ runner.arch }}-cargo-release
```
tomhoule, line 334: Not related to the fix, but for uniformity with the other jobs in the file.

fbjork (approved):

fbjork (approved):

PR #990, opened 2023-11-21
composition: implement @deprecated, @tag on fields

tomhoule:
This is the first PR that composes directives, so some of the work here is starting to lay down foundations for this.

What is implemented:

- `@deprecated` everywhere it is valid in the GraphQL 2021 spec (i.e. object fields and enum values)
- `@tag` on object fields. It can be in many other places, but I did not want to make this PR much larger than it already is.

Closes GB-5395
Part of GB-5396

# Type of change

- [ ] 💔 Breaking
- [x] 🚀 Feature
- [ ] 🐛 Fix
- [ ] 🛠️ Tooling
- [ ] 🔨 Refactoring
- [ ] 🧪 Test
- [ ] 📦 Dependency

### DIFF
--- a/engine/crates/composition/src/compose/context.rs
+++ b/engine/crates/composition/src/compose/context.rs
@@ -42,8 +42,13 @@ impl<'a> Context<'a> {
         self.ir.insert_enum(name)
     }
 
-    pub(crate) fn insert_enum_value(&mut self, enum_id: federated::EnumId, value: StringWalker<'_>) {
-        self.ir.insert_enum_value(enum_id, value);
+    pub(crate) fn insert_enum_value(
+        &mut self,
+        enum_id: federated::EnumId,
+        value: StringWalker<'_>,
+        deprecation: Option<Option<StringWalker<'_>>>,
+    ) {
+        self.ir.insert_enum_value(enum_id, value, deprecation);
     }
 
     pub(crate) fn insert_field(&mut self, ir: crate::composition_ir::FieldIr) {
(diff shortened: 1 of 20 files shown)


engine/crates/composition/src/composition_ir.rs
```diff
@@ -129,7 +129,27 @@ impl CompositionIr {
...
+
+        if let Some(deprecation) = deprecation {
+            let arguments = match deprecation {
+                Some(reason) => vec![(
+                    self.insert_static_str("reason"),
+                    federated::Value::String(self.insert_string(reason)),
+                )],
+                None => Vec::new(),
+            };
+            let name = self.insert_static_str("deprecated");
```
tomhoule, line 148: There is a bit of duplication here. We'll probably want a module for directive composition that handles the shared logic.

engine/crates/composition/src/compose/context.rs
```diff
@@ -42,8 +42,13 @@ impl<'a> Context<'a> {
         self.ir.insert_enum(name)
     }
 
-    pub(crate) fn insert_enum_value(&mut self, enum_id: federated::EnumId, value: StringWalker<'_>) {
-        self.ir.insert_enum_value(enum_id, value);
+    pub(crate) fn insert_enum_value(
+        &mut self,
+        enum_id: federated::EnumId,
+        value: StringWalker<'_>,
+        deprecation: Option<Option<StringWalker<'_>>>,
```
tomhoule, line 49: That should probably be an `Option<Deprecated>` with `struct Deprecated<'a> { reason: Option<StringWalker<'a>> }`, ping if you think that would be clearer and I'll change that.

Finistere: Yeah would be clearer :)

engine/crates/composition/src/emit_federated_graph/context.rs
```diff
@@ -29,6 +29,14 @@ impl<'a> Context<'a> {
         }
     }
 
+    // /// static string -> federated graph string.
```
Finistere, line 32: leftover?

engine/crates/composition/src/subgraphs/enums.rs
```diff
@@ -3,12 +3,24 @@ use super::*;
 #[derive(Default)]
 pub(super) struct Enums {
     values: BTreeSet<(DefinitionId, StringId)>,
+    deprecated: HashMap<(StringId, StringId), Option<StringId>>,
```
Finistere, line 6: not super obvious how `(StringId, StringId)` is a path for an enum

Finistere (approved):

engine/crates/composition/src/subgraphs/enums.rs
tomhoule: True. Will document.

engine/crates/composition/src/emit_federated_graph/context.rs
tomhoule: indeed!

PR #995, opened 2023-11-21
rust: merge cli/ and engine/ cargo workspaces

tomhoule:
The contents of cli/ and engine/ is nearly completely unchanged. Predictably, CI setup needed quite a bit of work, but I think the end result is simpler overall.


.github/workflows/cli.yml
```diff
@@ -70,6 +70,7 @@ jobs:
           npx prettier --check .
 
   lint:
+    name: Lint and run unit tests
```
tomhoule, line 73: The reasoning for grouping the two is that we'd be compiling again with a separate job for unit tests, but the downside is that it blocks the  other jobs that depend on lint until tests are done. Happy to change if that sounds better.

.github/workflows/cli.yml
```diff
@@ -427,3 +426,35 @@ jobs:
           footer: "<{run_url}|View Run>"
         env:
           SLACK_WEBHOOK_URL: ${{ secrets.SLACK_WEBHOOK_URL }}
+
+  cargo-integration-tests:
```
tomhoule, line 430: No change, just moved over from engine.yml

engine/crates/parser-openapi/src/graph/mod.rs
```diff
@@ -594,7 +594,7 @@ mod tests {
     fn test_graph_size() {
         // Our graph can end up with a ton of nodes & edges so it's important
         // that they don't get too big.
-        assert!(std::mem::size_of::<Node>() <= 48);
+        assert!(std::mem::size_of::<Node>() <= 88);
```
tomhoule, line 597: This is because of feature unification. The `preserve_order` feature on serde_json is in the workspace, so serde_json::Value can contain an indexmap, and that's larger than a HashMap.

tomhoule: The order changes in JSON in tests have the same cause.

.config/nextest.toml
```diff
@@ -0,0 +1,11 @@
+[profile.ci]
+slow-timeout = { period = "1m", terminate-after = 2 }
+retries = 2
+failure-output = "immediate-final"
+fail-fast = false
```
tomhoule, line 5: I simplified a bit and standardized on the engine one, but I'm happy to adjust.

engine/crates/integration-tests/src/lib.rs
```diff
@@ -26,7 +26,12 @@ thread_local! {
 
 pub fn runtime() -> &'static Runtime {
     static RUNTIME: OnceLock<Runtime> = OnceLock::new();
-    RUNTIME.get_or_init(|| Runtime::new().unwrap())
+    RUNTIME.get_or_init(|| {
+        tokio::runtime::Builder::new_current_thread()
+            .enable_all()
+            .build()
+            .unwrap()
```
tomhoule, line 33: Because of the rusoto situation. We were using upstream in the engine workspace and our patched version in cli. Since our patched version is about wasm, it only works single threaded (send_wrapper crate).

cli/crates/cli/tests/defer/main.rs
```diff
@@ -111,8 +111,8 @@ async fn defer_multipart_test() {
             ]
           }
         },
-        "path": [],
-        "hasNext": false
```
tomhoule, line 115: The CLI workspace had the `preserve_order` feature enabled on serde_json, and the engine workspace didn't. Had to standardize one way.

cli/crates/cli/tests/graphql-directive/snapshots/graphql_directive__namespaced-issue.snap
```diff
@@ -1,14 +1,14 @@
...
 data:
   gothub:
-    serverVersion: "1"
     pullRequestOrIssue:
       __typename: GothubIssue
-      title: Everythings fucked
       author:
         __typename: GothubUser
         email: pessimist@example.com
+      title: Everythings fucked
```
fbjork, line 12: use nicer wording here:)

tomhoule: Doing a quick round of fixes through the repo :p

tomhoule:
The windows test failures are happening on other PRs so it shouldn't block reviewing this one. I'll bisect to find a fix.

cli/crates/gateway/src/context.rs
```diff
@@ -45,7 +45,6 @@ impl gateway_core::RequestContext for Context {
     async fn wait_until(&self, fut: BoxFuture<'static, ()>) {
         self.wait_until_sender
             .send(fut)
-            .ok()
```
yoav-lavi, line 48: @tomhoule why did this change?

rust-toolchain.toml
```diff
@@ -1,9 +1,11 @@
 [toolchain]
+profile = "default"
 channel = "beta"
 targets = [
+    "aarch64-apple-darwin",
+    "x86_64-apple-darwin",
+    "x86_64-pc-windows-msvc",
     "x86_64-unknown-linux-musl",  # Tools
```
yoav-lavi, line 8: Will adding these extra targets to CLI mean that it'll also attempt to build / test for those if we pass `--all-targets`?

tomhoule: I guess so. I haven't noticed a big increase in CI runtime in this branch so it seems to be fine. On the other hand I'm not sure this is used in CI, since we only get the channel from this file in `init_rust_job`. Locally I usually only deal with the target for my dev machine and wasm32-unknown-unknown, the rest would be for cross compilation, right? Happy to change this as well.

yoav-lavi: As long as when running Clippy etc. we don't get errors in the CLI for wasm32-unknown-unknown it should be fine

yoav-lavi (approved):
Nice work!

PR #997, opened 2023-11-21
feat: engine-v2 add fragment support.

Finistere:
Fragments are now handled properly read and split by plan.

During planning, we first flatten all the fields to a single selection set (flattening fragments) of bound fields (so with duplicates) and plan all of those. We keep track of the attribution of fields which is later re-used as a filter when collecting fields during resolution.

- I've added again various walkers to traverse the bound operation and later the planned one
- Removed the "ResponseCompactObject" I was overzealous, it doesn't matter currently and is a bit tricky to get right.
- Added proper fragment support in the GraphQL query serialization


engine/crates/engine-v2/schema/src/field_set.rs
```diff
@@ -0,0 +1,99 @@
+use std::cmp::Ordering;
```
Finistere, line 1: Moved out from `lib.rs`

engine/crates/engine-v2/src/executor/introspection/resolver.rs
```diff
@@ -3,17 +3,17 @@ use std::cell::RefCell;
 use engine_value::ConstValue;
```
Finistere, line 3: Looks like a lot, but it's mostly the same thing. Just a different input that applies GraphQL resolution mechanism for fields.

engine/crates/engine-v2/src/engine.rs
```diff
@@ -9,20 +11,24 @@ use crate::{
 };
 
 pub struct Engine {
-    pub(crate) schema: Schema,
+    // We use an Arc for the schema to have a self-contained response which may still
+    // needs access to the schema strings
```
tomhoule, line 15: 100% agree with this move. Not saying that we have to deal with it right now, but have you thought about how to reload the schema on a running engine? (it might be as simple as `Arc<RwLock<Schema>>`)

engine/crates/engine-v2/src/request/path.rs
```diff
@@ -1,78 +1,85 @@
...
+    pub resolved_type_condition: Option<ResolvedTypeCondition>,
     pub name: StrId,
 }
 
-#[derive(Debug, Clone, PartialEq, Eq)]
-pub struct ResolvedTypeCondition(Vec<ObjectId>);
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+pub struct ResolvedTypeCondition {
+    // sorted to guarantee deterministic order
+    possible_types: Vec<ObjectId>,
```
tomhoule, line 18: Just to check my understanding: `...on MyObject` would be `vec![my_object_id]`,  `...on MyUnion` would be `union.possible_object_types().collect()`, and same on interface with the objects implementing the interface?

(all pseudocode ofc)

engine/crates/engine-v2/src/plan/attribution.rs
```diff
@@ -0,0 +1,76 @@
+use schema::Schema;
+
+use super::PlanId;
+use crate::request::{BoundFieldId, BoundSelectionSetId, BoundSelectionSetWalker, BoundSelectionWalker, Operation};
+
+/// Keeps track of which field is assigned to which plan. It's then used to filter out all
+/// irrelevant fields in a selection set for a given plan. We also track all plan ids for which a
+/// selection set has fields for. This helps us avoid creating empty selection sets when iterating
+/// over a planned selection set (associated with a specific plan).
+pub struct Attribution {
```
tomhoule, line 10: Nice abstraction!

engine/crates/engine-v2/src/request/bind.rs
```diff
@@ -35,148 +64,341 @@ impl From<BindError> for GraphqlError {
...
+                });
+            }
+        };
+        let possible_types = TypeCondition::from(parent)
+            .resolve(self.schema)
+            .possible_types()
+            .iter()
+            .copied()
+            .collect::<HashSet<_>>();
+        let frament_possible_types = type_condition
```
tomhoule, line 351: typo but no need to go through a whole CI cycle to fix it

tomhoule (approved):
I can't say I've absorbed every detail and nuance yet, but I've given this PR a close read and I think it's great. The abstractions all make sense, it looks like a very solid structure to build on.

engine/crates/engine-v2/src/engine.rs
Finistere: IMO a new engine should be built, it might need new services for connectors (ex: connection pooling) and the query plan cache shouldn't be re-used after schema changes. In my mind, the engine is only responsible for executing a request given a certain schema, it's immutable.

engine/crates/engine-v2/src/request/path.rs
Finistere: Yeah exactly. The main purpose of `ResolvedTypeCondition` is to be used when finding the root response objects on which a plan needs to act. So supposing `catalog { products { price } }`, if `price` comes from a different subgraph we need to find all the `products` objects. They serve as an "anchor" to know where to add data later and also from where the input data must be read. Now as we could have nested type conditions, we resolve them eagerly to a list of possible object IDs.

engine/crates/engine-v2/src/request/bind.rs
Finistere: ah good catch!

engine/crates/engine-v2/src/engine.rs
tomhoule: Yeah the reloading doesn't need to be at this level, the HTTP server can start a whole new engine, fair point.

PR #1003, opened 2023-11-22
fix(postgres): On a local context, we want to cache connections

pimeys:
# Description

This caches connections for postgres for local dev, storing them to a dumb hashmap. No pooling yet, but for quick and dirty dev runs this cuts hundreds of milliseconds from every query.

I seriously hope the factory changes don't hurt us in the API. We might want to have an extra command for a cached connections, if that is the case.

### DIFF
--- a/cli/crates/gateway/src/executor.rs
+++ b/cli/crates/gateway/src/executor.rs
@@ -6,6 +6,7 @@ use dynamodb::{DynamoDBBatchersData, DynamoDBContext};
 use engine::{registry::resolvers::graphql, RequestHeaders};
 use gateway_core::{RequestContext, StreamingFormat};
 use graphql_extensions::{authorization::AuthExtension, runtime_log::RuntimeLogExtension};
+use postgres_connector_types::transport::TcpTransport;
 use runtime_local::{Bridge, LocalPgTransportFactory, LocalSearchEngine, UdfInvokerImpl};
 
 pub struct Executor {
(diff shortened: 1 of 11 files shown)


cli/crates/gateway/src/executor.rs
```diff
@@ -106,13 +107,23 @@ impl Executor {
         let resolver_engine = UdfInvokerImpl::create_engine(self.bridge.clone());
         let search_engine = LocalSearchEngine::new(self.bridge.clone());
 
+        let mut postgres_transports = HashMap::new();
```
Finistere, line 110: This needs to be done earlier, `build_schema` is called at _every_ request because of `auth`.

Finistere:
> I seriously hope the factory changes don't hurt us in the API. We might want to have an extra command for a cached connections, if that is the case.

API doesn't use any of the code you touched (except the `Box<dyn Transport>` to `&dyn Transport` change in `runtiem` crate) if that's what you're concerned about

cli/crates/gateway/src/executor.rs
pimeys: yikes. done.

pimeys:
> > I seriously hope the factory changes don't hurt us in the API. We might want to have an extra command for a cached connections, if that is the case.
> 
> API doesn't use any of the code you touched (except the `Box<dyn Transport>` to `&dyn Transport` change in `runtiem` crate) if that's what you're concerned about
> 
> API uses the `gateway-core` package directly

Yeah I split it into two: `fetch_cached` and `try_new`. This will suffice without having a ton of changes everywhere, and the query times are down from 300ms to 40ms, which is a much better experience.

engine/crates/runtime/src/pg.rs
```diff
@@ -17,6 +19,8 @@ pub trait PgTransportFactoryInner {
         name: &str,
         database_definition: &DatabaseDefinition,
     ) -> PgTransportFactoryResult<Box<dyn Transport>>;
+
+    fn fetch_cached(&self, name: &str) -> PgTransportFactoryResult<&dyn Transport>;
```
Finistere, line 23: I think you could just change the `try_new` method instead. The `database_definition` method isn't used either in the cloud version: https://github.com/grafbase/api/blob/main/common/grafbase-runtime-cloud/src/postgres.rs#L33. Implementations will have access to the `DatabaseDefinition` when they're created, so wasn't really necessary I think.

Finistere (approved):
one comment, but not a blocker

engine/crates/runtime/src/pg.rs
pimeys: The difference here is that the other returns a reference and the other owned value.

Finistere: ah good point, didn't notice it. I guess an `Arc` could have been an alternative

PR #1004, opened 2023-11-23
feat(engine-v2): Add basic variable support

Finistere:
We don't validate nor bind anything, but works otherwise.
I moved stuff around to make it a bit more readable. I moved walkers
used during execution in `execution` module and moved `executor` one
under it.

# Description

# Type of change

- [ ] 💔 Breaking
- [ ] 🚀 Feature
- [ ] 🐛 Fix
- [ ] 🛠️ Tooling
- [ ] 🔨 Refactoring
- [ ] 🧪 Test
- [ ] 📦 Dependency


engine/crates/engine-v2/schema/src/walkers/mod.rs
```diff
@@ -41,7 +41,10 @@ where
         self.id
     }
 
-    pub fn walk<OtherId>(self, id: OtherId) -> SchemaWalker<'a, OtherId> {
+    pub fn walk<OtherId>(self, id: OtherId) -> SchemaWalker<'a, OtherId>
+    where
+        OtherId: Copy,
```
tomhoule, line 46: I don't understand why this bound, if it's not copy it won't change the method body, right?

tomhoule (approved):
:man_cook:

engine/crates/engine-v2/schema/src/walkers/mod.rs
Finistere: hum a good point, not sure why I've added it. Will remove it

PR #1005, opened 2023-11-23
fix(deploy): Attempt at fixing GB-5429

jakubadamw:
# Description

Use [tar](https://crates.io/crates/tar) rather than [async-tar](https://crates.io/crates/async-tar) to create the archive that gets pushed as part of the `grafbase deploy` flow.

# Type of change

- [ ] 💔 Breaking
- [ ] 🚀 Feature
- [X] 🐛 Fix
- [ ] 🛠️ Tooling
- [ ] 🔨 Refactoring
- [ ] 🧪 Test
- [ ] 📦 Dependency

### DIFF
--- a/cli/crates/backend/src/api/deploy.rs
+++ b/cli/crates/backend/src/api/deploy.rs
@@ -15,7 +15,6 @@ use std::ffi::OsStr;
 use std::path::Path;
 use tokio::fs::read_to_string;
 use tokio_util::codec::{BytesCodec, FramedRead};
-use tokio_util::compat::TokioAsyncReadCompatExt;
 use walkdir::{DirEntry, WalkDir};
 
 const ENTRY_BLACKLIST: [&str; 2] = ["node_modules", ".env"];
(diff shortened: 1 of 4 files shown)


cli/crates/backend/src/api/deploy.rs
```diff
@@ -41,42 +40,42 @@ pub async fn deploy() -> Result<(), ApiError> {
...
-        .into_parts();
+    let tar_file_path = tokio::task::spawn_blocking(|| {
+        let (tar_file, tar_file_path) = tempfile::NamedTempFile::new()
+            .map_err(ApiError::CreateTempFile)?
+            .into_parts();
 
-    let tar_file: tokio::fs::File = tar_file.into();
-    let tar_file = tar_file.compat();
+        let mut tar = tar::Builder::new(tar_file);
+        tar.mode(tar::HeaderMode::Deterministic);
```
yoav-lavi, line 49: Just so I'm not missing anything, this is the fix itself right? The other change is `spawn_blocking` which I'm assuming isn't related to the specific issue

yoav-lavi: Or actually it looks like we also switched to sync tar and that's the reason for blocking, ok

yoav-lavi (approved):
Nice!

PR #1014, opened 2023-11-24
cli: add hidden dump-config command

tomhoule:
For GB-5423. See [the internal notion doc](https://www.notion.so/grafbase/Netlify-Connect-c771eb679bf9498bbcd1156a553bd524) for the broader context.

### DIFF
--- a/engine/crates/engine/src/registry/mod.rs
+++ b/engine/crates/engine/src/registry/mod.rs
@@ -1421,8 +1421,6 @@ pub struct Registry {
     pub federation_entities: BTreeMap<String, FederationEntity>,
     #[serde(default)]
     pub enable_ai: bool,
-    #[serde(default)]
-    pub grafbase_cli_version: Option<String>,
 }
 
 impl Default for Registry {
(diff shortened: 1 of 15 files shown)


engine/crates/engine/src/registry/mod.rs
```diff
@@ -1421,6 +1421,8 @@ pub struct Registry {
     pub federation_entities: BTreeMap<String, FederationEntity>,
     #[serde(default)]
     pub enable_ai: bool,
+    #[serde(default)]
+    pub grafbase_cli_version: Option<String>,
```
tomhoule, line 1425: Alternatively, the version could be outside the registry, the CLI would write an object like `type RegistryWithVersion = { version: string, registry: Registry }`. Maybe that would be cleaner, do we need the version in the registry in other cases? What do you think?

pimeys (approved):

engine/crates/engine/src/registry/mod.rs
tomhoule: I'm going to make that change, I don't like how intrusive this PR is with the registry.

tomhoule: Done.

PR #1028, opened 2023-11-28
feat(composition): implement entity interfaces

tomhoule:
# Description

The implementation is pretty long, I would suggest getting familiar with
the docs and reading tests first to review.

This commit contains a change in control flow in how we compose fields
of objects. At a high level, the change here is the following:

- Before, we were iterating every object type in all subgraphs not
  thinking about their fields, then iterating all object fields in a
  second pass.
- Now, for every object, we iterate the fields for that object when
  composing the object in the first pass.

The reason for the previous behaviour was that we were expecting the second pass to need all the results from the first pass (e.g. to resolve field types), but it turns out we do well without, we don't really accumulate data during composition and composition IR is read only, which is a good property.

The reason for this change now is that for entity interfaces, we really want to compose the fields in a context where we know that we are dealing with fields that are part of an entity interface, so in this PR, we stop throwing away that context.

closes GB-5229

# Type of change

- [ ] 💔 Breaking
- [x] 🚀 Feature
- [ ] 🐛 Fix
- [ ] 🛠️ Tooling
- [ ] 🔨 Refactoring
- [ ] 🧪 Test
- [ ] 📦 Dependency

### DIFF
--- a/engine/crates/composition/src/compose/entity_interface.rs
+++ b/engine/crates/composition/src/compose/entity_interface.rs
@@ -1,0 +1,181 @@
+use super::*;
+use crate::composition_ir as ir;
+use std::collections::{BTreeMap, BTreeSet};
+
+pub(crate) fn merge_entity_interface_definitions(
+    ctx: &mut Context<'_>,
+    first: DefinitionWalker<'_>,
+    definitions: &[DefinitionWalker<'_>],
+) {
+    let interface_name = first.name();
+
+    let interface_defs = || definitions.iter().filter(|def| def.kind() == DefinitionKind::Interface);
+    let mut interfaces = interface_defs();
+
+    let Some(interface_def) = interfaces.next() else {
+        ctx.diagnostics.push_fatal(format!(
+            "The entity interface `{}` is not defined as an interface in any subgraph.",
+            interface_name.as_str()
+        ));
+        return;
+    };
+
+    // More than one interface in subgraphs.
+    if interfaces.next().is_some() {
+        let all_implementers: BTreeSet<_> = interface_defs()
+            .flat_map(|interface| {
+                interface
+                    .subgraph()
+                    .interface_implementers(interface_name.id)
+                    .map(|def| def.name().id)
+            })
+            .collect();
+
+        // All subsequent interfaces must have the same implementers.
+        for interface in interface_defs() {
+            let implementers: BTreeSet<_> = interface
+                .subgraph()
+                .interface_implementers(interface_name.id)
+                .map(|def| def.name().id)
+                .collect();
+
+            if implementers != all_implementers {
+                let subgraph_name = interface.subgraph().name().as_str();
+                let interface_name = interface_name.as_str();
+                let implementer_names = all_implementers
+                    .difference(&implementers)
+                    .map(|id| ctx.subgraphs.walk(*id).as_str())
+                    .join(", ");
+                ctx.diagnostics.push_fatal(format!(
+                r#"[{subgraph_name}]: Interface type "{interface_name}" has a resolvable key in subgraph "{subgraph_name}" but that subgraph is missing some of the supergraph implementation types of "{interface_name}". Subgraph "{subgraph_name}" should define types {implementer_names}."#
+                ));
+            }
+
+            if interface.is_interface_object() {
+                ctx.diagnostics.push_fatal(format!(
+                    "[{}] The @interfaceObject directive is not valid on interfaces (on `{}`).",
+                    interface.subgraph().name().as_str(),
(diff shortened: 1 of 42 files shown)


engine/crates/composition/src/compose/entity_interface.rs
```diff
@@ -0,0 +1,181 @@
+use super::*;
+use crate::composition_ir as ir;
+use std::collections::{BTreeMap, BTreeSet};
+
+pub(crate) fn merge_entity_interface_definitions(
```
tomhoule, line 5: Maybe some parts of this could be broken up in other functions. It wouldn't necessarily be any simpler, though.

engine/crates/composition/src/compose/entity_interface.rs
```diff
@@ -0,0 +1,181 @@
...
+    ctx: &mut Context<'_>,
+    first: DefinitionWalker<'_>,
+    definitions: &[DefinitionWalker<'_>],
+) {
+    let interface_name = first.name();
+
+    let iface_defs = || definitions.iter().filter(|def| def.kind() == DefinitionKind::Interface);
```
yoav-lavi, line 12: Not sure this is a standard abbreviation, personally I'd avoid it for clarity

engine/crates/composition/src/compose/entity_interface.rs
```diff
@@ -0,0 +1,181 @@
...
+
+            if implementers != all_implementers {
+                let subgraph_name = iface.subgraph().name().as_str();
+                let iface_name = interface_name.as_str();
+                let implementer_names = all_implementers
+                    .difference(&implementers)
+                    .map(|id| ctx.subgraphs.walk(*id).as_str())
+                    .join(", ");
+                ctx.diagnostics.push_fatal(format!(
+                r#"[{subgraph_name}] Interface type "{iface_name}" has a resolvable key in subgraph "{subgraph_name}" but that subgraph is missing some of the supergraph implementation types of "{iface_name}". Subgraph "{subgraph_name}" should define types {implementer_names}."#
```
yoav-lavi, line 50:
```suggestion
                r#"[{subgraph_name}]: Interface type "{iface_name}" has a resolvable key in subgraph "{subgraph_name}" but that subgraph is missing some of the supergraph implementation types of "{iface_name}". Subgraph "{subgraph_name}" should define types {implementer_names}."#
```

engine/crates/composition/src/compose/entity_interface.rs
```diff
@@ -0,0 +1,181 @@
...
+            }
+        }
+    }
+
+    let interface_id = ctx.insert_interface(interface_name);
+
+    let mut fields = BTreeMap::new();
+
+    for field in iface_def.fields() {
+        fields.entry(field.name().id).or_insert_with(|| ir::FieldIr {
```
yoav-lavi, line 69: Could probably implement default to shorten this but not critical

engine/crates/composition/src/compose/entity_interface.rs
```diff
@@ -0,0 +1,181 @@
...
+            field_type,
+            arguments,
+            resolvable_in,
+            provides,
+            requires,
+            composed_directives,
+        } in fields.values()
+        {
+            ctx.insert_field(ir::FieldIr {
+                parent_name: object.name().id,
```
yoav-lavi, line 166: A spread might be cleaner here

yoav-lavi (approved):
Nice!

engine/crates/composition/src/compose/entity_interface.rs
tomhoule: I agree, I use it all the time but it's lazy.

engine/crates/composition/src/compose/entity_interface.rs
tomhoule: Will try

engine/crates/composition/src/compose/entity_interface.rs
tomhoule: done

engine/crates/composition/src/compose/entity_interface.rs
tomhoule: It's a bit hard because there are required fields where default could be implemented (StringId::MIN for example) but that would always or nearly always be incorrect not to set them.

yoav-lavi: I see, ok

engine/crates/composition/src/compose/entity_interface.rs
tomhoule: Spreading is tricky here because of the dereferencing and the `.clone()`s.

PR #1030, opened 2023-11-28
composition: implement @override

tomhoule:
closes GB-5364

### DIFF
--- a/engine/crates/federated-graph/src/federated_graph.rs
+++ b/engine/crates/federated-graph/src/federated_graph.rs
@@ -108,6 +108,9 @@ pub struct Field {
     /// See [FieldRequires]
     pub requires: Vec<FieldRequires>,
 
+    /// See [Override].
+    pub overrides: Vec<Override>,
+
     pub arguments: Vec<FieldArgument>,
 
     /// All directives that made it through composition. Notably includes `@tag`.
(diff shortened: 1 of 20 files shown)


engine/crates/federated-graph/src/federated_graph.rs
```diff
@@ -185,6 +188,15 @@ pub struct FieldRequires {
     pub fields: FieldSet,
 }
 
+/// Represents an `@override(graph: .., from: ...)` directive on a field in a subgraph.
+#[derive(serde::Serialize, serde::Deserialize, Clone)]
+pub struct Override {
+    pub graph: SubgraphId,
+    /// Points to a subgraph referenced by name, but this is _not_ validated to allow easier field
+    /// migrations between subgraphs.
+    pub from: StringId,
```
Finistere, line 197: Not sure to follow the logic here. If the `from` graph doesn't exist, the overrides shouldn't be present in the federated graph imho. There's nothing else to do with it at runtime. 🤷

Finistere (approved):
Not really convinced by the optional override `name` in the federated graph 🤷

engine/crates/federated-graph/src/federated_graph.rs
tomhoule: I agree from the perspective of the router, but we need to still have them there in order to roundtrip to federation SDL, I checked and other implementations allow schemas that override fields that do not exist in the `from` schema. Alternatively, this could be an enum of `enum OverrideSource { Existing(SubgraphId), Missing(StringId) }`.

tomhoule: or a separate vec of unresolved overrides

Finistere: Good point, missed that. Well if it's easy for you would be better to have that enum IMHO, but not a big deal to do the mapping later.

tomhoule: I'll do the enum, should be very easy.

tomhoule: Done in 86ae7405b25918442d1468f6d1cd48990a613d15

Finistere: Thanks!

PR #1032, opened 2023-11-28
cargo: migrate clippy configuration from .config to Cargo.toml

tomhoule:
closes GB-5372


engine/crates/engine/src/lib.rs
```diff
@@ -1,44 +1,6 @@
...
-#![allow(clippy::match_same_arms)]
-#![allow(clippy::missing_const_for_fn)]
-#![allow(clippy::missing_errors_doc)]
-#![allow(clippy::module_name_repetitions)]
-#![allow(clippy::must_use_candidate)]
-#![allow(clippy::needless_borrow)]
-#![allow(clippy::needless_pass_by_value)]
-#![allow(clippy::needless_lifetimes)]
-#![allow(clippy::option_if_let_else)]
 #![allow(clippy::panic)]
```
fbjork, line 2: why leave this?

tomhoule: The ones that are left are those that need to be different in the engine crate (compared to the workspace rules). I could remove these three `allow`s but that would involve changing enum names, adding a bunch of lifetimes, and seeing how to replace the panics. I'm happy to do it if that sounds good. Especially the elided lifetimes I don't like, but it's probably a few hundred more lines of diff (`cargo clippy --fix` can't do it automatically,  I checked).

yoav-lavi (approved):
```toml
[lints]
 workspace = true
 ```

PR #1034, opened 2023-11-29
fix(parser-openapi): infer entities for resources with multiple schemas

tomhoule:
Some resources have multiple responses associated to them, e.g. for different status codes. Before this commit, we bailed when we encountered multiple response shapes. Now we try to pick the most likely (successful) one. Alternatively, we could infer a GraphQL union of these.

closes GB-5486

# Type of change

- [ ] 💔 Breaking
- [ ] 🚀 Feature
- [x] 🐛 Fix
- [ ] 🛠️ Tooling
- [ ] 🔨 Refactoring
- [ ] 🧪 Test
- [ ] 📦 Dependency

### DIFF
--- a/engine/crates/parser-openapi/src/parsing/grouping.rs
+++ b/engine/crates/parser-openapi/src/parsing/grouping.rs
@@ -25,7 +25,8 @@ pub fn determine_resource_relationships(ctx: &mut super::Context) {
 fn find_operation_resource_schema(ctx: &super::Context, operation_index: NodeIndex) -> Option<NodeIndex> {
     // We only want to look in the response fields
     let filtered_graph = EdgeFiltered::from_fn(&ctx.graph, |edge| {
-        matches!(edge.weight(), Edge::HasResponseType { .. } | Edge::HasField { .. })
+        matches!(edge.weight(), Edge::HasResponseType { status_code, .. } if status_code.is_success())
+            || matches!(edge.weight(), Edge::HasField { .. })
     });
 
     let mut schema_indices = Dfs::new(&filtered_graph, operation_index)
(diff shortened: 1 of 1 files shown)


engine/crates/engine/src/registry/resolvers/http/mod.rs
```diff
@@ -159,6 +159,13 @@ impl HttpResolver {
 }
 
 impl ExpectedStatusCode {
+    pub fn is_success(&self) -> bool {
+        match self {
+            ExpectedStatusCode::Exact(code) => 200 <= *code && *code < 300,
+            ExpectedStatusCode::Range(code_range) => code_range.contains(&200) && code_range.end < 300,
```
tomhoule, line 165: This feels a bit yolo.

tomhoule, line 165: ...yolo?

engine/crates/parser-openapi/src/parsing/grouping.rs
```diff
@@ -33,13 +33,22 @@ fn find_operation_resource_schema(ctx: &super::Context, operation_index: NodeInd
...
+        // If there is more than one schema associated to the operation, we have to rely on
+        // heuristics to pick one.
+        2.. => schema_indices.into_iter().find(|schema_index| {
+            ctx.graph
+                .edges_connecting(operation_index, *schema_index)
+                .any(|edge| match edge.weight() {
+                    // This is a/the sucess response: pick it.
+                    Edge::HasResponseType { status_code, .. } => status_code.is_success(),
+                    // ?
+                    Edge::HasField { .. } => true,
```
tomhoule, line 47: ?

tomhoule: This is very much me fumbling around in the crate, the structure is nice and clear but I'm very new to it. But I think I get what was wrong and the suggested change, I pushed a new commit, tests are still passing locally.

obmarg (approved):

PR #1036, opened 2023-11-29
composition: implement rest of @inaccessible

tomhoule:


engine/crates/composition/src/composition_ir.rs
```diff
@@ -202,7 +138,8 @@ pub(crate) struct FieldIr {
     pub(crate) parent_name: subgraphs::StringId,
     pub(crate) field_name: subgraphs::StringId,
     pub(crate) field_type: subgraphs::FieldTypeId,
-    pub(crate) arguments: Vec<(subgraphs::StringId, subgraphs::FieldTypeId)>,
+    /// (argument_name, argument_type, composed_directives)
```
Finistere, line 141: Worth a struct if you need a comment :)

Finistere (approved):

engine/crates/composition/src/composition_ir.rs
tomhoule: True :/ I'm lazy about creating structs, but there's no excuse here.

tomhoule: Fixed and pushed.

PR #1051, opened 2023-12-01
composition: implement description composition

tomhoule:
The strategy for composing descriptions seems to be picking the first one we find. There can be an optional lint to enforce that they are the same across subgraphs.

closes GB-5394

### DIFF
--- a/engine/crates/composition/src/compose/context.rs
+++ b/engine/crates/composition/src/compose/context.rs
@@ -23,7 +23,7 @@ impl<'a> Context<'a> {
         };
 
         for builtin_scalar in subgraphs.iter_builtin_scalars() {
-            context.insert_scalar(builtin_scalar, false);
+            context.insert_scalar(builtin_scalar, false, None);
         }
 
         context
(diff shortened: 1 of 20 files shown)


engine/crates/composition/src/compose/context.rs
```diff
@@ -64,8 +73,36 @@ impl<'a> Context<'a> {
         enum_id: federated::EnumId,
         value: StringWalker<'_>,
         deprecation: Option<Option<StringWalker<'_>>>,
+        description: Option<StringWalker<'_>>,
     ) {
-        self.ir.insert_enum_value(enum_id, value, deprecation);
+        let mut composed_directives = Vec::new();
```
tomhoule, line 78: The function moved to remove one level of unnecessary indirection. Only the description lines are new.

engine/crates/composition/src/compose/context.rs
```diff
@@ -135,12 +178,52 @@ impl<'a> Context<'a> {
...
+
+        // FIXME: Those roots probably shouldn't be hardcoded.
+        match object_name.as_str() {
+            "Query" => self.ir.query_type = Some(id),
+            "Mutation" => self.ir.mutation_type = Some(id),
+            "Subscription" => self.ir.subscription_type = Some(id),
+            _ => (),
+        }
+
+        id
```
tomhoule, line 216: Same, function just moved, only the bits about description are new.

yoav-lavi:
By default maybe it'd make sense to take the one that was updated most recently rather than it being random @tomhoule

PR #1053, opened 2023-12-01
feat: `@extendField` directive

obmarg:
Users might want to add federation directives to connector generated types. Doing this with the existing extend keyword is awkward because either:

1. We'd have to force users to write out the full field definition or
2. We'd have to just ignore the arguments & return type inside extensions - which is a bit ambigous, is the user adding a clashing field or adding a field just to extend its directives?

So I've added a `extendField` directive to objects to allow users to do this:

```graphql
extend type Blah
  @extendField(
    name: "foo"
    external: true
    override: {from: "Blah"}
    provides: {fields: "id"}
  )
```

Part of GB-5518


tomhoule (approved):
nice!

PR #1054, opened 2023-12-01
feat: connector types can be extended with federation keys

obmarg:
Missed this when adding federation to the SDK.  Will do a follow up PR with the ability to extend fields with attributes as well.

Part of GB-5518


tomhoule (approved):
I like the builder :+1:

PR #1066, opened 2023-12-04
composition: implement @composeDirective

tomhoule:
This is more permissive than the spec: it will compose directives without definition and without @link imports. It merges instances of the directive with the same arguments across subgraphs.

closes GB-5399


engine/crates/composition/src/ingest_subgraph/directives.rs
```diff
@@ -109,3 +125,358 @@ pub(super) fn ingest_keys(
...
+            ));
+        }
+    }
+
+    directive_matcher.composed_directives = composed_directives;
+
+    directive_matcher
+}
+
+/// This struct is the source of truth for matching federation directives by name when ingesting a
```
tomhoule, line 177: From here, this file is 99% moved. Changes: removed FederationDirectivesMatcher to DirectiveMatcher and added the two struct fields about composed directives.

engine/crates/composition/tests/composition/composed_directives_basic/federated.graphql
```diff
@@ -54,8 +54,8 @@ type ObserverDetails {
     observerType: ObserverType! @join__field(graph: OBSERVATIONS)
 }
 
-type BirdSighting {
-    bird: Bird! @join__field(graph: SIGHTINGS)
+type BirdSighting @deprecated(reason: "we haven't seen any birds in a while :(") {
```
tomhoule, line 57: Looks like we were missing deprecated here. Thanks to the recent reorganization that is prolonged in this PR, handling of directives is uniform across locations and things like that are less likely to happen.

engine/crates/composition/src/compose/directives.rs
```diff
@@ -0,0 +1,78 @@
+use super::*;
+
+pub(super) fn collect_composed_directives<'a>(
```
tomhoule, line 3: Logic gathered from different places and centralized.

engine/crates/composition/src/compose/directives.rs
```diff
@@ -0,0 +1,78 @@
...
+                    )
+                })
+                .map(|arg| vec![arg])
+                .unwrap_or_default(),
+        });
+    }
+
+    for container in containers {
+        tags.extend(container.tags().map(|t| t.id));
+        is_inaccessible |= container.inaccessible();
```
yoav-lavi, line 30: Might be a bit cleaner to do this separately or spell it out more, was a bit surprised to see this

engine/crates/composition/src/compose/directives.rs
```diff
@@ -0,0 +1,78 @@
...
+            name: ctx.insert_static_str("tag"),
+            arguments: vec![(ctx.insert_static_str("name"), federated::Value::String(name))],
+        });
+    }
+
+    composed_directives.extend(extra_directives);
+    composed_directives
+}
+
+fn subgraphs_value_to_federated_value(value: &subgraphs::Value, ctx: &mut ComposeContext<'_>) -> federated::Value {
```
yoav-lavi, line 62: Personally I'd expand single letter variables

yoav-lavi (commented):
Nice work! See comments

engine/crates/composition/src/compose/directives.rs
tomhoule: good point, it was a bit lazy, will do

engine/crates/composition/src/compose/directives.rs
tomhoule: yep, I'll see how to make it more readable. At the very least with a comment.

engine/crates/composition/src/compose/directives.rs
tomhoule: done

engine/crates/composition/src/compose/directives.rs
tomhoule: done

yoav-lavi (approved):
Nice work!

PR #1069, opened 2023-12-04
  composition: check implementers of composed interfaces 

tomhoule:
The rule is that a composed interface is the union of all the interfaces
in all the schemas, and objects implementing that interface in the
composed schema must implement all fields of the interfface. In this PR,
we validate that.

closes GB-5507

### DIFF
--- a/engine/crates/composition/src/compose/enums.rs
+++ b/engine/crates/composition/src/compose/enums.rs
@@ -80,11 +80,11 @@ fn merge_intersection<'a>(
     let enum_id = ctx.insert_enum(first.name(), description, composed_directives);
 
     for value in intersection {
-        let containers = definitions
+        let sites = definitions
             .iter()
             .filter_map(|enm| enm.enum_value_by_name(value))
             .map(|value| value.directives());
-        let composed_directives = collect_composed_directives(containers, ctx);
+        let composed_directives = collect_composed_directives(sites, ctx);
         ctx.insert_enum_value(enum_id, first.walk(value), None, composed_directives);
     }
 }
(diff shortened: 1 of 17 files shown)


engine/crates/composition/src/compose/enums.rs
```diff
@@ -3,23 +3,22 @@ use std::collections::HashSet;
...
+pub(super) fn merge_enum_definitions<'a>(
+    first: &DefinitionWalker<'a>,
+    definitions: &[DefinitionWalker<'a>],
+    ctx: &mut Context<'a>,
 ) {
     let enum_name = first.name().id;
-    let is_inaccessible = definitions
-        .iter()
-        .any(|definition| definition.directives().inaccessible());
+    let directive_containers = definitions.iter().map(|def| def.directives());
```
Finistere, line 12: Not related to this PR, but why is it named `DirectiveContainerId`, why not `DirectiveId`? Does the `container` notion for a directive mean something specific?

engine/crates/composition/src/compose/enums.rs
```diff
@@ -78,41 +77,58 @@ fn merge_intersection(
...
+        let name = all_values[start].0;
+        let end = all_values[start..].partition_point(|(n, _)| *n == name) + start;
+        let containers = all_values[start..end]
+            .iter()
+            .map(|(_, directives)| first.walk(*directives));
+        let composed_directives = collect_composed_directives(containers, ctx);
+
+        ctx.insert_enum_value(enum_id, first.walk(name), None, composed_directives);
+
+        if end == all_values.len() {
```
Finistere, line 119: `end` could be mutable and you would be able to do a simpler `while end < all_value.len()` loop

Finistere: or what you did later with mutable `start` :D

Finistere (approved):

engine/crates/composition/src/compose/enums.rs
Finistere: I would use `DIrectiveLocationId` mapping to a list of directives from your description

tomhoule: It's something I introduced so we can handle directives uniformly across different kinds of schema items. It really payed off in ease of implementation of the last few composition features.

tomhoule: Renamed to DirectiveSite, even shorter.

PR #1070, opened 2023-12-05
composition: implement strict matching rules for shareable objects

tomhoule:
The fields must match between subgraphs, except those marked with `@inaccessible`.

closes GB-5508

# Type of change

- [ ] 💔 Breaking
- [x] 🚀 Feature
- [ ] 🐛 Fix
- [ ] 🛠️ Tooling
- [ ] 🔨 Refactoring
- [ ] 🧪 Test
- [ ] 📦 Dependency

### DIFF
--- a/engine/crates/composition/src/compose.rs
+++ b/engine/crates/composition/src/compose.rs
@@ -65,6 +65,7 @@ fn merge_object_definitions<'a>(
         ctx.diagnostics.push_fatal(format!(
             "Cannot merge {first_kind:?} with {second_kind:?} (`{name}` in `{first_subgraph}` and `{second_subgraph}`)",
         ));
+        return;
     }
 
     let first_is_entity = first.is_entity();
@@ -98,6 +99,10 @@ fn merge_object_definitions<'a>(
         .filter(|key| key.is_resolvable())
     {
         ctx.insert_resolvable_key(object_id, key.id);
+    }
+
+    if is_shareable {
+        object::validate_shareable_object_fields_match(definitions, ctx);
     }
 
     ctx.subgraphs.iter_field_groups(first.name().id, |fields| {
--- a/engine/crates/composition/src/compose/object.rs
+++ b/engine/crates/composition/src/compose/object.rs
@@ -213,3 +213,33 @@ fn collect_overrides(fields: &[FieldWalker<'_>], ctx: &mut Context<'_>) -> Vec<f
 
     overrides
 }
+
+pub(crate) fn validate_shareable_object_fields_match(
+    definitions: &[DefinitionWalker<'_>],
+    ctx: &mut ComposeContext<'_>,
+) {
+    let all_fields: BTreeSet<StringId> = definitions
+        .iter()
+        .flat_map(|def| def.fields())
+        .map(|field| field.name().id)
+        .collect();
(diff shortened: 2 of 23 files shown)


engine/crates/composition/src/compose.rs
```diff
@@ -100,6 +101,10 @@ fn merge_object_definitions<'a>(
         ctx.insert_resolvable_key(object_id, key.id);
     }
 
+    if is_shareable {
```
tomhoule, line 104: This is not the complete solution: we should check shareable fields inside non-shareable objects, take into accounts provides and requires, etc. It gets very complicated, but this is an 80% solution as a first step.

engine/crates/composition/src/compose/object.rs
```diff
@@ -216,3 +216,28 @@ fn collect_overrides(fields: &[FieldWalker<'_>], ctx: &mut Context<'_>) -> Vec<f
...
+) {
+    let all_fields: BTreeSet<StringId> = definitions
+        .iter()
+        .flat_map(|def| def.fields())
+        .filter(|field| !field.directives().inaccessible())
+        .map(|field| field.name().id)
+        .collect();
+
+    for field in all_fields {
+        for definition in definitions {
```
tomhoule, line 232: I'd like to avoid doing (field.len() × definitions.len()) BTree lookups here, but there's no super easy fix at hand. Suggestions welcome.

engine/crates/composition/tests/composition/inaccessible_basic/federated.graphql
```diff
@@ -56,9 +56,9 @@ type Series {
 
 type New {
     name: String! @inaccessible
```
tomhoule, line 58: The changes in the existing tests are there to avoid these schemas to become invalid with the new rule.

engine/crates/composition/src/compose/object.rs
Finistere: I think the best would be to do the outer loop by definition, have `all_fields` be a sorted array of fields, and iterate over the definitions fields. The definition fields would come in sorted order, so you can just move forward an index in the `all_fields` array until you find the matching field. It should also be more friendly with regards to caching.

Finistere (approved):

engine/crates/composition/src/compose/object.rs
```diff
@@ -216,3 +216,28 @@ fn collect_overrides(fields: &[FieldWalker<'_>], ctx: &mut Context<'_>) -> Vec<f
...
 }
+
+pub(crate) fn validate_shareable_object_fields_match(
+    definitions: &[DefinitionWalker<'_>],
+    ctx: &mut ComposeContext<'_>,
+) {
+    let all_fields: BTreeSet<StringId> = definitions
+        .iter()
+        .flat_map(|def| def.fields())
+        .filter(|field| !field.directives().inaccessible())
```
tomhoule, line 227: This is wrong: if _any_ of the definitions of the field is inaccessible, that's enough for the field to be inaccessible. There should be a test for that.

engine/crates/composition/src/compose/object.rs
tomhoule: I'll try!

PR #1071, opened 2023-12-05
feat(engine-v2): Add basic @key support

Finistere:
I'm adding basic support for `@key` joins. Currently, all fields must be present in the query for it to work, we don't add requirements to parent plans yet.

I've changed the planning workflow to make all of this work with type conditions:
- During the planning phase, we don't fully plan children yet. At plan boundaries, we only plan the missing fields to ensure we have all the required fields, eventually adding those.
- At each of those boundaries we track all response objects with their `ResponsePath` and `ObjectId`.
- A plan may have multiple boundaries. For each of those, we have a list of response objects and child plans after execution. For each child plan, we extract the matching response objects (correct `ObjectId`) and fully plan its expected output/selection set if it's non-empty. 

Currently, a single plan generates a single executor, later it should be a bit more complex to allow merging GraphQL queries to the same endpoint. But well it is what it is for now.

I've also refactored a few parts of the code base:

- In the schema data sources have been simplified, there is no `DataSourceId` anymore but a struct `DataSources` with all of them.
- `Names` has been slightly simplified
- I had two sets of walkers in engine v2 (before and after planning), I've now merged them into a single with some generic type parameters.

I'm not super satisfied with some naming (`ResponseBoundaryItem`), but for now don't have a better idea.


engine/crates/engine-v2/src/plan/planner.rs
```diff
@@ -1,254 +1,481 @@
...
+            &FlatSelectionSet {
+                ty: SelectionSetType::Object(self.operation.root_object_id),
+                fields: vec![],
+            },
+            None,
+            FlatSelectionSet {
+                fields: missing,
+                ..flat_selection_set
+            },
+        )?;
```
tomhoule, line 103: nice

tomhoule (approved):
It's hard to review in the details since it's a bit of a moving target and I don't have a very clear mental model yet, but it does look good. The changes in the planner module make sense.

engine/crates/engine-v2/src/response/error.rs
```diff
@@ -3,7 +3,7 @@ use std::collections::HashMap;
 use super::ResponsePath;
 use crate::request::Pos;
 
-#[derive(Debug)]
+#[derive(Debug, Default)]
```
tomhoule, line 6: Why Default on error?

tomhoule: ah yes, that makes sense :+1:

PR #1072, opened 2023-12-05
chore(engine-v2): Remove cf-workers feature flag and move it behind a trait

Finistere:
feature flags are a pain to deal with we've seen that with the previous
sqlite vs dynamodb for example. So environment specific logic should be
abstracted away in `runtime` crate. This also solves the problem of
sharing CF worker specific http request logic across the whole engine.

# Description

# Type of change

- [ ] 💔 Breaking
- [ ] 🚀 Feature
- [ ] 🐛 Fix
- [ ] 🛠️ Tooling
- [ ] 🔨 Refactoring
- [ ] 🧪 Test
- [ ] 📦 Dependency

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2009,26 +2009,20 @@ version = "3.0.31"
 dependencies = [
  "anyhow",
  "async-runtime",
- "bytes",
  "derive_more",
  "engine",
  "engine-parser",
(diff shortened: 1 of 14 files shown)


jakubadamw (approved):
Let's do it !

PR #1077, opened 2023-12-06
GB-5348 - Rework how to enable subgraphs

yoav-lavi:
# Description

## Features

- Tweaks the subgraph API for the SDK

# Type of change

- [ ] 💔 Breaking
- [ ] 🚀 Feature
- [x] 🐛 Fix
- [ ] 🛠️ Tooling
- [ ] 🔨 Refactoring
- [ ] 🧪 Test
- [ ] 📦 Dependency


packages/grafbase-sdk/src/grafbase-schema.ts
```diff
@@ -60,7 +60,9 @@ export interface IntrospectParams {
   namespace?: boolean
 }
 
-export class SingleGraph {
+const FEDERATION_VERSION = '2.3'
+
+export class Graph {
```
fbjork, line 65: shouldn't this be StandaloneGraph?

packages/grafbase-sdk/tests/unit/federation.test.ts
```diff
@@ -2,7 +2,7 @@ import { config, graph } from '../../src/index'
 import { describe, expect, it, beforeEach } from '@jest/globals'
 import { renderGraphQL } from '../utils'
 
-const g = graph.Single()
+const g = graph.Standalone({ subgraph: true })
```
fbjork, line 5: I wonder if we should future-proof subgraph and make it an object that takes a few fields like enable, version, etc. Might be overthinking it.

yoav-lavi: I don't think it'd be a breaking change to add those later

packages/grafbase-sdk/src/grafbase-schema.ts
yoav-lavi: We can change it, internally not sure it matters since graph is pretty descriptive

yoav-lavi: It might be good since the other variant also has the word graph and isn't an extension but we could iterate on it later

PR #1084, opened 2023-12-06
feat: variable coercion in engine-v2

obmarg:
This implements variable coercion in engine-v2 along with a bunch of tests.  It also includes some validation of the actual values of variables - though it doesn't implement most of the other variable validations from the spec.

Part of GB-5406

### DIFF
--- a/engine/crates/integration-tests/tests/federation/basic/variables.rs
+++ b/engine/crates/integration-tests/tests/federation/basic/variables.rs
@@ -1,5 +1,9 @@
 use engine_v2::Engine;
-use integration_tests::{federation::EngineV2Ext, mocks::graphql::EchoSchema, runtime, MockGraphQlServer};
+use integration_tests::{
+    federation::{EngineV2Ext, GraphqlResponse},
+    mocks::graphql::EchoSchema,
+    runtime, MockGraphQlServer,
+};
 use serde::Serialize;
 use serde_json::json;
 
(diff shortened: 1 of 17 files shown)


engine/crates/integration-tests/tests/federation/basic/variables.rs
```diff
@@ -44,42 +48,398 @@ fn lists() {
...
+    );
 
-            engine.execute(query).variables(json!({"input": input})).await
-        }
+    let query = "query($input: [String!]!) { listOfStrings(input: $input) }";
+    assert_eq!(run_query(query, &input).into_data()["listOfStrings"], json!(["hello"]));
+
+    let query = "query($input: InputObj!) { inputObject(input: $input) }";
+    let input = json!({
+        "input": {"annoyinglyOptionalStrings": "hello", "recursiveObjectList": {"id": "1"}}
```
tomhoule, line 158: TIL single values coerce to a one-item list. I didn't know that part of the spec.

engine/crates/integration-tests/tests/federation/basic/variables.rs
```diff
@@ -44,42 +48,398 @@ fn lists() {
...
+    ]
+    "###);
+    insta::assert_json_snapshot!(error_test("listOfStrings", "[String!]!", json!([1])), @r###"
+    [
+      "Variable $input got an invalid value: found a Integer value where we expected a String scalar at $input.0"
+    ]
+    "###);
+    insta::assert_json_snapshot!(error_test("listOfStrings", "[String!]!", json!([null])), @r###"
+    [
+      "Variable $input got an invalid value: found a null where we expected a non-null type at $input.0"
```
tomhoule, line 314: Nice error :+1:

engine/crates/integration-tests/tests/federation/basic/variables.rs
```diff
@@ -44,42 +48,398 @@ fn lists() {
...
+        .iter()
+        .map(|error| error["message"].as_str().expect("message to be a string").to_string())
+        .collect::<Vec<_>>();
+
+    insta::assert_json_snapshot!(errors, @r###"
+    [
+      "Variable $one got an invalid value: found a Boolean value where we expected a String scalar at $one",
+      "Variable $two got an invalid value: found a String value where we expected a Int scalar at $two"
+    ]
+    "###);
```
tomhoule, line 442: Just in case it's an omission, I'm not sure I've seen cases about enum value coercions

tomhoule (approved):
:100:

engine/crates/integration-tests/tests/federation/basic/variables.rs
obmarg: Ah yes, good point.  I'll add some of those

PR #1085, opened 2023-12-06
feat: GraphQL Query Normalizer

pimeys:
Implements a normalizer to calculate distinct query ids.

Part of: GB-5519

### DIFF
--- a/engine/crates/operation-normalizer/Cargo.toml
+++ b/engine/crates/operation-normalizer/Cargo.toml
@@ -1,0 +1,21 @@
+[package]
+name = "operation-normalizer"
+version.workspace = true
+edition.workspace = true
+license.workspace = true
+homepage.workspace = true
+keywords.workspace = true
+repository.workspace = true
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+anyhow = "1.0.75"
+graphql-parser = { git = "https://github.com/graphql-rust/graphql-parser", version = "0.4.0" }
+
+[lints]
+workspace = true
+
+[dev-dependencies]
+expect-test = "1.4.1"
+indoc = "2.0.4"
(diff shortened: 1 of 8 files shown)


engine/crates/operation-normalizer/Cargo.toml
```diff
@@ -0,0 +1,21 @@
...
+license.workspace = true
+homepage.workspace = true
+keywords.workspace = true
+repository.workspace = true
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+anyhow = "1.0.75"
+graphql-parser = { git = "https://github.com/graphql-rust/graphql-parser", version = "0.4.0" }
```
Finistere, line 14: Shouldn't we re-use the `engine_parser` that is used by both engine v1 & v2 currently anyway? @obmarg any opinion?

pimeys: Can it also serialize back to string pretty GraphQL?

pimeys: And if yes, can we guarantee that this output never changes to something different, changing the operation_ids as we go?

engine/crates/operation-normalizer/src/lib.rs
```diff
@@ -0,0 +1,88 @@
+//! # Grafbase Operation Normalizer
+//!
+//! A library to modify a incoming query so we can match similar queries to be the
+//! same, even if they have differences in their string representation. This is achieved
+//! by doing the following steps:
+//!
+//! - Removal of all hard-coded variables for the following types:
```
obmarg, line 7: Nitpicking, but:

```suggestion
//! - Removal of all hard-coded arguments for the following types:
```

engine/crates/operation-normalizer/src/arguments.rs
```diff
@@ -0,0 +1,26 @@
+use graphql_parser::{query::Number, schema::Value};
+
+pub(super) fn normalize<'a>(arguments: &mut [(&'a str, Value<'a, &'a str>)]) {
+    arguments.sort_by(|(a, _), (b, _)| dbg!(a).cmp(dbg!(b)));
```
obmarg, line 4: Guessing you don't need those `dbg!`s anymore?

PR #1088, opened 2023-12-06
use latest SDK syntax

fbjork:
* Updates templates to use latest SDK syntax for standalone graphs
* Removes SDL templates
* Adds Postgres template


yoav-lavi (approved):
Nice!

PR #1094, opened 2023-12-07
GB-5538 - Remove db and search

yoav-lavi:
# Description

## Breaking
 
- Removes the embedded SQLite database from the CLI (previously deprecated)
- Removes search functionality from the CLI (previously deprecated)

# Type of change

- [x] 💔 Breaking
- [ ] 🚀 Feature
- [ ] 🐛 Fix
- [ ] 🛠️ Tooling
- [ ] 🔨 Refactoring
- [ ] 🧪 Test
- [ ] 📦 Dependency


obmarg:
So are we just deleting all the CLI tests?  That doesn't seem ideal...

yoav-lavi:
@obmarg - @fbjork asked to do this, we'll return some of the ones that test features unrelated to the DB around Christmas

yoav-lavi:
@obmarg it's not all of them though, just the ones that use the DB

obmarg:
Have we made a list of the ones we're intending to resurrect?

yoav-lavi:
@obmarg Essentially:



auth
caching
coercion
compilation_error
concurrency?
resolvers
default
watch
home 
length
owner
reserved_dates?
scalars

obmarg:
> @obmarg it's not all of them though, just the ones that use the DB

Sure, but there's some core functionality I see deleted.  `@defer`, the `dev` command, the `start` command, compilation errors, caching

yoav-lavi:
@obmarg yes, all of these currently rely on the DB and we need a local alternative that we can use to replace it for these tests

obmarg:
Could w mark the ones we're planning to resurrect with an `#[ignore]` rather than just deleting?  Seems like it'd make the job of resurrecting them (and reviewing them upon resurrection) easier?

yoav-lavi:
@obmarg They're going to be massively changed, I'm not sure it'd actually make the diff much smaller. But I can look into it if we think it will

https://linear.app/grafbase/issue/GB-5641/restore-the-removed-cli-tests-from-removing-the-db

tomhoule:
Yeah `@ignore` will still require changes because they still have to compile (I think?). Alternatively, we could comment them out.

yoav-lavi:
@obmarg done

cli/crates/cli/tests/caching.rs
```diff
@@ -100,6 +102,8 @@ async fn model_and_field_caching() {
     assert_eq!(header(&response, GRAFBASE_CACHE_HEADER), Some("HIT"));
     assert_eq!(response.headers().typed_get::<CacheControl>(), None);
 }
+
+#[ignore]
```
yoav-lavi, line 106:
```suggestion
#[ignore]
```

obmarg (approved):
Nice work

PR #1095, opened 2023-12-07
cli: add introspect --dev command to dump project schema 

tomhoule:
The use case is schema checks. You want the SDL schema of the current
project as it is locally, not the deployed version. The current best
solution is to start a dev server and introspect it. This commit
introduces a much easier solution: `grafbase introspect --dev`.

### DIFF
--- a/cli/crates/cli/src/introspect.rs
+++ b/cli/crates/cli/src/introspect.rs
@@ -1,9 +1,35 @@
-use crate::{cli_input::IntrospectCommand, errors::CliError};
+use crate::{cli_input::IntrospectCommand, errors::CliError, output::report};
 use tokio::runtime::Runtime;
 
 pub(crate) fn introspect(command: &IntrospectCommand) -> Result<(), CliError> {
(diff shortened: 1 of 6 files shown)


cli/crates/cli/src/introspect.rs
```diff
@@ -2,8 +2,34 @@ use crate::{cli_input::IntrospectCommand, errors::CliError};
...
+    }
+}
+
+fn introspect_local() -> Result<(), CliError> {
+    match server::introspect_local().map_err(CliError::ServerError)? {
+        server::IntrospectLocalOutput::Sdl(schema) => {
+            println!("{schema}");
+        }
+        server::IntrospectLocalOutput::EmptyFederated => {
+            eprintln!("⚠️ The introspected schema is empty. Introspecting federated graphs is not implemented yet.")
```
tomhoule, line 24: should this go to the report module?

tomhoule: done

PR #1096, opened 2023-12-07
fix: don't remove built in scalars in composition

obmarg:
Composition was filtering out built in scalars that weren't present in any subgraph. This meant that engine-v2 wasn't aware of them, so you'd get `Unknown type: "ID"` errors if you tried to use them.  Shouldn't really matter for valid queries, but still worth fixing.

Fixes GB-5576

### DIFF
--- a/engine/crates/composition/src/subgraphs.rs
+++ b/engine/crates/composition/src/subgraphs.rs
@@ -23,7 +23,6 @@ use itertools::Itertools;
 use std::collections::{BTreeMap, BTreeSet, HashSet};
 
 /// A set of subgraphs to be composed.
-#[derive(Default)]
 pub struct Subgraphs {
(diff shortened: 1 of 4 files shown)


engine/crates/composition/src/emit_federated_graph.rs
```diff
@@ -33,6 +33,8 @@ pub(crate) fn emit_federated_graph(mut ir: CompositionIr, subgraphs: &Subgraphs)
         field_types: vec![],
     };
 
+    dbg!(out.scalars.len());
```
fbjork: remove?

obmarg: Yep, done

fbjork (approved):

yoav-lavi (approved):
Nice!

PR #1097, opened 2023-12-07
cli: release 0.48.1

tomhoule:
Features

- The `grafbase introspect` command can now take a `--dev` flag to introspect the local project's schema. It is mutually exclusive with the `--url` flag. (#1095)

Fixes

- Due to the recent added support for configuration at the repository root, the generated TS code for typed resolvers was generated at the repository root, even when the configuration lived in the ./grafbase directory. This releases fixes that behaviour: the `generated` directory will be at the same level as the configuration. (#1093)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2208,7 +2208,7 @@ dependencies = [
 
 [[package]]
 name = "federated-dev"
-version = "0.48.0"
+version = "0.48.1"
 dependencies = [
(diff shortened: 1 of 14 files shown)


obmarg (approved):
Missing the top level changelog edit?

tomhoule:
Indeed. Fixed and pushed.

PR #1098, opened 2023-12-07
ci: skip the cli workflow on markdown-only PRs

tomhoule:


obmarg:
I approved this, though I suppose one problem is that `prettier` is part of the cli workflow, and I think it cares about markdown.  So suppose this increases the changes of someone accidentally causing problems around that.

Maybe we could move prettier into it's own workflow to avoid?

.github/workflows/prettier.yml
```diff
@@ -0,0 +1,39 @@
...
+    runs-on: buildjet-8vcpu-ubuntu-2204
+    steps:
+      - name: Get sources
+        uses: actions/checkout@v3
+
+      - name: Run prettier on cli/
+        shell: bash
+        run: |
+          cd cli
+          npx prettier --check .
```
tomhoule, line 38: We should move towards formatting the entire repo, but there are configuration differences between cli/ and packages/, it will take some work to fix.

PR #1099, opened 2023-12-08
feat: Expose errors & execution metdata from engine v2 and merge several error types

Finistere:
Exposing errors & some execution metadata (operation type) for the gateway to add those in the response headers. Also replaced `ServerError`, `EngineError` with simply `GraphqlError`. There wasn't a need for the former ones.

Later we'll likely add more stuff into `ExecutionMetdata` or change that, so just tried to have the smallest API possible for now. In the gateway we check only whether there are errors currently, that's it.


pimeys (approved):
awesome

PR #1100, opened 2023-12-08
feat: variable validations in engine-v2

obmarg:
This implements & tests all the [GraphQL variable validations][1] in engine-v2 _except_ "All Variable Usages are Allowed" (because that one looks a bit more in-depth so will be in it's own PR/issue).

This should prevent an issue I found yesterday where undeclared variables cause a panic.

FIxes GB-5569

[1]: http://spec.graphql.org/October2021/#sec-Validation.Variables

### DIFF
--- a/engine/crates/engine-v2/src/request/bind.rs
+++ b/engine/crates/engine-v2/src/request/bind.rs
@@ -45,11 +45,25 @@ pub enum BindError {
     #[error("Leaf field '{name}' must be a scalar or an enum, but is a {ty}.")]
     LeafMustBeAScalarOrEnum { name: String, ty: String, location: Pos },
     #[error(
-        "Variable named '{name}' does not have a valid input type. Can only be a scalar, enum or input object. Found: '{ty}'."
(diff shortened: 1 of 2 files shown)


engine/crates/engine-v2/src/request/bind.rs
```diff
@@ -145,39 +173,56 @@ impl<'a> Binder<'a> {
...
                 let name = node.name.node.to_string();
+                let name_location = node.name.pos;
+
+                if seen_names.contains(&name) {
+                    return Err(BindError::DuplicateVariable {
+                        name,
+                        location: name_location,
+                    });
+                }
+                seen_names.insert(name.clone());
```
yoav-lavi, line 189: Modifying a mutable variable within a map might be something we want to avoid

yoav-lavi (commented):
See comment, looks good otherwise!

engine/crates/engine-v2/src/request/bind.rs
obmarg: I could rewrite it as a for, but I didn't see much benefit?

yoav-lavi (approved):

PR #1102, opened 2023-12-08
feat: add engine-v2 config to put headers in

obmarg:
This is an alternative approach to #1101.

As we discovered in the demo this morning, we need a way to specify headers that should be sent to a subgraph in engine-v2.  I'd initially planned to add these into the `federated-graph` structure, but on review two possible issues came up:

1. Putting secrets into the federated-graph might not be the best idea. We'll likely want to dispaly the federated-graph SDL to users at some point, and we may also want to generate a FederatedGraph out of its corresponding SDL.  If we put headers (and therefore secrets) into the FederatedGraph then the SDL would end up with secrets in it, which isn't great.
2. We've open sourced the federated-graph & composition libraries, as they're  currently relatively generic.  My original change would start adding grafbase specific things into otherwise generic libraries, which isn't ideal.

This PR proposes an alternative approach of adding a `engine-v2/config` crate that wraps FederatedGraph with engine-specific configuration.  We can use this to store headers for forwarding, and potentially any additional auth/caching/whatever configuration.

Part of GB-5590

[ci skip]


Finistere (approved):

tomhoule (approved):

engine/crates/engine-v2/config/src/lib.rs
```diff
@@ -0,0 +1,44 @@
...
+pub enum VersionedConfig {
+    /// The initial version of our configuration only contained the FederatedGraph.
+    V1(FederatedGraphV1),
+    /// V2 introduced some other configuration concerns
+    V2(v2::Config),
+}
+
+impl VersionedConfig {
+    /// Converts a config of any version into whatever the latest version is.
+    pub fn into_latest(self) -> latest::Config {
```
Finistere, line 33: nice!

Finistere (approved):

engine/crates/engine-v2/config/src/lib.rs
obmarg: Can you tell I've done this before?

Finistere: would be surprised if you didn't^^

PR #1105, opened 2023-12-08
feat: parse subgraph headers from SDL, pass to engine-v2

obmarg:
- Added a `@subgraph` directive to parser-sdl, which lets you add subgraph specific config to federated graphs.  Could probably have done this as  a parameter to `@graph` but I think a separate directive is nicer, and I assume most users will be using TS for this anyway.  Though open to discussions if anyone wants to do it differently.
- This doesn't cover headers shared amongst all subgraphs, I'll do that in a future commit.
- Added an `engine-config-builder` crate that can take the output from `parser-sdl` & `federated-graph` and combine it into an `engine-v2::VersionedConfig`.  Wasn't entirely sure about adding a new crate just for this, but I didn't really like the idea of having engine-v2-config pull in parser-sdl, or cluttering `engine-v2-config` with any extra stuff that's just used for building it.  If anyone has any better suggestions let me know.

Fixes GB-5603


jakubadamw (approved):
Awesome work!

PR #1115, opened 2023-12-11
fix: composition should short circuit error if composing with no graphs

pimeys:
This gives a much better error compared to whatever it is with missing query.

Part of: GB-5588


tomhoule (approved):
Neat!

PR #1120, opened 2023-12-11
test: allow engine-v2 tests to use engine-v1 as subgraph

obmarg:
I wanted to test some engine<->engine-v2 interactions, so figured I might as well add the code to integration-tests that allows us to do this there.

Currently only tests introspection, because that's what I needed to test - if you try making calls you _might_ need to fix some more stuff.

### DIFF
--- a/engine/crates/integration-tests/src/engine/mod.rs
+++ b/engine/crates/integration-tests/src/engine/mod.rs
@@ -189,13 +189,6 @@ impl crate::mocks::graphql::Schema for Engine {
 
         // Not sure this will work but lets see
         async_graphql::Response::deserialize(serde_json::to_value(response).unwrap()).unwrap()
-        // async_graphql::Response {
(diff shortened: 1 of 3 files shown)


engine/crates/integration-tests/src/engine/mod.rs
```diff
@@ -165,3 +166,39 @@ where
...
+
+        // Not sure this will work but lets see
+        async_graphql::Response::deserialize(serde_json::to_value(response).unwrap()).unwrap()
+        // async_graphql::Response {
+        //     data: async_graphql::Value::from_json(response.data.to_json_value().unwrap()).unwrap(),
+        //     extensions: response.extensions,
+        //     cache_control: response.cache_control.into(),
+        //     errors: response.errors,
+        //     http_headers: response.http_headers,
+        // }
```
obmarg, line 198:
```suggestion
```

Finistere (approved):

PR #1123, opened 2023-12-12
fix: a bad deprecation message in the sdk

obmarg:


yoav-lavi (approved):
Is this the only example of this? I think we added these before changing the export

obmarg:
@yoav-lavi Only one I could find (I grepped for `single)

PR #1127, opened 2023-12-12
feat(engine-v2): Execute mutations sequentially

Finistere:
For mutations, I'm now executing executors one at a time. The executor future returns then:
```
struct ExecutorFutResult<'ctx> {
    result: ExecutorResult<ExecutorOutput>,
    next_executors: VecDeque<Executor<'ctx>>,
}
```
So the actual result to update the response and plan the children. And the next executors, which are only used for mutation root fields. Ensuring they're executed one after another. Error management is more hacky than I thought during planning, but well good enough for now.

### DIFF
--- a/engine/crates/integration-tests/tests/federation/basic/mutation.rs
+++ b/engine/crates/integration-tests/tests/federation/basic/mutation.rs
@@ -1,0 +1,177 @@
+use engine_v2::Engine;
+use integration_tests::{federation::EngineV2Ext, mocks::graphql::StateMutationSchema, runtime, MockGraphQlServer};
+
+#[test]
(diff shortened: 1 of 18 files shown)


engine/crates/integration-tests/tests/federation/basic/mutation.rs
```diff
@@ -0,0 +1,177 @@
...
+        insta::assert_json_snapshot!(response, @r###"
+        {
+          "data": {
+            "first": 1,
+            "second": 2,
+            "third": 14,
+            "fourth": 3,
+            "fifth": 33
+          }
+        }
```
obmarg, line 42: Nice

obmarg (approved):

PR #1137, opened 2023-12-14
chore: rewrite two of the `@defer` cli tests

obmarg:
We're removing the DB in #1094 - this currently means we're ignoring a bunch of tests that rely on DB features.

I wanted to see how hard it'd be to rewrite them, so I've rewritten two of the `@defer` ones as an experiement.


cli/crates/cli/tests/utils/async_client.rs
```diff
@@ -29,7 +29,7 @@ impl AsyncClient {
             playground_endpoint,
             client: reqwest::Client::builder()
                 .connect_timeout(Duration::from_secs(1))
-                .timeout(Duration::from_secs(5))
+                .timeout(Duration::from_secs(20))
```
obmarg, line 32: Had to up this because resolver compiles can mean requests take a bit longer

fbjork (approved):

PR #1141, opened 2023-12-15
fix(messaging): GB-5625: Handle the “branch does not exist” error in the CLI's publish command

jakubadamw:
# Description

Handle the “branch does not exist” error in the CLI's publish command.

# Type of change

- [ ] 💔 Breaking
- [ ] 🚀 Feature
- [X] 🐛 Fix
- [ ] 🛠️ Tooling
- [ ] 🔨 Refactoring
- [ ] 🧪 Test
- [ ] 📦 Dependency

### DIFF
--- a/cli/crates/backend/src/api/publish.rs
+++ b/cli/crates/backend/src/api/publish.rs
@@ -2,7 +2,10 @@ use super::{
     client::create_client,
     consts::API_URL,
     errors::ApiError,
-    graphql::mutations::{FederatedGraphCompositionError, PublishPayload, SubgraphCreateArguments, SubgraphPublish},
+    graphql::mutations::{
+        BranchDoesNotExistError, FederatedGraphCompositionError, PublishPayload, SubgraphCreateArguments,
+        SubgraphPublish,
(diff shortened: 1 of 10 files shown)


cli/crates/backend/src/api/publish.rs
```diff
@@ -35,6 +38,9 @@ pub async fn publish(
         Some(PublishPayload::FederatedGraphCompositionError(FederatedGraphCompositionError { messages })) => {
             Ok(Err(messages.clone()))
         }
+        Some(PublishPayload::BranchDoesNotExistError(BranchDoesNotExistError { .. })) => {
+            Ok(Err(vec!["Provided branch does not exist in the project.".to_owned()]))
```
tomhoule, line 42: I think the caller will print that with something like "Published with composition errors: - Provided branch does not exist....". If it's the case, we have to introduce an enum.

jakubadamw: Addressed. 🙂

tomhoule (approved):

PR #1143, opened 2023-12-15
parser-sdl: fix extending type causing duplicate type definition error

tomhoule:
Roughly, there are three scenarios for extend type:

- extend type Query / extend type Mutation -> these are special cases, there are tests
- extend types for types from connectors -> also tested
- extend type for types defined in the config / sdl -> that's where the issue applies

In that last point, two subcases: extending a model and extending a type that is returned by a resolver extending a connector type. The first use case is being sunset. The second use case is tested and fixed in this PR.

closes GB-5067

# Type of change

- [ ] 💔 Breaking
- [ ] 🚀 Feature
- [x] 🐛 Fix
- [ ] 🛠️ Tooling
- [ ] 🔨 Refactoring
- [ ] 🧪 Test
- [ ] 📦 Dependency

### DIFF
--- a/engine/crates/parser-sdl/src/rules/basic_type.rs
+++ b/engine/crates/parser-sdl/src/rules/basic_type.rs
@@ -42,7 +42,8 @@ impl<'a> Visitor<'a> for BasicType {
         let directives = &type_definition.node.directives;
 
         if ["Query", "Mutation"].contains(&type_definition.node.name.node.as_str())
-            | directives.iter().any(|directive| directive.is_model())
+            || directives.iter().any(|directive| directive.is_model())
+            || type_definition.node.extend
         {
             return;
         }
--- a/engine/crates/parser-sdl/src/rules/basic_type.rs
+++ b/engine/crates/parser-sdl/src/rules/basic_type.rs
@@ -131,43 +131,7 @@ impl<'a> Visitor<'a> for BasicType {
         // final schema.
         add_input_type_non_primitive(ctx, object, &type_name);
 
-        // We also need to parse any @key directives
-        let key_directives = directives
-            .iter()
-            .filter(|directive| directive.node.name.node == "key")
-            .collect::<Vec<_>>();
-
-        let (oks, errors) = key_directives
-            .into_iter()
-            .map(|directive| {
-                Ok((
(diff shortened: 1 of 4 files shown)


engine/crates/parser-sdl/src/rules/basic_type.rs
```diff
@@ -42,7 +42,8 @@ impl<'a> Visitor<'a> for BasicType {
         let directives = &type_definition.node.directives;
 
         if ["Query", "Mutation"].contains(&type_definition.node.name.node.as_str())
-            | directives.iter().any(|directive| directive.is_model())
+            || directives.iter().any(|directive| directive.is_model())
+            || type_definition.node.extend
```
obmarg, line 46: I think some of my federation directive code was kind of relying on the fact that extensions went through this code path (which is probably what's causing the errors in tests).

Probably need to make sure all the federation directive stuff is present in `ExtendConnectorTypes` as well now

tomhoule: Good to know, I'll do some discovery and maybe write tests to make sure that doesn't break.

tomhoule: It looks like the federation directives _should_ be handled here: https://github.com/grafbase/grafbase/blob/6a5ec45803b4d698f217936fcc80593f1aadcf4e/engine/crates/parser-sdl/src/rules/extend_connector_types.rs#L47

I'll write at least one test to confirm though.

tomhoule: Done in [`7774665` (#1143)](https://github.com/grafbase/grafbase/pull/1143/commits/777466549ddc3be5c31f178430274a886209f0af)

Happy to dig in more if that doesn't build enough confidence.

tomhoule:
The integration test failure might be related.

obmarg:
> The integration test failure might be related.

[This code here](https://github.com/grafbase/grafbase/blob/main/engine/crates/parser-sdl/src/rules/basic_type.rs#L133-L169) probably also needs run in the ExtendConnectorTypes visitor now.

obmarg (approved):

PR #1144, opened 2023-12-15
feat(caching): support caching in federated graph

hfgbarrigas:
# Description

Adds initial support for global caching at the federated graph level. 

- Adjusted the TS SDK to support cache in federated config 
- Engine v2:
  - Adds a new `cache_config` field in `v2::Config` (latest)
  - Adjusts the conversion from `Config` to `Schema` so it deals with `cache_config` accordingly
  - Adds a `cache_config` field in `Operation`. This field is set after *binding* an unbound operation to a `Schema`
    - the result of this field is the merge of all `cache_config`'s for the query selection set (fields and objects)
  - Adds a `cache_config` to `ResponseMetadata`

A follow up PR will contain the necessary changes to extract the cache present in `gateway-core` to its own independent crate and adjust it so it's abstract enough to deal with engine v1 and v2.

Note: For now caching in v2 doesn't deal with caching scopes and mutation invalidation. Planning on adding scopes when auth is dealt with.

# Type of change

- [ ] 💔 Breaking
- [x] 🚀 Feature
- [ ] 🐛 Fix
- [ ] 🛠️ Tooling
- [ ] 🔨 Refactoring
- [ ] 🧪 Test
- [ ] 📦 Dependency


engine/crates/parser-sdl/src/rules/visitor/context.rs
```diff
@@ -219,6 +219,9 @@ impl<'a> VisitorContext<'a> {
             required_udfs.insert((UdfKind::Authorizer, name.clone()));
         }
 
+        // lets make sure we add the global rules to the federated graph config as well
```
hfgbarrigas, line 222: not entirely happy with this though

engine/crates/integration-tests/tests/federation/basic/caching.rs
```diff
@@ -0,0 +1,324 @@
+//! Tests of caching behaviour
```
Finistere, line 1: nice

engine/crates/parser-sdl/src/rules/visitor/context.rs
Finistere: You could probably avoid it by changing the places that build the config as the parent most likely has the config. But probably not that important for now. 🤷

Finistere (approved):
Nice work. I would be more explicit about the unit for `engine_v2_config::CacheConfig`. I find it a bit ambiguous what `max_age: usize` is supposed to be.

engine/crates/engine-v2/schema/src/cache.rs
```diff
@@ -0,0 +1,46 @@
+#[derive(Debug, Default, Clone, Copy, Eq, PartialEq)]
+pub struct CacheConfig {
+    pub max_age: usize,
+    pub stale_while_revalidate: usize,
+}
+
+pub trait Merge<T> {
```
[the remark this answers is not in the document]
hfgbarrigas: This is merely for me to use `merge` on `Option<CacheConfig>` [here](https://github.com/grafbase/grafbase/pull/1144/files#diff-44954d605b4942ddafcd1a5fec1a9dd7f825aeeec834ba135938cfe9a763d5a0R65), [here](https://github.com/grafbase/grafbase/pull/1144/files#diff-44954d605b4942ddafcd1a5fec1a9dd7f825aeeec834ba135938cfe9a763d5a0R114) and [here](https://github.com/grafbase/grafbase/pull/1144/files#diff-44954d605b4942ddafcd1a5fec1a9dd7f825aeeec834ba135938cfe9a763d5a0R97).

engine/crates/engine-v2/schema/src/cache.rs
```diff
@@ -0,0 +1,46 @@
...
+
+impl Merge<CacheConfig> for CacheConfig {
+    fn merge(self, right: CacheConfig) -> CacheConfig {
+        CacheConfig {
+            max_age: if self.max_age == 0 {
+                right.max_age
+            } else if right.max_age == 0 {
+                self.max_age
+            } else {
+                self.max_age.min(right.max_age)
```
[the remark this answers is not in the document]
hfgbarrigas: min is enough. Remnants from v1 logic.

engine/crates/engine-v2/schema/src/cache.rs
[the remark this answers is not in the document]
Finistere: Ah ok, didn't see the `Option` implementation, never mind then :)

engine/crates/engine-v2/config/src/lib.rs
```diff
@@ -2,6 +2,7 @@ use federated_graph::FederatedGraphV1;
 
 // The specific version modules should be kept private, users of this crate
 // should only access types via `latest`
+mod cache_config;
```
Finistere, line 5: should be part of `v2`

engine/crates/engine-v2/config/src/lib.rs
```diff
@@ -38,6 +39,7 @@ impl VersionedConfig {
                 headers: Default::default(),
                 default_headers: Default::default(),
                 subgraph_configs: Default::default(),
+                cache_config: Default::default(),
```
Finistere, line 42: Could just be named `cache`, it's part of `Config` struct. So it's natural for it to be configuration.

engine/crates/engine-v2/config/src/v2.rs
```diff
@@ -14,6 +15,10 @@ pub struct Config {
 
     /// Additional configuration for our subgraphs
     pub subgraph_configs: BTreeMap<SubgraphId, SubgraphConfig>,
+
+    /// Caching configuration
+    #[serde(default)]
+    pub cache_config: BTreeMap<CacheConfigTarget, CacheConfig>,
```
Finistere, line 21: should be a struct imho, like:
```
pub struct CacheConfig {
  rules: BTreeMap<CacheConfigTarget, CacheControl>
}
```
Would provide more flexibility in case of changes later.

Finistere (commented):
A few additional comments on the configuration

PR #1145, opened 2023-12-15
fix: correctly handle reloads in federated dev 

obmarg:
First off: sorry to whoever reviews this.  Might have got a bit carried away with the refactoring.

The dev command watches for changes in files.  Prior to this PR when it saw a change it would just restart everything it was running.  This has been fine for ages as we've had little internal state.  However, the federated dev command does have internal state, and this restart drops all of that state - leaving users to re-publish all their graphs whenever they change some config.  Not very good.

My solution for this is to read the config on startup and either go into federated mode (which does not restart on changes) or go into the standalone dev mode, which will restart things as it did before.  

I've changed _a lot_ of things to get this working right though: some changes I needed to make, some I just thought were worthwhile.  Hopefully nobody disagrees 😓 

1. I've stopped using the `event_bus` for file updates, instead they go on their own broadcast channel.
2. I've added a `ConfigActor` abstraction on top of that - it'll watch for changes to config files, and run through `node`/`parser-sdl` to get the current config.  `ConfigActor` exposes these via async `Stream`s to the various places that need config.
4. `federated-dev` now uses watch channels for `FederatedGraphConfig` & `FederatedGraph`.  These are like regular channels, but they cache their last value.  Very nice for this case.
5. The `Watcher` channel & `ConfigActor` channels made the `Event::Reload` & `Event::NewSdlFromTsConfig` a bit redundant, so I rewrote things to just use the relevant channel.
6. At this point the event_bus was mostly being used for fairly local signalling - so I reworked its remaining uses.

And probably a bunch of other small changes - if anything is unclear (or looks stupid) let me know.

Fixes GB-5607


cli/crates/cli/tests/openapi/introspection_headers.rs
```diff
@@ -34,7 +34,7 @@ fn schema(address: &SocketAddr) -> String {
         r#"
           extend schema
           @openapi(
-            namespace: "petstore",
+            name: "petstore",
```
obmarg, line 37: No idea why this started failing here: looks like it's been broken for a while?

engine/crates/federated-graph/src/federated_graph.rs
```diff
@@ -9,7 +9,7 @@
 /// Does not guarantee:
 ///
 /// - The ordering of items inside each `Vec`.
-#[derive(serde::Serialize, serde::Deserialize)]
+#[derive(serde::Serialize, serde::Deserialize, Clone)]
```
tomhoule, line 12: nitpicking that may be irrelevant since I haven't read the rest of the diff yet, but an `Arc<FederatedGraph>` would be cheaper to clone than what I'm assuming is cloning the entire data structure here.

cli/crates/server/src/node.rs
```diff
@@ -4,41 +4,56 @@ use tokio::process::Command;
...
+    /// returned if node is not in the user $PATH
+    #[error("Node.js does not seem to be installed")]
+    NodeInPath,
+
+    /// returned if the installed version of node is unsupported
+    #[error("Node.js version {0} is unsupported")]
+    OutdatedNode(String, String),
+
+    /// returned if the installed version of node could not be retreived
+    #[error("Could not retrive the installed version of Node.js")]
```
tomhoule, line 20:
```suggestion
    #[error("Could not retrieve the installed version of Node.js")]
```

cli/crates/server/src/servers.rs
```diff
@@ -435,22 +469,29 @@ pub async fn get_random_port_unchecked() -> Result<u16, ServerError> {
...
+            PortSelection::Specific(port) => TcpListener::bind((Ipv4Addr::LOCALHOST, port))
+                .await
+                .map_err(|_| ServerError::PortInUse(port))?
+                .into_std()
+                .map_err(|_| ServerError::PortInUse(port)),
+        }
     }
 }
 
+/// determines if a port or port range are available
```
tomhoule, line 493: This looks like an orphan comment

tomhoule (approved):
Much cleaner!

engine/crates/federated-graph/src/federated_graph.rs
obmarg: Yeah, I considered that.  But the engine wants to take an owned FederatedGraph so we'd still need to be able to clone the contents.

cli/crates/server/src/servers.rs
```diff
@@ -435,22 +469,29 @@ pub async fn get_random_port_unchecked() -> Result<u16, ServerError> {
...
+
```
obmarg, line 494:
```suggestion
```

PR #1154, opened 2023-12-19
composition: implement non-default root type support

tomhoule:
Concretely, we now take into account the schema definitions (e.g. `schema { query: MyQuery }`).

There are subtleties in the rules:

- If there is a custom mutation root (or query, or subscription) and a type that has the default name for the root (Query, Mutation or Subscription), it is a hard error, to align with expectations.
- If a type is a root in some subgraphs but not in others, it will be handled as a regular type in the subgraphs where it isn't the root and as the corresponding root where it is. We are better behaved than other implementations in that regard.

There are many changes in this PR that look unrelated to the matter at hand. This is because until now, we could assume that all the types with the same name compose together (and their fields too) across subgraphs. This is no longer the case, take these two subgraphs for example:

```graphql

# Subgraph A

schema {
  query: CocoQuery
}

type CocoQuery {
  coconutYoghurtVariants: [String]
}

# Subgraph B

type Query {
  soyYoghurtVariants: [String]
}

```

The `CocoQuery` and `Query` type will compose together as the `Query` type of the federated schema. This broke a number of assumptions about composition, so it had to be addressed. Among other things, special treatment for root types reorders them in the federated schema, hence the snapshot changes.

closes GB-5673


engine/crates/composition/src/compose/context.rs
```diff
@@ -156,15 +149,7 @@ impl<'a> Context<'a> {
...
             .definitions_by_name
-            .insert(object_name.id, federated::Definition::Object(id));
-
-        // FIXME: Those roots probably shouldn't be hardcoded.
-        match object_name.as_str() {
-            "Query" => self.ir.query_type = Some(id),
-            "Mutation" => self.ir.mutation_type = Some(id),
-            "Subscription" => self.ir.subscription_type = Some(id),
-            _ => (),
-        }
```
tomhoule, line 167: Now they aren't anymore :tada:

engine/crates/composition/src/compose/context.rs
```diff
@@ -241,6 +225,28 @@ impl<'a> Context<'a> {
...
+
+    pub(crate) fn set_mutation(&mut self, id: federated::ObjectId) {
+        self.ir.mutation_type = Some(id);
+    }
+
+    pub(crate) fn set_subscription(&mut self, id: federated::ObjectId) {
+        self.ir.subscription_type = Some(id);
+    }
+
+    pub(crate) fn insert_object_field_from_entity_interface(
```
tomhoule, line 241: This had to be split as a cascading change, because of string interning constants. I find the new implementation more straightforward regardless.

engine/crates/federated-graph/src/render_sdl.rs
```diff
@@ -88,6 +88,8 @@ pub fn render_sdl(graph: &FederatedGraph) -> Result<String, fmt::Error> {
                 write_field(field.field_id, graph, &mut sdl)?;
             }
             writeln!(sdl, "}}\n")?;
+        } else {
+            sdl.push_str("\n\n");
```
tomhoule, line 92: Empty types look weird otherwise. They should not happen, but it doesn't cost anything to render them properly anyway.

engine/crates/composition/src/compose/roots.rs
```diff
@@ -0,0 +1,59 @@
...
+    root: &'static str,
+    definitions: &[subgraphs::DefinitionWalker<'a>],
+    ctx: &mut Context<'a>,
+) -> Option<federated::ObjectId> {
+    if definitions.is_empty() {
+        return None;
+    }
+
+    let type_name = ctx.insert_static_str(root);
+    let directives = collect_composed_directives(definitions.iter().map(|def| def.directives()), ctx);
```
tomhoule, line 49: We don't filter, we take all the directives as if the root were a regular object, so you can have `type Query @shareable` for example, which makes little sense. We can refine later.

engine/crates/composition/src/composition_ir.rs
```diff
@@ -32,8 +32,12 @@ pub(crate) struct CompositionIr {
 
     pub(crate) strings: StringsIr,
     pub(crate) fields: Vec<FieldIr>,
-    pub(crate) union_members: BTreeSet<(subgraphs::StringId, subgraphs::StringId)>,
+    pub(crate) union_members: BTreeSet<(federated::StringId, federated::StringId)>,
     pub(crate) resolvable_keys: Vec<KeyIr>,
+
+    /// Fields on implementers of an interface entity's implementers that are contributed by other
+    /// subgraphs.
```
tomhoule, line 39: Reformulate. "Fields of an interface entity that are contributed by other subgraphs and must be added to the interface's implementers in the federated schema".

tomhoule:
This PR made me realize I have to clean up string interning, currently the assumption that two strings that are the same have the same id can be broken in edge cases in the federated graph, that should be fixed.

engine/crates/composition/src/ingest_subgraph.rs
```diff
@@ -49,13 +53,33 @@ fn ingest_top_level_definitions(
...
+                                subgraphs.set_query_type(subgraph_id, definition_id);
+                            }
+                            RootTypeMatch::Mutation => {
+                                subgraphs.set_mutation_type(subgraph_id, definition_id);
+                            }
+                            RootTypeMatch::Subscription => {
+                                subgraphs.set_subscription_type(subgraph_id, definition_id);
+                            }
+                            RootTypeMatch::NotRootButHasDefaultRootName => {
+                                subgraphs.push_ingestion_diagnostic(subgraph_id, format!("The {type_name} type has the default name for a root but is itself not a root. This is not valid in a federation context."));
```
obmarg, line 76: Just curious: did we define this rule or is it standard elsewhere?

obmarg (approved):

engine/crates/composition/src/ingest_subgraph.rs
tomhoule: It's a rule replicated from the implementations we compare with. The message is not exactly the same but it's the same rule.

PR #1156, opened 2023-12-19
cli: fix ts code generation

tomhoule:
It broke as a result of e867890. We started sending the path to the schema as a string instead of the schema to the codegen function.

The solution requires merging the two tasks, because they share `last_seen_sdl`.

closes GB 5676

# Type of change

- [ ] 💔 Breaking
- [ ] 🚀 Feature
- [x] 🐛 Fix
- [ ] 🛠️ Tooling
- [ ] 🔨 Refactoring
- [ ] 🧪 Test
- [ ] 📦 Dependency

### DIFF
--- a/cli/crates/server/src/codegen_server.rs
+++ b/cli/crates/server/src/codegen_server.rs
@@ -6,7 +6,7 @@ use std::{
 
 use common::environment::{Project, SchemaLocation};
 use tokio::task::JoinHandle;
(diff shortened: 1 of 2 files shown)


obmarg (approved):
Oops, my bad.

tomhoule:
>  Oops, my bad.

I reviewed and I missed it, so no reason to feel bad :p

PR #1158, opened 2023-12-19
composition: restore broken assumption about interned strings

tomhoule:
A while ago, we introduced a separate string interning map for static strings that may not show up in the subgraphs but are part of the federated graph and as such, need to be in `FederatedGraph.strings`.

The problem is that the same string can end up in both the map for strings from the subgraphs, and, if the static version was inserted before, in the second map for static strings.

The important consequence of this is that you could end up with two `StringId`s that are not equal representing the same string, which defeats the point of one big reason for interning in the first place: cheap and reliable comparisons.

This commit fixes that by not trying to be smart about converting `subgraphs::StringId`s into `federated::StringId`s and rehashing the string themselves instead.

### DIFF
--- a/engine/crates/composition/src/emit_federated_graph.rs
+++ b/engine/crates/composition/src/emit_federated_graph.rs
@@ -20,7 +20,7 @@ pub(crate) fn emit_federated_graph(mut ir: CompositionIr, subgraphs: &Subgraphs)
         unions: mem::take(&mut ir.unions),
         scalars: mem::take(&mut ir.scalars),
         input_objects: mem::take(&mut ir.input_objects),
-        strings: mem::take(&mut ir.strings.strings),
+        strings: Vec::new(),
         subgraphs: vec![],
         root_operation_types: RootOperationTypes {
             query: ir.query_type.unwrap(),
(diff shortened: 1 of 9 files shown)


engine/crates/composition/src/emit_federated_graph.rs
```diff
@@ -20,7 +20,7 @@ pub(crate) fn emit_federated_graph(mut ir: CompositionIr, subgraphs: &Subgraphs)
         unions: mem::take(&mut ir.unions),
         scalars: mem::take(&mut ir.scalars),
         input_objects: mem::take(&mut ir.input_objects),
-        strings: mem::take(&mut ir.strings.strings),
+        strings: Vec::new(),
```
tomhoule, line 23: This is another bug we just hadn't ran into: we were moving the contents of StringIr out, then kept using it in context, so duplicate string ids again.

engine/crates/composition/src/emit_federated_graph/context.rs
```diff
@@ -54,3 +50,9 @@ impl<'a> Context<'a> {
             .push(federated::InterfaceField { interface_id, field_id });
     }
 }
+
+impl Drop for Context<'_> {
+    fn drop(&mut self) {
+        self.out.strings = std::mem::take(&mut self.strings_ir).into_federated_strings();
```
tomhoule, line 56: We could assert that `self.out.strings` is empty at this point, but since the drop impls run during unwinding, it would be a panic in a panic, and these are terrible.

Other possible solutions:

- Regular `finalize()` method instead of Drop.
- Put the IndexSet directly inside `FederatedGraph`. It should be backwards compatible, and there would be no more room for error.

obmarg (approved):

PR #1160, opened 2023-12-20
GB-5629 - Scalar and create exports for the SDK

yoav-lavi:
# Description

## Features

- Adds `scalar` and `define` exports to the SDK for external type definitions
- Adds an `add` function to `Graph` to attach external types to the graph

# Type of change

- [ ] 💔 Breaking
- [x] 🚀 Feature
- [ ] 🐛 Fix
- [ ] 🛠️ Tooling
- [ ] 🔨 Refactoring
- [ ] 🧪 Test
- [ ] 📦 Dependency


yoav-lavi:
May add a bit more to the test but ready for review

packages/grafbase-sdk/src/define.ts
```diff
@@ -0,0 +1,141 @@
...
+import { Query, QueryInput } from './query'
+import { Type, TypeFields } from './type'
+import { EnumDefinition } from './typedefs/enum'
+import { InputDefinition } from './typedefs/input'
+import { ReferenceDefinition } from './typedefs/reference'
+import { Union } from './union'
+
+export default {
+  /**
+   * Creates a new composite type.
```
tomhoule, line 13: "Object type" would be closer to the standard imo, if I'm understanding the purpose right. But just an opinion, no need to change.

tomhoule (approved):
The code is very nice and easy to follow :100:

pimeys (approved):
Super nice code, a good change also. Maybe we should have some tests with the new syntax too?

packages/grafbase-sdk/src/define.ts
obmarg: Agree on this.  The spec does use "composite type" in a few places, but its [definition includes input objects](http://spec.graphql.org/October2021/#sec-The-__Type-Type.Input-Object) rather than just objects

yoav-lavi: I used the existing comment here but don't mind changing it

yoav-lavi, line 13:
```suggestion
   * Creates a new object type.
```

yoav-lavi: Done

PR #1163, opened 2023-12-20
feat(sdl): Add support for operation limits to the SDL

jakubadamw:
# Description

Initial PR in anticipation of further work on the platform.

# Type of change

- [ ] 💔 Breaking
- [X] 🚀 Feature
- [ ] 🐛 Fix
- [ ] 🛠️ Tooling
- [ ] 🔨 Refactoring
- [ ] 🧪 Test
- [ ] 📦 Dependency


engine/crates/parser-sdl/src/rules/operation_limits_directive.rs
```diff
@@ -0,0 +1,164 @@
...
+    use futures::executor::block_on;
+
+    use crate::connector_parsers::MockConnectorParsers;
+
+    #[test]
+    fn parsing_operation_limits_directive() {
+        let schema = r#"
+            extend schema
+              @operationLimits(
+                depth: 5,
```
obmarg, line 134: Not directly related to this PR, just a thought i had while reading: pretty sure a `depth: 5` setting would break the standard introspection query quite horribly.

Wonder if we need to care about that...

obmarg (approved):
Nice.

engine/crates/parser-sdl/src/rules/operation_limits_directive.rs
jakubadamw: This is a great point, thanks! We can figure this out when I implement the actual enforcement of these limits, but I'd be leaning towards excluding introspection-shaped queries from limits.

PR #1162, opened 2023-12-20
parser-sdl: make mongodb connector produce valid schemas 

tomhoule:
The MongoDB connector was producing registries with many fields referencing types that do not exist. The changes in this PR make it produce more valid GraphQL schemas, where the types of input fields exist.

Additionally, schemas with relations (fields on a model that have another model as their type) silently produced invalid schemas previously, now they produce a readable validation error.

closes GB-5677

# Type of change

- [ ] 💔 Breaking
- [ ] 🚀 Feature
- [x] 🐛 Fix
- [ ] 🛠️ Tooling
- [ ] 🔨 Refactoring
- [ ] 🧪 Test
- [ ] 📦 Dependency
- [ ] 📖 Requires documentation update

PR #1169, opened 2023-12-21
cli: release 0.52.1

tomhoule:
Fixes

- Make MongoDB connector produce valid schemas when introspected (#1162)
- Fix a bug in the rendering of GraphQL schemas where the interface description would be printed after its name (#1166)

### DIFF
--- a/cli/changelog/0.52.1.md
+++ b/cli/changelog/0.52.1.md
@@ -1,0 +1,4 @@
+### Fixes
+
+- Make MongoDB connector produce valid schemas when introspected (#1162)
+- Fix a bug in the rendering of GraphQL schemas where the interface description would be printed after its name (#1166)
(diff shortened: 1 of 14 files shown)


fbjork (approved):

cli/changelog/0.52.1.md
```diff
@@ -0,0 +1,4 @@
+### Fixes
+
+- Make MongoDB connector produce valid schemas when introspected (#1162)
+- Fix a bug in the rendering of GraphQL schemas where the interface description would be printed after its name (#1166)
```
obmarg, line 4: Don't think this bug would ever have manifested in the CLI - probably doesn't need to be in the changelog

obmarg (approved):

cli/changelog/0.52.1.md
tomhoule: Hmm, the other one would also only show up in `dev`, so do we need this release at all?

PR #1172, opened 2023-12-22
feat(engine): Enforce operation limits in engine v1

jakubadamw:
# Description

Add enforcement of operation limits as declared in the SDL to the 1st version of the engine.

# Type of change

- [ ] 💔 Breaking
- [X] 🚀 Feature
- [ ] 🐛 Fix
- [ ] 🛠️ Tooling
- [ ] 🔨 Refactoring
- [ ] 🧪 Test
- [ ] 📦 Dependency

### DIFF
--- a/engine/crates/engine/src/schema.rs
+++ b/engine/crates/engine/src/schema.rs
@@ -30,8 +30,6 @@ pub struct SchemaBuilder {
     validation_mode: ValidationMode,
     registry: Registry,
     data: Data,
-    complexity: Option<usize>,
-    depth: Option<usize>,
     extensions: Vec<Box<dyn ExtensionFactory>>,
 }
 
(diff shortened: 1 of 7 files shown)


engine/crates/engine/src/schema.rs
```diff
@@ -186,8 +172,10 @@ impl Deref for SchemaEnv {
 #[doc(hidden)]
 pub struct SchemaInner {
     pub(crate) validation_mode: ValidationMode,
-    pub(crate) complexity: Option<usize>,
-    pub(crate) depth: Option<usize>,
+    pub(crate) max_complexity: Option<usize>,
```
Finistere, line 175: Could also be in a `OperationLimits` struct with a `Default`

engine/crates/engine/src/validation/mod.rs
```diff
@@ -32,6 +32,12 @@ pub struct ValidationResult {
 
     /// Query depth
     pub depth: usize,
+
+    /// Query height
```
Finistere, line 36: Not super obvious what `height` meant before reading the tests to be honest^^

Finistere (approved):

engine/crates/engine/src/schema.rs
jakubadamw: Good idea!

engine/crates/engine/src/validation/mod.rs
jakubadamw: Added a comment.

PR #1178, opened 2023-12-27
chore: a bit of refactor on gateway-core and its caching bits

hfgbarrigas:
Refactors gateway-core a bit and extracts some caching bits that should be generic and independent of the specific runtime.

# Type of change

- [ ] 💔 Breaking
- [ ] 🚀 Feature
- [ ] 🐛 Fix
- [ ] 🛠️ Tooling
- [x] 🔨 Refactoring
- [ ] 🧪 Test
- [ ] 📦 Dependency


engine/crates/engine-v2/src/response/mod.rs
```diff
@@ -23,6 +25,20 @@ pub enum Response {
 // Our internal error struct shouldn't be accessible. It'll also need some context like
 // ResponseKeys to even just present paths correctly.
 pub struct Error<'a>(&'a GraphqlError);
+impl<'a> Error<'a> {
+    pub fn message(&self) -> &String {
```
obmarg, line 29: Nitpicking a bit, but any reason for this not to be `&str`?

```suggestion
    pub fn message(&self) -> &str {
```

engine/crates/engine-v2/src/response/mod.rs
```diff
@@ -23,6 +25,20 @@ pub enum Response {
...
+        &self.0.message
+    }
+    pub fn locations(&self) -> &Vec<Pos> {
+        &self.0.locations
+    }
+    pub fn path(&self) -> &Option<ResponsePath> {
+        &self.0.path
+    }
```
obmarg, line 37: Similar here:

```suggestion
    pub fn path(&self) -> Option<&ResponsePath> {
        self.0.path.as_ref()
    }
```

engine/crates/gateway-core/src/lib.rs
```diff
@@ -109,23 +110,48 @@ where
...
+
+                return Response::engine(Arc::new(result));
+            }
+
+            match build_cache_key(&self.cache_config, ctx.as_ref(), &request, &auth) {
+                Ok(cache_key) => {
+                    let execution_fut = Arc::clone(&self.executor)
+                        .execute(Arc::clone(ctx), auth, request)
+                        .instrument(info_span!("execute"))
+                        .map(move |res| res.map(Arc::new));
```
obmarg, line 127: Don't think this needs to be `move`?
```suggestion
                        .map(|res| res.map(Arc::new));
```

engine/crates/gateway-core/src/lib.rs
```diff
@@ -109,23 +110,48 @@ where
...
+                    )
+                    .await;
+
+                    cache::process_execution_response(ctx.as_ref(), cached_execution)
+                }
+                Err(_) => {
+                    let result = Arc::clone(&self.executor)
+                        .execute(Arc::clone(ctx), auth, request)
+                        .instrument(info_span!("execute"))
+                        .map(move |res| res.map(Arc::new))
```
obmarg, line 144: Or this
```suggestion
                        .map(|res| res.map(Arc::new))
```

engine/crates/runtime/src/cache/cached.rs
```diff
@@ -0,0 +1,895 @@
...
+
+#[derive(Debug, PartialEq, Eq)]
+enum CacheResponse<T> {
+    Hit(T),
+    Miss(T),
+    Bypass(T),
+    Stale { response: T, updating: bool },
+}
+
+impl<T: Cacheable> From<CacheResponse<Arc<T>>> for CachedExecutionResponse<Arc<T>> {
```
obmarg, line 22: Just asking, but why do we need both `CacheResponse` and `CachedExecutionResponse`?  They seem very similar just from glancing at this code.

engine/crates/runtime/src/cache/cached.rs
```diff
@@ -0,0 +1,895 @@
...
+                tracing::info!(ray_id, "Purging global cache by tags: {:?}", purge_cache_tags);
+
+                ctx.wait_until(
+                    async_runtime::make_send_on_wasm(async move {
+                        if let Err(err) = cache
+                            .purge_by_tags(purge_cache_tags)
+                            .instrument(tracing::info_span!("cache_tags_purge", ray_id))
+                            .await
+                        {
+                            tracing::error!(ray_id, "Error global cache purge by tags: {}", err);
```
obmarg, line 155: Do we know what order workers-rs runs the wait_until futures in?  Just wondering if this will end up running before or after the wait_until we use for logging...

obmarg (approved):

engine/crates/runtime/src/cache/cached.rs
hfgbarrigas: You're correct, there _was_ a reason but completely unnecessary.

engine/crates/runtime/src/cache/cached.rs
hfgbarrigas: couldn't find any details about `wait_until` future execution. But, looking at our code any `wait_until` future will run before any post request clean up, [here](https://github.com/grafbase/api/blob/main/workers/gateway/src/lib.rs#L97) and [here](https://github.com/grafbase/api/blob/main/workers/executor/src/lib.rs#L84).

obmarg: Ah yes cool - forgot we'd done our own `wait_until` thing on top of the CF one.  Nice one.

PR #1180, opened 2023-12-28
chore(deps): GB-5699: Upgrade axum, hyper, http and friends in the CLI

jakubadamw:
# Description

Upgrade axum, hyper, http and other HTTP-related dependencies.

# Type of change

- [ ] 💔 Breaking
- [ ] 🚀 Feature
- [ ] 🐛 Fix
- [ ] 🛠️ Tooling
- [ ] 🔨 Refactoring
- [ ] 🧪 Test
- [X] 📦 Dependency

### DIFF
--- a/cli/crates/gateway/src/executor.rs
+++ b/cli/crates/gateway/src/executor.rs
@@ -1,6 +1,5 @@
-use std::{collections::HashMap, sync::Arc};
+use std::{collections::HashMap, str::FromStr, sync::Arc};
 
-use axum::response::IntoResponse;
 use common_types::auth::ExecutionAuth;
 use dynamodb::{DynamoDBBatchersData, DynamoDBContext};
 use engine::{registry::resolvers::graphql, RequestHeaders};
@@ -168,10 +167,21 @@ impl gateway_core::Executor for Executor {
     ) -> Result<Self::Response, crate::Error> {
         let schema = self.build_schema(&ctx, auth).await?;
         let payload_stream = Box::pin(schema.execute_stream(request));
+        let mut response_builder = axum::response::Response::builder();
         let (headers, bytes_stream) =
             gateway_core::encode_stream_response(ctx.as_ref(), payload_stream, streaming_format).await;
-        Ok((headers, axum::body::StreamBody::new(bytes_stream))
-            .into_response()
+        response_builder
+            .headers_mut()
+            .unwrap()
+            .extend(headers.into_iter().map(|(name, value)| {
+                (
+                    name.map(|name| axum::http::HeaderName::from_str(name.as_str()).expect("must be a valid name")),
+                    axum::http::HeaderValue::from_bytes(value.as_bytes()).expect("must be a valid value"),
+                )
+            }));
+        Ok(response_builder
+            .body(axum::body::Body::from_stream(bytes_stream))
+            .unwrap()
             .into())
     }
 }
--- a/cli/crates/gateway/src/executor.rs
+++ b/cli/crates/gateway/src/executor.rs
@@ -1,4 +1,4 @@
-use std::{collections::HashMap, str::FromStr, sync::Arc};
+use std::{collections::HashMap, sync::Arc};
 
 use common_types::auth::ExecutionAuth;
 use dynamodb::{DynamoDBBatchersData, DynamoDBContext};
@@ -165,23 +165,13 @@ impl gateway_core::Executor for Executor {
         request: engine::Request,
         streaming_format: StreamingFormat,
     ) -> Result<Self::Response, crate::Error> {
+        use axum::response::IntoResponse;
         let schema = self.build_schema(&ctx, auth).await?;
         let payload_stream = Box::pin(schema.execute_stream(request));
-        let mut response_builder = axum::response::Response::builder();
         let (headers, bytes_stream) =
             gateway_core::encode_stream_response(ctx.as_ref(), payload_stream, streaming_format).await;
-        response_builder
-            .headers_mut()
-            .unwrap()
-            .extend(headers.into_iter().map(|(name, value)| {
-                (
-                    name.map(|name| axum::http::HeaderName::from_str(name.as_str()).expect("must be a valid name")),
-                    axum::http::HeaderValue::from_bytes(value.as_bytes()).expect("must be a valid value"),
-                )
-            }));
-        Ok(response_builder
-            .body(axum::body::Body::from_stream(bytes_stream))
-            .unwrap()
+        Ok((headers, axum::body::Body::from_stream(bytes_stream))
+            .into_response()
             .into())
     }
 }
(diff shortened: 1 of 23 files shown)


cli/crates/gateway/src/executor.rs
```diff
@@ -168,10 +167,21 @@ impl gateway_core::Executor for Executor {
...
         let schema = self.build_schema(&ctx, auth).await?;
         let payload_stream = Box::pin(schema.execute_stream(request));
+        let mut response_builder = axum::response::Response::builder();
         let (headers, bytes_stream) =
             gateway_core::encode_stream_response(ctx.as_ref(), payload_stream, streaming_format).await;
-        Ok((headers, axum::body::StreamBody::new(bytes_stream))
-            .into_response()
+        response_builder
+            .headers_mut()
+            .unwrap()
```
yoav-lavi, line 175: Would it be possible to either add an error here or an expect depending on whether it's supposed to be fallible? Same for the other unwrap

cli/crates/server/src/proxy.rs
```diff
@@ -84,12 +88,14 @@ async fn start_inner(listener: TcpListener) -> Result<(), ServerError> {
...
-        .map_err(ServerError::StartProxyServer)?
-        .http1_preserve_header_case(true)
-        .serve(router.into_make_service())
-        .await
-        .map_err(ServerError::StartProxyServer)?;
+    axum::serve(
+        tokio::net::TcpListener::from_std(listener).map_err(ServerError::StartProxyServer)?,
+        router,
+    )
+    // FIXME: Bring back!
```
yoav-lavi, line 95: Might be breaking, we had various tests that expected cased headers (which is why I preserved the behavior here). Not sure it's critical though

cli/crates/server/src/proxy.rs
```diff
@@ -120,11 +126,15 @@ async fn graphql_inner(
     mut req: Request<Body>,
     path: &str,
 ) -> Result<impl IntoResponse, impl IntoResponse> {
+    const BODY_LIMIT: usize = 1_024 * 1_024 * 512;
```
yoav-lavi, line 129: What's this based on?

cli/crates/server/src/servers.rs
```diff
@@ -371,16 +379,19 @@ async fn spawn_servers(
         port
     };
 
-    let gateway_server = axum::Server::bind(&"127.0.0.1:0".parse().unwrap()).serve(
+    let tcp_listener = tokio::net::TcpListener::bind("127.0.0.1:0").await.unwrap();
```
yoav-lavi, line 382: See earlier comment on unwrap

yoav-lavi (changes requested):
Looks good, see comments

cli/crates/server/src/proxy.rs
jakubadamw: Yes, this is definitely something to address. Sadly, the new `axum::serve()` interface doesn't allow any configuration, and opting out of it would require a lot of boilerplate. I'm thinking, however, of doing something like [this](https://github.com/grafbase/grafbase/commit/a8c785a592cc03caf951a38f7a640111842ee17b). This is based on a [proposed change](https://github.com/grafbase/axum/commit/f9f1e4299ce247604ef2975171d0662e0fff5008) to `axum` from our fork. Do you think it makes sense? If it does, I'd try to send it upstream.

cli/crates/server/src/proxy.rs
jakubadamw: An arbitrary number, really.

cli/crates/server/src/servers.rs
jakubadamw: Yes, this one is an oversight!

cli/crates/server/src/proxy.rs
yoav-lavi: Does CF have a max body size? Would probably be more accurate to use that

cli/crates/server/src/proxy.rs
yoav-lavi: It'd work, but perhaps 
```rs
    axum::build().preserve_header_case(true).serve(
         tokio::net::TcpListener::from_std(listener).map_err(ServerError::StartProxyServer)?,
         router
     )
```
?

yoav-lavi: (or `Builder` or something similar)

cli/crates/server/src/proxy.rs
yoav-lavi: Cloudflare Plan	Maximum body size
Free	100 MB
Pro	100 MB
Business	200 MB
Enterprise	500 MB (by default)

yoav-lavi: We could pin it to whatever we're using since it'd mimic production (and possibly add a flag to make it larger for on-prem)

yoav-lavi: Although looks like it's already around the largest size so can be left as is I guess

cli/crates/server/src/proxy.rs
jakubadamw: That configuration option is specific to HTTP1, though, hence why it's under `http1()`. What they're saying in the documentation for `serve()` is that if one wants to customise the server's configuration, one ought to use hyper's `Server` directly, which I am not a fan of. Hmm…

jakubadamw: @yoav-lavi what do you think we should do in the mean time? I feel like proceeding with the fork is odd since they're unlikely to accept the change. And if they could accept it, I wouldn't want to block on that happening. Is this setting critical enough?

BTW, you said:
> we had various tests that expected cased headers
To be fair, all the tests passed before I implemented this new way to turn that setting on, so I wonder where it would be critical?

cli/crates/server/src/proxy.rs
jakubadamw: @yoav-lavi hah, that's funny, I wasn't following Cloudflare there. I'll add a comment referring to the numbers you quoted. Thanks so much for investigating!

cli/crates/server/src/proxy.rs
jakubadamw: Actually, this PR also doesn't pull in that fork (I did it in a separate branch), and it does pass.

yoav-lavi: @jakubadamw I think we disabled those once we removed the DB (they were mostly auth related), but I was thinking more of client implementations that don't take into account that headers can be lowercase and may break. We might be able to do that and let them know though.

yoav-lavi: Since for now on prem isn't used yet we probably can proceed without the fork but we may want to check

yoav-lavi (approved):
Nice work!

jakubadamw:
> Nice work!

Thank you, sir!

PR #1191, opened 2024-01-04
 operation-checks: initial crate with breaking up of operations into fields

tomhoule:
This PR adds a crate in engine/crates. The crate is called `operation-checks` and is meant to hold the business logic for operation checks.

At a high level, within a schema check, the operation check step does the following work:

1. Gathers the fields that are actively used by API consumers, with thresholds based on user settings. This can be done in two ways:

   a. Taking already aggregated field data directly from an analytics database.

   b. Taking aggregated operation-level data, and splitting it into field-level data.

2. Diff the old and the new API schema. The diffing part is already implemented in the graphql-schema-diff crate.

3. Determine if any of the breaking changes in the diff would apply to fields actively used by clients (determined in (1.)).

This PR implements bullet point 1.b.

Two tradeoffs lead to this approach:

- No splitting of field-level data for now because it turned out more challenging than expected, and we are already aggregating at the operation level, so the fine grained approach may not turn out to be necessary. The approach in this PR is also significantly less work to get a first version of the feature out.
- Not depending on engine-v2 for the splitting, but on GraphQL ASTs instead: the API is less of a moving target and I did not want to burden engine-v2 with another consumer with its own needs. For example, we want to error if a query selects a field not present in the schema in engine v2, but it is a completely expected situation in operation checks, since we work with queries that were sent to past versions of the API.

Next PR in this crate will implement part 3 as a function taking the diff and the usage counts from this PR as arguments. It will output operation check warnings and errors.

closes GB-5692

### DIFF
--- a/engine/crates/operation-checks/src/aggregate_field_usage.rs
+++ b/engine/crates/operation-checks/src/aggregate_field_usage.rs
@@ -1,0 +1,473 @@
+mod async_graphql;
+
+use std::collections::HashMap;
+
+#[derive(Copy, Clone, Debug, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub struct SelectionId(usize);
+
+#[derive(Copy, Clone, Debug, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub struct FieldId(usize);
+
+/// Usage count of a field in a query.
+#[derive(Debug)]
+pub struct FieldUsage<'a> {
+    pub increment: u64,
+    /// field id -> usage count
+    pub count_per_field: &'a mut HashMap<FieldId, u64>,
+}
+
+impl FieldUsage<'_> {
+    pub fn with_increment(self, new_increment: u64) -> Self {
+        Self {
+            increment: new_increment,
+            count_per_field: self.count_per_field,
+        }
+    }
+
+    /// Register a field usage.
+    fn register(&mut self, field_id: FieldId) {
+        if let Some(count) = self.count_per_field.get_mut(&field_id) {
+            *count += self.increment;
+        } else {
+            self.count_per_field.insert(field_id, self.increment);
+        }
+    }
+}
+
+/// (type name, field name) -> field type
+#[derive(Debug)]
+pub struct Schema {
+    pub fields: Vec<SchemaField>,
+    pub query_type_name: String,
+    pub mutation_type_name: String,
+    pub subscription_type_name: String,
+}
+
+#[derive(Debug, PartialEq, Eq, PartialOrd, Ord)]
+pub struct SchemaField {
+    type_name: String,
+    field_name: String,
+    /// The type fo the field without any wrapping type (! and []).
+    base_type: String,
+}
+
+impl Schema {
+    fn find_field(&self, type_name: &str, field_name: &str) -> Option<FieldId> {
+        self.fields
+            .binary_search_by_key(
+                &(type_name, field_name),
+                |SchemaField {
+                     type_name, field_name, ..
+                 }| { (type_name, field_name) },
+            )
+            .map(FieldId)
+            .ok()
+    }
+}
+
+impl std::ops::Index<FieldId> for Schema {
+    type Output = SchemaField;
+
+    fn index(&self, index: FieldId) -> &Self::Output {
+        &self.fields[index.0]
+    }
+}
+
+#[derive(Debug)]
+pub struct Query {
+    /// fragment name -> fragment
+    pub fragments: HashMap<String, Fragment>,
+
+    pub operation_type: OperationType,
+    pub root_selection: SelectionId,
+
+    /// (parent selection, selection)
+    pub selections: Vec<(SelectionId, Selection)>,
+}
+
+#[derive(Debug)]
+pub struct Fragment {
+    pub type_condition: String,
+    pub selection: SelectionId,
+}
+
+#[derive(Debug)]
+pub enum OperationType {
+    Query,
+    Mutation,
+    Subscription,
+}
+
+#[derive(Debug)]
+pub enum Selection {
+    Field {
+        field_name: String,
+        subselection: Option<SelectionId>,
+    },
+    FragmentSpread {
+        fragment_name: String,
+    },
+    InlineFragment {
+        on: Option<String>,
+        selection: SelectionId,
+    },
+}
+
+/// Given a GraphQL query and the corresponding schema, count the number of times each schema field is used.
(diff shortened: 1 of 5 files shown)


engine/crates/operation-checks/src/aggregate_field_usage.rs
```diff
@@ -0,0 +1,421 @@
+mod async_graphql;
+
+use std::collections::HashMap;
+
+#[derive(Debug)]
+pub struct FieldUsage<'a> {
+    pub increment: u64,
+    /// field id -> usage count
+    pub counts: &'a mut HashMap<usize, u64>,
```
yoav-lavi, line 9: Might be clearer to name this something like `count_per_field`

tomhoule: Good catch, will rename. I am doing a round of documenting and making APIs clearer now (sorry I put this back into draft — feel free to continue to read though, no major change expected).

engine/crates/operation-checks/src/aggregate_field_usage.rs
```diff
@@ -0,0 +1,421 @@
...
+        OperationType::Query => &schema.query_type_name,
+        OperationType::Mutation => &schema.mutation_type_name,
+        OperationType::Subscription => &schema.subscription_type_name,
+    };
+    aggregate_field_usage_inner(query.root_selection, ty, query, schema, usage);
+}
+
+fn aggregate_field_usage_inner(
+    selection_id: usize,
+    ty: &str,
```
yoav-lavi, line 81: Might be clearer as `r#type`

engine/crates/operation-checks/src/aggregate_field_usage.rs
```diff
@@ -0,0 +1,421 @@
...
+) {
+    let start = query.selections.partition_point(|(id, _)| *id < selection_id);
+    let selection_set = query.selections[start..]
+        .iter()
+        .take_while(|(id, _)| *id == selection_id);
+
+    for (_, selection) in selection_set {
+        match selection {
+            Selection::Field(field_name, subselection_id) => {
+                dbg!(field_name);
```
yoav-lavi, line 94: Is this meant to be here?

engine/crates/operation-checks/src/aggregate_field_usage.rs
```diff
@@ -0,0 +1,421 @@
...
+                let Some(field_id) = schema.find_field(ty, field_name) else {
+                    dbg!("here");
```
yoav-lavi, line 96: Same

engine/crates/operation-checks/src/aggregate_field_usage.rs
tomhoule: renamed to parent_type in upcoming commit

engine/crates/operation-checks/src/aggregate_field_usage.rs
tomhoule: deleted in upcoming commit

engine/crates/operation-checks/src/aggregate_field_usage.rs
```diff
@@ -0,0 +1,449 @@
...
+        } else {
+            self.count_per_field.insert(field_id, self.increment);
+        }
+    }
+}
+
+/// (type name, field name) -> field type
+#[derive(Debug)]
+pub struct Schema {
+    pub fields: Vec<(String, String, String)>,
```
tomhoule, line 40: This should be a struct

engine/crates/operation-checks/src/aggregate_field_usage.rs
```diff
@@ -0,0 +1,461 @@
...
+    pub mutation_type_name: String,
+    pub subscription_type_name: String,
+}
+
+#[derive(Debug, PartialEq, Eq, PartialOrd, Ord)]
+pub struct SchemaField {
+    type_name: String,
+    field_name: String,
+    /// The type fo the field without any wrapping type (! and []).
+    field_type: String,
```
yoav-lavi, line 51: I think we use `base_type` in the engine for this definition

tomhoule: good catch, let's align the terminology

engine/crates/operation-checks/src/aggregate_field_usage/async_graphql.rs
```diff
@@ -0,0 +1,183 @@
...
+                let item = ingest_selection(counter, &item.node, selections);
+                selections.push((selection_id, item));
+            }
+
+            super::Selection::InlineFragment {
+                on: inline_fragment
+                    .node
+                    .type_condition
+                    .as_ref()
+                    .unwrap() // TODO: what?
```
yoav-lavi, line 94: Is this meant to be here? Perhaps we could use `expect` if it can't fail?

yoav-lavi (approved):
Nice work!

engine/crates/operation-checks/src/aggregate_field_usage/async_graphql.rs
tomhoule: Another good catch :see_no_evil: I meant to come back to that but forgot. We probably shouldn't unwrap, it just wasn't clear to me how the type condition could be missing on an inline fragment, I guess I have to re-read that part of the spec.

tomhoule: The answer is here: https://github.com/graphql/graphql-spec/blame/main/spec/Section%202%20--%20Language.md#L696-L698

engine/crates/operation-checks/src/aggregate_field_usage.rs
tomhoule: done

engine/crates/operation-checks/src/aggregate_field_usage/async_graphql.rs
tomhoule: fixed

engine/crates/operation-checks/src/aggregate_field_usage.rs
```diff
@@ -0,0 +1,449 @@
...
+    type Output = (String, String, String);
+
+    fn index(&self, index: FieldId) -> &Self::Output {
+        &self.fields[index.0]
+    }
+}
+
+#[derive(Debug)]
+pub struct Query {
+    pub fragments: HashMap<String, (String, SelectionId)>,
```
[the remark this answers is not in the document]
tomhoule: done

PR #1195, opened 2024-01-04
chore: Remove Deref on walkers and clarify OperationWalker.with_ext()

Finistere:
Initially I thought walkers would just provide utility methods, but now I realize
they should just be the default way to access anything. So Deref on those was confusing
as you would have access to both `.name` (StringId) and `.name()` (str).

I also clarified the `ext` in the OperationWalker, renaming it to `ctx: ExecutorWalkContextOrUnit`
and adding an explanation.

I also renamed `wrapped` to `item`, which provides a bit more consistent naming overall

# Description

# Type of change

- [ ] 💔 Breaking
- [ ] 🚀 Feature
- [ ] 🐛 Fix
- [ ] 🛠️ Tooling
- [ ] 🔨 Refactoring
- [ ] 🧪 Test
- [ ] 📦 Dependency

### DIFF
--- a/engine/crates/engine-v2/schema/src/walkers/mod.rs
+++ b/engine/crates/engine-v2/schema/src/walkers/mod.rs
@@ -28,20 +28,20 @@ pub use union::UnionWalker;
 
 #[derive(Clone, Copy)]
 pub struct SchemaWalker<'a, I> {
-    // 'wrapped' instead of 'inner' to avoid confusion with TypeWalker.inner()
-    pub(crate) wrapped: I,
+    // 'item' instead of 'inner' to avoid confusion with TypeWalker.inner()
+    pub(crate) item: I,
     pub(crate) schema: &'a Schema,
     pub(crate) names: &'a dyn Names,
 }
 
 impl<'a, I> SchemaWalker<'a, I> {
-    pub fn new(wrapped: I, schema: &'a Schema, names: &'a dyn Names) -> Self {
-        Self { wrapped, schema, names }
+    pub fn new(item: I, schema: &'a Schema, names: &'a dyn Names) -> Self {
+        Self { item, schema, names }
     }
 
-    pub fn walk<Other>(self, wrapped: Other) -> SchemaWalker<'a, Other> {
+    pub fn walk<Other>(self, item: Other) -> SchemaWalker<'a, Other> {
         SchemaWalker {
-            wrapped,
+            item,
             schema: self.schema,
             names: self.names,
         }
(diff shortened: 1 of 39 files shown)


engine/crates/engine-v2/schema/src/walkers/mod.rs
```diff
@@ -52,22 +52,15 @@ impl<'a, Id: Copy> SchemaWalker<'a, Id>
 where
     Schema: std::ops::Index<Id>,
 {
-    pub fn get(&self) -> &'a <Schema as std::ops::Index<Id>>::Output {
-        &self.schema[self.wrapped]
+    // Clippy complains because it's ambiguous with AsRef. But AsRef doesn't allow us to add the 'a
+    // lifetime. I could rename to `to_ref()` or `ref()`, but doesn't feel better than `as_ref()`.
+    #[allow(clippy::should_implement_trait)]
```
tomhoule, line 57: There's examples of `as_ref()` methods without trait impl even in the std library (https://doc.rust-lang.org/std/option/enum.Option.html#method.as_ref) so I agree this is the right name here, clippy is being unnecessarily pedantic.

engine/crates/engine-v2/src/request/walkers/mod.rs
```diff
@@ -34,15 +34,20 @@ use super::{
 };
 
 #[derive(Clone, Copy)]
-pub struct OperationWalker<'a, Walkable = (), SchemaId = (), Extension = ()> {
+pub struct OperationWalker<'a, Item = (), SchemaItem = (), ExecutorContextOrUnit = ()> {
     /// The operation MUST NOT be accessible directly in any form. Elans may have additional
```
tomhoule, line 38: s/Elans/Plans/ (I think?)

engine/crates/engine-v2/src/request/walkers/mod.rs
```diff
@@ -34,15 +34,20 @@ use super::{
...
     pub(super) operation: &'a Operation,
-    pub(super) schema_walker: SchemaWalker<'a, SchemaId>,
-    pub(super) ext: Extension,
-    pub(super) wrapped: Walkable,
+    pub(super) schema_walker: SchemaWalker<'a, SchemaItem>,
+    /// Walkers are used in two different situations:
+    /// - during planning
+    /// - by executors, with plan metadata like which fields are attributed to them.
+    /// In practice this type is a `() | ExecutorContext` as some methods only make sense in either case
+    /// and others are common.
```
tomhoule, line 48: :+1:

tomhoule (approved):
Nice!

engine/crates/engine-v2/src/request/walkers/mod.rs
Finistere: ah indeed, thanks :)

PR #1198, opened 2024-01-05
feat: update ExecutorCoordinator with subscription flow

obmarg:
I'm working on adding subscriptions to our federation offering.  The flow of subscriptions is fairly similar to a normal federation execution but with a few of important differences:

1. There can only be a single subscription field at the root.
2. That subscription field will translate into a subscription call to a subgraph.
3. We build up a fresh GraphQL response from each response we get back from that initial subscription.

This PR updates the ExecutorCoordinator to support the subscription flow as well as the normal flow.  I want to return a stream for subscriptions, ideally without spawning an async task, which needed a lot of refactoring.  Mostly removing references where I could, as they 

I needed to do a surprising amount of refactoring to get this working.  Mostly turning references into owned things where it made sense, and removing the mutable parts of ExecutorCoordinator into local variables instead.

Most of the changes in this PR are still dead code - my next PR will hook them up.

Part of GB-5708

### DIFF
--- a/engine/crates/engine-v2/src/execution/coordinator.rs
+++ b/engine/crates/engine-v2/src/execution/coordinator.rs
@@ -1,18 +1,21 @@
-use std::collections::{BTreeMap, VecDeque};
+use std::{
+    collections::{BTreeMap, VecDeque},
+    sync::Arc,
+};
 
 use async_runtime::make_send_on_wasm;
 use engine::RequestHeaders;
 use engine_parser::types::OperationType;
-use futures_util::{future::BoxFuture, stream::FuturesUnordered, FutureExt, StreamExt};
+use futures_util::{future::BoxFuture, stream::FuturesUnordered, Future, FutureExt, Stream, StreamExt};
 
 use crate::{
     execution::{ExecutionContext, Variables},
-    plan::{PlanBoundary, Planner, PlanningError},
+    plan::{Plan, PlanBoundary, Planner, PlanningError},
     request::{Operation, QueryPath},
     response::{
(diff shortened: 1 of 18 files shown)


engine/crates/engine-v2/src/execution/coordinator.rs
```diff
@@ -1,94 +1,139 @@
...
             variables,
             request_headers,
         }
     }
 
-    pub async fn execute(&mut self) {
-        let mut futures = FuturesUnordered::<BoxFuture<'_, ExecutorFutResult>>::new();
+    pub async fn execute(self) -> Response {
+        let planner = Planner::new(&self.engine.schema, &self.operation);
+        let mut planner = planner;
```
Finistere, line 43: nitpick: redundant

engine/crates/engine-v2/src/execution/coordinator.rs
```diff
@@ -97,109 +142,152 @@ impl<'ctx> ExecutorCoordinator<'ctx> {
...
-    // ugly... should be sent back through a stream to support defer.
-    pub fn into_response(self) -> Response {
-        self.response.build(
-            self.engine.schema.clone(),
-            self.operation.response_keys.clone(),
-            ExecutionMetadata::build(self.operation),
-        )
+pub struct ExecutorFutureSet<'a>(FuturesUnordered<BoxFuture<'a, ExecutorFutResult>>);
+
+impl<'a> ExecutorFutureSet<'a> {
```
Finistere, line 267: nice!

Finistere (approved):

engine/crates/engine-v2/src/execution/coordinator.rs
obmarg: 🤦

PR #1199, opened 2024-01-05
feat: expose `execute_stream` from engine-v2

obmarg:
A follow up to #1198 - exposing the changes I made there to consumers of `engine-v2`.  I've not hooked this up in the CLI yet because I want to get subscriptions working first.   This does allow the code to be called in tests though.

I've added a passing test of queries & mutations via this function just to confirm.

Part of GB-5708

### DIFF
--- a/engine/crates/engine-v2/src/engine.rs
+++ b/engine/crates/engine-v2/src/engine.rs
@@ -1,10 +1,13 @@
 use std::sync::Arc;
 
 use engine::RequestHeaders;
+use engine_parser::types::OperationType;
+use futures::channel::mpsc;
+use futures_util::{future::BoxFuture, Stream};
 use schema::Schema;
 
 use crate::{
-    execution::{ExecutorCoordinator, Variables},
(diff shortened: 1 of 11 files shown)


engine/crates/engine-v2/src/engine.rs
```diff
@@ -29,22 +35,97 @@ impl Engine {
...
+        }
+
+        coordinator.execute().await
+    }
+
+    pub fn execute_stream(
+        &self,
+        request: engine::Request,
+        headers: RequestHeaders,
+    ) -> impl Stream<Item = Response> + '_ {
```
Finistere, line 57: Considering how the cli federation dev is structured you'll most likely need to make it `'static` with an `Arc<Engine>` I think

Finistere (approved):
Nice!

engine/crates/engine-v2/src/engine.rs
obmarg: Yeah, you may be right - I'll deal with it when I get to the inevitable compiler errors.

PR #1211, opened 2024-01-08
fix: default values in graphql connector

obmarg:
The GraphQL connector introspects the remote server and uses that to build up a Registry.  This commit fixes it's handling of default values.

Default values in introspection are returned as a string where the contents are the default value, encoded as it would be in a GraphQL document.  The GraphQL connector was taking this and converting it into a `ConstValue::String` of that string, which is not quite right.

This updates the conversion to use `engine-parser` to interpret the string first, so we end up with the correct `ConstValue` in the registry.

Looks like this fixes quite a few potential issues from the various snapshots this has updated.

Fixes GB-5752

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4163,6 +4163,7 @@ dependencies = [
  "cynic",
  "cynic-introspection",
  "engine",
+ "engine-parser",
  "http 0.2.11",
  "insta",
  "reqwest",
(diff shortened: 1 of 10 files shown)


fbjork (approved):

tomhoule (approved):
nice!

PR #1212, opened 2024-01-08
nix: derivation for the cli

tomhoule:
This lets anybody with nix installed build the CLI from source with `nix build github:grafbase/grafbase#cli`, or run it with `nix run github:grafbase/grafbase#cli`, or build it locally with `nix build .#cli`, etc. This is also useful documentation on all the inputs required to build the Grafbase CLI.

This commit also introduces flake-parts to the repo. It is just a convenient way to organize the derivations that reuses the same module system any nixos or home-manager user will be familiar with.

One caveat though: the version and URL of the assets.tar.gz file are hardcoded. Ideally, we would also build these from source, but that can be done in a separate PR.

### DIFF
--- a/cli/crates/server/build.rs
+++ b/cli/crates/server/build.rs
@@ -39,8 +39,9 @@ Please see the instructions in cli/crates/server/README.md.
 fn decompress_assets() -> io::Result<tempfile::TempDir> {
     use flate2::bufread::GzDecoder;
     let dir = tempfile::tempdir()?;
-    eprintln!("Decompressing the assets at `{ASSETS_GZIP_PATH}` from `server` crate build script.");
-    let file_reader = io::BufReader::new(fs::File::open(ASSETS_GZIP_PATH)?);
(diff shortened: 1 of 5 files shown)


cli/crates/server/build.rs
```diff
@@ -39,8 +39,9 @@ Please see the instructions in cli/crates/server/README.md.
 fn decompress_assets() -> io::Result<tempfile::TempDir> {
     use flate2::bufread::GzDecoder;
     let dir = tempfile::tempdir()?;
-    eprintln!("Decompressing the assets at `{ASSETS_GZIP_PATH}` from `server` crate build script.");
-    let file_reader = io::BufReader::new(fs::File::open(ASSETS_GZIP_PATH)?);
+    let assets_gzip_path = env::var("ASSETS_GZIP_PATH").unwrap_or_else(|_| ASSETS_GZIP_PATH.to_owned());
```
yoav-lavi, line 42: Might be good to prefix this with grafbase

yoav-lavi (approved):
Nice!

cli/crates/server/build.rs
tomhoule: Changed and pushed

PR #1215, opened 2024-01-08
chore: GB-5753: Remove all the remnants of the database and search features from the engine

jakubadamw:
# Description

As above.

# Type of change

- [X] 💔 Breaking
- [ ] 🚀 Feature
- [ ] 🐛 Fix
- [ ] 🛠️ Tooling
- [ ] 🔨 Refactoring
- [ ] 🧪 Test
- [ ] 📦 Dependency


cli/crates/cli/tests/utils/environment.rs
```diff
@@ -58,52 +56,20 @@ fn get_free_port() -> u16 {
 impl Environment {
     #[allow(clippy::needless_return, clippy::unused_async)]
```
obmarg, line 57: Can we get rid of this allow now? (and the others in this file)

cli/crates/gateway/src/executor.rs
```diff
@@ -126,7 +63,7 @@ impl Executor {
         let search_engine = LocalSearchEngine::new(self.bridge.clone());
 
         Ok(engine::Schema::build(engine::Registry::clone(&self.registry))
-            .data(dynamodb_batchers_data)
+            .data(engine::TraceId(ctx.ray_id().to_string()))
```
obmarg, line 66: This doesn't match with the `.data::<Box<TraceId>>()` you're doing elsewhere.

engine/crates/engine/src/context/ext.rs
```diff
@@ -64,9 +63,8 @@ pub trait ContextExt<'a>: Context<'a> {
     }
 
     fn trace_id(&self) -> String {
-        self.data::<Arc<DynamoDBBatchersData>>()
-            .map(|x| x.ctx.trace_id.clone())
-            .ok()
+        self.data::<Box<TraceId>>()
```
obmarg, line 66: Should this just be `TraceId` instead of `Box<TraceId>`?  Looks like you're not putting it in a box when you add it to the context elsewhere...

obmarg (approved):
🚮

engine/crates/engine/src/context/ext.rs
jakubadamw: You're right, fixed!

cli/crates/cli/tests/utils/environment.rs
jakubadamw: Definitely, good catch!

PR #1216, opened 2024-01-08
feat: implements `Cacheable` for `v2::Response`

hfgbarrigas:
# Description

This PR implements `Cacheable` for `v2::Response`. With this we can re-use `runtime::cache` for v2.

# Type of change

- [ ] 💔 Breaking
- [x] 🚀 Feature
- [ ] 🐛 Fix
- [ ] 🛠️ Tooling
- [ ] 🔨 Refactoring
- [ ] 🧪 Test
- [ ] 📦 Dependency

### DIFF
--- a/engine/crates/engine-v2/src/response/cacheable.rs
+++ b/engine/crates/engine-v2/src/response/cacheable.rs
@@ -53,7 +53,7 @@ impl Cacheable for CacheableResponse {
                 .metadata
                 .operation_type
                 .map(|operation_type| {
-                    operation_type != OperationType::Mutation && operation_type != OperationType::Subscription
+                    operation_type == OperationType::Query
                 })
                 .unwrap_or_default()
     }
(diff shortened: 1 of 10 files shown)


engine/crates/engine-v2/src/response/cacheable.rs
```diff
@@ -0,0 +1,60 @@
...
+        false // to be added when mutation invalidation is supported in v2
+    }
+
+    fn should_cache(&self) -> bool {
+        !self.has_errors
+            && self
+                .metadata
+                .operation_type
+                .map(|operation_type| {
+                    operation_type != OperationType::Mutation && operation_type != OperationType::Subscription
```
obmarg, line 56: Nitpicking a bit, but wouldn't it be easier to do:

```suggestion
                    operation_type == OperationType::Query
```

obmarg (approved):

engine/crates/engine-v2/src/response/cacheable.rs
hfgbarrigas: ofc it would :)

Finistere (approved):

PR #1217, opened 2024-01-09
feat: allow SDK users to add fields to connector types

obmarg:
I added the ability to extend connector generated types with federation directives to the SDK in 81fd711.  One thing that still wasn't added then was the ability to add fields onto connector generated types - oops.

This does that.


tomhoule (approved):
The implementation looks good, but I'm not sure how this works downstream from there. You would have to attach a custom resolver, right? Otherwise you can define any field that isn't backed by the connector.

obmarg:
If it's a connector backed tye then yes you have to do _something_ that makes it resolvable.  At the moment that means either adding a resolver or (as the user I'm adding this for wanted to do) adding a join on that field.

PR #1220, opened 2024-01-10
feat: fragment cycle detection

Finistere:
# Description

# Type of change

- [ ] 💔 Breaking
- [ ] 🚀 Feature
- [ ] 🐛 Fix
- [ ] 🛠️ Tooling
- [ ] 🔨 Refactoring
- [ ] 🧪 Test
- [ ] 📦 Dependency

### DIFF
--- a/engine/crates/engine-v2/src/request/bind.rs
+++ b/engine/crates/engine-v2/src/request/bind.rs
@@ -5,6 +5,7 @@ use std::{
 
 pub use engine_parser::types::OperationType;
 use engine_parser::Positioned;
+use itertools::Itertools;
 use schema::{Definition, FieldWalker, Schema};
 
 use crate::response::GraphqlError;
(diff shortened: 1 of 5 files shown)


obmarg (approved):

engine/crates/engine-v2/src/request/bind.rs
```diff
@@ -173,6 +178,7 @@ pub struct Binder<'a> {
     // We also need to use the global request position so that merged selection sets are still
     // in the right order.
     next_response_position: usize,
+    current_fragments_stack: VecDeque<String>,
```
tomhoule, line 181: If I'm reading the file right, it looks like we're only pushing and popping at the back, so a `Vec<String>` should work too?

tomhoule (approved):

engine/crates/engine-v2/src/request/bind.rs
Finistere: good point, you're right

PR #1228, opened 2024-01-10
chore: SDK release 0.14.1

Finistere:
# Description

# Type of change

- [ ] 💔 Breaking
- [ ] 🚀 Feature
- [ ] 🐛 Fix
- [ ] 🛠️ Tooling
- [ ] 🔨 Refactoring
- [ ] 🧪 Test
- [ ] 📦 Dependency


packages/grafbase-sdk/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+## [0.14.1] - Tue Jan 10 2024
```
obmarg, line 3:
```suggestion
## [0.14.1] - Mon Jan 15 2024
```

packages/grafbase-sdk/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@grafbase/sdk",
-  "version": "0.14.0",
+  "version": "0.14.1",
```
obmarg, line 3: new features means a minor bump usually, no?

obmarg (approved):

packages/grafbase-sdk/package.json
Finistere: Good question. Re-reading the semver guidance, it only states:

> Major version zero (0.y.z) is for initial development. Anything MAY change at any time. The public API SHOULD NOT be considered stable.

And everything else only applies to `x > 0`. Until now thought `y` was breaking changes and `z` everything else for `0.y.z`

Finistere: ok will use `0.15` as that's what we did before

obmarg: I usually read that "Major version zero..." point as permission to make breaking changes in version 0 but otherwise mostly stick to the rules in the rest of the spec.  Though maybe that's just me

Finistere: yeah makes more sense, don't know where I got my interpretation from 🤷

PR #1230, opened 2024-01-11
operation-checks: make sure we handle wrapper types correctly

tomhoule:
We were only looking at the outermost wrapper type previously and not really handling lists, not to speak of nested lists. This commit takes the wrapper types (! and []) into account on field type changes in operation checks.

In addition, expose CheckDiagnostic and Severity in public API. It was
an omission.

closes GB-5749

### DIFF
--- a/engine/crates/operation-checks/src/check/rules.rs
+++ b/engine/crates/operation-checks/src/check/rules.rs
@@ -28,7 +28,7 @@ pub(super) fn remove_field(
         let field_is_required = check_params
             .source
             .find_field(type_name, field_name)
-            .map(|field_id| check_params.source[field_id].type_is_required)
+            .map(|field_id| check_params.source[field_id].is_required())
             .unwrap_or_default();
 
         let diagnostic = if field_is_required {
@@ -112,7 +112,7 @@ pub(super) fn add_field(
         return None;
     };
 
-    if !check_params.target[field_id].type_is_required {
+    if !check_params.target[field_id].is_required() {
         return None;
     }
 
@@ -135,8 +135,6 @@ pub(super) fn add_field(
 
 /// Changing the type of an argument or removing an argument is safe iff the argument is not in
 /// use or if it was required and became optional (keeping the same inner type).
-///
-/// FIXME: we do not handle lists great here and in other places => GB-5749
 pub(super) fn change_field_argument_type(
     CheckArgs {
         change, check_params, ..
@@ -151,6 +149,7 @@ pub(super) fn change_field_argument_type(
 
     let arg_in_src = check_params.source.find_argument((type_name, field_name, arg_name));
     let arg_in_target = check_params.source.find_argument((type_name, field_name, arg_name));
+
     'refine: {
         if let Some((src_id, target_id)) = arg_in_src.zip(arg_in_target) {
             let src_arg = &check_params.source[src_id];
@@ -160,25 +159,25 @@ pub(super) fn change_field_argument_type(
                 break 'refine; // type change overrides arity change
             }
 
-            if !src_arg.is_required && target_arg.is_required {
-                return Some(CheckDiagnostic {
-                    message: format!(
-                        "The argument `{}` became required, but clients are not providing it.",
-                        change.path
-                    ),
-                    severity: Severity::Error,
-                });
-            } else if src_arg.is_required && !target_arg.is_required {
-                return None;
+            match src_arg.wrappers.compare(&target_arg.wrappers) {
+                crate::schema::WrapperTypesComparison::RemovedNonNull => return None,
+                crate::schema::WrapperTypesComparison::AddedNonNull => {
+                    return Some(CheckDiagnostic {
+                        message: format!(
+                            "The argument `{}` became required, but clients are not providing it.",
+                            change.path
+                        ),
+                        severity: Severity::Error,
+                    });
+                }
+                crate::schema::WrapperTypesComparison::NoChange
+                | crate::schema::WrapperTypesComparison::NotCompatible => break 'refine,
             }
         }
     }
 
     Some(CheckDiagnostic {
-        message: format!(
-            "The argument `{}` changed type but it is still used by clients.",
-            change.path
-        ),
+        message: format!("The argument `{}` changed type but it is used by clients.", change.path),
         severity: Severity::Error,
     })
 }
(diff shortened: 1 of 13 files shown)


engine/crates/operation-checks/src/schema/wrapper_types.rs
```diff
@@ -0,0 +1,104 @@
...
+                        WrapperType::REQUIRED_LIST => Some(WrapperType::RequiredList),
+                        _ => unreachable!(),
+                    }
+                }),
+                &[],
+            ),
+            WrapperTypes::Large(large) => ([None; 4], large.as_ref()),
+        };
+
+        first_four.into_iter().flatten().chain(rest.iter().copied())
```
tomhoule, line 49: The dance with the array is here to avoid having to box the iterator. Open to better ideas.

engine/crates/operation-checks/src/schema/wrapper_types.rs
```diff
@@ -0,0 +1,104 @@
...
+    pub(crate) fn compare(&self, target: &WrapperTypes) -> WrapperTypesComparison {
+        use WrapperType::*;
+        use WrapperTypesComparison::*;
+
+        let mut src_wrappers = self.iter_wrappers();
+        let mut target_wrappers = target.iter_wrappers();
+        let mut end_state = NoChange;
+
+        loop {
+            match (src_wrappers.next(), target_wrappers.next()) {
```
tomhoule, line 68: This is surprisingly complicated...

engine/crates/operation-checks/src/check/rules.rs
```diff
@@ -160,25 +159,25 @@ pub(super) fn change_field_argument_type(
...
-                        "The argument `{}` became required, but clients are not providing it.",
-                        change.path
-                    ),
-                    severity: Severity::Error,
-                });
-            } else if src_arg.is_required && !target_arg.is_required {
-                return None;
+            match src_arg.wrappers.compare(&target_arg.wrappers) {
+                crate::schema::WrapperTypesComparison::BecameLessRequired => return None,
+                crate::schema::WrapperTypesComparison::BecameMoreRequired => {
```
yoav-lavi, line 164: Why less and more required specifically? Isn't it a binary status?

engine/crates/operation-checks/src/schema/async_graphql.rs
```diff
@@ -98,9 +103,59 @@ impl From<ServiceDocument> for schema::Schema {
...
+fn extract_type(top_level_ty: &async_graphql_parser::types::Type) -> (String, WrapperTypes) {
+    let mut ty = top_level_ty;
+    let mut small = 0u8;
+
+    // Fast path that should be taken nearly all the time.
+    for i in 0..4 {
+        match &ty.base {
+            async_graphql_parser::types::BaseType::Named(name) => {
+                if !ty.nullable {
+                    small |= (WrapperType::Required as u8) << (i * 2);
```
yoav-lavi, line 115: I'd add a comment on the reasoning here and below

engine/crates/operation-checks/src/schema/wrapper_types.rs
```diff
@@ -0,0 +1,104 @@
...
+}
+
+impl WrapperTypes {
+    /// Iterate wrapper types from outermost to innermost.
+    pub(crate) fn iter_wrappers(&self) -> impl Iterator<Item = WrapperType> + '_ {
+        let (first_four, rest): ([Option<WrapperType>; 4], &[WrapperType]) = match self {
+            WrapperTypes::Small(small) => (
+                [0, 1, 2, 3].map(move |i| {
+                    let shift = i * 2;
+                    let bits = (small >> shift) & 0b11;
```
yoav-lavi, line 34: I'd add a comment here on why this is being done

yoav-lavi (approved):
Looks good, see comments

engine/crates/operation-checks/src/check/rules.rs
tomhoule: The idea of the "more"/"less" required is that `[[String]!]` is optional, but "more required" than `[[String]]`. It's not well defined, let's try to come up with something better. Maybe `AddedNonNull`/`RemovedNonNull`?

tomhoule:
Realizing now that the WrapperTypes would have been better modeled by separating the list from the non-list bits like I've done in other places, but tests are passing, so let's not re-engineer that now.

engine/crates/operation-checks/src/schema/async_graphql.rs
tomhoule: Just pushed something.

engine/crates/operation-checks/src/schema/wrapper_types.rs
```diff
@@ -0,0 +1,107 @@
...
+#[derive(Debug, PartialEq, Eq)]
+pub enum WrapperTypes {
+    /// A compact representation for up to four levels of wrapper types. Each pair of bits
+    /// corresponds to an Option<WrapperType> (0b00 = None, 0b01 = Some(Required), 0b10 =>
+    /// Some(List), 0b11 => Some(RequiredList)). The least significant pair of bits corresponds to
+    /// the outermost wrappers.
+    Small(u8),
+    /// This variant is for extreme cases of list fields with more than three levels of nested
+    /// lists. From outermost to innermost.
+    Large(Box<[WrapperType]>),
```
obmarg, line 24: Just a thought, but would it be worth making this a `struct WrapperTypes(u32)` and always doing bitwise encoding?

Gives you space for 16 lists and still uses less memory than this will

tomhoule: Yes even with this type, there's more room than the u8 if we want. Even though lists nested 16 level deep isn't a realistic use case, I had some worries about not supporting it. I'll open a PR when there's some downtime with the change mentioned in my last comment and this, that should simplify things a lot.

tomhoule: Actually, since I'm waiting on other PRs now, let's try right away.

engine/crates/operation-checks/src/schema/wrapper_types.rs
```diff
@@ -1,62 +1,60 @@
...
-    const REQUIRED_LIST: u8 = WrapperType::RequiredList as u8;
-}
+/// The [wrapping types](http://spec.graphql.org/October2021/#sec-Wrapping-Types) for a given
+/// instance of a type.
+///
+/// Implementation: this is a 64 bit integer. Layout is the following, from highest to lowest bits:
+///
+/// - 6 bits containing an integer representing the number of list wrapping types.
+/// - 1 bit representing whether the innermost type is required.
+/// - 57 bits representing the list wrappers. The lowest bit is the outermost list wrapping type.
```
tomhoule, line 15: This should clarify that for each bit, 0 means List, and 1 means RequiredList.

engine/crates/operation-checks/src/schema/wrapper_types.rs
tomhoule: 0e5649e1f37dccb4027011a2055543bec6a4779f reimplements the same API but with a single integer for the wrappers. I'd like to improve the API though, doing this now.

tomhoule:
@yoav-lavi @obmarg I pushed a new commit, passing the same tests, with a single-integer implementation of the wrapping types, with a simpler API too.

engine/crates/operation-checks/src/schema.rs
```diff
@@ -71,23 +71,25 @@ pub struct SchemaField {
...
+                ListType::NonNullList => format!("[{result}]!"),
             }
         }
 
-        result
+        if self.wrappers.inner_is_required() {
+            format!("{result}!")
+        } else {
+            result
+        }
```
obmarg, line 92: Should this not be done before the list wrappers if it's for the inner type?

tomhoule: yes :see_no_evil: Weird that tests didn't catch it, I'll write that does.

tomhoule: done and pushed

PR #1231, opened 2024-01-11
chore: rework federated dev a bit

obmarg:
The existing setup in federated dev has the gateway living inside `Router` with all requests passed in as messages using `RequestSender`.  

This works fine at the moment, but I'm currently adding subscriptions which don't fit into the same request<->response model as well.  Each request requires multiple responses _and_ (at least with websockets) auth is no necessarily longer tied to a graphql operation - we can auth connections and every operation after that runs without auth.  The websocket system also needs to observe configuration updates so it can end subscriptions when they occur.

Updating the router to support all this was getting messy so I've refactored stuff instead.  The `Router` actor has been replaced with `GatewayNanny` - the only job of which is to combine the config & graph into a `Gateway` and publish that on a tokio watch channel.

The rest of the system can then use that watch channel to access the `Gateway` & react to updates.

Part of GB-5715


cli/crates/federated-dev/src/dev/bus.rs
```diff
@@ -35,14 +37,13 @@ pub(crate) type ComposeSender = mpsc::Sender<ComposeMessage>;
...
-pub(crate) type RequestReceiver = mpsc::Receiver<(engine::Request, RequestHeaders, ResponseSender)>;
+/// Send half of the gateway watch channel
+pub(crate) type GatewaySender = watch::Sender<Option<Arc<Gateway>>>;
 
-/// Send half of channel for the router actor to send responses
-pub(crate) type ResponseSender = oneshot::Sender<RouterResult<engine_v2::CacheableResponse>>;
+/// Receive half of the gateway watch channel.
+///
+/// Anything part of the system that needs access to the gateway can use this
+pub(crate) type GatewayReceiver = watch::Receiver<Option<Arc<Gateway>>>;
```
Finistere, line 46: maybe `GatewayWatcher`? Receiver sounds a bit weird to me as you do `receiver.borrow().clone()` to "receive" something

cli/crates/federated-dev/src/dev.rs
```diff
@@ -45,7 +44,7 @@ pub(super) async fn run(port: u16, expose: bool, config: ConfigReceiver) -> Resu
     let (graph_sender, graph_receiver) = watch::channel(None);
     let (refresh_sender, refresh_receiver) = mpsc::channel(16);
     let (compose_sender, compose_receiver) = mpsc::channel(16);
-    let (request_sender, request_receiver) = mpsc::channel(16);
+    let (gateway_sender, gateway) = watch::channel(None);
```
Finistere, line 47: nice

Finistere (approved):

tomhoule (approved):
:100:

cli/crates/federated-dev/src/dev/bus.rs
obmarg: Yeah, wasn't happy with the name either - good idea

PR #1235, opened 2024-01-12
feat: implement graphql-over-websockets in cli

obmarg:
Implements [the GraphqlOverWebSocket specification](https://github.com/graphql/graphql-over-http/blob/main/rfcs/GraphQLOverWebSocket.md).

Fixes GB-5715


cli/crates/federated-dev/src/dev/websockets/messages.rs
```diff
@@ -0,0 +1,45 @@
...
+#[serde(tag = "type", rename_all = "snake_case")]
+pub enum Event {
+    ConnectionInit { payload: Option<serde_json::Value> },
+    Subscribe { id: String, payload: engine_v2::Request },
+    Complete { id: String },
+    Ping { payload: Option<serde_json::Value> },
+    Pong { payload: Option<serde_json::Value> },
+}
+
+// TODO: Clean up whichever ones of these we don't need.
```
obmarg, line 15:
```suggestion
```

cli/crates/federated-dev/src/dev/websockets/messages.rs
```diff
@@ -0,0 +1,44 @@
+//! Message definitions for the [GraphQLOverWebsocket protocol][1]
```
Finistere, line 1: Could be in the gateway-v2 crate later as this would be shared between api & cli. Supposing we support it in api^^

cli/crates/federated-dev/src/dev/websockets.rs
```diff
@@ -0,0 +1,234 @@
...
+    }
+
+    pub async fn handler(mut self) {
+        while let Some(connection) = self.sockets.recv().await {
+            let gateway = self.gateway.clone();
+
+            tokio::spawn(async move {
+                let Some((websocket, _init_payload)) = accept_websocket(connection).await else {
+                    log::warn!("Failed to accept websocket connection");
+                    return;
```
Finistere, line 38: Shouldn't an error be returned?

Finistere (approved):
Nice!

cli/crates/federated-dev/src/dev/websockets.rs
obmarg: Returned where?  The only way to get to this point at the moment is if the connection is closed.

Finistere: If you get an `Event::Subscribe` event for example, the `accept_websocket` function will return `None`.  It's an edge case though obviously

obmarg: If it gets an `Event::Subscribe` it will ignore it and continue looping.  It only returns `None` if the websocket stream ends - which means there's nowhere to return errors.

I do need to add timeouts on this accept call but I'm going to do that in [another issue](https://linear.app/grafbase/issue/GB-5792/handle-ws-accept-failures-in-the-cli)

Finistere: ah my bad, didn't notice the `while`.

PR #1240, opened 2024-01-15
feat:  pg-pooling

hfgbarrigas:
# Description

Adds connection pooling for pg connections under feature flag `pooling`.

- Renames `TcpTransport` to `DirectTcpTransport`. This type implements `Transport` and is meant as a direct connection to pg.
- Adds a new `PooledTcpTransport` that when instantiated configures a connection pool. It contains a `connection` method that gets a connection from the pool. The returned type implements `Transport`
- Adds a new trait extension to `Transport`, called `TransportTransactionExt`. Allows implementors to create/begin a tx and return it. Its a separate trait and not in `Transport` directly because not all implementors of `Transport` can create tx (e.g: hyperdrive)

# Type of change

- [ ] 💔 Breaking
- [x] 🚀 Feature
- [ ] 🐛 Fix
- [ ] 🛠️ Tooling
- [ ] 🔨 Refactoring
- [ ] 🧪 Test
- [ ] 📦 Dependency


pimeys (approved):
Looks like all the pooling code I've ever written in Rust. Approved.

PR #1244, opened 2024-01-16
chore: add federated-graph-schema arg for federated start

Finistere:
Adding an option to read an Apollo `supergraphql.graphql` (`FederatedGraph`) and use that as the graph statically. I'm not support any subgraph changes when doing so.

### DIFF
--- a/cli/crates/federated-dev/src/dev/bus/admin.rs
+++ b/cli/crates/federated-dev/src/dev/bus/admin.rs
@@ -5,7 +5,7 @@ use url::Url;
 
 pub(crate) enum AdminBus {
     DynamicGraph { compose_sender: ComposeSender },
-    StatisGraph,
+    StaticGraph,
 }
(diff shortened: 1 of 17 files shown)


cli/crates/federated-dev/src/dev/bus/admin.rs
```diff
@@ -3,13 +3,18 @@ use crate::{dev::admin::Header, error::Error};
 use async_graphql_parser::types::ServiceDocument;
 use url::Url;
 
-pub(crate) struct AdminBus {
-    compose_sender: ComposeSender,
+pub(crate) enum AdminBus {
+    DynamicGraph { compose_sender: ComposeSender },
+    StatisGraph,
```
obmarg, line 8:
```suggestion
    StaticGraph,
```

obmarg (approved):

PR #1245, opened 2024-01-16
chore: reworked streaming transport code

obmarg:
The existing streaming transport code made use of `wait_until` to pump messages from one place to the other.  This was causing me lifetime problems when integrating into engine-v2, so I've reworked everything to happen inside the returned `Stream`.

I'm hoping this will work on production as well - we had to be careful about where the work was done previously so it was definitely inside the request context.  But I think doing the work inside the returned stream _should_ count as inside the context.  I'll test once this is merged though.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2263,6 +2263,7 @@ dependencies = [
  "mime",
  "multipart-stream",
  "runtime",
+ "serde",
  "serde_json",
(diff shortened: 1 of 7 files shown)


tomhoule (approved):
I'm not the most qualified to review, but lgtm, reasoning and implementation both make sense.

PR #1248, opened 2024-01-17
feat: add sse & multipart transports to federated dev

obmarg:
Adds [Incremental Delivery][1] & [GraphQL Over SSE][2] support to federated dev.

We already had code for these in the CLI, so I'm mostly just hooking things up.

Fixes GB-5797

[1]: https://github.com/graphql/graphql-over-http/blob/main/rfcs/IncrementalDelivery.md
[2]: https://github.com/graphql/graphql-over-http/blob/main/rfcs/GraphQLOverSSE.md

### DIFF
--- a/engine/crates/async-runtime/src/stream.rs
+++ b/engine/crates/async-runtime/src/stream.rs
@@ -1,0 +1,49 @@
+use futures_util::{
+    future::{self, BoxFuture},
+    stream::{self, BoxStream},
+    Future, FutureExt, Stream, StreamExt,
+};
+
+/// Combines a stream and a "producer" future into a single stream
+///
+/// This can be used when you have the receivng side of a channel and a future that sends
+/// on that channel - combining the two into a single stream that'll run till the channel
+/// is exhausted.  If you drop the stream you also cancel the underlying process.
+pub fn producer_stream<'a, S, P, Item>(stream: S, producer: P) -> impl Stream<Item = Item> + 'a
+where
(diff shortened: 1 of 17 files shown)


engine/crates/async-runtime/src/stream.rs
```diff
@@ -0,0 +1,49 @@
...
+    stream::{self, BoxStream},
+    Future, FutureExt, Stream, StreamExt,
+};
+
+/// Combines a stream and a "producer" future into a single stream
+///
+/// This can be used when you have the receivng side of a channel and a future that sends
+/// on that channel - combining the two into a single stream that'll run till the channel
+/// is exhausted.  If you drop the stream you also cancel the underlying process.
+pub fn producer_stream<'a, S, P, Item>(stream: S, producer: P) -> impl Stream<Item = Item> + 'a
```
obmarg, line 12: Any name suggestions for this welcome, it's an awkward concept to describe succinctly

Finistere (approved):

engine/crates/async-runtime/src/stream.rs
Finistere: I think I would create an extension like:
```
impl Stream {
  fn with_producer_future(&self, producer: P)
}
```

Finistere: Or maybe `join_with_future`

Finistere: "join" feels a bit more idiomatic imho

PR #1249, opened 2024-01-17
cli: handle warnings in `grafbase check`

tomhoule:
We now have warnings in addition to errors, we have to handle them.

closes GB-5812

### DIFF
--- a/cli/crates/cli/src/output/report.rs
+++ b/cli/crates/cli/src/output/report.rs
@@ -375,6 +375,12 @@ pub fn create_success(name: &str, urls: &[String]) {
     }
 }
 
+pub(crate) fn check_warnings<'a>(operation_check_warnings: impl Iterator<Item = &'a str>) {
+    for warning in operation_check_warnings {
+        watercolor::output!("⚠️  Warning: {warning}", @BrightYellow);
+    }
+}
+
 pub(crate) fn check_success() {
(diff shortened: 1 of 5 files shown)


cli/crates/cli/src/output/report.rs
```diff
@@ -402,8 +405,15 @@ pub(crate) fn check_errors<'a>(
...
-        for error in operation_errors {
-            watercolor::output!("- {error}", @BrightRed);
+        for (message, severity) in operation_errors {
+            match severity {
+                SchemaCheckErrorSeverity::Error => {
+                    watercolor::output!("- {message}", @BrightRed);
+                }
+                SchemaCheckErrorSeverity::Warning => {
+                    watercolor::output!("- {message}", @BrightYellow);
+                }
```
obmarg, line 415: Should we maybe differentiate these on more than just the colour?  I assume this command would commonly be run on CI where colour support isn't universally available.

obmarg (approved):

cli/crates/cli/src/output/report.rs
tomhoule: Good point. I also realize we're not showing warnings when error_count = 0, now. We can/should probably output the warnings separately, before the success / error message.

tomhoule: done

PR #1251, opened 2024-01-18
chore: streaming transport utilities in cli tests

obmarg:
Currently using SSE or multipart in the cli tests is quite laborious.  I'm about to add a few more of these tests, so I've extracted out some utilities to make doing this less laborious.


pimeys (approved):
nice

PR #1252, opened 2024-01-18
feat: hook subscriptions up to auth

obmarg:
Up till now subscriptions weren't authenticated.  This implements that, and adds some tests in the CLI.  

Websockets made this a tiny bit awkward - they commonly get authenticated at the point of connection rather than when an actual subscription comes in.  I've reworked the `Gateway` a bit to handle this - rather than doing auth inside the `execute` calls, it now exposes an `authenticate` function that handles auth and returns a `Session` that exposes the `execute` & `execute_stream` functions.

Also had to fake up an idp server in order to test this in the CLI tests.

Fixes GB-5801
Fixes GB-5717


cli/crates/federated-dev/src/dev/websockets.rs
```diff
@@ -170,20 +149,32 @@ async fn subscription_loop(
...
 }
 
-async fn accept_websocket(mut websocket: WebSocket) -> Option<(WebSocket, Option<serde_json::Value>)> {
+async fn accept_websocket(mut websocket: WebSocket, gateway: &GatewayWatcher) -> Option<(WebSocket, Session)> {
     while let Some(event) = websocket.recv_graphql().await {
         match event {
             Event::ConnectionInit { payload } => {
+                let Some(gateway) = gateway.borrow().clone() else {
+                    websocket
+                        .send(Message::close(4403, "Forbidden").try_into().unwrap())
```
Finistere: Internal server error (503) maybe?

Finistere (approved):
Nice!

engine/crates/engine-v2/src/response/read/ser.rs
```diff
@@ -32,7 +32,6 @@ impl serde::Serialize for Response {
             }
             Response::RequestError(RequestErrorResponse { errors, .. }) => {
                 let mut map = serializer.serialize_map(Some(1))?;
-                map.serialize_entry("data", &serde_json::Value::Null)?;
```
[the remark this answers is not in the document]
obmarg: From [this bit](http://spec.graphql.org/October2021/#sec-Response-Format)

> If the request included execution, the response map must contain an entry with key data. The value of this entry is described in the “Data” section. *If the request failed before execution, due to a syntax error, missing information, or validation error, this entry must not be present.*

cli/crates/federated-dev/src/dev/websockets.rs
obmarg: This is a websocket error code not an http error code - they're different.

(This being forbidden & 4403 was a mistake, but I've fixed that now)

PR #1254, opened 2024-01-19
chore: Add unresolvable keys in FederatedGraph

Finistere:
# Description

# Type of change

- [ ] 💔 Breaking
- [ ] 🚀 Feature
- [ ] 🐛 Fix
- [ ] 🛠️ Tooling
- [ ] 🔨 Refactoring
- [ ] 🧪 Test
- [ ] 📦 Dependency


engine/crates/federated-graph/src/federated_graph.rs
```diff
@@ -59,8 +59,7 @@ pub struct Object {
 
     pub implements_interfaces: Vec<InterfaceId>,
 
-    /// All _resolvable_ keys.
-    pub resolvable_keys: Vec<Key>,
+    pub keys: Vec<Key>,
```
tomhoule, line 62: That would break the JSON format, I'm not up to date on whether this is going to be a problem, wdyt?

tomhoule (approved):

engine/crates/federated-graph/src/federated_graph.rs
Finistere: oh yeah, you're right! forgot about that, good catch!

PR #1256, opened 2024-01-19
Fix cargo warning on main

tomhoule:
```
prisma-desktop.tom.src/gh/grafbase/grafbase λ cargo check
warning: /home/tom/src/gh/grafbase/grafbase/engine/crates/graphql-mocks/Cargo.toml: unused manifest key: package.private
```

The option that was meant is likely "publish"
(https://doc.rust-lang.org/cargo/reference/manifest.html#the-publish-field). This commit replaces private = true with publish = false.


obmarg (approved):
🤦

PR #1258, opened 2024-01-20
fix(engine-v2): better errors when upstream returns null data

Finistere:
With this PR, when `data` is null in the GraphQL response, we now only show the upstream error if there is any.

I simplified the use of `GraphqlResponseSeed` by re-using the `SeedContext`, which holds a `RefCell` of the mutable output, used for deserializing data into the response graph. I also created an inner struct wrapped in a `Rc` to make it simpler to share.

I separated it from the `ResponseObjectWriter` which is a broken abstraction only used for introspection as of today and intend to remove at some point.

### DIFF
--- a/engine/crates/engine-v2/src/execution/context.rs
+++ b/engine/crates/engine-v2/src/execution/context.rs
@@ -5,7 +5,7 @@ use super::Variables;
 use crate::{
     plan::PlanOutput,
     request::{ExecutorWalkContext, OperationWalker, PlanOperationWalker, VariablesWalker},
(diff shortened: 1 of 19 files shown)


obmarg:
> With this PR, when `data` is null in the GraphQL response or we fail to deserialize, we now only show the upstream error if there is any.

What do we do if data is null and there are no errors?

Finistere:
We always try to deserialize data. It's just that now the `DeserializeSeed` for the data is wrapped by another one to return whether the data was null or not. Allowing us to replace errors with upstream ones only if data was null. If there are no upstream errors we don't replace anything and we're left with deserialization errors if any.

obmarg (approved):

PR #1259, opened 2024-01-21
feat(engine): Enforce operation limits in engine v2

jakubadamw:
# Description

Work in progress.

# Type of change

- [ ] 💔 Breaking
- [X] 🚀 Feature
- [ ] 🐛 Fix
- [ ] 🛠️ Tooling
- [ ] 🔨 Refactoring
- [ ] 🧪 Test
- [ ] 📦 Dependency


jakubadamw:
I am yet to write tests for this.

engine/crates/engine-v2/src/request/mod.rs
```diff
@@ -37,6 +39,49 @@ pub struct Operation {
 }
 
 impl Operation {
+    fn enforce_operation_limits(&self, schema: &Schema) -> Result<(), OperationLimitExceededError> {
+        // FIXME: Implement.
+
+        let walker = self.walker_with(schema.walker()).walk(self.root_selection_set_id);
```
Finistere, line 45: nitpick, would have named it `selection_set` actually

engine/crates/engine-v2/src/request/walkers/operation_limits.rs
```diff
@@ -0,0 +1,86 @@
...
+            .max()
+            .expect("must be defined")
+    }
+
+    pub(crate) fn alias_count(&self) -> u16 {
+        (*self)
+            .into_iter()
+            .map(|selection| match selection {
+                BoundSelectionWalker::Field(field) => {
+                    (field.definition().as_field().expect("must be a field").name() == field.response_key_str()) as u16
```
Finistere, line 31: `as_field()` returns `None` for `__typename` fields. Actually, you should add the following to the `BoundAnyFieldDefinitionWalker` (`walkers/field_definition`):

```
    pub fn schema_name(&self) -> &'a str {
        match self.as_ref() {
            BoundAnyFieldDefinition::TypeName(_) => "__typename",
            BoundAnyFieldDefinition::Field(definition) => self.schema_walker.walk(definition.field_id).name(),
        }
    }
```

and compare that against the `response_key_str()` like you do here.

engine/crates/engine-v2/src/request/walkers/operation_limits.rs
```diff
@@ -0,0 +1,86 @@
...
+            })
+            .sum()
+    }
+
+    pub(crate) fn height(&self, fields_seen: &mut HashSet<StringId>) -> u16 {
+        (*self)
+            .into_iter()
+            .map(|selection| match selection {
+                BoundSelectionWalker::Field(field) => {
+                    (if fields_seen.insert(field.definition().as_field().expect("must be a field").name_string_id()) {
```
Finistere, line 72: Similar, will break on `__typename`. Probably worth using the `FieldId` instead of the name imho, no?
```
        if let Some(field) = self.as_field() {
            field.id();
        }
```
`__typename` doesn't have a `FieldId`, so I would just ignore those.

engine/crates/engine-v2/src/request/mod.rs
jakubadamw: Done!

engine/crates/engine-v2/src/request/walkers/operation_limits.rs
jakubadamw: Thanks 🙂

engine/crates/engine-v2/src/request/walkers/operation_limits.rs
jakubadamw: Huh, should use `FieldId`, one sec

engine/crates/engine-v2/src/request/walkers/operation_limits.rs
```diff
@@ -0,0 +1,87 @@
...
+                        .map(|selection_set| selection_set.max_depth())
+                        .unwrap_or_default()
+                        + 1
+                }
+                BoundSelectionWalker::InlineFragment(inline) => inline.selection_set().max_depth(),
+                BoundSelectionWalker::FragmentSpread(spread) => spread.selection_set().max_depth(),
+            })
+            .max()
+            .expect("must be defined")
+    }
```
pimeys, line 22: Would these functions be nicer if they'd take `self` by-value instead of a reference? That would make using them much easier too, as long as `Self: Copy`.

pimeys (approved):
cool

engine/crates/engine-v2/src/request/walkers/operation_limits.rs
jakubadamw: But `Self` itself isn't `Copy,` `*self` invokes the `Deref` implementation, and the result of that implement `IntoIterator`. `BoundSelectionSetWalker` isn't supposed to be `Copy`. But @Finistere would know more. 🙂

jakubadamw:
@Finistere @pimeys thank you, gentlemen!

engine/crates/engine-v2/src/request/walkers/operation_limits.rs
Finistere: No he's right, it is `Copy` and we could use `self`. Sounds so obvious I don't know why I missed it

jakubadamw: @Finistere @pimeys ah, okay, sorry, I stand corrected

PR #1264, opened 2024-01-22
feat: Handle sibling plan dependencies

Finistere:
This PR adapts the planner to take into account sibling plan dependencies. Supposing a subgraph A returns a `Product`, we might retrieve fields from a subgraph B which depends on fields coming from a different subgraph C. This typically happens when the `key` used between `A` and `B` aren't the same.

I've hanged mutations root field logic to also rely on this, making the plan for a root field a dependency of the next one.

All of this code is more complex than needed and I understood why doing this. The tricky parts of GraphQL planning are type conditions and the `skip`/`include` as this generates a lot of conditions for the shape of the queried data. However this only really impacts us when generating the expectations (how to read/validate the upstream response) but not that much the attribution (which plan/resolver is responsible for which fields and add any extra field). Currently, I'm doing both at the same time, but attribution is complex because it happens after the generation of `Operation` which we can't touch anymore during the planning phase.

So that's something I'll try to improve upon when working on the `@skip` & `@include` moving the attribution to the operation build phase. Also allowing us to cache it more easily. It should make the planning _a lot_ easier to understand afterwards.

In the meantime, the planning works roughly like this now:

Given a set of missing fields for a selection set (flattened, no fragments anymore)
- generate all possible candidate resolvers and the fields they could resolve
- if the resolver/field has requirements, we try to plan them immediately, adding extra fields if necessary to the parent plan or any previous child plan. It doesn't matter whether this candidate will be used or not at this stage, planning them immediately makes it overall a lot simpler to manage nested requirements etc.
- Currently, we suppose that If a root field of the current selection set is planned by a Plan A, all of its subselection can also be retrieved by A. So if a resolver requires `author { organizations { ... } }` and A can provide `author`, we also expect it to provide `organizations`. That's limiting, but tricky to do more for now.
- Once we have generated all possible candidates, we select the one with the most fields and generate its input (what will be read from the response). For any extra field we read, we mark them. That ensures we won't retrieve more data than necessary even though we may plan more fields than necessary.
- If there are still any missing fields, we loop again.

### DIFF
--- a/engine/crates/engine-v2/Cargo.toml
+++ b/engine/crates/engine-v2/Cargo.toml
@@ -15,7 +15,6 @@ workspace = true
 
 [dependencies]
 async-runtime.workspace = true
-derive_more = "0.99"
 futures.workspace = true
 im = "15"
 indexmap.workspace = true
@@ -31,6 +30,7 @@ thiserror.workspace = true
 futures-util.workspace = true
 hex = "0.4.3"
 url.workspace = true
+tracing.workspace = true
 
 config = { package = "engine-v2-config", path = "./config" }
 engine-value = { path = "../engine/value" }
--- a/engine/crates/engine-v2/src/execution/coordinator.rs
+++ b/engine/crates/engine-v2/src/execution/coordinator.rs
@@ -1,4 +1,4 @@
-use std::collections::BTreeMap;
+use std::collections::HashMap;
 
 use async_runtime::make_send_on_wasm;
 use engine::RequestHeaders;
@@ -11,7 +11,7 @@ use futures_util::{
 
 use crate::{
     execution::{ExecutionContext, Variables},
-    plan::{Plan, PlanBoundary, Planner, PlanningError},
+    plan::{Plan, PlanBoundary, PlanId, Planner},
     request::Operation,
     response::{ExecutionMetadata, ExecutorOutput, GraphqlError, Response, ResponseBoundaryItem, ResponseBuilder},
     sources::{Executor, ExecutorResult, ResolverInput, SubscriptionExecutor, SubscriptionResolverInput},
@@ -47,50 +47,30 @@ impl<'ctx> ExecutorCoordinator<'ctx> {
     }
 
     pub async fn execute(self) -> Response {
+        if matches!(self.operation.ty, OperationType::Subscription) {
+            unreachable!("execute shouldnt be called for subscriptions")
+        }
+
         let mut planner = Planner::new(&self.engine.schema, &self.operation);
         let mut response = ResponseBuilder::new(self.operation.root_object_id);
 
-        // Mutation root fields need to be executed sequentially. So we're tracking for each
-        // executor whether it was for one and if so execute the next executor in the queue.
-        // Keeping the queue outside of the FuturesUnordered also ensures the future is static
-        // which wasm target somehow required. (not entirely sure why though)
-        let mut mutation_root_fields_executors = vec![];
         let response_boundary = vec![response
             .root_response_boundary()
             .expect("a fresh response should always have a root")];
 
-        let initial_executors = match planner.generate_root_plan_boundary() {
+        let (executors, plans_with_dependencies) = match planner.generate_root_plan_boundary() {
             Ok(boundary) => self.generate_executors(vec![(boundary, response_boundary)], &mut planner, &mut response),
             Err(error) => {
-                response.push_error(error.into_graphql_error());
-
-                vec![]
+                response.push_error(error);
+                (vec![], vec![])
             }
         };
 
         let mut futures = ExecutorFutureSet::new();
-
-        match self.operation.ty {
-            OperationType::Query => {
-                for executor in initial_executors {
-                    futures.execute(executor);
-                }
-            }
-            OperationType::Mutation => {
-                // Reverse the initial_executors so we can pop them in the order they should run
-                mutation_root_fields_executors = initial_executors;
-                mutation_root_fields_executors.reverse();
-
-                if let Some(executor) = mutation_root_fields_executors.pop() {
-                    futures.execute_mutation_root(executor);
-                }
-            }
-            OperationType::Subscription => {
-                unreachable!("execute shouldnt be called for subscriptions")
-            }
+        for executor in executors {
+            futures.execute(executor);
         }
-
-        self.execute_once(futures, response, mutation_root_fields_executors, &mut planner)
+        self.execute_once(futures, response, &mut planner, plans_with_dependencies)
             .await
     }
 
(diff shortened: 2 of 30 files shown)


engine/crates/engine-v2/Cargo.toml
```diff
@@ -31,6 +30,7 @@ thiserror.workspace = true
 futures-util.workspace = true
 hex = "0.4.3"
 url.workspace = true
+tracing.workspace = true
```
obmarg, line 33: Should we be using `log` since `tracing` doesn't work on production?

engine/crates/engine-v2/src/execution/coordinator.rs
```diff
@@ -135,14 +122,19 @@ impl<'ctx> ExecutorCoordinator<'ctx> {
...
-        }) = futures.next().await
-        {
+        let mut outgoing_edges = HashMap::<PlanId, Vec<PlanId>>::new();
+        let mut plans = HashMap::<PlanId, Plan>::new();
+        for plan in plans_with_dependencies {
+            for dependency in &plan.sibling_dependencies {
+                outgoing_edges.entry(*dependency).or_default().push(plan.id);
+            }
+            plans.insert(plan.id, plan);
+        }
```
obmarg, line 135: Feel like there might be some way to wrap this (and your other changes in this function) up so they're easier to follow.

engine/crates/engine-v2/src/plan/planner/boundary_planner.rs
```diff
@@ -0,0 +1,592 @@
...
+pub(in crate::plan) struct ExtraBoundaryField {
+    pub extra_field: ExtraField<ExtraBoundarySelectionSet>,
+    pub read: bool,
+}
+
+#[derive(Debug, Clone)]
+pub(in crate::plan) struct ExtraBoundarySelectionSet {
+    pub ty: SelectionSetType,
+    pub fields: HashMap<FieldId, ExtraBoundaryField>,
+}
```
obmarg, line 66: A docstring or 6 would be nice, not clear to me what any of these things are

engine/crates/engine-v2/src/plan/planner/boundary_planner.rs
```diff
@@ -0,0 +1,592 @@
...
+        let walker = planner.default_operation_walker();
+        PlanBoundaryChildrenPlanner {
+            planner,
+            walker,
+            maybe_parent,
+            children: vec![],
+        }
+    }
+
+    pub fn plan_children(
```
obmarg, line 79: Given that this function is like 200 lines long (and I assume it calls into the other 400 lines of this file) a docstring summarising its purpose would be useful

engine/crates/integration-tests/src/federation/builder.rs
```diff
@@ -53,7 +53,7 @@ impl FederationGatewayBuilder {
         let graph = graphql_composition::compose(&subgraphs)
             .into_result()
             .expect("schemas to compose succesfully");
-
+        println!("{}", graph.to_sdl().unwrap());
```
obmarg, line 56: This left here intentionally?

Finistere: ah no mistake, thx

engine/crates/engine-v2/src/plan/attribution.rs
```diff
@@ -129,181 +132,121 @@ impl<'a> std::ops::Deref for ExtraFieldWalker<'a> {
...
-                    }),
-            });
-            self[extra_selection_set_id].fields.insert(field_id, extra_field_id);
-            extra_field_id
-        } else {
-            self[extra_selection_set_id].fields[&field_id]
+                            },
+                        };
+                        let id = ExtraFieldId::from(self.extra_fields.len());
+                        self.extra_fields.push(field);
```
pimeys, line 229: I find this a bit too confusing to mutate `self` in an iterator map. Maybe convert this whole structure to a nice for loop for clarity?

engine/crates/engine-v2/src/plan/mod.rs
```diff
@@ -61,4 +67,8 @@ pub struct ChildPlan {
     pub resolver_id: ResolverId,
     pub input_selection_set: ReadSelectionSet,
     pub root_selection_set: FlatSelectionSet<EntityType>,
+    pub sibling_dependencies: HashSet<PlanId>,
+    // Only includes extra fields necessary for other child plans within the same
+    // plan boundary.
+    extra_selection_sets: HashMap<BoundSelectionSetId, planner::ExtraBoundarySelectionSet>,
```
pimeys, line 73: Maybe explain this with a better docstring. Why are we indexing selection sets?

engine/crates/engine-v2/Cargo.toml
Finistere: For now, I'm only using those with `DEBUG` for stuff we would want to log in production imho and tracing provides nice formatting in tests. Kinda tempted to keep that as is for now to be honest...

obmarg: Assuming you mean stuff we would not want to log in production then ok

engine/crates/engine-v2/src/plan/planner/boundary_planner.rs
pimeys: Maybe even splitting it into its own module with flat functions, smaller functions and a healthy amount of rustdoc?

engine/crates/engine-v2/Cargo.toml
Finistere: yeah, missed the "not" 😅

engine/crates/engine-v2/src/execution/coordinator.rs
```diff
@@ -135,14 +122,19 @@ impl<'ctx> ExecutorCoordinator<'ctx> {
...
-        mut mutation_root_fields_executors: Vec<Executor<'ctx>>,
         planner: &mut Planner<'ctx>,
+        plans_with_dependencies: Vec<Plan>,
     ) -> Response {
-        while let Some(ExecutorFutResult {
-            result,
-            is_mutation_root_field,
-        }) = futures.next().await
-        {
+        let mut outgoing_edges = HashMap::<PlanId, Vec<PlanId>>::new();
```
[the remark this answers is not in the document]
Finistere: Did the first point. I agree with the second point overall, but that's something I intend to change later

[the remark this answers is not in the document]
Finistere: I've separated the concerns a bit better now, but yeah not hashing could be avoided later.

engine/crates/engine-v2/src/execution/coordinator.rs
Finistere: I think it did better, not great, but ok for now 🤷

engine/crates/engine-v2/src/plan/attribution.rs
```diff
@@ -49,16 +52,16 @@ impl Attribution {
     }
 }
 
-#[derive(Debug)]
-pub struct ExtraField {
+#[derive(Debug, Clone)]
+pub struct ExtraField<SelectionSet = ExtraSelectionSetId> {
```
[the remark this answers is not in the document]
Finistere: `Clone` because a child plan needs to be cloneable for subscriptions which clones them at each new data input to continue the execution. Should be avoidable later. The generic is there because during planning I need a mapping from `FieldId` to fields to satisfy requirements but after this is an id.

engine/crates/engine-v2/src/execution/coordinator.rs
obmarg: Much better, thanks

obmarg (approved):

PR #1267, opened 2024-01-22
cli: handle types from connectors in resolver codegen

tomhoule:
The decisive change to get the GraphQL schema from the registry instead of the config. (Internal) RFC:
https://www.notion.so/grafbase/Grafbase-GraphQL-clients-integration-68b7a475f5d94222aba4c9e6a405b2d7

closes GB-5256

### DIFF
--- a/cli/crates/typed-resolvers/src/lib.rs
+++ b/cli/crates/typed-resolvers/src/lib.rs
@@ -5,44 +5,33 @@ mod check_resolver;
 mod codegen;
 mod error;
 
-pub use self::analyze::{analyze as analyze_schema, AnalyzedSchema};
+pub use self::{
+    analyze::{analyze as analyze_schema, AnalyzedSchema},
+    check_resolver::check_resolver,
+};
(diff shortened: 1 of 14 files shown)


cli/crates/typed-resolvers/src/lib.rs
```diff
@@ -62,7 +64,7 @@ pub fn check_resolvers(resolvers_root: &Path, schema: &analyze::AnalyzedSchema<'
     AnalyzedResolvers { errs }
 }
 
-fn experimental_codegen_is_enabled(parsed_schema: &engine_parser::types::ServiceDocument) -> bool {
+pub fn experimental_codegen_is_enabled(parsed_schema: &engine_parser::types::ServiceDocument) -> bool {
```
tomhoule, line 67: Technically, we could move this to the registry now, but this works and it will go away in the not too far away future, so I decided not to make the change.

obmarg (approved):

PR #1268, opened 2024-01-22
chore(tests): Enable the 'preserve_order' feature flag for serde_json globally

jakubadamw:
# Description

This will help avoid map ordering issues relative to what directory the tests are ran from.

# Type of change

- [ ] 💔 Breaking
- [ ] 🚀 Feature
- [ ] 🐛 Fix
- [ ] 🛠️ Tooling
- [ ] 🔨 Refactoring
- [X] 🧪 Test
- [ ] 📦 Dependency

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3903,6 +3903,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "more-asserts"
+version = "0.3.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "1fafa6961cabd9c63bcd77a45d7e3b7f3b552b70417831fb0f56db717e72407e"
+
+[[package]]
 name = "multer"
 version = "3.0.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -4498,6 +4504,7 @@ dependencies = [
  "indexmap 1.9.3",
  "insta",
  "itertools 0.12.0",
+ "more-asserts",
  "once_cell",
  "openapi",
  "openapiv3",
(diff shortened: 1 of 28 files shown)


Finistere:
You have a bunch of unrelated commits in it^^

obmarg (approved):
Nice, thanks for handling this

jakubadamw:
image

Quite an odd failure to be coming out of this

obmarg:
> image
> Quite an odd failure to be coming out of this

Ah fuck.  That is actually expected.  `preserve_order` uses indexmap which is a few times larger than `BTreeMap`.

Now, I guess the question is will that bump in size start causing memory problems again (the reason I put that test there)

jakubadamw:
I changed the test to use `more_asserts::assert_le`.

```
    #[test]
    fn test_graph_size() {
        // Our graph can end up with a ton of nodes & edges so it's important
        // that they don't get too big.
        more_asserts::assert_le!(std::mem::size_of::<Node>(), 48);
        more_asserts::assert_le!(std::mem::size_of::<Edge>(), 48);
    }
```

Somehow, this PR caused the memory size of `Node` to increase to 88.

```
--- STDERR:              parser-openapi graph::tests::test_graph_size ---
thread 'graph::tests::test_graph_size' panicked at engine/crates/parser-openapi/src/graph/mod.rs:597:9:
assertion failed: `(left <= right)`
  left: `88`,
 right: `48`
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

jakubadamw:
It's ultimately up to the compiler to decide on the memory layout, so I assume this effect isn't strictly related. @obmarg WDYT? Safe to increase?

jakubadamw:
> That is actually expected. `preserve_order` uses indexmap which is a few times larger than `BTreeMap`.
> 
> Now, I guess the question is will that bump in size start causing memory problems again (the reason I put that test there)

Ah, of course, makes sense!

PR #1269, opened 2024-01-22
chore: Various engine-v2-schema and federated-graph fixes

Finistere:
- Ignoring empty entity keys in engine schema
- Ignoring schema directives in federated-graph for now. The one used in gateway-benchmarks has:

```graphql
schema
  https://github.com/link(url: "https://specs.apollo.dev/link/v1.0")
  https://github.com/link(url: "https://specs.apollo.dev/join/v0.3", for: EXECUTION)
{
  query: Query
}
```

- Ignoring `@join_field` when `external: true`. Not sure to see any point in them.

### DIFF
--- a/engine/crates/engine-v2/schema/src/conversion.rs
+++ b/engine/crates/engine-v2/schema/src/conversion.rs
@@ -68,6 +68,10 @@ impl From<Config> for Schema {
 
             for key in object.keys {
                 let subgraph_id = key.subgraph_id.into();
+                // Some SDL are generated with empty keys, they're useless to us.
+                if key.fields.is_empty() {
+                    continue;
+                }
                 if key.resolvable {
                     let resolver_id = ResolverId::from(schema.resolvers.len());
                     schema
--- a/engine/crates/federated-graph/src/from_sdl.rs
+++ b/engine/crates/federated-graph/src/from_sdl.rs
@@ -39,9 +39,9 @@ struct State<'a> {
     strings: IndexSet<String>,
     field_types: indexmap::IndexSet<FieldType>,
 
-    query_type: Option<ObjectId>,
-    mutation_type: Option<ObjectId>,
-    subscription_type: Option<ObjectId>,
+    query_type_name: Option<String>,
+    mutation_type_name: Option<String>,
+    subscription_type_name: Option<String>,
 
     definition_names: HashMap<&'a str, Definition>,
     selection_map: HashMap<(Definition, &'a str), FieldId>,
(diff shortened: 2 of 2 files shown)


engine/crates/engine-v2/schema/src/conversion.rs
```diff
@@ -68,6 +68,10 @@ impl From<Config> for Schema {
 
             for key in object.keys {
                 let subgraph_id = key.subgraph_id.into();
+                // Some SDL are generated with empty keys, they're useless to us.
+                if key.fields.is_empty() {
+                    continue;
+                }
```
obmarg, line 74: Why is stuff generated with empty keys?  That kinda sounds like a bug, though maybe there's something I'm missing

Finistere: Ours doesn't, but theirs does: https://github.com/the-guild-org/gateways-benchmark/blob/main/federation-v1/gateways/apollo-router/supergraph.graphql#L59

engine/crates/federated-graph/src/from_sdl.rs
```diff
@@ -138,9 +138,7 @@ fn ingest_fields<'a>(parsed: &'a ast::ServiceDocument, state: &mut State<'a>) ->
     for definition in &parsed.definitions {
         match definition {
             ast::TypeSystemDefinition::Schema(_) => {
-                return Err(DomainError(
-                    "Not implemented: schema definitions in federated schema".to_owned(),
-                ))
+                continue;
```
obmarg, line 141: What's the thinking here?  Seems to me like this error would be useful rather than silently ignoring schema definitions and potentially doing the wrong thing

Finistere: Fair point. My goal was to ignore the following:
https://github.com/the-guild-org/gateways-benchmark/blob/main/federation-v1/gateways/apollo-router/supergraph.graphql#L1

```
schema
  @link(url: "https://specs.apollo.dev/link/v1.0")
  @link(url: "https://specs.apollo.dev/join/v0.3", for: EXECUTION)
{
  query: Query
}
```

I didn't want to bother it more than necessary. I could just ignore `@link` directives and handle the custom query names though. Ok will do that

engine/crates/engine-v2/schema/src/conversion.rs
obmarg: Ah, annoying. But ok

engine/crates/federated-graph/src/from_sdl.rs
tomhoule: FWIW I think that the composition implementations I've looked at (including ours) will always name the composed query type `Query`, the composed mutation type `Mutation` and so on and so on.

Finistere: Agreed it's most likely useless today, probably forever. Anyway did it quickly, ignoring `@link` directive and raising an error for other directives.

PR #1278, opened 2024-01-25
feat: add websocket urls into the federation config

obmarg:
I'd hard coded this in previous commits so I could do testing without messing about with configuration - this is a more "proper" implementation.  By default we use the normal subgraph URL, but users can override with a specific URL to use for websockets.

Part of GB-5711


tomhoule (approved):

obmarg:
Actually, I can't get the tests to pass without adding tsconfig support into this PR so i'll be doing that here as well.

PR #1280, opened 2024-01-25
Add two basic engine benchmarks

Finistere:
Adding two simple engine benchmarks just to have a baseline when working on it.


engine/crates/engine-v2/src/sources/graphql/mod.rs
```diff
@@ -84,6 +84,7 @@ impl<'ctx> GraphqlExecutor<'ctx> {
             })
             .await?
             .bytes;
+        tracing::debug!("{}", String::from_utf8_lossy(&bytes));
```
fbjork, line 87: still need this?

fbjork (approved):

engine/crates/engine-v2/src/sources/graphql/mod.rs
Finistere: half the time it's useful I'd say. Not sure, but it has no cost at runtime if not activated. So I'll leave it for now I think.

PR #1283, opened 2024-01-26
fix: send headers in subscription init payload

obmarg:
I've been tesitng things with hasura, which requires auth - but was getting a
lot of "forbidden" errors.  After far too much digging, turns out this is
because I forgot to pass headers in the `connection_init` payload.

Also enabled TLS in `async_tungstenite` since it turns out to not be 
enabled by default.

Fixes GB-5879

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -420,8 +420,11 @@ dependencies = [
  "futures-util",
  "log 0.4.20",
  "pin-project-lite",
+ "rustls-pki-types",
  "tokio",
+ "tokio-rustls 0.25.0",
  "tungstenite",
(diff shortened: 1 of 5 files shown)


tomhoule (commented):
Is that payload part of the spec?

tomhoule (approved):

obmarg:
> Is that payload part of the spec?

Nope, as far as I'm aware auth is unspecified for websockets.  I can't remember precisely where I took this approach from (it's how we're doing auth in our websocket implementation) but it's also what Hasura expect, so does seem a common-ish approach.

PR #1287, opened 2024-01-27
chore: rework engine-v2 introspection

Finistere:
The primary goal was to remove a bunch of code that is fundamentally useless and broken. I added
a sort of Response builder initially supporting error propagation for
cases when we write data ourselves. But this case actually only happens
for introspection. Everything else receives data from somewhere and
needs to validate it. Furthermore there aren't any possible errors
during introspection, they're either caught during query validation or
when building the schema.

So removing all of that code to build the response manually making it
easier for later changes.

As a bonus it improves introspection speed by ~30%, now it spends most
of its time preparing the operation.

# Description

# Type of change

- [ ] 💔 Breaking
- [ ] 🚀 Feature
- [ ] 🐛 Fix
- [ ] 🛠️ Tooling
- [ ] 🔨 Refactoring
- [ ] 🧪 Test
- [ ] 📦 Dependency


engine/crates/engine-v2/schema/src/walkers/definition.rs
```diff
@@ -51,15 +51,15 @@ impl<'a> DefinitionWalker<'a> {
         }
     }
 
-    pub fn interfaces(&self) -> Option<Box<dyn Iterator<Item = InterfaceWalker<'a>> + 'a>> {
+    pub fn interfaces(&self) -> Option<Box<dyn ExactSizeIterator<Item = InterfaceWalker<'a>> + 'a>> {
```
Finistere, line 54: not really necessary anymore, but kept it 🤷

tomhoule: yeah it's likely going to be useful again at some point

tomhoule (approved):
Nice!

PR #1288, opened 2024-01-28
chore: remove field_definition in engine-v2 request walkers

Finistere:
Simplifying engine-v2 request walkers by removing the field definition.
Representing it as a `BoundField` enum is simpler to understand.

Also adding `BoundFragmentSpreadId` and `BoundInlineFragmentId`, it doesn't change a lot of the code and makes the `BoundSelection` enum smaller.


engine/crates/integration-tests/tests/federation/basic/operation_limits.rs
```diff
@@ -3,7 +3,7 @@ use graphql_mocks::{FakeGithubSchema, MockGraphQlServer};
 use integration_tests::{federation::GatewayV2Ext, runtime};
 
 #[rstest::rstest]
-#[case(
+#[case( // 1
```
obmarg, line 6: You know you can give these better names with `case::insert_name_here`?

obmarg (approved):

engine/crates/integration-tests/tests/federation/basic/operation_limits.rs
Finistere: good point, should have done that instead.

PR #1300, opened 2024-01-30
composition: federated graph v2

tomhoule:
This PR was prompted by GB-5894, but there are other changes to FederatedGraph that were discussed with @Finistere , so they are bundled here in a new FederatedGraph version.

Specifically:

- `Value::Float` becomes an f64 to align with the spec.
- Arguments of output fields and fields of input objects are unified into InputValueDefinition for code reuse and, more importantly, common identifiers for both, so we can treat them uniformly.
- Directives are treated more uniformly across the schema and stored separately. That had to happen because we have a new Directive type in V2 because of the new Value type containing Float(f64). That change trickled down to all the structs with a Vec<Directive>, and made the upgrade from V1 to V2 painful. Future changes to directives will be much smoother with this decoupling.

We also wanted to represent defaults of arguments but 1. we would need to check how they are composed, 2. we can add them backwards compatibly later and this PR is already large enough.

closes GB-5894


engine/crates/engine-v2/schema/tests/conversion.rs
```diff
@@ -1,101 +1,101 @@
...
-}
-"#;
-
-#[test]
-fn should_not_fail() {
-    let FederatedGraph::V1(graph) = FederatedGraph::from_sdl(SCHEMA).unwrap();
-    let config = config::VersionedConfig::V1(graph).into_latest();
-    let _schema = Schema::from(config);
-}
+// use engine_v2_schema::Schema;
```
tomhoule, line 1: This test is broken, because `from_sdl()` and `to_sdl()`  now work with FederatedGraph::V2, so we need the engine v2 conversion code to unbreak it.

engine/crates/federated-graph/src/federated_graph.rs
tomhoule, line 1: The code here moved to v1.rs, for the most part.

engine/crates/integration-tests/src/federation/builder/bench.rs
```diff
@@ -37,7 +37,9 @@ impl<'a> FederationGatewayBench<'a> {
                 .collect(),
             dummy_responses_index.clone(),
         );
-        let FederatedGraph::V1(federated_graph) = FederatedGraph::from_sdl(schema).unwrap();
+        let FederatedGraph::V1(federated_graph) = FederatedGraph::from_sdl(schema).unwrap() else {
+            todo!("handle v2");
```
tomhoule, line 41: Handing over there, can work on it if that feels better.

engine/crates/federated-graph/src/federated_graph/v2.rs
```diff
@@ -0,0 +1,517 @@
...
+    pub field_types: Vec<FieldType>,
+
+    /// All composed directive instances (not definitions) in a federated graph.
+    pub directives: Vec<Directive>,
+}
+
+#[derive(serde::Serialize, serde::Deserialize, Clone, PartialEq, Eq, PartialOrd, Ord)]
+pub struct InputValueDefinition {
+    pub name: StringId,
+    pub field_type_id: FieldTypeId,
```
Finistere, line 52: Up to you, I've named it `TypeId` in the `engine-v2-schema` because it mixes Input & Output types. An input value can also be an argument, so field_type  feels a bit weird imho

engine/crates/federated-graph/src/federated_graph/v2.rs
```diff
@@ -0,0 +1,517 @@
...
+    pub description: Option<StringId>,
+}
+
+#[derive(serde::Serialize, serde::Deserialize, Clone, PartialEq, PartialOrd)]
+pub enum Value {
+    String(StringId),
+    Int(i64),
+    Float(f64),
+    Boolean(bool),
+    EnumValue(StringId),
```
Finistere, line 63: as I had asked why you did it, it's worth a comment here specifying why it's not the same as `String(StirngId)` with the example you mentionned

engine/crates/federated-graph/src/federated_graph/v2.rs
```diff
@@ -0,0 +1,517 @@
...
+#[derive(serde::Serialize, serde::Deserialize, Clone, PartialEq, PartialOrd)]
+pub struct Directive {
+    pub name: StringId,
+    pub arguments: Vec<(StringId, Value)>,
+}
+
+#[derive(serde::Serialize, serde::Deserialize, Clone)]
+pub struct Enum {
+    pub name: StringId,
+    pub values: Vec<EnumValue>,
```
Finistere, line 77: IMHO it's worth changing it immediately to an `EnumValueId` and storing `EnumValue`s in an array, I'll need it to support renaming those efficiently in default values.

engine/crates/federated-graph/src/federated_graph/v2.rs
```diff
@@ -0,0 +1,517 @@
...
+    /// See [FieldRequires]
+    pub requires: Vec<FieldRequires>,
+
+    /// See [Override].
+    pub overrides: Vec<Override>,
+
+    /// All directives that made it through composition. Notably includes `@tag`.
+    pub composed_directives: Directives,
+
+    #[serde(default)]
```
Finistere, line 125: if you keep `#[serde(default)]` in the v2, you might want to skip to use `#[serde(default, skip_serializing_if = "Option::is_none")]`

engine/crates/federated-graph/src/federated_graph/v2.rs
```diff
@@ -0,0 +1,517 @@
...
+    pub description: Option<StringId>,
+}
+
+/// A (start, len) range in FederatedSchema.
+pub type Directives = (DirectiveId, usize);
```
Finistere, line 199: all of those `usize` could be a u32 since ids are u32, but not that important

engine/crates/federated-graph/src/federated_graph/v2.rs
```diff
@@ -0,0 +1,517 @@
...
+
+pub const NO_DIRECTIVES: Directives = (DirectiveId(0), 0);
+
+#[derive(serde::Serialize, serde::Deserialize, Clone, Copy, PartialEq, Eq, PartialOrd, Ord)]
+pub struct InputValueDefinitionId(pub usize);
+
+/// A (start, len) range in FederatedSchema.
+pub type InputValueDefinitions = (InputValueDefinitionId, usize);
+
+pub const NO_INPUT_VALUE_DEFINITION: InputValueDefinitions = (InputValueDefinitionId(0), 0);
```
Finistere, line 212: smart

engine/crates/federated-graph/src/lib.rs
```diff
@@ -16,15 +16,28 @@ pub use from_sdl::{from_sdl, DomainError};
...
+    #[deprecated(note = "Use into_sdl() instead")]
+    pub fn to_sdl(&self) -> Result<String, std::fmt::Error> {
+        self.clone().into_sdl()
+    }
+
     pub fn from_sdl(sdl: &str) -> Result<FederatedGraph, DomainError> {
         from_sdl(sdl)
     }
+
+    pub fn into_v2(self) -> FederatedGraphV2 {
```
Finistere, line 37: `into_latest()` ? could be worth doing the same as Graeme did in `engine-v2-config`

engine/crates/engine-v2/schema/tests/conversion.rs
Finistere: IMHO you should just an `engine_v2_config::Config::from_graph(federated_graph::latest::FederatedGraph)`

Finistere (commented):
nice work already!

engine/crates/federated-graph/src/federated_graph/v2.rs
tomhoule: Ah good point, I agree it's probably a lot less pain in the future, doing that.

engine/crates/federated-graph/src/lib.rs
tomhoule: Ah I didn't have that in mind. `into_latest()` makes more sense and is more consistent so I'll do that.

engine/crates/engine-v2/schema/tests/conversion.rs
tomhoule: I left the conversion code alone (so far) in this PR, so it still expects a federated graph v1 — we have to change that, probably even before merging this branch or tests are going to be broken.

engine/crates/federated-graph/src/federated_graph/v2.rs
tomhoule: Yes, it was to avoid the confusion between an identifier for a top level type and an identifier for the type of a field (which this is). Taking from the spec, maybe `TypeRefId` would be even better, but let's go with TypeId since it's shorter and the other interpretation of typeid already exists and it's called `Definition`.

engine/crates/engine-v2/schema/tests/conversion.rs
Finistere: ahh right, I can push a commit in your branch for that if you want once you're done, up to you

engine/crates/federated-graph/src/federated_graph/v2.rs
tomhoule: done in next commit

engine/crates/federated-graph/src/federated_graph/v2.rs
tomhoule: done in next commit

engine/crates/federated-graph/src/federated_graph/v2.rs
tomhoule: done in next commit

engine/crates/federated-graph/src/lib.rs
tomhoule: done in next commit

engine/crates/composition/src/compose/entity_interface.rs
```diff
@@ -188,3 +164,28 @@ pub(crate) fn merge_entity_interface_definitions<'a>(
...
+            directives,
+            description: None,
+        });
+
+        if let Some((_start, len)) = &mut ids {
+            *len += 1;
+        } else {
+            ids = Some((id, 1));
+        }
+    }
```
tomhoule, line 188: I find these loops a bit awkward. Worth adding an abstraction?

Finistere: What I do for these kind of thinds in engine-v2 is:
```
let start = array.len();
// insert stuff
let end = array.len();
(start, end - start)
```

you could also have a `insert_input_value_definitions()` taking an `Iterator` for argument

engine/crates/engine-v2/schema/src/builder/mod.rs
```diff
@@ -502,16 +489,17 @@ impl From<federated_graph::Value> for Value {
...
             federated_graph::Value::EnumValue(s) => Value::EnumValue(s.into()),
             federated_graph::Value::Object(fields) => Value::Object(
                 fields
+                    .into_vec()
                     .into_iter()
                     .map(|(id, value)| (id.into(), value.into()))
                     .collect(),
             ),
-            federated_graph::Value::List(l) => Value::List(l.into_iter().map(Into::into).collect()),
+            federated_graph::Value::List(l) => Value::List(l.into_vec().into_iter().map(Into::into).collect()),
```
Finistere, line 502: why `into_vec()`?

engine/crates/engine-v2/schema/src/lib.rs
```diff
@@ -124,10 +126,23 @@ pub struct Object {
     pub description: Option<StringId>,
     pub interfaces: Vec<InterfaceId>,
     /// All directives that made it through composition. Notably includes `@tag`.
-    pub composed_directives: Vec<Directive>,
+    pub composed_directives: Directives,
```
Finistere, line 129: In my planning PR I've added `IdRange` in `crate::utils` for this exact purpose, so you should use `IdRange<DirectiveId>` here, same for the other ones. Avoids the need for the extra index and provides a way to iterate over the ids.

engine/crates/engine-v2/schema/tests/conversion.rs
```diff
@@ -95,7 +95,17 @@ type User
 
 #[test]
 fn should_not_fail() {
-    let FederatedGraph::V1(graph) = FederatedGraph::from_sdl(SCHEMA).unwrap();
-    let config = config::VersionedConfig::V1(graph).into_latest();
+    let graph = FederatedGraph::from_sdl(SCHEMA).unwrap().into_latest();
```
Finistere, line 98: Will add a function `config::latest::Config::from_schema()` if you don't^^

engine/crates/engine-v2/schema/src/sources/introspection.rs
```diff
@@ -592,6 +600,16 @@ impl<'a> IntrospectionSchemaBuilder<'a> {
         }
     }
 
+    fn insert_field_argument(&mut self, field_id: FieldId, argument: InputValueId) {
```
Finistere, line 603: IMHO we should have a function `insert_all_field_arguments()` to make sure we do add everything at once. Here it's fine because introspection likely won't change. But it's easy to misuse this function

engine/crates/engine-v2/schema/src/builder/mod.rs
```diff
@@ -469,21 +451,26 @@ impl From<federated_graph::Object> for Object {
...
     fn from(directive: federated_graph::Directive) -> Self {
-        Directive {
-            name: directive.name.into(),
-            arguments: directive
-                .arguments
-                .into_iter()
-                .map(|(id, value)| (id.into(), value.into()))
-                .collect(),
+        match directive {
+            federated_graph::Directive::Inaccessible => Directive::Inaccessible,
```
Finistere, line 463: I'm not a fan of keeping `Inaccessible`, but removing those fields would make all id conversions a lot more complex, so well, let's keep it.

But this isn't enough, an inaccessible field will be exposed through introspection _and_ be accessible through queries here.

Finistere (approved):
I think would be worth using `IdRange` and for the inaccessible stuff, for another PR^^

engine/crates/engine-v2/schema/src/builder/mod.rs
tomhoule: `Box::<[T]>::into_iter()` gives you an iterator over `&T`s, but we were expecting owned iterators in these spots. `into_vec()` does that without copy / allocation. Happy to change if there's a better option though.

engine/crates/engine-v2/schema/src/lib.rs
tomhoule: Will do :+1:

engine/crates/engine-v2/schema/tests/conversion.rs
tomhoule: Yes it was already tedious here, can do in this branch.

engine/crates/engine-v2/schema/src/sources/introspection.rs
tomhoule: true, that came to mind too but I didn't act on it. Let's enforce this.

engine/crates/federated-graph/src/federated_graph/v2.rs
```diff
@@ -0,0 +1,560 @@
...
+                }
+            }
+        )*
+    }
+}
+
+id_newtypes! {
+    EnumId + enums + Enum,
+    FieldId + fields + Field,
+    TypeId + field_types + FieldType,
```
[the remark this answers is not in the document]
tomhoule: I think we're reusing `FieldType` from v1 but it shouldn't break in the json format, let's do this.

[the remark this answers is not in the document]
tomhoule: True true, will make it consistent.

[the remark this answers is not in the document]
tomhoule: By the way, probably a change for federated graph v3 but what do you think of making FieldType more like [this](https://github.com/grafbase/grafbase/blob/main/engine/crates/operation-checks/src/schema/wrapping_types.rs)? So instead of interning, use a `DefinitionId` and encode the wrapping types in a `u64`.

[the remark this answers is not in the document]
Finistere: haha did the same https://github.com/grafbase/grafbase/pull/1324 xD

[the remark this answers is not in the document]
Finistere: I'm not sure to be honest. We could but we would need a `UnpackedDefinition` for `match` like I did for this: https://github.com/grafbase/grafbase/blob/main/engine/crates/engine-v2/src/response/path.rs#L65

not sure how great that is, but why not 🤷 I'm already reserving the 4 upper bits in all ids in`engine-v2-schema` :D

engine/crates/engine-v2/schema/tests/conversion.rs
tomhoule: done

PR #1303, opened 2024-01-31
cli: add GraphQL syntax highlighting in introspect

tomhoule:
- Implemented using [syntect](https://crates.io/crates/syntect).
- Only for `grafbase introspect` at first.
- Comes with a `--no-color` flag to disable.
- Takes into account the [`NO_COLOR` standard](https://no-color.org/)
- Does not add ansi colors when used in pipes
- Adds nearly exactly 1MB to the release binaries
- Uses the solarized dark theme because that's the only builtin dark theme that works well on my light background terminal. We could look into [bat's themes](https://github.com/sharkdp/bat/tree/master/assets/themes) or make our own for alternative options

closes GB-5888

### DIFF
--- a/cli/crates/cli/src/introspect.rs
+++ b/cli/crates/cli/src/introspect.rs
@@ -5,9 +5,9 @@ pub(crate) fn introspect(command: &IntrospectCommand) -> Result<(), CliError> {
     match (command.url(), command.dev) {
         (Some(url), _) => {
             let headers = command.headers().collect::<Vec<_>>();
-            introspect_remote(url, &headers)
+            introspect_remote(url, &headers, command.no_color)
         }
-        (None, true) => introspect_local(),
+        (None, true) => introspect_local(command.no_color),
         (None, false) => {
             eprintln!("Error: Either the --url or the --dev argument must be provided.");
             std::process::exit(1);
@@ -15,10 +15,10 @@ pub(crate) fn introspect(command: &IntrospectCommand) -> Result<(), CliError> {
     }
 }
 
-fn introspect_local() -> Result<(), CliError> {
+fn introspect_local(no_color: bool) -> Result<(), CliError> {
     match server::introspect_local().map_err(CliError::ServerError)? {
         server::IntrospectLocalOutput::Sdl(schema) => {
-            println!("{schema}");
+            print_introspected_schema(&schema, no_color);
         }
         server::IntrospectLocalOutput::EmptyFederated => {
             report::federated_schema_local_introspection_not_implemented();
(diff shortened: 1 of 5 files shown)


cli/crates/cli/src/introspect.rs
```diff
@@ -28,11 +28,52 @@ fn introspect_local() -> Result<(), CliError> {
...
+        }
         Err(e) => Err(CliError::Introspection(e)),
     }
 }
+
+fn print_introspected_schema(sdl: &str, no_color: bool) {
+    use syntect::util::{as_24_bit_terminal_escaped, LinesWithEndings};
+
+    // No highlighting when stdout is not a tty (likely a pipe) or when explicitly requested.
+    if no_color || !atty::is(atty::Stream::Stdout) || no_color_env_var() {
```
yoav-lavi, line 47: Would this also follow `NO_COLOR` being an env var?

cli/crates/cli/src/introspect.rs
```diff
@@ -28,11 +28,52 @@ fn introspect_local() -> Result<(), CliError> {
...
+        println!("{sdl}");
+        return;
+    }
+
+    const GRAMMAR: &str = include_str!("introspect/graphql_grammar.yml");
+
+    let mut builder = syntect::parsing::SyntaxSetBuilder::new();
+    builder.add(syntect::parsing::SyntaxDefinition::load_from_str(GRAMMAR, false, Some("graphql")).unwrap());
```
yoav-lavi, line 55: Could we make this an expect or a user facing error?

cli/crates/cli/src/introspect.rs
```diff
@@ -28,11 +28,52 @@ fn introspect_local() -> Result<(), CliError> {
...
+            .expect("line to be highlightable");
+
+        let escaped = as_24_bit_terminal_escaped(&ranges[..], false);
+        print!("{}", escaped);
+    }
+}
+
+/// https://no-color.org/
+fn no_color_env_var() -> bool {
+    std::env::var("NO_COLOR").is_ok_and(|s| !s.is_empty())
```
yoav-lavi, line 78: Would it be possible to use a longer var name here?

yoav-lavi (approved):
Approved, nice work, see comments

cli/crates/cli/src/introspect.rs
tomhoule: Addressed in commit (with expect, since there's really no variability, it always loads exactly the same grammar definition baked into the binary).

cli/crates/cli/src/introspect.rs
tomhoule: changed to `value`.

pimeys:
Screenshot would've been cool.

PR #1306, opened 2024-02-01
fix: handle websocket error conditions better in cli

obmarg:
A few things I knowingly skipped from my initial implementation:

- Timeouts if the user doesn't init the connection
- Returning the `Error` message on fatal errors.
- Erroring out if the user tries to subscribe before init

Fixes GB-5792

### DIFF
--- a/cli/crates/federated-dev/src/dev/websockets.rs
+++ b/cli/crates/federated-dev/src/dev/websockets.rs
@@ -1,6 +1,7 @@
 use std::collections::HashMap;
 
 use ::axum::extract::ws::{self, WebSocket};
+use engine_v2::Response;
 use futures_util::{pin_mut, stream::SplitStream, SinkExt, StreamExt};
 use gateway_v2::{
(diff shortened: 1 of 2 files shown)


cli/crates/federated-dev/src/dev/websockets.rs
```diff
@@ -31,16 +34,32 @@ impl WebsocketAccepter {
...
+                match accept_future.await {
+                    Ok(Some(session)) => websocket_loop(connection, session).await,
+                    Ok(None) => {
+                        log::warn!("Failed to accept websocket connection");
+                    }
+                    Err(_) => {
+                        log::info!("Connection wasn't initialised on time, dropping");
+                        connection
+                            .send(
+                                Message::close(4408, "Connection initialisation timeout")
```
tomhoule, line 55: I haven't dug in the api so maybe it's obvious, but what is 4408 here? (a port?)

tomhoule (approved):

PR #1308, opened 2024-02-01
graphql-schema-validation: expand README, fix typo in error

tomhoule:
There was a typo in one validation error message, this commit fixes it.

The README is made more complete by adding an `Example` section and a `Status` section.

### DIFF
--- a/engine/crates/validation/tests/validation_errors/unknown_type_in_schema_def.errors.txt
+++ b/engine/crates/validation/tests/validation_errors/unknown_type_in_schema_def.errors.txt
@@ -1,4 +1,4 @@
-  × Cannot set schema query root to unknown type type `Dunno`
+  × Cannot set schema query root to unknown type `Dunno`
    ╭─[1:1]
  1 │ schema {
    · ──────
@@ -6,7 +6,7 @@
    ╰────
 
 
-  × Cannot set schema mutation root to unknown type type `WhatdoIknow`
+  × Cannot set schema mutation root to unknown type `WhatdoIknow`
    ╭─[1:1]
  1 │ schema {
    · ──────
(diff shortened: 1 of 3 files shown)


fbjork (approved):

engine/crates/validation/tests/validation_errors/unknown_type_in_schema_def.errors.txt
```diff
@@ -1,20 +1,20 @@
-  × Cannot set schema query root to unknown type type `Dunno`
+  × Cannot set schema query root to unknown type `Dunno`
    ╭─[1:1]
  1 │ schema {
    · ──────
  2 │     query: Dunno
    ╰────
```
obmarg, line 6: Unrelated to this PR, but the span we've chosen for this error message is a bit confusing - should we not be underlining `Dunno`?  (Potentially with a label if we wanted to make it even more helpful)

tomhoule: True, opening a linear issue to do it when there is time for a small issue

tomhoule: Ah I think I maybe remember that these don't have spans in the async graphql parser AST. But I'll double check.

obmarg: That would explain it if so - odd choice from `async_graphql` though, I usually remember having to do `.node` about 1000 times to get through all the `Positioned<T>` it uses

tomhoule: I know the pain :p

PR #1310, opened 2024-02-01
GB-5915 - A few fixes and improvements for Bun

yoav-lavi:
# Description

## Fixes

- Adds a few miscellaneous fixes around Bun and tests
- Updates Bun

# Type of change

- [ ] 💔 Breaking
- [ ] 🚀 Feature
- [x] 🐛 Fix
- [ ] 🛠️ Tooling
- [ ] 🔨 Refactoring
- [ ] 🧪 Test
- [ ] 📦 Dependency


cli/crates/server/src/bridge/udf.rs
```diff
@@ -211,12 +211,24 @@ impl UdfRuntime {
...
+                .try_next()
+                .await
+                .ok()
+                .flatten()
+                .ok_or(lines_skipped_over)
+        };
+        let bound_port = match bound_port {
+            Ok(port) => port,
+            Err(skipped_over_lines) => {
+                let outcome = bun.wait_with_output().await.unwrap();
```
yoav-lavi, line 224: I currently kept this as `unwrap` to match stdout but we should turn these into `expect`

yoav-lavi: This isn't `map_err` due to the `await` here

tomhoule (approved):
nice!

PR #1319, opened 2024-02-05
Fix Bun flakiness issue and remove additional build

yoav-lavi:
# Description

## Fixes

- Fix Bun flakiness issue
- Remove additional Bun build
- Update Bun
- Remove a few unnecessary file copies and writes

# Type of change

- [ ] 💔 Breaking
- [ ] 🚀 Feature
- [x] 🐛 Fix
- [ ] 🛠️ Tooling
- [ ] 🔨 Refactoring
- [ ] 🧪 Test
- [ ] 📦 Dependency

### DIFF
--- a/cli/crates/server/src/udf_builder.rs
+++ b/cli/crates/server/src/udf_builder.rs
@@ -1,7 +1,9 @@
 use common::environment::Environment;
 use common::types::UdfKind;
 
+use futures_util::FutureExt;
 use itertools::Itertools;
+use std::panic::RefUnwindSafe;
 use std::path::{Path, PathBuf};
 use std::process::Stdio;
 use tokio::io::AsyncWriteExt;
@@ -198,6 +200,12 @@ pub(crate) async fn build(
         .map_err(|err| UdfBuildError::CreateNotWriteToTemporaryFile(temp_file_path.to_path_buf(), err))?;
 
     let entrypoint_js_path = dist_path.join(ENTRYPOINT_SCRIPT_FILE_NAME);
+
+    temp_file
+        .flush()
+        .await
+        .map_err(|err| UdfBuildError::CreateNotWriteToTemporaryFile(temp_file_path.to_path_buf(), err))?;
+
     tokio::fs::copy(temp_file_path, &entrypoint_js_path)
         .await
         .map_err(|err| UdfBuildError::CreateUdfArtifactFile(entrypoint_js_path.clone(), udf_kind, err))?;
--- a/cli/crates/server/src/udf_builder.rs
+++ b/cli/crates/server/src/udf_builder.rs
@@ -210,6 +210,8 @@ pub(crate) async fn build(
         .await
         .map_err(|err| UdfBuildError::CreateUdfArtifactFile(entrypoint_js_path.clone(), udf_kind, err))?;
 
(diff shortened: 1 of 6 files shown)


fbjork (approved):

cli/crates/server/src/udf_builder.rs
```diff
@@ -253,30 +186,38 @@ pub(crate) async fn build(
...
+        .map_err(|err| UdfBuildError::CreateNotWriteToTemporaryFile(temp_file_path.to_path_buf(), err))?;
+    temp_file
+        .write_all(&tokio::fs::read(&udf_build_entrypoint_path).await.expect("must succeed"))
+        .await
+        .map_err(|err| UdfBuildError::CreateNotWriteToTemporaryFile(temp_file_path.to_path_buf(), err))?;
+
     let entrypoint_js_path = dist_path.join(ENTRYPOINT_SCRIPT_FILE_NAME);
+
+    temp_file
+        .flush()
```
obmarg, line 191: Just reading the docs on `fs::File`:

> Note that this [flush] does not ensure that the file has been fully written to disk; the operating system might keep the changes around in an in-memory buffer. See the [sync_all](https://docs.rs/tokio/latest/tokio/fs/struct.File.html#method.sync_all) method for telling the OS to write the data to disk.

Might want to do a sync_all?

yoav-lavi: >pub async fn [sync_all](https://docs.rs/tokio/latest/tokio/fs/struct.File.html#method.sync_all)(&self) -> [Result](https://doc.rust-lang.org/nightly/std/io/error/type.Result.html)<[()](https://doc.rust-lang.org/nightly/std/primitive.unit.html)>
Attempts to sync all OS-internal metadata to disk.
This function will attempt to ensure that all in-core data reaches the filesystem before returning.

I was looking at that but the `tokio` version claims it's metadata only

obmarg: Although reading a bit more it seems like flush might be enough for it to be visible elsewhere - maybe not super important

yoav-lavi: Although it's super weird that this happened on this PR, since this issue should have existed before. I think it's a race condition that's dependent on the file size so before without any dependencies it wasn't reproducible

yoav-lavi: I'm also not entirely sure we need this temp file rather than writing it directly to the resolver dir. @jakubadamw what was the original reasoning for it?

yoav-lavi: Theoretically the first write could also be removed since we write the file and then read it again

obmarg: > I was looking at that but the tokio version claims it's metadata only

I think that might be bad documentation, because I copied my quote above from the tokio docs (and there's also a `sync_data` function that claims to be a subset of `sync_all` "without syncing metadata".

yoav-lavi: Removed the unnecessary writes

obmarg: > Any reason to add it though?

Probs not

PR #1326, opened 2024-02-06
various small engine-v2 optimizations

Finistere:
Made various quick small improvements from a profile. Added a solid 5% performance with wrapping bein Copy.
Mostly avoid some unecessary work.

- renamed `AttributionLogic` to `PlanningLogic` and `field_attribution` to `bound_field_to_plan_id`
- using a mutable Vec to keep track of the response path rather than `im::Vector`
- factorized `CollectedSelectionSetSeed` to be used for both the planning and runtime generated ones.
- avoid adding `__typename` in the query when not necessary and simplified ObjectId detection
- bunch of micro improvements
- added an actual error test for federation


engine/crates/engine-v2/src/response/path.rs
```diff
@@ -167,15 +177,17 @@ impl std::ops::Index<BoundResponseKey> for ResponseKeys {
     type Output = str;
 
     fn index(&self, key: BoundResponseKey) -> &Self::Output {
-        self.0.resolve(&ResponseKey::from(key))
+        // SAFETY: We only generate a BoundResponseKey for a key that exist.
+        unsafe { self.0.resolve_unchecked(&ResponseKey::from(key)) }
```
obmarg, line 181: Chatted about this on slack, seems risky for now

engine/crates/engine-v2/src/response/write/deserialize/selection_set/collected.rs
```diff
@@ -6,20 +6,50 @@ use std::{
...
+    request::SelectionSetType,
     response::{
         write::deserialize::{key::Key, FieldSeed, SeedContextInner},
-        ResponseBoundaryItem, ResponseEdge, ResponseObject, ResponsePath, ResponseValue,
+        ResponseBoundaryItem, ResponseEdge, ResponseObject, ResponseValue,
     },
 };
 
-pub(crate) struct ConcreteCollectionSetSeed<'ctx, 'parent> {
+pub(crate) struct CollectedSelectionSetSeed<'ctx, 'parent> {
```
obmarg, line 17: What does collected mean in this context?

obmarg (approved):

PR #1333, opened 2024-02-07
feat: support nested joins in engine-v1

obmarg:
We added joins a while back, but one key limitation was that you could only join onto top level fields: `user(id: $id)` and not nested fields: `contentful { user(id: $id) }`.  This makes joins useless if you're using namespaces in our connectors, which is quite a big limitation.

This PR removes that limitation. You can now select a field nested to any depth using normal GraphQL syntax, provided the query makes sense (i.e. it must be a path to a single field, you can't have any list fields in the path, all required arguments must be provided).

This involves a breaking change to the registry, so I'm going to need to bump the gateway version when its merged there.

Part of GB-5335


engine/crates/engine/src/context/field.rs
```diff
@@ -80,6 +80,21 @@ impl<'a> ContextField<'a> {
         }
     }
 
+    pub fn with_joined_selection_set(&self, selection_set: &'a Positioned<SelectionSet>) -> ContextSelectionSet<'a> {
```
obmarg, line 83: think this was pointless in the end

engine/crates/engine/src/resolver_utils/field.rs
```diff
@@ -64,6 +64,57 @@ pub async fn resolve_field(
...
+    //      let current_value = run_field_resolver(resolver, last_value);
+    //      let context = dunno_but_do_this_somehow()
+    //      let last_value = current_value;
+    //   }
+    // }
+    //
+    // at the end of the loop current_value _should_ be the value we use for the current field.
+    // (e.g. the return value of run_field_resolver, assuming we were able to implement
+    // this in there (can we - there's a question?))
+
```
obmarg, line 115: Get rid of this shit

obmarg:
Note to self (and others): this breaks the registry so will require a bump when pulled into `api`.  I'll try and do immediately after merging.

engine/crates/engine/src/resolver_utils/joins.rs
```diff
@@ -0,0 +1,165 @@
...
+    let mut resolved_value = ResolvedValue::default();
+    let mut current_type = ctx
+        .schema_env()
+        .registry
+        .root_type(engine_parser::types::OperationType::Query);
+
+    while let Some((name, _)) = field_iter.next() {
+        let meta_field = current_type
+            .field(name)
+            .ok_or_else(|| Error::new(format!("Internal error: could not find joined field {}", &name)))?;
```
tomhoule, line 35: nitpick, but it would be nice to have the parent type name in the error (`<current_type>.<field_name>`)

tomhoule (approved):
:100:

engine/crates/engine/src/resolver_utils/joins.rs
obmarg: Yeah, true.   This error shouldn't _really_ happen but I'll improve the error anyway (wouldn't be the first time something that shouldn't happen does actually happen in the end)

engine/crates/parser-sdl/src/validations.rs
```diff
@@ -144,44 +145,114 @@ fn validate_join(
...
+
+        for (name, argument) in &field.args {
+            if argument.ty.is_non_null() && !join_arguments.contains_argument(name) {
+                errors.push(RuleError::new(
+                    vec![],
+                    format!(
+                        "{coord} is trying to join with {}.{}, but does not provide the non-nullable argument {name}",
+                        current_type.name(),
+                        &field.name,
+                    ),
```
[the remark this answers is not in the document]
obmarg: Adding a little bit of validation for this - don't want to be too strict (still punting on type checking for example) but think rejecting stuff that doesn't exist is a good idea

PR #1334, opened 2024-02-07
nix: fix cli derivation

tomhoule:
udf wrapper needs to be packaged

### DIFF
--- a/cli/nix/udf-wrapper.nix
+++ b/cli/nix/udf-wrapper.nix
@@ -1,0 +1,18 @@
+{ pkgs, ... }:
+
+{
+  packages.udf-wrapper = pkgs.buildNpmPackage {
+    src = ../udf-wrapper;
+    name = "udf-wrapper";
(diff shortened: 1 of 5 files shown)


cli/udf-wrapper/package-lock.json
tomhoule, line 1: switch from package-lock v3 to v2 (tooling hasn't caught up to v3 yet)

cli/nix/udf-wrapper.nix
```diff
@@ -0,0 +1,18 @@
+{ pkgs, ... }:
+
+{
+  packages.udf-wrapper = pkgs.buildNpmPackage {
```
Finistere, line 4: Does it use the build script defined in `package.json`?

tomhoule: I think it does, yes

Finistere (approved):
Nice work!

PR #1335, opened 2024-02-07
feat: APQ in engine-v1

Finistere:
Implementing APQ: https://www.apollographql.com/docs/apollo-server/performance/apq/


cli/crates/cli/tests/apq.rs
```diff
@@ -0,0 +1,143 @@
...
+    let execute = |query: &'static str, extensions: &Value| {
+        let request = client.gql::<Value>(query).extensions(extensions).into_reqwest_builder();
+        async move { request.send().await.unwrap().json::<Value>().await.unwrap() }
+    };
+
+    let apq_ext = serde_json::json!({
+        "persistedQuery": {
+        "version": 1,
+        "sha256Hash": sha256(query)
+    }
```
obmarg, line 60:
```suggestion
            "version": 1,
            "sha256Hash": sha256(query)
        }
```

cli/crates/cli/tests/apq.rs
```diff
@@ -0,0 +1,143 @@
...
+            todo(id: "1") {
+                id
+                title
+            }
+        }
+    "#;
+
+    let execute = |query: &'static str, extensions: &Value| {
+        let request = client.gql::<Value>(query).extensions(extensions).into_reqwest_builder();
+        async move { request.send().await.unwrap().json::<Value>().await.unwrap() }
```
obmarg, line 53: There a reason this can't just be `client.gql::<Value>(query).extension(extensions).await.unwrap()`?

engine/crates/engine/src/request.rs
```diff
@@ -34,19 +34,38 @@ pub struct Request {
...
+    pub custom: HashMap<String, Value>,
+}
+
+#[serde_with::serde_as]
+#[derive(Debug, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct PersistedQueryRequestExtension {
+    pub version: u32,
+    #[serde_as(as = "serde_with::hex::Hex")]
+    pub sha256_hash: Vec<u8>,
```
obmarg, line 55: Could have encoded this as a versioned enum.  No big deal though

obmarg (approved):

engine/crates/engine/src/request.rs
Finistere: I tried, but versioned enum only work with strings, not integers :/
https://github.com/serde-rs/serde/issues/745

cli/crates/cli/tests/apq.rs
Finistere: assert_json_snapshot expect a `impl Serialize` which the `reqwest::Response` isn't

engine/crates/engine/src/request.rs
Finistere: could make a custom Deserialize, but felt way overkill :D

obmarg: Ah ok, never mind then

PR #1340, opened 2024-02-09
GB-5690 - Configurable introspection

yoav-lavi:
# Description

## Features

- Supports enabling or disabling introspection by config

# Type of change

- [ ] 💔 Breaking
- [x] 🚀 Feature
- [ ] 🐛 Fix
- [ ] 🛠️ Tooling
- [ ] 🔨 Refactoring
- [ ] 🧪 Test
- [ ] 📦 Dependency


packages/grafbase-sdk/src/config.ts
```diff
@@ -81,15 +88,19 @@ export class GraphConfig {
       : ''
     const cache = this.cache ? this.cache.toString() : ''
     const experimental = this.experimental ? this.experimental.toString() : ''
+    const introspection = this.introspection
+      ? `extend schema @introspection(enable: true)\n\n`
+      : `extend schema @introspection(enable: false)\n\n`
```
yoav-lavi, line 93: Debating whether always to add this or only if explicitly set

cli/crates/server/src/config/mod.rs
```diff
@@ -171,6 +176,14 @@ async fn parse_and_generate_config_from_ts(ts_config_path: &Path) -> Result<Stri
     validate_node().await?;
     let node_command = Command::new("node")
         .args(args)
+        .env(
+            "GRAFBASE_ENV",
+            match environment_name {
```
yoav-lavi, line 181: TODO make this an imp

engine/crates/engine-v2/config/src/lib.rs
```diff
@@ -29,6 +30,8 @@ pub enum VersionedConfig {
     V2(v2::Config),
     /// V3 is like V2 but with FederatedGraphV2
     V3(v3::Config),
+    /// V4 is like V3 but with disable_introspection
```
Finistere, line 33: You don't need to add a `V4` for that, it's not a breaking change. Just add a `#[serde(default)]` or `#[serde(default = "func")]` and you're good ;)

engine/crates/engine-v2/src/request/mod.rs
```diff
@@ -102,25 +102,31 @@ impl Operation {
...
                 .map(|cache_config_id| schema[cache_config_id]);
-
-            let selection_set_cache_config = operation
+            let selection_set = operation
                 .walker_with(schema.walker())
-                .walk(operation.root_selection_set_id)
-                .cache_config();
+                .walk(operation.root_selection_set_id);
+
+            if let Some(location) = selection_set.find_introspection_field_location() {
```
Finistere, line 123: conditions could be reversed, no need to search for location if it's not disabled^^

yoav-lavi: Done

engine/crates/engine-v2/config/src/lib.rs
yoav-lavi: Done

cli/crates/server/src/config/mod.rs
```diff
@@ -171,6 +176,14 @@ async fn parse_and_generate_config_from_ts(ts_config_path: &Path) -> Result<Stri
...
+                EnvironmentName::Production => "production",
```
fbjork, line 182: missing Preview envs

yoav-lavi: Those don't exist for the moment in the CLI

yoav-lavi: And this specific distinction is only for `dev` vs `prod` but we could add them

Finistere (commented):
Nice work! Some tests would be great though, especially with the logic based on the environment in the CLI. An integration test for the engine v2 would also be nice (based on the schema config) like here for auth: https://github.com/grafbase/grafbase/blob/main/engine/crates/integration-tests/tests/federation/auth/jwt.rs#L17

engine/crates/engine/src/request.rs
```diff
@@ -11,6 +11,15 @@ use crate::{Data, ParseRequestError, UploadValue, Value, Variables};
 mod query;
 pub use query::QueryParamRequest;
 
+#[derive(Serialize, Deserialize, Default, PartialEq, Eq, Debug, Clone, Copy, Hash)]
+#[serde(rename_all = "camelCase")]
+pub enum IntrospectionState {
```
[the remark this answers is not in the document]
yoav-lavi: I agree with your earlier comment, I think it's clearer when using it if it's an enum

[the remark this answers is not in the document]
Finistere: not really, both are fine IMHO. It mostly just comes down to how it would feel in terms of code. When specifying `override` it feels natural to me to have an `Option<bool>` here but it's a long struct field name. Using an option would allow you to use something like `request.is_introspection_allowed_override.unwrap_or(config.is_introspection_disabled)` instead of a match. But a match is also fine

PR #1344, opened 2024-02-11
chore: Add default values in introspection

Finistere:
Adding `InputValues` to store default input values & directive arguments in
the schema and later re-use the same type for input values for an operation.

As those are stored with the `InputValueDefinitionId` and `EnumValueId`
rather than strings this will allow an engine-v2 Resolver to be entirely
unaware of schema names, meaning we could rename any part of the schema.

And also adds the default values in the introspection now :D

### DIFF
--- a/engine/crates/engine-v2/schema/Cargo.toml
+++ b/engine/crates/engine-v2/schema/Cargo.toml
@@ -16,6 +16,12 @@ strum_macros.workspace = true
 url.workspace = true
 fnv = "1"
 indexmap.workspace = true
+serde.workspace = true
+thiserror.workspace = true
 
 config = { package = "engine-v2-config", path = "../config" }
 federated-graph = { package = "graphql-federated-graph", path = "../../federated-graph" }
+
+[dev-dependencies]
+serde_json.workspace = true
+insta.workspace = true
--- a/engine/crates/engine-v2/schema/src/ids.rs
+++ b/engine/crates/engine-v2/schema/src/ids.rs
@@ -1,19 +1,21 @@
+use std::ops::Range;
+
 /// Isolating ids from the rest to prevent misuse of the NonZeroU32.
 /// They can only be created by From<usize>
 use crate::{
     sources::federation::{DataSource as FederationDataSource, Subgraph},
-    CacheConfig, Definition, Directive, Enum, EnumValue, Field, Header, InputObject, InputValue, Interface, Object,
-    Resolver, Scalar, Schema, Type, Union,
+    CacheConfig, Definition, Directive, Enum, EnumValue, Field, Header, InputObject, InputValueDefinition, Interface,
+    Object, Resolver, Scalar, Schema, Type, Union,
 };
 use url::Url;
 
 /// Reserving the 4 upper bits for some fun with bit packing. It still leaves 268 million possible values.
 /// And it's way easier to increase that limit if needed that to reserve some bits later!
 /// Currently, we use the two upper bits of the FieldId for the ResponseEdge in the engine.
-const MAX_ID: usize = (1 << 29) - 1;
+pub(crate) const MAX_ID: usize = (1 << 29) - 1;
 
 macro_rules! id_newtypes {
-    ($($ty:ident.$field:ident[$name:ident] => $out:ident unless $msg:literal,)*) => {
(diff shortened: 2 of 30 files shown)


engine/crates/engine-v2/schema/Cargo.toml
```diff
@@ -16,6 +16,12 @@ strum_macros.workspace = true
...
 indexmap.workspace = true
+serde.workspace = true
+thiserror.workspace = true
 
 config = { package = "engine-v2-config", path = "../config" }
 federated-graph = { package = "graphql-federated-graph", path = "../../federated-graph" }
+
+[dev-dependencies]
+serde_json.workspace = true
+insta = { version = "1", features = ["json"] }
```
obmarg, line 27: Unimportant, but this seems like it should be a workspace dep as well if it's not already

engine/crates/engine-v2/schema/src/ids.rs
```diff
@@ -43,6 +45,14 @@ macro_rules! id_newtypes {
...
             }
 
+            impl std::ops::IndexMut<crate::ids::IdRange<$name>> for $ty {
+                fn index_mut(&mut self, range: crate::ids::IdRange<$name>) -> &mut Self::Output {
+                    let crate::ids::IdRange { start, end } = range;
+                    let start = usize::from(start);
+                    let end = usize::from(end);
+                    &mut self.$field[start..end]
+                }
+            }
```
obmarg, line 55: What's this being used for?  Wouldn't have thought we should be editing the schema generally....

engine/crates/engine-v2/schema/src/input_value/de.rs
```diff
@@ -0,0 +1,88 @@
...
+                Ok(map)
+            }
+        }
+    }
+
+    forward_to_deserialize_any! {
+        bool i8 i16 i32 i64 i128 u8 u16 u32 u64 u128 f32 f64 char str string
+        bytes byte_buf option unit unit_struct newtype_struct seq tuple
+        tuple_struct map struct enum identifier ignored_any
+    }
```
obmarg, line 76: I faintly recall running into problems doing `forward_to_deserialize_any` like this on option in the past, as some Deserialize impl was expecting a call to `visit_some` on the `Some` path.

Might be worth a unit test to check its ok?

engine/crates/engine-v2/schema/src/input_value/display.rs
```diff
@@ -0,0 +1,98 @@
+use std::fmt::{Display, Formatter, Result, Write};
+
+use crate::{InputValue, InputValuesContext};
+
+pub struct DisplayableInpuValue<'ctx, Str, Ctx> {
```
obmarg, line 5: What format is this printing to?  JSON?  GraphQL syntax?  Arbitrary strings for a human to read?

Could use a comment to say.

obmarg (approved):

engine/crates/engine-v2/schema/Cargo.toml
Finistere: agreed, was lazy, but adding it

engine/crates/engine-v2/schema/src/ids.rs
Finistere: We might not in this PR, but I'm sorting the enum values by their name in a further PR. Same for field arguments. Currently the federated graph doesn't guarantee that, should likely be changed at some point.

engine/crates/engine-v2/schema/src/input_value/display.rs
Finistere: ah yes fair point, it's GraphQL syntax

engine/crates/engine-v2/schema/src/input_value/de.rs
Finistere: I've added a test with a struct here: https://github.com/grafbase/grafbase/pull/1344/files/fd1b73a719fb1b3449dda70caa1ca8a73fd99e9e#diff-ccc4120bf5ff96d2a86233a1448eb1debb054c1d8c469870cef9b6c1d2ecbfefR313

Will add an `Option<String>` in a case where there is actually a string

PR #1351, opened 2024-02-13
fix: Deserialize properly GET requests for the engine

Finistere:
GET request have a specific format, values are encoded in JSON, but the root fields are url
 encoded. To ensure that the HTTP framework doesn't mess with it, we're first trying to
 deserialize all values as strings.

 ```sh
 curl --get http://localhost:4000/graphql \
   --header 'content-type: application/json' \
   --data-urlencode 'query={__typename}' \
   --data-urlencode 'extensions={"persistedQuery":{"version":1,"sha256Hash":"ecf4edb46db40b5132295c0291d62fb65d6759a9eedfa4d5d612dd5ec54a6b38"}}'
 ```

### DIFF
--- a/cli/crates/gateway/src/serving.rs
+++ b/cli/crates/gateway/src/serving.rs
@@ -1,4 +1,4 @@
-use std::collections::HashMap;
+use std::{collections::HashMap, sync::Arc};
 
 use axum::{
     extract::{Query, State},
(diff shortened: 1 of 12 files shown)


tomhoule (approved):
Makes sense

engine/crates/engine/src/request/query.rs
```diff
@@ -0,0 +1,77 @@
...
+impl<'de> serde::Deserialize<'de> for QueryParamRequest {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: Deserializer<'de>,
+    {
+        let query_params = QueryParams::deserialize(deserializer)?;
+        // Avoid unecessary serde round trips by creating a JSON string manually.
+        // This also ensures we stay as close as possible from the original Request deserialization
+        // behavior.
+        let mut request: Request = serde_json::from_str(&query_params.variables_and_extensions_as_json_string())
```
obmarg, line 34: Not sure how I feel about this...

Finistere: 😅

cli/crates/gateway/src/serving.rs
```diff
@@ -56,31 +56,42 @@ async fn post_graphql(
...
+            .get(X_API_KEY_HEADER)
+            .and_then(|value| value.to_str().ok().map(ToString::to_string))
+            .or_else(|| params.x_api_key.clone()),
+        authorization_header: headers
+            .get(AUTHORIZATION_HEADER)
+            .and_then(|value| value.to_str().ok().map(ToString::to_string))
+            .or_else(|| params.authorization.clone()),
+        headers,
+        wait_until_sender: sender,
+    });
```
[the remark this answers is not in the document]
Finistere: `Context::new()` expected all the params in a hashmap, could just create one though here.

PR #1354, opened 2024-02-13
feat: graphql batching in cli standalone graphs

obmarg:
Implements [this RFC](https://github.com/graphql/graphql-over-http/blob/main/rfcs/Batching.md) for standalone graphs in the CLI.

The RFC is pretty silent on whether requests should be run in parallel or in series.  Even the implementations linked to from the RFC can't agree: half of them do things in parallel, half in series.  I've settled on doing things serially for now as it's easy and definitely won't break anything.  Can revisit later if we need to.

Part of GB-5931

### DIFF
--- a/cli/crates/cli/tests/utils/async_client.rs
+++ b/cli/crates/cli/tests/utils/async_client.rs
@@ -56,7 +56,6 @@ impl AsyncClient {
         self
     }
 
-    // TODO: update this one as well...
     pub fn gql<Response>(&self, query: impl Into<String>) -> GqlRequestBuilder<Response>
     where
         Response: serde::de::DeserializeOwned + 'static,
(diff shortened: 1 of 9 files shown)


cli/crates/cli/tests/utils/async_client.rs
```diff
@@ -190,6 +189,31 @@ impl AsyncClient {
...
             .unwrap()
     }
+
+    /// Makes a batch GraphQL request.
+    ///
+    /// At the moment this is way less functional than the non-batch request builder
+    /// but is enough to test.
+    pub async fn batch_gql<T>(&self, queries: impl IntoIterator<Item = T>) -> serde_json::Value
+    where
+        T: AsRef<str>,
```
Finistere, line 199: Could have been `Into<GqlRequest>` otherwise

Finistere (approved):

cli/crates/cli/tests/utils/async_client.rs
obmarg: Oh, I don't think I had that when I was writing this

PR #1361, opened 2024-02-15
cli: avoid crashing when stdout becomes unavailable in introspect

tomhoule:
This is the well known issue of `print!()` & co from the standard library panicking if stdout/stderr becomes unavailable. This commit fixes the issue in `grafbase introspect`.

Issue illustrated:

The panic message isn't shown anymore after this commit because introspect handles the error gracefully.

Also, update the npm deps hash in udf-wrapper.nix so `nix build .#cli` works again. 

closes GB-5788


yoav-lavi (approved):

obmarg (approved):
Sweet

PR #1367, opened 2024-02-18
feat: Validate all inputs part 2

Finistere:
This PR only includes the engine-v2 changes on top of the part 1.

With the next PR it allows us to fully isolate:
- all the GraphQL-specific input processing: default values, undefined variables, coercion
- schema names from connector/data sources names. Virtually we can
  trivially rename anything now as we keep track of input value
  definition ids and enum value ids. And executors have everything
  automatically renamed by the provided SchemaWalker.

Input values, which will be accessed through a `PlanInputValue`, can be
used through:
- Serialize to serialize the data immediately
- Deserializer to map arguments to a custom type
- Display to format the value with GraphQL syntax
- InputValue<'_> which provides a friendlier format than RawInputValue
  which also abstracts away the differences between schema default
  values and operation input values.

In all cases, the values have always already been coerced and validated.


engine/crates/engine-v2/src/plan/walkers/input_value.rs
```diff
@@ -0,0 +1,133 @@
+use schema::{InputValue, InputValueSerdeError, RawInputValuesContext, SchemaWalker};
```
pimeys, line 1: This could benefit some quick unit tests to show what gets deserialized etc.

engine/crates/engine-v2/src/request/bind/coercion/const_value.rs
```diff
@@ -0,0 +1,291 @@
...
+
+        match ty.inner {
+            Definition::Scalar(scalar) => self.coerce_scalar(self.schema.walk(scalar), value),
+            Definition::Enum(r#enum) => self.coerce_enum(self.schema.walk(r#enum), value),
+            Definition::InputObject(input_object) => self.coerce_input_objet(self.schema.walk(input_object), value),
+            _ => unreachable!("Cannot be an output type."),
+        }
+    }
+
+    fn coerce_input_objet(
```
pimeys, line 124: typo?

engine/crates/engine-v2/src/request/walkers/variables.rs
```diff
@@ -1,29 +0,0 @@
...
-                ListWrapping::NullableList => format!("[{name}]"),
-            }
-        }
-        name
-    }
-
-    pub fn default_value(&self) -> Option<&ConstValue> {
-        self.as_ref().default_value.as_ref()
-    }
-}
```
pimeys, line 29: :+1:

engine/crates/integration-tests/tests/federation/basic/variables.rs
```diff
@@ -192,32 +192,32 @@ fn list_coercion() {
 fn invalid_ints() {
     insta::assert_json_snapshot!(error_test("int", "Int!", 1.5), @r###"
     [
-      "Variable $input got an invalid value: found a Float value where we expected a Int scalar at $input"
+      "Variable $input has an invalid value. Found value 1.5 which cannot be coerced into a Int scalar"
```
pimeys, line 195:
```suggestion
      "Variable $input has an invalid value. Found value 1.5 which cannot be coerced into an Int scalar"
```

et.al. for the rest... Annoying yes, but it's a public error :)

pimeys (approved):
Nice work. Maybe @tomhoule could have a read too?

engine/crates/engine-v2/src/plan/walkers/input_value.rs
Finistere: You mean to check serialize, deserialize, etc?
I have some sanity unit-tests here: https://github.com/grafbase/grafbase/blob/81f1bca1bc5c86c07830d8ace8dbaa002612a5bf/engine/crates/engine-v2/schema/src/input_value/raw.rs#L255
in the previous PR

engine/crates/integration-tests/tests/federation/basic/variables.rs
Finistere: yeah well, giving up on this one. Too annoying for me and tons of places where this would be needed even with enum names.

PR #1372, opened 2024-02-19
engine-v2-schema: remove inaccessible items

tomhoule:
Since they are not part of the API of the engine, they should be excluded as early as possible. They can fall through the cracks if they are included in the schema, so we remove them at conversion time.

The approach taken is to remove inaccessible output fields, input value definitions, union members and enum values. Since inaccessible types cause the fields returning them to also be inaccessible (composition rule), that automatically deals with inaccessible types.

closes GB-5949

### DIFF
--- a/engine/crates/engine-v2/schema/src/builder/ids.rs
+++ b/engine/crates/engine-v2/schema/src/builder/ids.rs
@@ -6,14 +6,6 @@
 
 use crate::ids::*;
 
-pub(super) trait MappableId: Copy {
-    type Counterpart: Copy + From<usize>;
-
-    fn get_idx(self) -> usize;
-    fn get_map(mapper: &IdMapper) -> &[usize];
-    fn get_map_mut(mapper: &mut IdMapper) -> &mut Vec<usize>;
-}
-
 macro_rules! mapped_ids {
     ($($name:ident : $from:ty => $to:ty)*) => {
 
@@ -26,20 +18,24 @@ macro_rules! mapped_ids {
         }
 
         $(
-            impl MappableId for $from {
-                type Counterpart = $to;
+            #[allow(unused)]
+            pub(super) mod $name {
+                use super::*;
 
-                fn get_idx(self) -> usize {
-                    self.0
+                /// Mark an id as skipped in the target schema. The element has to be actually filtered out, separately.
+                pub(crate) fn skip(mapper: &mut IdMapper, id: $from) {
+                    super::skip_impl(&mut mapper.$name, id.0)
                 }
 
-                fn get_map(mapper: &IdMapper) -> &[usize] {
-                    &mapper.$name
+                /// Map a federated_graph id to an engine_schema id taking the skipped IDs into account.
+                pub(crate) fn map(mapper: &IdMapper, id: $from) -> Option<$to> {
+                    super::map_impl(&mapper.$name, id.0)
                 }
 
-                fn get_map_mut(mapper: &mut IdMapper) -> &mut Vec<usize> {
-                    &mut mapper.$name
+                pub(crate) fn map_range(mapper: &IdMapper, range: ($from, usize)) -> IdRange<$to> {
+                    super::map_range_impl(&mapper.$name, (range.0.0, range.1))
                 }
+
             }
         )*
     };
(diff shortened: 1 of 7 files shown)


engine/crates/engine-v2/schema/src/builder/ids.rs
```diff
@@ -0,0 +1,186 @@
...
+        let skipped = Id::get_map(self);
+
+        // How many ids were skipped before the range.
+        let skipped_before = skipped.partition_point(|skipped| *skipped < start_idx);
+
+        // How many ids were skipped inside the range.
+        let skipped_inside = skipped[skipped_before..]
+            .iter()
+            .take_while(|skipped| **skipped < (start_idx + len))
+            .count();
```
tomhoule, line 94: We can do a binary search here too, theoretically the complexity would be lower but I'd bet iterating and counting is going to be a lot faster in all realistic cases.

engine/crates/engine-v2/schema/src/builder/ids.rs
```diff
@@ -0,0 +1,186 @@
...
+            mapper.map_range(range)
+        );
+    }
+
+    #[test]
+    #[should_panic(expected = "assertion failed: last_entry < idx")]
+    fn skip_out_of_order() {
+        let mut mapper = IdMapper::default();
+        mapper.skip(federated_graph::InputValueDefinitionId(5));
+        mapper.skip(federated_graph::InputValueDefinitionId(5)); // boom
```
tomhoule, line 184: Maybe we should add 3 here too

engine/crates/integration-tests/tests/federation/introspection.rs
```diff
@@ -819,7 +819,6 @@ fn introspection_on_multiple_federation_subgraphs() {
     }
 
     type Picture {
-      altText: String!
```
tomhoule, line 822: This one is `@inaccessible`, so it's expected that it doesn't show up in introspection anymore.

engine/crates/engine-v2/schema/src/builder/ids.rs
```diff
@@ -0,0 +1,196 @@
...
+        entries.push(idx);
+    }
+
+    /// Map a federated_graph id to an engine_schema id taking the skipped IDs into account.
+    pub(super) fn map<Id: MappableId>(&self, id: Id) -> Option<Id::Counterpart> {
+        let idx = id.get_idx();
+        let map = Id::get_map(self);
+        let skipped = map.partition_point(|skipped| *skipped <= idx);
+
+        if let Some(last) = map[..skipped].last().copied() {
```
Finistere, line 73: so `map.get(skipped - 1)`?

engine/crates/engine-v2/schema/src/builder/ids.rs
```diff
@@ -0,0 +1,196 @@
+//! This module is responsible for mapping identifiers between federated schema and the engine v2
+//! schema when the mapping is not 1:1.
+//!
+//! As of 3b56f12c95d334ce6cb46f4f8654ce531a69f975, this only happens when @inaccessible items
+//! are removed.
+
+use crate::ids::*;
+
+pub(super) trait MappableId: Copy {
```
Finistere, line 9: IMHO this could probably be simpler with a few changes:
- You could avoid `get_idx` by implementing `usize: From<OriginId>` for the federated ids
- Instead of having an array of existing ids, I would use `Vec<Option<TargetId>` with the same size as the number of `OriginId`. So `fields_mapping = vec![None; federated_graph.fields.len()]` Since the ids are `NonZeroU32` in the schema it's free. It's simpler and smaller than a `[usize]`. You only need custom logic for the ranges.

With this you could just implement:
```
impl Index<OriginId> for IdMapper { type Target = Option<TargetId> }

impl Index<(OriginId, usize)> for IdMapper { type Target = IdRange<TargetId> }
```

and have a nicer API to work with later

engine/crates/engine-v2/schema/src/builder/mod.rs
```diff
@@ -88,23 +79,75 @@ impl SchemaBuilder {
...
+            })
+            .collect();
+
+        builder
     }
 
     fn insert_directives(&mut self, config: &mut Config) {
         let mut directives = Vec::with_capacity(config.graph.directives.len());
-        for directive in take(&mut config.graph.directives) {
+        for directive in &config.graph.directives {
```
Finistere, line 139: why not `take`?

Finistere (approved):
nice

engine/crates/engine-v2/schema/src/builder/ids.rs
Finistere: I don't think you would need a trait with this either, just a macro implementing it for all id pairs like almost everything related to ids would work imho

tomhoule: The idea behind this representation is that you have Vec's that will grow to `len() == inaccessible_count` instead of the same size as the vec of elements. But yeah, I'll try again (I tried this but the branch got messy with other things so ended up reverted), it's worth trying to simplify.

engine/crates/engine-v2/schema/src/builder/mod.rs
tomhoule: Since `insert_directives()` is called before we insert the object fields, we ended up not being able to know which object fields were inaccessible because the directives vec was empty. But now I grouped the inaccessible checks in `initialize()`, so it's likely we can revert this line back to `take()`, I'll try.

engine/crates/engine-v2/schema/src/builder/ids.rs
Finistere: ah, didn't understood it

Finistere: Ah smart. What's unclear is `get_map` and `get_map_mut` then. The variables names `skipped` felt weird indeed, make a lot more sense now.
I think you could still avoid the trait, but up to you.

I would rename `map` to `skipped_ids`, would be a lot clearer IMHO

engine/crates/engine-v2/schema/src/builder/ids.rs
```diff
@@ -0,0 +1,196 @@
...
+
+    pub(super) fn map_range<Id: MappableId>(&self, (start_id, len): (Id, usize)) -> crate::IdRange<Id::Counterpart>
+    where
+        usize: From<Id::Counterpart>,
+    {
+        let start_idx = start_id.get_idx();
+        let skipped = Id::get_map(self);
+
+        // How many ids were skipped before the range.
+        let skipped_before = skipped.partition_point(|skipped| *skipped < start_idx);
```
Finistere, line 90: would rename `skipped_ids_before_start`

engine/crates/engine-v2/schema/src/builder/ids.rs
```diff
@@ -0,0 +1,196 @@
...
+
+        // How many ids were skipped inside the range.
+        let skipped_inside = skipped[skipped_before..]
```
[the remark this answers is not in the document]
tomhoule: done

engine/crates/engine-v2/schema/src/builder/ids.rs
tomhoule: done

engine/crates/engine-v2/schema/src/builder/ids.rs
```diff
@@ -0,0 +1,179 @@
+//! This module is responsible for mapping identifiers between federated schema and the engine v2
+//! schema when the mapping is not 1:1.
+//!
+//! As of 3b56f12c95d334ce6cb46f4f8654ce531a69f975, this only happens when @inaccessible items
+//! are removed.
+
+use crate::ids::*;
+
+macro_rules! mapped_ids {
+    ($($name:ident : $from:ty => $to:ty)*) => {
```
Finistere, line 10: Probably you could have avoided the macro stuff by making it generic `IdMapper<Origin: Into<usize>, Target: From<usize>>` and just instantiating the right id mappers in the schema builder. A bit late as a comment though^^

engine/crates/engine-v2/schema/src/builder/mod.rs
```diff
@@ -103,8 +146,8 @@ impl SchemaBuilder {
                     name: name.into(),
                     arguments: {
                         let map = arguments
-                            .into_iter()
-                            .map(|(id, value)| (id.into(), self.insert_value(value)))
+                            .iter()
```
Finistere, line 149: could be `into_iter()` again then I think

Finistere (approved):
Nice!

engine/crates/engine-v2/schema/src/builder/ids.rs
tomhoule: Hmm true

engine/crates/engine-v2/schema/src/builder/mod.rs
tomhoule: fixed

PR #1373, opened 2024-02-19
fix: add `@link` into federation "enhanced" introspection

obmarg:


yoav-lavi (approved):
Is this not needed for the normal (non export) schema?

tomhoule (approved):

obmarg:
> Is this not needed for the normal (non export) schema?

Which schema do you mean?

yoav-lavi:
> Which schema do you mean? 

I think I'd assumed `export_sdl` is used only for the export command rather than introspection

obmarg:
> > Which schema do you mean?
> 
> I think I'd assumed `export_sdl` is used only for the export command rather than introspection

I think that particular code path is only hit on "enhanced" introspection and when running `grafbase introspect --dev` on a federated graph, and those are the only places where we'd want the link attribute.  So I think this is fine.

PR #1374, opened 2024-02-19
feat: Validate all inputs part 1

Finistere:
This PR only includes the engine-v2-schema changes which were
necessary to re-use `RawInputValues` for operation input values and
make them pleasant to work with.

With the next PR it allows us to fully isolate:
- all the GraphQL-specific input processing: default values, undefined variables, coercion
- schema names from connector/data sources names. Virtually we can
  trivially rename anything now as we keep track of input value
  definition ids and enum value ids. And executors have everything
  automatically renamed by the provided SchemaWalker.

Input values, which will be accessed through a `PlanInputValue`, can be
used through:
- Serialize to serialize the data immediately
- Deserializer to map arguments to a custom type
- Display to format the value with GraphQL syntax
- InputValue<'_> which provides a friendlier format than RawInputValue
  which also abstracts away the differences between schema default
  values and operation input values.

In all cases, the values have always already been coerced and validated.

### DIFF
--- a/engine/crates/engine-v2/schema/src/input_value/raw.rs
+++ b/engine/crates/engine-v2/schema/src/input_value/raw.rs
@@ -1,0 +1,522 @@
+use crate::{
+    EnumValueId, IdRange, InputValueDefinitionId, RawInputKeyValueId, RawInputObjectFieldValueId, RawInputValueId,
+    SchemaInputValueId,
+};
+
+/// Holds input values for the Schema and for an Operation during execution.
+/// It's generic over 'Str', the string representation:
+/// - StringId for the Schema (default input values and directive arguments)
+/// - Box<str> for the Operation
+///
+/// This allow us to share the code for:
+/// - Display: used to print default values in introspection and add input values in subgraph query strings.
+/// - Serialize: used to serialize variables/arguments.
+/// - Deserializer: used by resolvers to deserialize arguments into a specific struct.
+///
+/// Data is stored in flat arrays for hopefully faster serialization and as as bonus smaller InputValue
+/// footprint on x64 architectures: u64 + usize (padding) rather than 3 * usize (Box<[]> + padding).
+#[derive(Clone)]
+pub struct RawInputValues<Str> {
+    /// Inidividual input values and list values
+    pub(super) values: Vec<RawInputValue<Str>>,
+    /// InputObject's fields
+    pub(super) input_fields: Vec<(InputValueDefinitionId, RawInputValue<Str>)>,
+    /// Object's fields (for JSON)
+    pub(super) key_values: Vec<(Str, RawInputValue<Str>)>,
+}
+
+impl<Str> Default for RawInputValues<Str> {
+    fn default() -> Self {
+        Self {
+            // Reserve the first slot for undefined values.
+            // This allows us in the engine to create from scratch a PlanInputValue that is
+            // undefined. Used when requesting a field argument that wasn't provided.
+            values: vec![RawInputValue::Undefined],
+            input_fields: Default::default(),
+            key_values: Default::default(),
+        }
+    }
+}
+
+/// Represents any possible input value for field arguments, operation variables and directive
+/// arguments.
+/// Used for storage and as input format for operation input values to support default values,
+/// variables, etc.
+#[derive(Debug, Clone)]
+pub enum RawInputValue<Str> {
+    Null,
+    String(Str),
+    EnumValue(EnumValueId),
+    Int(i32),
+    BigInt(i64),
+    Float(f64),
+    Boolean(bool),
+    InputObject(IdRange<RawInputObjectFieldValueId<Str>>),
+    List(IdRange<RawInputValueId<Str>>),
+
+    /// for JSON
+    Map(IdRange<RawInputKeyValueId<Str>>),
+    U64(u64),
+
+    /// Directive arguments have unknown enum values. Likely doesn't make sense, but not sure.
+    UnknownEnumValue(Str),
+
+    /// Primary purpose is for operation to reference the value of a variable which is initialized
+    /// to the default or an arbitrary value first. It might also be used later to share the same
+    /// InputValue at different places.
+    Ref(RawInputValueId<Str>),
+
+    /// Used to reference default values for operation input values. It's tricky without as default
+    /// values also need to be taken into account for nested input object fields.
+    SchemaRef(SchemaInputValueId),
+
+    /// https://spec.graphql.org/October2021/#sec-Input-Objects.Input-Coercion
+    /// An input { a: $var, b: 123 } without any variable should be interpreted as { b: 123 } if a
+    /// is nullable.
+    /// However for an operation variable we reserve an InputValue slot for its future value.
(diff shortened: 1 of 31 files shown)


engine/crates/engine-v2/schema/src/input_value/raw.rs
```diff
@@ -0,0 +1,522 @@
...
+/// - StringId for the Schema (default input values and directive arguments)
+/// - Box<str> for the Operation
+///
+/// This allow us to share the code for:
+/// - Display: used to print default values in introspection and add input values in subgraph query strings.
+/// - Serialize: used to serialize variables/arguments.
+/// - Deserializer: used by resolvers to deserialize arguments into a specific struct.
+///
+/// Data is stored in flat arrays for hopefully faster serialization and as as bonus smaller InputValue
+/// footprint on x64 architectures: u64 + usize (padding) rather than 3 * usize (Box<[]> + padding).
```
obmarg, line 17: The last bit of this docstring seems like it either needs corrected or clarified?  I can't see how 3 Vecs would only have a single u64 of footprint?  I'm also not clear on what the alternative that has `3 * usize` is?  

Also why a whole usize of padding?  That seems sus to me.

engine/crates/engine-v2/schema/src/input_value/raw.rs
```diff
@@ -0,0 +1,522 @@
...
+    Boolean(bool),
+    InputObject(IdRange<RawInputObjectFieldValueId<Str>>),
+    List(IdRange<RawInputValueId<Str>>),
+
+    /// for JSON
+    Map(IdRange<RawInputKeyValueId<Str>>),
+    U64(u64),
+
+    /// Directive arguments have unknown enum values. Likely doesn't make sense, but not sure.
+    UnknownEnumValue(Str),
```
obmarg, line 62: This comment could do with an elaboration.  Why do directive arguments have unknown enum values?  What does it mean to be unknown.  In what way doesn't it make sense?

(I do think I know the answer to most of these questions, but others might not)

engine/crates/engine-v2/schema/src/input_value/raw.rs
```diff
@@ -0,0 +1,522 @@
...
+
+    /// Primary purpose is for operation to reference the value of a variable which is initialized
+    /// to the default or an arbitrary value first. It might also be used later to share the same
+    /// InputValue at different places.
+    Ref(RawInputValueId<Str>),
```
obmarg, line 67: I don't get this at all.  If it's just a variable: ok, I probably get it (though I don't think it'd hurt to have the word in the variant name somewhere).

But I assume from the comment it's not: but for what purpose do we want to "use later to share the same InputValue at different places"?  Why do we only store variables "which are initialized to the default or an arbitrary value first"?  Do we put variables that aren't initialised somewhere else?  Am I just misunderstanding entirely?

engine/crates/engine-v2/schema/src/input_value/walker.rs
```diff
@@ -0,0 +1,123 @@
...
+    }
+
+    /// Undefined values within InputObjects are silently removed according to the GraphQL spec.
+    /// But we can't remove an undefined argument
+    pub fn is_undefined(&self) -> bool {
+        match self.value {
+            RawInputValue::Undefined => true,
+            RawInputValue::Ref(id) => self.ctx.walk(*id).is_undefined(),
+            // I don't think this should ever happen, so should maybe be an unreachable!()
+            RawInputValue::SchemaRef(id) => self.ctx.schema_walk(*id).is_undefined(),
```
obmarg, line 25: I assume no, but just incase: is there ever a chance of `Ref` or `SchemaRef` becoming cyclical?

engine/crates/engine-v2/schema/src/lib.rs
```diff
@@ -268,12 +269,30 @@ impl From<InputObjectId> for Definition {
...
 
-#[derive(Debug)]
+#[derive(Debug, Clone, Copy)]
 pub struct Type {
     pub inner: Definition,
     pub wrapping: Wrapping,
 }
 
+impl Type {
+    /// Determines whether a varia
```
obmarg, line 279: This comment needs finished.  a link to the list coercion bit of the spec would probably be good to include if you weren't already going to.

obmarg (approved):

engine/crates/engine-v2/schema/src/input_value/raw.rs
Finistere: Well interestingly Rust does some optimization when only a single `Vec` is present in the possibilities:

```rust
enum First {
    A(u64),
    B(Vec<u64>),
}

enum Second {
    A(u64),
    B(Vec<u64>),
    D(Vec<u128>)
}

struct IdRange<T> {
    start: T,
    end: T
}

enum Third {
    A(u64),
    B(IdRange<u32>),
    C(IdRange<u32>),
}

fn main() {
    println!("First size: {} align: {}", std::mem::size_of::<First>(), std::mem::align_of::<First>());
    println!("Second size: {} align: {}", std::mem::size_of::<Second>(), std::mem::align_of::<Second>());
    println!("Third size: {} align: {}", std::mem::size_of::<Third>(), std::mem::align_of::<Third>());
}
```
which outputs:
```text
First size: 24 align: 8
Second size: 32 align: 8
Third size: 16 align: 8
```

> I can't see how 3 Vecs would only have a single u64 of footprint?

With `IdRange` we're using two `u32` for start and end. We just need the enum variant tag + alignment padding. I may have extrapolated it to be `usize` though to be fair. I expect it to be a padding of `u32` for wasm32, but might be wrong.

Without `IdRange` we would have a least `Box<[T]>`  (`usize` for pointer + `usize` for len) + alignment padding. And obviously with the downside of having many small arrays rather than a single one

engine/crates/engine-v2/schema/src/input_value/raw.rs
Finistere: yeah well. Coming from `FederatedGraph` which aren't associated with anything in the Schema, will update rustdoc

engine/crates/engine-v2/schema/src/input_value/raw.rs
Finistere: well ok, not that clear. with a proper name it would be `VariablePlaceholderRef` which felt a bit weird to put inside `engine-v2-chema`. But that might be my whole problem, trying to avoid duplicating the serde/deser/etc. implementation by having a more complex struct playing the role of both `engine::Value` and `engine::ConstValue`.

And I thought somehow it would be smart to update input values we parse from the query with the variables to avoid more back and forth, but it's just more complex in the end. I should probably keep the `ConstValue` vs `Value` separation and just duplicate code

engine/crates/engine-v2/schema/src/input_value/walker.rs
Finistere: Well the structure doesn't prevent it. With the current construction no. `Ref` is used as a variable placeholder and `SchemaRef` for default value. so fair point.

Finistere:
Coming back to those comments later this week. Merging both input value PRs for now

PR #1375, opened 2024-02-19
chore: Cleaning up the request module in engine-v2

Finistere:
This PR only moves around code, no business logic is changed.

I'm mostly simplifying the `Operation` build logic, exposing only a `Operation::build` to engine which also does the parsing and separating all the validation logic inside a module. I'm also updating it to traverse the selection sets through `fields()` which is a lot simpler.


engine/crates/engine-v2/src/request/validation/mod.rs
```diff
@@ -0,0 +1,43 @@
...
+    }
+}
+
+pub(crate) fn validate_operation(
+    schema: &Schema,
+    operation: &Operation,
+    request: &engine::Request,
+) -> Result<(), ValidationError> {
+    enforce_operation_limits(schema, operation, request)?;
+    ensure_introspection_is_accepted(schema, operation, request)?;
```
tomhoule, line 40: This reads nicely

tomhoule (approved):
It's easier to follow, nice!

PR #1381, opened 2024-02-22
GB-5967 - Use bun for config

yoav-lavi:
# Description

## Features

- Uses Bun for the TS config

# Type of change

- [ ] 💔 Breaking
- [x] 🚀 Feature
- [ ] 🐛 Fix
- [ ] 🛠️ Tooling
- [ ] 🔨 Refactoring
- [ ] 🧪 Test
- [ ] 📦 Dependency


cli/crates/server/src/bun.rs
```diff
@@ -0,0 +1,174 @@
...
+        Err(err) => Err(NpmCommandError::WorkingDirectoryCannotBeRead(
+            current_directory.to_owned(),
+            Arc::new(err),
+        )),
+    }?;
+    trace!("running '{command_string}'");
+
+    // Use `which` to work-around weird path search issues on Windows.
+    // See https://github.com/rust-lang/rust/issues/37519.
+    let program_path = which::which("npm").map_err(NpmCommandError::NotFound)?;
```
obmarg, line 84: Does this play nice if a user is using yarn/pnpm?

obmarg (approved):
Nice, apart from that one question

cli/crates/server/src/bun.rs
yoav-lavi: This is removed in the next PR, but we required node at this point so yes

PR #1394, opened 2024-02-29
fix: respect `.gitignore` in `grafbase deploy`

obmarg:
A user was running into issues with `grafbase deploy` yesterday, where deployments failed with a bizzare error very similar to [this one][1].

We eventually tracked this down to files in the users `.direnv` or `.devenv` folders - based on the tar-rs issue _probably_ hard links of some description. While it would be good to not choke on hardlinks, I'd argue we shouldn't have been archiving these files in the first place: they're clearly meant to stay local to the users machine.

So this PR switches `walkdir` for `ignore`, which will respect `.gitignore` files etc, providing users with a sensible (and in most cases zero effort) way to exclude files from the archive we build on `grafbase deploy`.  I think this is a sensible default.

This _could_ have ramifications for anyone doing a `grafbase deploy` after some kind of local build step, though I don't know how likely that is really.  I suggest we try it and see - if anyone complains we can revisit, maybe adding a CLI flag to change the behaviour.

Fixes GB-6107

[1]: https://github.com/alexcrichton/tar-rs/issues/313

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -662,6 +662,16 @@ dependencies = [
 ]
 
 [[package]]
+name = "bstr"
+version = "1.9.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "05efc5cfd9110c8416e471df0e96702d58690178e206e61b7173706673c93706"
+dependencies = [
+ "memchr",
+ "serde",
+]
+
+[[package]]
 name = "bumpalo"
 version = "3.14.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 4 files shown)


yoav-lavi (approved):

yoav-lavi:
I think the original reasoning was build steps, but we could try this. Nice!

yoav-lavi:
@obmarg although we could also explicitly ignore those dirs

obmarg:
> @obmarg although we could also explicitly ignore those dirs

We could, but the set of files/directories we probably shouldn't be uploading is _large_. I don't think a blacklist approach is reasonable, without some way for users to add to that blacklist.

I think having miscellaneous files/folders that shouldn't be uploaded is a far more common case than having an additional build step, so prioritising that makes sense to me.  Though happy to discuss if anyone agrees.

PR #1399, opened 2024-03-01
fix: Add missing upstream error for federation subgraph response.

Finistere:
We wouldn't return upstream errors if data wasn't `null`, a stupid mistake.

We can do better with a custom Deserialize for federation
entities rather than re-using the generic GraphQL response one, but it
does require more work. So doing the MVP for now.

### DIFF
--- a/engine/crates/integration-tests/tests/federation/basic/errors.rs
+++ b/engine/crates/integration-tests/tests/federation/basic/errors.rs
@@ -5,7 +5,7 @@ const SCHEMA: &str = include_str!("../../../data/federated-graph-schema.graphql"
 
 #[test]
 fn simple_error() {
-    let bench = FederationGatewayWithoutIO::new(
+    let gateway = FederationGatewayWithoutIO::new(
         SCHEMA,
         r#"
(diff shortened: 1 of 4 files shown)


engine/crates/integration-tests/tests/federation/basic/errors.rs
```diff
@@ -95,3 +95,56 @@ fn simple_error() {
...
+    );
+    let response = integration_tests::runtime().block_on(gateway.unchecked_execute());
+
+    let json = serde_json::from_slice::<serde_json::Value>(&response.bytes).unwrap();
+    insta::assert_json_snapshot!(json, @r###"
+    {
+      "data": null,
+      "errors": [
+        {
+          "message": "Upstream response error: Missing required field named 'reviews'",
```
obmarg, line 127: Not necessarily one for this PR, but `Error decoding response from upstream` might be clearer than `Upstream response error`?

"Upstream response error" is a bit ambiguous: is it a response error coming from the upstream or an error with the upstream response...?

obmarg (approved):

PR #1401, opened 2024-03-01
chore: clean up graph-entities

obmarg:
There's a lot of code in graph-entities that was really only there to support live queries.  That code stuck around when we removed live queries because it was also quite tied into the dynamo code.  Now that dynamo is also gone there's really no reason to keep this stuff around.  That and, once again, it's getting in the way of the fix I want to make.

As a nice bonus this likely reduces the serialised and in memory size of a bunch of these structs.


engine/crates/runtime-local/src/search.rs
```diff
@@ -33,13 +35,14 @@ impl SearchEngineInner for LocalSearchEngine {
...
-            .normalized(|ulid| NodeID::new(&index, &ulid.to_string()).to_string())
+        todo!("reinstate this if you need it")
+        // self.bridge
+        //     .request::<QueryRequest, QueryResponse>("search", request)
+        //     .await
+        //     .map_err(|error| {
+        //         log::error!(ctx.ray_id(), "Search Request failed with: {}", error);
+        //         QueryError::ServerError
+        //     })?
+        //     .normalized(|ulid| NodeID::new(&index, &ulid.to_string()).to_string())
```
obmarg, line 46: This whole file is removed in #1400, just needed to make it compile for this PR

tomhoule (approved):
satisfyingly red

Finistere (approved):

PR #1402, opened 2024-03-01
feat: self-hosted gateway

pimeys:
# Description

Closes: GB-6090

# Type of change

- [ ] 💔 Breaking
- [x] 🚀 Feature
- [ ] 🐛 Fix
- [ ] 🛠️ Tooling
- [ ] 🔨 Refactoring
- [ ] 🧪 Test
- [ ] 📦 Dependency


Cargo.lock
```diff
@@ -528,6 +535,41 @@ dependencies = [
...
+ "arc-swap",
+ "bytes",
+ "futures-util",
+ "http",
+ "http-body",
+ "http-body-util",
+ "hyper",
+ "hyper-util",
+ "pin-project-lite",
+ "rustls 0.21.10",
```
tomhoule, line 565: Is there a way to avoid bundling multiple rustls versions?

obmarg (approved):

cli/crates/production-server/src/server/engine.rs
```diff
@@ -0,0 +1,147 @@
...
+}
+
+async fn wait(mut receiver: UnboundedReceiver<BoxFuture<'static, ()>>) {
+    // Wait simultaneously on everything immediately accessible
+    future::join_all(std::iter::from_fn(|| receiver.try_recv().ok())).await;
+
+    // Wait sequentially on the rest
+    while let Some(fut) = receiver.recv().await {
+        fut.await;
+    }
```
tomhoule, line 146: I've never seen this pattern, what is the reasoning?

Cargo.lock
pimeys: Yes. Pushing a fix. Needed to patch axum-server, but I have a new version now with the same version as everywhere else.

cli/crates/production-server/src/server/engine.rs
pimeys: Copied from dev. Ping @Finistere

Finistere: On CF for background tasks we need to use `wait_until` to ensure the worker doesn't stop just after returning the response. The current cache implementation directly relies on it and the `RequestContext` exposes this notion with a channel. So this function waits for all futures put into it in parallel first and then sequentially, as initial futures can themselves use `wait_until` in theory. But I'm removing that. We should just use `tokio::spawn` in native.

PR #1403, opened 2024-03-02
chore: Merge v1 auth into gateway-v2-auth

Finistere:
Merging V1 auth into the V2 to have a single AccessToken.

# Description

# Type of change

- [ ] 💔 Breaking
- [ ] 🚀 Feature
- [ ] 🐛 Fix
- [ ] 🛠️ Tooling
- [ ] 🔨 Refactoring
- [ ] 🧪 Test
- [ ] 📦 Dependency


engine/crates/gateway-v2/auth-config/src/lib.rs
```diff
@@ -0,0 +1,8 @@
+pub mod v1;
+pub mod v2;
+
+#[derive(serde::Serialize, serde::Deserialize)]
+pub enum VersionedAuthConfig {
+    V1(v1::AuthConfig),
+    V2(v2::AuthConfig),
```
obmarg, line 7: Is "AuthConfig" for engine-v1 vs engine-v2 or is this just versioning for backwards compat in storage?  Guessing the latter, but mIght be worth a comment saying which. I've been wondering for the whole PR.

obmarg (approved):

engine/crates/gateway-v2/auth-config/src/lib.rs
Finistere: Is there a difference? In the end, the only use for this struct will be in the cloud where the config worker will return this instead of the whole registry/federation config. So could be in `api`, just convenient to be here

obmarg: > Is there a difference?

One of them would imply that either engine can take either `AuthConfig`, the other would imply that `v1::AuthConfig` is always for engine-v1 and `v2::AuthConfig` is always for engine-v2...

Finistere: well the gateway will be able to take either, the goal would be to have a single gateway worker for both engine. We probably would have one executor for the two engines now also, don't know

PR #1407, opened 2024-03-04
feat: support tracing customisation

hfgbarrigas:
# Description

This PR adds support for tracing customisation. Through the new `toml` configuration, one can configure how to export tracing data and define its level of granularity.

## Summary

In a nutshell, there's a new crate under `engine/tracing` that exposes tracing capabilities. Such as, `otel` tracing layers, a `tower` layer and typed spans that represent specific moments of handling incoming requests.
There are three spans to consider:

- `gateway`: the root span for each incoming request. Set when leveraging the `tower` layer and contains attributes specific to the http protocol, like status code, body sizes, ...
- `graphql`: represents a graphql request and contains attributes specific to `graphql` like query, operation type, ...
- `subgraph`: represents a request made to a subgraph when in federation mode. Contains attributes similar to `graphql` and the subgraph name

These spans are set in different places:

- `gateway` and `gateway-core` crates for engine v1
- `gateway-v2` and `engine-v2` for engine v2

These spans apart from containing specific attributes, have a specific `target=grafbase`. This will come into play when configuring the filters of the provided layers.

## Tracing configuration

The new `tracing` crate does not, as it should, configure the application tracing subscriber.
Configuring the subscriber is done at the CLI using the following steps:

- configure the global subscriber as soon as the binary runs with a `fmt_layer`, a global `EnvFilter` that can be customised via `--trace` and potentially a `noop` otel layer. The global env filter defaults to `info`.
- the noop otel layer is needed because the subscriber can't mutate after being set, or better put, can't accept new things but allows the existing things to change shape
- if using the `dev` or `start` commands, we set the `noop` layer and *after* starting the `tokio` runtime we *reload* the noop with a `batch` layer. You might be wondering why the `batch` layer isn't set from the get go. Two reasons: it needs an async runtime, `Tokio` in this case, and we only read the `config.toml` when dealing with `dev` or `start` commands.

## Highlighs

- the exposed tracing config is defined on the new crate
- the `batch` layer is subject to _per layer filtering_ and has two filters by default: 
  - a *sampling filter* with a `0.15` default value
  - a *level filter* with a `grafbase=info,off` default value. This means that by default, this layer will only capture spans that have a `target=grafbase` *and* `level=info`. Can be tweaked at the config level. *the global filter influences this filter* 
- span sampling is done at the root level, `span_name=gateway`. Any other span will *not* be sampled *unless* its root got sampled. This is achieved by leveraging tracing span extensions

## Full tracing configuration
None of the following is mandatory to be defined and is subject to defaults.

```toml
[telemetry]
service_name = "abc"

[telemetry.tracing]
enabled = false
sampling = 0.15
filter = "grafbase=info,off"

[telemetry.tracing.collect]
max_events_per_span = 128
max_attributes_per_span = 128
max_links_per_span = 128
max_attributes_per_event = 128
max_attributes_per_link = 128

[telemetry.tracing.exporters.otlp]
enabled = true
endpoint = "http://localhost:1234"
protocol = "grpc" # http
timeout = 60

[telemetry.tracing.exporters.otlp.batch_export]
scheduled_delay = 5
max_queue_size = 2048
max_export_batch_size = 512
max_concurrent_exports = 1

[telemetry.tracing.exporters.otlp.grpc.tls]
domain_name = "my_domain"
key = "/certs/grafbase.key"
ca = "/certs/ca.crt"
cert = "/certs/grafbase.crt"

[exporters.otlp.grpc.headers]
header1 = "header1"

[exporters.otlp.http.headers]
header1 = "header1"

[exporters.stdout]
enabled = true
timeout = 60

[exporters.stdout.batch_export]
scheduled_delay = 5
max_queue_size = 2048
max_export_batch_size = 512
max_concurrent_exports = 1
```

## Example
I added an example [here](https://github.com/grafbase/grafbase/pull/1407/files#diff-c049a7bb0393f3b979dcf931a2fa186641117ba2a92c28c5c7075a85a6b2757f) to test things out.

*NOTE:* Still working integration tests and crate docs

# Type of change

- [ ] 💔 Breaking
- [x] 🚀 Feature
- [ ] 🐛 Fix
- [ ] 🛠️ Tooling
- [ ] 🔨 Refactoring
- [x] 🧪 Test
- [ ] 📦 Dependency


engine/crates/engine-v2/src/sources/graphql/mod.rs
```diff
@@ -84,6 +85,15 @@ pub(crate) struct GraphqlExecutor<'ctx> {
 impl<'ctx> GraphqlExecutor<'ctx> {
     #[tracing::instrument(skip_all, fields(plan_id = %self.plan.id(), federated_subgraph = %self.subgraph.name()))]
     pub async fn execute(mut self) -> ExecutionResult<ResponsePart> {
+        let operation_name = self.plan.operation().as_ref().name.as_ref();
+        let operation_type = self.plan.operation().as_ref().ty.as_ref();
+
+        let subgraph_request_span = grafbase_tracing::span::subgraph::SubgraphRequestSpan::new(self.subgraph.name())
+            .with_operation_type(operation_type)
+            .with_operation_name(operation_name.map(|s| s.as_str()))
+            .with_document(self.json_body.as_str())
```
Finistere, line 94: operation name isn't useful here, it's the one for the root request. We don't generate operation names for sub requests anymore.
For the document, I'm mitigated. you're including variables in the `json_body` which is something we shouldn't keep around. The query string makes more sense. That's something that could be kept separated in the `GraphqlExecutor`. So would add a `FIXME:` here and use an empty string for the document. Can do it later

engine/crates/engine-v2/src/sources/graphql/mod.rs
```diff
@@ -118,6 +128,11 @@ impl<'ctx> GraphqlExecutor<'ctx> {
             &mut serde_json::Deserializer::from_slice(&bytes),
         );
 
+        subgraph_request_span.record_gql_response(GqlResponseAttributes {
+            has_errors: self.response_part.has_errors(),
+            operation_type: Some(operation_type),
```
Finistere, line 133: it's weird to have `operation_type` here, it's not something coming from the response and was already set beforehands

engine/crates/gateway-v2/src/lib.rs
```diff
@@ -106,6 +109,11 @@ pub struct Session {
 
 impl Session {
     pub async fn execute(self, ctx: &impl RequestContext, request: Request) -> Response {
+        let gql_request_span = grafbase_tracing::span::gql::GqlRequestSpan::new()
+            .with_operation_name(request.operation_name())
```
Finistere, line 113: operation name needs to be added later actually, after parsing the request. I'll fix that later, I'll rework some of the gateway code anyway

engine/crates/tracing/src/span/gql.rs
```diff
@@ -0,0 +1,57 @@
...
+    pub fn new() -> Self {
+        Default::default()
+    }
+
+    pub fn with_document(mut self, document: impl Into<Option<&'a str>>) -> Self {
+        self.document = document.into();
+        self
+    }
+
+    pub fn with_operation_name(mut self, operation_name: impl Into<Option<&'a str>>) -> Self {
```
Finistere, line 30: operation name & operation type would always be added in the same step, we need to parse the query to be sure to have them

engine/crates/engine-v2/src/sources/graphql/mod.rs
hfgbarrigas: The spans and their attributes follow the Apollo docs. We can discuss it from a product perspective but as is, anything graphql related should be in the graphql span which is for that purpose - the root is more for the protocol.
you're 100% right on the variables, we shouldn't be considering those.

engine/crates/engine-v2/src/sources/graphql/mod.rs
hfgbarrigas: yep this is probably a left over, removing.

engine/crates/tracing/src/config.rs
```diff
@@ -0,0 +1,649 @@
...
+    pub(crate) max_concurrent_exports: usize,
+}
+
+fn deserialize_duration<'de, D>(deserializer: D) -> Result<chrono::Duration, D::Error>
+where
+    D: Deserializer<'de>,
+{
+    let input = i64::deserialize(deserializer)?;
+
+    Ok(chrono::Duration::seconds(input))
```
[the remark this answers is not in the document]
hfgbarrigas: nice catch!

engine/crates/tracing/src/config.rs
```diff
@@ -0,0 +1,649 @@
...
+    /// The maximum links per span before discarding.
+    /// The default is 128.
+    pub max_links_per_span: u32,
+    /// The maximum attributes per event before discarding.
+    /// The default is 128.
+    pub max_attributes_per_event: u32,
+    /// The maximum attributes per link before discarding.
+    /// The default is 128.
+    pub max_attributes_per_link: u32,
+}
```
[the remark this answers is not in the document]
hfgbarrigas: nice catch! adjusted and added corresponding tests

engine/crates/engine-v2/src/sources/graphql/mod.rs
hfgbarrigas: Added the `FIXME:` and removed it from the time being.

Cargo.lock
```diff
@@ -470,21 +470,49 @@ version = "1.1.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "d468802bab17cbc0cc575e9b053f41e72aa36bfa6b7f55e3529ffa43161b97fa"
 
+[[package]]
+name = "axum"
+version = "0.6.20"
```
pimeys, line 475: Ping @jakubadamw

cli/crates/cli/src/main.rs
```diff
@@ -111,12 +126,24 @@ fn try_main(args: Args) -> Result<(), CliError> {
                 process::exit(exitcode::OK);
             });
 
+            // TODO: what's the story for dev and the toml config?
```
pimeys, line 129: No config for now. Also, we're moving the start command to a new place in my PR:

```
grafbase federated start
```

We're going to do a bigger split here in the near future:

https://linear.app/grafbase/issue/GB-6147/separate-federatedstandalone-into-subcommands

cli/crates/cli/src/start.rs
```diff
@@ -62,7 +64,15 @@ where
 
     let handle = thread::spawn(move || {
         #[allow(clippy::ignored_unit_patterns)]
-        tokio::runtime::Runtime::new().unwrap().block_on(async {
+        let runtime = tokio::runtime::Runtime::new().unwrap();
+        runtime.block_on(async {
```
pimeys, line 68: The new gateway is in another castle. We have a crate `production-server` where we need to add this too when we merge my PR.

cli/crates/cli/src/start/configuration.rs
```diff
@@ -492,4 +499,24 @@ mod tests {
         invalid value: integer `1000000000000000000`, expected u16
         "###);
     }
+
+    #[test]
+    fn telemetry() {
```
pimeys, line 504: Maybe we need a few more tests for this?

cli/crates/cli/src/start/configuration/telemetry.rs
```diff
@@ -0,0 +1,9 @@
+#[derive(Debug, PartialEq, Clone, serde::Deserialize)]
+#[serde(deny_unknown_fields)]
+pub struct TelemetryConfig {
```
pimeys, line 3: I wonder do we even need to have this in the gateway level config? You can load the same file from your own config struct, and it will just skip all options it doesn't recognize.

engine/crates/engine-v2/src/sources/graphql/mod.rs
```diff
@@ -84,6 +85,16 @@ pub(crate) struct GraphqlExecutor<'ctx> {
 impl<'ctx> GraphqlExecutor<'ctx> {
     #[tracing::instrument(skip_all, fields(plan_id = %self.plan.id(), federated_subgraph = %self.subgraph.name()))]
     pub async fn execute(mut self) -> ExecutionResult<ResponsePart> {
+        let operation_name = self.plan.operation().as_ref().name.as_ref();
+        let operation_type = self.plan.operation().as_ref().ty.as_ref();
+
+        let subgraph_request_span = grafbase_tracing::span::subgraph::SubgraphRequestSpan::new(self.subgraph.name())
```
pimeys, line 91: It's a bit long line, maybe `use` would be better for the `SubgraphRequestSpan`?

engine/crates/gateway-core/Cargo.toml
```diff
@@ -23,6 +23,7 @@ common-types.workspace = true
 engine.workspace = true
 engine-value.workspace = true
 futures-util.workspace = true
+grafbase-tracing = { path = "../../../engine/crates/tracing"}
```
pimeys, line 26: Add this to the workspace and do `grafbase-tracing.workspace = true`.

examples/tracing/docker-compose.yaml
```diff
@@ -0,0 +1,54 @@
+version: "3"
```
pimeys, line 1: Could we add all of these to the docker-compose.yml in the project root?

cli/crates/cli/src/main.rs
hfgbarrigas: ok, I'll leave a FIXME in place for future reference and leave a note

cli/crates/cli/src/start/configuration.rs
hfgbarrigas: hm, there are [unit tests](https://github.com/grafbase/grafbase/pull/1407/files#diff-7596b966e4a4dab502426e39482aa121c31b2e1e667e5f141cdd98dccbec9cafR397) and [integration-tests](https://github.com/grafbase/grafbase/pull/1407/files#diff-08894959da040b5465cb8a63a4a0f4efc1144bacfb04b58ac961f42685d61274R1). Anything else you'd like to see?

cli/crates/cli/src/start/configuration/telemetry.rs
hfgbarrigas: Hm I'd rather leave the file read part out of the tracing crate.

examples/tracing/docker-compose.yaml
hfgbarrigas: hm, I think so. When I created these I had this tracing specific use case. I'll move these to the root in a different PR and review how it can integrate with the cli code or grafbase's docker image.

pimeys (approved):
go for it, it requires a bit more troublesome rebase to the production-server, but this should go first.

PR #1423, opened 2024-03-11
chore: log info when a gateway is created/updated

pimeys:
# Description

We should log when we successfully created/updated a new gateway.

# Type of change

- [ ] 💔 Breaking
- [ ] 🚀 Feature
- [x] 🐛 Fix
- [ ] 🛠️ Tooling
- [ ] 🔨 Refactoring
- [ ] 🧪 Test
- [ ] 📦 Dependency

### DIFF
--- a/cli/crates/production-server/src/server/graph_fetch_method.rs
+++ b/cli/crates/production-server/src/server/graph_fetch_method.rs
@@ -6,6 +6,7 @@ use crate::config::{AuthenticationConfig, OperationLimitsConfig};
 use ascii::AsciiString;
 use std::sync::Arc;
 use tokio::sync::watch;
+use tracing::Level;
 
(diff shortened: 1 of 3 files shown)


cli/crates/production-server/src/server/graph_fetch_method.rs
```diff
@@ -62,6 +63,14 @@ impl GraphFetchMethod {
                 });
             }
             GraphFetchMethod::FromLocal { federated_schema } => {
+                tracing::event!(
```
jakubadamw, line 66: Should we be using tracing events for stdout? In the regular dev server we use `println!()`, if I recall correctly.

jakubadamw (approved):
One question. Looks good otherwise!

PR #1426, opened 2024-03-11
fix: custom field entity mutation invalidation cache tag

hfgbarrigas:
# Description

Fixes a bug that happened when computing response cache tags that were not present when using mutation invalidation of type `entity` with a custom field.

# Type of change

- [ ] 💔 Breaking
- [ ] 🚀 Feature
- [x] 🐛 Fix
- [ ] 🛠️ Tooling
- [ ] 🔨 Refactoring
- [ ] 🧪 Test
- [ ] 📦 Dependency

### DIFF
--- a/engine/crates/tracing/src/filter.rs
+++ b/engine/crates/tracing/src/filter.rs
@@ -1,2 +1,0 @@
-/// Filter to sample spans based on a ratio
-pub mod ratio_sampling;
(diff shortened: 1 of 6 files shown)


engine/crates/tracing/src/filter.rs
hfgbarrigas, line 1: this is  leftover from the tracing effort.

tomhoule (approved):
:100:

PR #1430, opened 2024-03-12
feat: allow setting uplink, api and auth urls with env vars

pimeys:
# Description

We need three different env vars:

- `GRAFBASE_API_URL` which is the API endpoint. Defaults to https://api.grafbase.com/graphql
- `GRAFBASE_GDN_URL` for starting the federated router, and getting the latest federated graph. Defaults to https://gdn.grafbase.com

Closes: GB-6209

# Type of change

- [ ] 💔 Breaking
- [x] 🚀 Feature
- [ ] 🐛 Fix
- [ ] 🛠️ Tooling
- [ ] 🔨 Refactoring
- [ ] 🧪 Test
- [ ] 📦 Dependency

### DIFF
--- a/cli/crates/backend/src/api/consts.rs
+++ b/cli/crates/backend/src/api/consts.rs
@@ -1,6 +1,18 @@
+use std::sync::OnceLock;
+
 pub const CREDENTIALS_FILE: &str = "credentials.json";
-pub const AUTH_URL: &str = "https://app.grafbase.com/auth/cli";
-pub const API_URL: &str = "https://api.grafbase.com/graphql";
 pub const PACKAGE_JSON: &str = "package.json";
 pub const TAR_CONTENT_TYPE: &str = "application/x-tar";
 pub const GRAFBASE_ACCESS_TOKEN_ENV_VAR: &str = "GRAFBASE_ACCESS_TOKEN";
+pub const AUTH_URL: &str = "https://app.grafbase.com/auth/cli";
+
+const API_URL: &str = "https://api.grafbase.com/graphql";
+
+pub fn api_url() -> &'static str {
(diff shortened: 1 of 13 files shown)


cli/crates/backend/src/api/consts.rs
```diff
@@ -1,6 +1,27 @@
...
-pub const AUTH_URL: &str = "https://app.grafbase.com/auth/cli";
-pub const API_URL: &str = "https://api.grafbase.com/graphql";
 pub const PACKAGE_JSON: &str = "package.json";
 pub const TAR_CONTENT_TYPE: &str = "application/x-tar";
 pub const GRAFBASE_ACCESS_TOKEN_ENV_VAR: &str = "GRAFBASE_ACCESS_TOKEN";
+
+const API_URL: &str = "https://api.grafbase.com/graphql";
+const AUTH_URL: &str = "https://app.grafbase.com/auth/cli";
+
+pub fn auth_url() -> &'static str {
```
fbjork, line 11: perhaps `dashboard_url()` is better? to avoid having one env var for every url that leads to the dashboard.
`GRAFBASE_DASHBOARD_URL`

yoav-lavi:
Do we want these as env vars rather than a configuration? Seems like it might be a bit brittle and third parties could use this maliciously potentially

cli/crates/backend/src/api/consts.rs
pimeys: We only use it for authentication.

fbjork: Thinking about this so more, why is the federated graph concerned with an auth url anyway?

fbjork (approved):

PR #1433, opened 2024-03-12
chore: replace jsonwebtoken with jwt-compact

Finistere:
jsonwebtoken doesn't compile to wasm locally because of ring (who knows
why it works in the CI).
For the v1 auth we rely on jwt-compact anyway and WASM-compatbility is
an explicit goal.

# Description

# Type of change

- [ ] 💔 Breaking
- [ ] 🚀 Feature
- [ ] 🐛 Fix
- [ ] 🛠️ Tooling
- [ ] 🔨 Refactoring
- [ ] 🧪 Test
- [ ] 📦 Dependency


tomhoule (approved):

engine/crates/gateway-v2/auth/src/jwt.rs
```diff
@@ -82,93 +120,122 @@ impl Authorizer for JwtProvider {
...
-        for key in decoding_keys(&metadata.jwks, &jose_header) {
-            if let Ok(TokenData { claims, .. }) = jsonwebtoken::decode(token_str, &key, &validation) {
-                use base64::{engine::general_purpose::URL_SAFE_NO_PAD, Engine as _};
-                let signature = URL_SAFE_NO_PAD
-                    .decode(token_str.rsplit('.').next().expect("valid jwt"))
-                    .expect("valid jwt");
-                return Some(AccessToken::Jwt(JwtToken { claims, signature }));
+        if let Some(expected) = self.config.jwks.audience.as_ref() {
+            let audience = token.claims().custom.audience.as_ref()?;
+            if audience.iter().all(|aud| aud != expected) {
```
obmarg, line 156: `all` on an empty iterator returns true - I don't think that's what you want here?

Finistere: damn, good catch, just at the moment I'm merging xD
thanks!

Finistere: ah no that's what I want. I do test it. If there is nothing, it's true and I'm returning `None`

Finistere: not the clearest, have to admit it

obmarg: Oh shit yes, cool.  Ignore me then.

PR #1434, opened 2024-03-12
feat: separate docker setup for standalone/federated modes

pimeys:
# Description

Changes how our docker images are built, now we create two separate images:

- `grafbase/federated` for federated gateway
- `grafbase/standalone` is the current docker image we build

The builds for standalone was broken since forever (if built without a cache), because we need bun and bun needs glibc. This fixes it for both builds.

Both setups take a listen address instead of a port, and the docker images should default to `0.0.0.0:4000`.

An example docker-compose.yml for the federated build:

```yaml
version: '3'
services:
  grafbase:
    image: grafbase/federated:latest
    restart: always
    volumes:
      - ./grafbase.toml:/etc/grafbase.toml
    environment:
      GRAFBASE_GRAPH_REF: "graph-ref@branch"
      GRAFBASE_ACCESS_TOKEN: "ACCESS_TOKEN_HERE"
    ports:
      - '4000:4000'
```

The configuration in `cli/crates/production-server/config/grafbase.toml` is used as the default configuration. Copying it to the current directory will make it as the config for the federated gateway.

Closes: GB-6218

### DIFF
--- a/.github/workflows/cli.yml
+++ b/.github/workflows/cli.yml
@@ -51,6 +51,7 @@ jobs:
           filters: |
             build:
               - Dockerfile
+              - Dockerfile.*
               - 'cli/**/*'
               - 'engine/**/*'
               - '.github/workflows/cli.yml'
@@ -140,24 +141,33 @@ jobs:
           username: ${{ github.actor }}
           password: ${{ secrets.GITHUB_TOKEN }}
 
-      - name: Build docker image
+      - name: Build standalone docker image
         run: |
-          docker build -t ghcr.io/grafbase/grafbase:$SHA_TAG .
+          docker build -f Dockerfile.standalone -t ghcr.io/grafbase/standalone:$SHA_TAG .
         env:
           SHA_TAG: ${{ github.sha }}
 
-      - name: Tag released version
+      - name: Build federated docker image
+        run: |
+          docker build -f Dockerfile.federated -t ghcr.io/grafbase/federated:$SHA_TAG .
+        env:
+          SHA_TAG: ${{ github.sha }}
+
+      - name: Tag released versions
         if: ${{ startsWith(github.ref, 'refs/tags') }}
         run: |
-          docker tag ghcr.io/grafbase/grafbase:$SHA_TAG ghcr.io/grafbase/grafbase:latest
-          docker tag ghcr.io/grafbase/grafbase:$SHA_TAG ghcr.io/grafbase/grafbase:$(echo "$REF" | sed -e "s/^refs\/tags\/cli-//")
+          docker tag ghcr.io/grafbase/standalone:$SHA_TAG ghcr.io/grafbase/standalone:latest
+          docker tag ghcr.io/grafbase/standalone:$SHA_TAG ghcr.io/grafbase/standalone:$(echo "$REF" | sed -e "s/^refs\/tags\/cli-//")
+          docker tag ghcr.io/grafbase/federated:$SHA_TAG ghcr.io/grafbase/federated:latest
+          docker tag ghcr.io/grafbase/federated:$SHA_TAG ghcr.io/grafbase/federated:$(echo "$REF" | sed -e "s/^refs\/tags\/cli-//")
         env:
           SHA_TAG: ${{ github.sha }}
           REF: ${{ github.ref }}
 
       - name: Push docker image
         run: |
-          docker push --all-tags ghcr.io/grafbase/grafbase
+          docker push --all-tags ghcr.io/grafbase/standalone
+          docker push --all-tags ghcr.io/grafbase/federated
 
   # windows:
   #   needs: [lint]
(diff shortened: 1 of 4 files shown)


Dockerfile.standalone
```diff
@@ -12,17 +12,33 @@ COPY ./engine ./engine
...
+RUN npx --yes pnpm run build
+
 WORKDIR /grafbase
 
 RUN cargo build -p grafbase --release
 
+# Remove bun dependencies
+RUN rm -rf /etc/apk/keys/sgerrand.rsa.pub
+RUN apk del glibc
+RUN rm -rf glibc-2.28-r0.apk
```
tomhoule, line 40: If it's dynamically linked, isn't this going to be a problem?

Dockerfile.federated
```diff
@@ -0,0 +1,67 @@
...
+
+RUN adduser -g wheel -D grafbase -h "/data" && mkdir -p /data && chown grafbase: /data
+USER grafbase
+
+COPY --from=build /grafbase/target/release/grafbase /bin/grafbase
+COPY --from=build /grafbase/cli/crates/production-server/config/grafbase.toml /etc/grafbase.toml
+
+ENTRYPOINT ["/bin/grafbase"]
+
+# these args should be set so the binary can start. they have to be changed for successfully running the gatewa
```
tomhoule, line 58: y

tomhoule (approved):

Dockerfile.standalone
pimeys: It's only to run bun, we don't need bun with the federated gateway during runtime. Eventually we want to get rid of all the javascript dependencies for the federated gateway. This might mean splitting it to a separate binary, or feature flags.

Finistere: You're removing in both federated and standalone though, no?

pimeys: Yes. Both builds also need bun right now, we need to change that in the future.

.github/workflows/cli.yml
```diff
@@ -140,24 +141,33 @@ jobs:
...
-      - name: Build docker image
+      - name: Build standalone docker image
         run: |
-          docker build -t ghcr.io/grafbase/grafbase:$SHA_TAG .
+          docker build -f Dockerfile.standalone -t ghcr.io/grafbase/standalone:$SHA_TAG .
         env:
           SHA_TAG: ${{ github.sha }}
 
-      - name: Tag released version
+      - name: Build federated docker image
```
fbjork, line 150:
```suggestion
      - name: Build gateway Docker image
```

.github/workflows/cli.yml
```diff
@@ -140,24 +141,33 @@ jobs:
           username: ${{ github.actor }}
           password: ${{ secrets.GITHUB_TOKEN }}
 
-      - name: Build docker image
+      - name: Build standalone docker image
         run: |
-          docker build -t ghcr.io/grafbase/grafbase:$SHA_TAG .
+          docker build -f Dockerfile.standalone -t ghcr.io/grafbase/standalone:$SHA_TAG .
```
fbjork, line 146:
```suggestion
          docker build -f Dockerfile.standalone -t ghcr.io/grafbase/grafbase:$SHA_TAG .
```

.github/workflows/cli.yml
```diff
@@ -140,24 +141,33 @@ jobs:
...
         env:
           SHA_TAG: ${{ github.sha }}
 
-      - name: Tag released version
+      - name: Build federated docker image
+        run: |
+          docker build -f Dockerfile.federated -t ghcr.io/grafbase/federated:$SHA_TAG .
```
fbjork, line 152:
```suggestion
          docker build -f Dockerfile.gateway -t ghcr.io/grafbase/gateway:$SHA_TAG .
```

.github/workflows/cli.yml
```diff
@@ -140,24 +141,33 @@ jobs:
...
+          docker tag ghcr.io/grafbase/federated:$SHA_TAG ghcr.io/grafbase/federated:latest
+          docker tag ghcr.io/grafbase/federated:$SHA_TAG ghcr.io/grafbase/federated:$(echo "$REF" | sed -e "s/^refs\/tags\/cli-//")
         env:
           SHA_TAG: ${{ github.sha }}
           REF: ${{ github.ref }}
 
       - name: Push docker image
         run: |
-          docker push --all-tags ghcr.io/grafbase/grafbase
+          docker push --all-tags ghcr.io/grafbase/standalone
```
fbjork, line 169:
```suggestion
          docker push --all-tags ghcr.io/grafbase/grafbase
```

.github/workflows/cli.yml
```diff
@@ -140,24 +141,33 @@ jobs:
...
+          docker push --all-tags ghcr.io/grafbase/federated
```
fbjork, line 170:
```suggestion
          docker push --all-tags ghcr.io/grafbase/gateway
```

Dockerfile.federated
```diff
@@ -0,0 +1,67 @@
+# Build
```
fbjork, line 1: rename this file to Dockerfile.gateway

Dockerfile.standalone
```diff
@@ -12,17 +12,33 @@ COPY ./engine ./engine
 COPY ./packages/grafbase-sdk/package.json ./packages/grafbase-sdk
```
fbjork, line 12: rename back to just Dockerfile

PR #1436, opened 2024-03-13
Implement trusted documents in self-hosted engine

tomhoule:
# Description

This plugs in the last bits of trusted documents logic into the self-hosted engine, in hybrid mode only. We now fetch documents from the R2 bucket the API uploads them to, and we respect the `bypass_header` configuration.

closes GB-6169
closes GB-6164

# Type of change

- [ ] 💔 Breaking
- [x] 🚀 Feature
- [ ] 🐛 Fix
- [ ] 🛠️ Tooling
- [ ] 🔨 Refactoring
- [ ] 🧪 Test
- [ ] 📦 Dependency


engine/crates/engine/src/headers.rs
```diff
@@ -24,7 +24,7 @@ impl RequestHeaders {
         )
     }
 
-    pub fn find(&self, expected_name: &str) -> Option<&str> {
+    pub fn find<'a>(&'a self, expected_name: &str) -> Option<&'a str> {
```
tomhoule, line 27: Not necessary in this PR anymore, but it doesn't hurt to be more specific and extend the lifetime.

engine/crates/integration-tests/tests/federation/trusted_documents.rs
```diff
@@ -21,12 +21,6 @@ const TRUSTED_DOCUMENTS: &[TestTrustedDocument] = &[
         document_id: "favourite-repo-query-doc-id",
         document_text: "query { __typename }",
     },
-    TestTrustedDocument {
```
tomhoule, line 24: The API of the trait changed in such a way that we can't test the wrong branch case anymore.

cli/crates/federated-dev/src/dev/gateway_nanny.rs
```diff
@@ -57,7 +57,7 @@ pub(super) fn new_gateway(graph: Option<FederatedGraph>, config: &FederatedGraph
         EngineEnv {
             fetcher: runtime_local::NativeFetcher::runtime_fetcher(),
             cache: cache.clone(),
-            trusted_documents: runtime_noop::trusted_documents::NoopTrustedDocuments.into(),
+            trusted_documents: Box::new(runtime_noop::trusted_documents::NoopTrustedDocuments),
```
Finistere, line 60: For most runtime stuff I usually have a struct like:
```
struct TrustedDocuments(Arc<dyn TrustedDocumentsInner>)

impl Defer for TrustedDocuments {
   type Target = dyn TrustedDocumentsInner;
}
```

to hide the dynamic part

cli/crates/production-server/src/config.rs
```diff
@@ -107,7 +107,7 @@ pub struct TlsConfig {
     pub key: PathBuf,
 }
 
-#[derive(Debug, PartialEq, serde::Deserialize, Default)]
+#[derive(Debug, PartialEq, serde::Deserialize, Default, Clone)]
 #[serde(deny_unknown_fields)]
 pub struct TrustedDocumentsConfig {
```
Finistere, line 112: bypass_header_name & value should be merged into a tuple imho it they only make sense together

Finistere (approved):

cli/crates/production-server/src/config.rs
Finistere: ah I guess that's the toml config. Should be a struct with `#[serde(flatten)]` then

cli/crates/federated-dev/src/dev/gateway_nanny.rs
Finistere: makes it easier to manipulate elsewhere

cli/crates/production-server/src/config.rs
tomhoule: done

cli/crates/federated-dev/src/dev/gateway_nanny.rs
tomhoule: done

cli/crates/production-server/src/config.rs
```diff
@@ -107,18 +107,26 @@ pub struct TlsConfig {
...
+#[derive(Debug, serde::Deserialize, Default, Clone)]
 #[serde(deny_unknown_fields)]
 pub struct TrustedDocumentsConfig {
     /// If true, the engine will only accept trusted document queries. Default: false.
     #[serde(default)]
     pub enabled: bool,
-    /// Optional name of the header that can be set to bypass trusted documents enforcement, when `enabled = true`. Only meaningful in combination with `bypass_header_value`.
+    /// See [BypassHeader]
+    #[serde(flatten)]
+    pub bypass_header: BypassHeader,
```
Finistere, line 118: bypass_header should be `Option` and `bypass_header_name` shouldn't, no?

tomhoule: That's where I started but there are two downsides to that:

- small one: errors are worse (it's still the case with the current solution)
- big one: since `flatten` will swallow the error and put a `None` on the field if deserialization fails, you don't get an error when the env var is missing in a situation like ` bypass_header_value = "secret-{{ env.TEST_HEADER_SECRET }}" ` (there's a test for it).

Finistere: ok, fair points 👍

PR #1448, opened 2024-03-15
fix: coerce enums when handling join variables

obmarg:
We let users provide variables when joining, and those variables get read from
the current ResolvedValue in order to perform the join.  However, ResovledValue
is based on JSON, and arguments are based on `engine::Value`.  In JSON an enum
value is represented as a String, but in `engine::Value` there's a dedicated
variant for enums.

When serializing to JSON the difference doesn't matter, but it can when
serializing to GraphQL syntax - some servers will reject strings in enum
position.

This PR updates the join code to pass variables through input processing, which
should make sure that fields which are enums are represented by
`engine::Value::Enum` and not `engine::Value::String`.


tomhoule (approved):
:100:

PR #1455, opened 2024-03-17
feat(tracing): Add query & operation type for subgraph requests

Finistere:
In order to log the query we're sending to the subgraph I'm replacing
all input values in the query string with variables. I also realized how
much allocations I was doing in the query building, so improved that.

I've also simplified the query generation for named fragments, instead
of trying to keep them I just transform them into inline ones. And I'm
only keeping type conditions which are useful. There is little downside
except probably a bit bigger query strings.

The operation type is also now better handled. the `PlanWalker` doesn't
expose `operation` anymore. It was the root operation and was easily
confused with the subgraph graphql operation. As subscription executors
also relied on that, I've also changed the coordinator/subscription
executor interface.

# Description

# Type of change

- [ ] 💔 Breaking
- [ ] 🚀 Feature
- [ ] 🐛 Fix
- [ ] 🛠️ Tooling
- [ ] 🔨 Refactoring
- [ ] 🧪 Test
- [ ] 📦 Dependency


engine/crates/engine-v2/src/sources/graphql/query.rs
```diff
@@ -232,9 +301,19 @@ impl std::ops::DerefMut for Buffer {
...
+    fn write_indent(&mut self) {
+        let mut indent = self.indent;
+        loop {
+            if indent < INDENTS {
+                self.inner.push_str(&SPACES[..self.indent * INDENT_SIZE]);
+                return;
+            }
+            self.inner.push_str(&SPACES[..INDENTS * INDENT_SIZE]);
+            indent -= INDENTS
+        }
```
obmarg, line 317: Seems like it'd be a lot easier to just do `self.inner.push(' ')` in a loop...

Finistere: hum.... indeed...

engine/crates/engine-v2/src/sources/graphql/query.rs
```diff
@@ -127,65 +166,87 @@ impl QueryBuilder {
...
-                .fragment_name_to_last_id
-                .entry(fragment.name().to_string())
-                .and_modify(|id| *id += 1)
-                .or_default();
-            format!("{}_{}", fragment.name(), id)
-        });
-        buffer.indent_write(&format!("...{name}\n"))?;
+        let ty = fragment.ty();
+        if parent_ty != ty {
+            indent_write!(buffer, "... on {}", ty.name())?;
```
obmarg, line 202: Intentionally skipping directives here?

engine/crates/engine-v2/src/sources/graphql/query.rs
```diff
@@ -127,65 +166,87 @@ impl QueryBuilder {
...
+        } else {
+            // We don't create a nested selection set if the type condition is equivalent to the parent
+            // type
+            for selection in spread.selection_set() {
+                match selection {
+                    PlanSelection::Field(field) => self.write_field(buffer, field)?,
+                    PlanSelection::FragmentSpread(spread) => self.write_fragment_spread(buffer, ty, spread)?,
+                    PlanSelection::InlineFragment(fragment) => self.write_inline_fragment(buffer, ty, fragment)?,
+                }
+            }
```
obmarg, line 213: What's the thinking behind this branch?  I get that we don't need the type condition, but it'd be simpler to just send it anyway and I can't see much downside.

Also same question about directives here (which would be one reason to make this an inline fragment rather than flattening)

engine/crates/engine-v2/src/sources/graphql/query.rs
```diff
@@ -127,65 +166,87 @@ impl QueryBuilder {
...
+            // type
+            for selection in fragment.selection_set() {
+                match selection {
+                    PlanSelection::Field(field) => self.write_field(buffer, field)?,
+                    PlanSelection::FragmentSpread(spread) => self.write_fragment_spread(buffer, parent_ty, spread)?,
+                    PlanSelection::InlineFragment(fragment) => {
+                        self.write_inline_fragment(buffer, parent_ty, fragment)?
+                    }
+                }
+            }
```
obmarg, line 238: Same question here - is there any need for this extra complexity?

obmarg (approved):
Nice.  

I am slightly skeptical that making every argument into a variable just so we can log things is a good idea, but I have no concrete objections so never mind I guess.

engine/crates/engine-v2/src/sources/graphql/query.rs
Finistere: Well, to answer all your comments:
- we don't support directives anywhere to be honest. I would expect supported directives to be defined by the subgraph SDL and for us to validate them. The distinction in the GraphQL spec seems a bit useless to me... In case it's a `FRAGMENT_DEFINITION` only directive I would need to create a fragment. So when adding directive support I would always write a fragment here.
- I'm flattening the selection sets because it makes more readable queries and friendlier to debug imho. I could probably factorize a bit the code though.

PR #1465, opened 2024-03-19
fix: error handling in graphql connector

obmarg:
The GraphQL connector uses data loaders to group concurrent queries into a single query that can be sent to the downstream server.  The handling of errors for this case was extremely naive:  all the errors from the response were just cloned and sent back to the individual ungrouped resolvers, each of which would then record all the errors in the context.

This is particularly problematic when combined with joins on list fields: the GraphQL connector will join every subquery on the list elements into a single query.  If that response contains errors on all its field, then you end up with N^2 errors in the GraphQL response.  In one case this seems to have led to ~70MB responses being sent back to Pathfinder.

I've updated the code to be smarter about errors:

1. Deduplicating errors before sending back to resolvers.
2. Rewriting the path inside errors so its correct in the context of our API.
3. More thorough tests of error handling


yoav-lavi (approved):
Looks great!
I'd consider if it makes sense to send above X number of errors though rather than indicating that there's additional errors to be solved after the current batch.

PR #1471, opened 2024-03-20
feat: arguments on join fields

obmarg:
A customer wants to forward an argument on a join field into the query that we use to perform the join.  This adds support for that, using the same variable syntax we already use.  

Any arguments on the join field are automatically available under `$argumentName`.  Currently arguments will always shadow fields of the parent object.  So if there's a `foo` argument then `$foo` will _always_ refer to that argument, and any field on the parent type named `foo` will be inaccessible.  The alternative to this would have been to use some special syntax to differentiate fields & arguments, but that seemed like a lot more work, and I'm not sure how likely it is for users to need that functionality.  We can always revisit later.

I'm not doing much validation on these arguments at parsing time currently.  I'll do that in a follow up PR, just wanted to unblock the customer as quick as possible.

Fixes GB-6304

### DIFF
--- a/engine/crates/parser-sdl/src/schema_coord.rs
+++ b/engine/crates/parser-sdl/src/schema_coord.rs
@@ -1,0 +1,47 @@
+/// Helper enum for specifying the location of errors
+#[derive(Clone, Copy, Debug)]
+pub enum SchemaCoord<'a> {
+    Field(&'a str, &'a str),
+    Entity(&'a str, &'a str),
+}
+
+impl std::fmt::Display for SchemaCoord<'_> {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            SchemaCoord::Field(ty, field) => {
+                write!(f, "{ty}.{field}")
+            }
+            SchemaCoord::Entity(ty, key) => {
+                write!(f, "federation key `{key}` on the type {ty}")
+            }
+        }
+    }
+}
+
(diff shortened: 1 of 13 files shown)


engine/crates/parser-sdl/src/schema_coord.rs
```diff
@@ -0,0 +1,47 @@
+/// Helper enum for specifying the location of errors
+#[derive(Clone, Copy, Debug)]
+pub enum SchemaCoord<'a> {
+    Field(&'a str, &'a str),
```
Finistere, line 4: might be worth naming the field?

engine/crates/parser-sdl/src/schema_coord.rs
```diff
@@ -0,0 +1,47 @@
...
+            }
+            SchemaCoord::Entity(ty, key) => {
+                write!(f, "federation key `{key}` on the type {ty}")
+            }
+        }
+    }
+}
+
+impl SchemaCoord<'_> {
+    pub fn into_owned(self) -> OwnedSchemaCoord {
```
Finistere, line 22: an alternative could be using `Cow` to store the value and return a `'static` with `into_owned()`

Finistere (approved):

engine/crates/parser-sdl/src/schema_coord.rs
obmarg: I almost went down that route, but it makes constructing them a bit more annoying so I took this approach instead

PR #1474, opened 2024-03-20
composition: allow arguments in `@requires` selection sets

tomhoule:
closes GB-6295

### DIFF
--- a/engine/crates/composition/src/subgraphs/keys.rs
+++ b/engine/crates/composition/src/subgraphs/keys.rs
@@ -53,8 +53,26 @@ impl Subgraphs {
                 .map(|item| match &item.node {
                     ast::Selection::Field(item) => {
                         let field = subgraphs.strings.intern(item.node.name.node.as_str());
+                        let arguments = item
+                            .node
+                            .arguments
+                            .iter()
+                            .map(|(name, value)| {
+                                let name = subgraphs.strings.intern(name.node.as_str());
+                                let value = crate::ast_value_to_subgraph_value(
+                                    &value.node.clone().into_const().unwrap(),
+                                    subgraphs,
+                                );
+
+                                (name, value)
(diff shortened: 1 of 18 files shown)


engine/crates/composition/src/subgraphs/keys.rs
```diff
@@ -121,15 +139,16 @@ impl NestedKeyFields {
...
 pub(crate) struct Key {
     selection_set: Vec<Selection>,
     resolvable: bool,
 }
 
-#[derive(PartialEq, Eq, PartialOrd, Ord, Debug)]
+#[derive(PartialEq, PartialOrd, Debug)]
 pub(crate) struct Selection {
     pub(crate) field: StringId,
+    pub(crate) arguments: Vec<(StringId, Value)>,
```
Finistere, line 151: Hum, unless I'm mistaken you're not validating the arguments, no?
IMHO composition should validate those and not bother with GraphQL coercion rules, it should just match.
Doesn't sound obvious to me that you'll be able to store it in a better format (`Vec<(InputValueDefinitionId, Value)>` typically) considering you're reading from SDL. But well

Though saying this but when providing a SDL schema in the CLI it'll bypass composition, so I'll have to validate in all cases. Errors at composition would be a lot friendlier for the user at least

Finistere (approved):
Nice

engine/crates/composition/src/subgraphs/keys.rs
tomhoule: Yes I agree it's probably easier to validate at composition time, and worth the extra effort to put an InputValueDefinitionId there. I'll work on that.

tomhoule: Done & pushed, let's see what the tests say.

PR #1477, opened 2024-03-20
chore(engine-v2): Simplify SchemaInputValue

Finistere:
Now that input values for are separated I can simplify the input values
stored in the schema.

I'm removing the `UnknownEnumValue` case and the `Other` directives.
Those are schema directives, if we can't handle them they won't be used
anywhere else.


engine/crates/engine-v2/Cargo.toml
```diff
@@ -10,6 +10,9 @@ readme = "README.md"
 repository = "https://github.com/grafbase/grafbase"
 version = "3.0.31"
 
+[lib]
```
Finistere, line 13: Just annoyed at `cargo test -p engine-v2` failing because cargo is trying to look for tests when they aren't any.

engine/crates/engine-v2/schema/src/input_value/tests.rs
```diff
@@ -0,0 +1,363 @@
+use std::borrow::Cow;
+
+use serde::Deserialize;
+use wrapping::Wrapping;
+
+use crate::{EnumValue, InputValue, InputValueDefinition, Schema, StringId, Type};
+
+use super::*;
+
+fn create_schema() -> Schema {
```
Finistere, line 10: moved existing stuff

engine/crates/engine-v2/Cargo.toml
obmarg: Just curious here, but what kind of failure?  Literally never seen this

obmarg (approved):

engine/crates/engine-v2/Cargo.toml
Finistere: if you run `cargo test -p engine-v2` it'll try to run doctest on stuff like this https://github.com/grafbase/grafbase/blob/main/engine/crates/engine-v2/src/response/key/mod.rs#L4
and fails

but also realizing that I could just wrap the relevant parts in "```text"... will remove it 😅

obmarg: You can also do ```ignore

obmarg: (or ```no_run if you want to compile it but you don't want to run it)

PR #1481, opened 2024-03-22
composition: implement progressive override

tomhoule:
This consists of a new `label` argument where you can specify a percentage
of traffic to direct to the overriding subgraph.

closes GB-6126

### DIFF
--- a/engine/crates/federated-graph/src/federated_graph/v1.rs
+++ b/engine/crates/federated-graph/src/federated_graph/v1.rs
@@ -230,6 +230,41 @@ pub struct Override {
     /// Points to a subgraph referenced by name, but this is _not_ validated to allow easier field
     /// migrations between subgraphs.
     pub from: OverrideSource,
+    #[serde(default)]
+    pub label: OverrideLabel,
+}
+
+#[derive(serde::Serialize, serde::Deserialize, Clone, Default, Debug)]
+pub struct OverrideLabel {
+    pub percent: Option<u8>,
+}
+
+impl std::fmt::Display for OverrideLabel {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        if let Some(percent) = self.percent {
+            f.write_str("percent(")?;
+            percent.fmt(f)?;
+            f.write_str(")")?;
+        }
+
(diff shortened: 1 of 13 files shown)


engine/crates/composition/tests/composition/override_basic/federated.graphql
```diff
@@ -23,9 +23,9 @@ enum join__Graph {
 
 type Mammoth {
     tuskLength: Int
-    weightGrams: Int @join__field(graph: mangrove, overrides: "steppe")
+    weightGrams: Int @join__field(graph: mangrove, override: "steppe")
```
tomhoule, line 26: We were rendering that wrong. You'll see in from_sdl() that we still accept "overrides" for backward compatibility, but "override" is correct.

engine/crates/federated-graph/src/federated_graph/v1.rs
```diff
@@ -230,6 +230,41 @@ pub struct Override {
     /// Points to a subgraph referenced by name, but this is _not_ validated to allow easier field
     /// migrations between subgraphs.
     pub from: OverrideSource,
+    #[serde(default)]
+    pub label: OverrideLabel,
```
Finistere, line 234: would use an enum here, hard to know what will be in this `label` in the future.

Finistere (approved):
Nice 👍

engine/crates/federated-graph/src/federated_graph/v1.rs
tomhoule: Yes this struct felt a bit awkward in different places, let's do that.

Finistere: I mean transform `OverrideLabel` into an enum, maybe doesn't matter

tomhoule: yep that's how I understood it. I'm wondering if we should already have a fallback variant for forward compatibility.

Finistere: Up to you, won't do anything with it engine-side 🤷 
would avoid from_sdl -> to_sdl to change anything though

PR #1489, opened 2024-03-22
fix: more thorough validation of join arguments

obmarg:
Prior to this change the arguments we pass in to a join were not validated very much.  The assumption was that any errors would be caught when someone tried to run the join.  But, now that joins are getting more complicated and variables present in them can come from more places, it makes sense to put the effort into improving this a bit.

And, that's what this commit does.  We now check that variables exist, check that they have compatible types and also similarly check that any literals in the string approximately make sense.  I don't expect that we'll catch every single error with this, but it should catch the dumbest of them.  In particular I've been quite lax on validating scalars - I don't want to stop block things that are technically wrong but in reality fine (e.g. passing a String to an ID field or vice versa).  It could probably be made a little stricter, but lets get this out there and see how it does.

One thing to note is that i haven't done list coercion here.  Mostly because it's complex enough as it is and I don't want to add that extra complexity.  If someone really wants it we can revisit.

Fixes GB-6305

### DIFF
--- a/engine/crates/parser-sdl/src/validations.rs
+++ b/engine/crates/parser-sdl/src/validations.rs
@@ -309,7 +309,7 @@ fn validate_join_argument(
         let error_prefix = JoinArgumentErrorPrefix(join_coord, argument_definition.name.as_str(), &path);
         match value {
             SerializableArgument::Variable(variable) => {
-                let Some(expected_type) = available_variables.get(variable.as_str()) else {
+                let Some(variable_type) = available_variables.get(variable.as_str()) else {
                     errors.push(RuleError::new(
                         vec![],
                         format!(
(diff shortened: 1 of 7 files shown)


engine/crates/parser-sdl/src/validations.rs
```diff
@@ -462,14 +462,28 @@ fn types_are_compatible(mut actual: &Type, mut expected: &Type, registry: &Regis
...
+                    Some((MetaType::Scalar(_), MetaType::Scalar(_))) => {
                         // If the names don't match but both sides are scalars we'll say they're compatible.
                         // This maybe isn't strictly correct but a bit of flexibility around
                         // passsing strings to ID fields and handling mismatched custom scalars
                         // seems sensible.
                         return true;
                     }
+                    Some(
+                        (MetaType::InputObject(_), MetaType::Object(_))
+                        | (MetaType::Object(_), MetaType::InputObject(_)),
```
obmarg, line 479: Might need to do something similar around enums here to handle trip advisors case :|

Finistere (approved):

obmarg:
I don't think this can be merged in its current form, so making draft till I get time to fix.

PR #1494, opened 2024-03-24
chore(engine-v2): improve schema builder

Finistere:
Extracting most definitions to `Graph` struct which is built with introspection data sources. Avoids the need to have an `Option` for the introspection data source even though it's always present. And cleaned up a bunch of code.

### DIFF
--- a/engine/crates/engine-v2/schema/src/builder/coerce/mod.rs
+++ b/engine/crates/engine-v2/schema/src/builder/coerce/mod.rs
@@ -252,7 +252,7 @@ impl<'a> InputValueCoercer<'a> {
 
     fn type_name(&self, ty: Type) -> String {
         let mut s = String::new();
-        for _ in ty.wrapping.list_wrappings().rev() {
+        for _ in 0..ty.wrapping.list_wrappings().len() {
             s.push('[');
         }
         s.push_str(match ty.inner {
--- a/engine/crates/engine-v2/schema/src/builder/coerce/mod.rs
+++ b/engine/crates/engine-v2/schema/src/builder/coerce/mod.rs
@@ -2,7 +2,7 @@ mod error;
 mod path;
 
 use crate::{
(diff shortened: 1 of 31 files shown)


engine/crates/engine-v2/schema/src/builder/external_sources.rs
```diff
@@ -0,0 +1,50 @@
+use std::mem::take;
+
+use config::latest::Config;
+
+use crate::sources;
+
+use super::BuildContext;
+
+pub struct ExternalDataSources {
```
Finistere, line 9: Basically every data sources except Introspection which we need to build after.

engine/crates/engine-v2/schema/src/builder/ids.rs
```diff
@@ -30,11 +31,45 @@ where
     }
 }
 
-#[derive(Default)]
-pub struct IdMaps {
+pub(super) struct IdMaps {
```
Finistere, line 34: Initializing `IdMaps` (federated_graph ids -> engine_v2 ids) immediately with the right values with `new()`. Avoids the pitfall of trying to get the associated id but the idmap wasn't fully initialized yet.

engine/crates/engine-v2/schema/src/builder/coerce/mod.rs
```diff
@@ -235,12 +245,37 @@ impl<'a> InputValueCoercer<'a> {
...
             actual: value.into(),
-            expected: scalar.name().to_string(),
+            expected: self.ctx.strings[self.graph[scalar_id].name].to_string(),
             path: self.path(),
         })
     }
 
+    fn type_name(&self, ty: Type) -> String {
+        let mut s = String::new();
+        for _ in ty.wrapping.list_wrappings().rev() {
```
yoav-lavi, line 255: This `.rev()` can be removed

PR #1497, opened 2024-03-25
validation: revert newly added validation against empty objects, interfaces, input objects

tomhoule:
The validation is prescribed by the GraphQL spec, but we may be breaking existing user deployments with it. We should introduce it more carefully. This commit can be reverted to re-enable the validation.


tomhoule:
Change of plan. Closing this.

PR #1499, opened 2024-03-25
Typo fixes

yoav-lavi:
# Description

## Fix

- Typos

# Type of change

- [ ] 💔 Breaking
- [ ] 🚀 Feature
- [x] 🐛 Fix
- [ ] 🛠️ Tooling
- [ ] 🔨 Refactoring
- [ ] 🧪 Test
- [ ] 📦 Dependency


cli/crates/cli/src/check.rs
```diff
@@ -108,7 +108,7 @@ fn git_sha() -> Option<String> {
 }
 
 fn git_branch() -> Option<String> {
-    let output = Command::new("git").arg("branch").arg("--show-curent").output().ok()?;
+    let output = Command::new("git").arg("branch").arg("--show-current").output().ok()?;
```
tomhoule, line 111: Hmm so this probably wasn't working

cli/crates/backend/src/api/graphql/api.graphql
```diff
@@ -229,7 +229,7 @@ type BranchOperationCheckConfiguration {
   """
   enabled: Boolean!
   """
-  The time range in days to consider for operation checks. Operations older than the specificied
```
obmarg, line 232: This file is generated iirc, probably not the right place to fix?

cli/crates/cli/tests/graphql-directive/server.rs
```diff
@@ -64,7 +64,7 @@ impl Query {
             }));
         } else if id == "3" {
             return Some(PullRequestOrIssue::Issue(Issue {
-                title: "Everythings fine".into(),
+                title: "Everything's fine".into(),
```
tomhoule, line 67: nice

tomhoule (approved):

cli/crates/cli/src/check.rs
```diff
@@ -108,7 +108,7 @@ fn git_sha() -> Option<String> {
 }
 
 fn git_branch() -> Option<String> {
-    let output = Command::new("git").arg("branch").arg("--show-curent").output().ok()?;
```
obmarg, line 111: Clearly this needs better tests

cli/crates/cli/tests/graphql-directive/server.rs
```diff
@@ -64,7 +64,7 @@ impl Query {
             }));
         } else if id == "3" {
             return Some(PullRequestOrIssue::Issue(Issue {
-                title: "Everythings fine".into(),
```
obmarg, line 67: This kind of thing is why I have reservations about installing this tool on CI.

This is "wrong" but in a way that doesn't matter in the slightest.  I'd get extremely annoyed if I had to wait another 30 minutes on CI because it caught this.

cli/crates/cli/tests/graphql-directive/server.rs
yoav-lavi: This is the false fix I mentioned, it fixed it to "Everything fine" (which is ok but not exact)

cli/crates/cli/tests/owner.rs
```diff
@@ -113,7 +113,7 @@ mod global {
                     .send()
                     .await
             );
-            // an attemt by user2 to delete the todo should fail.
```
obmarg, line 116: Same comment here, its wrong but in a "who cares" kind of way

obmarg (approved):

cli/crates/cli/tests/graphql-directive/server.rs
yoav-lavi: I commented in Slack, it would be better as a periodic check since it has false positives in the engine. But we should run it occasionally

yoav-lavi: Note that most of these were automatic fixes

PR #1501, opened 2024-03-25
feat(engine-v2): add argument support for @requires part2

Finistere:
This PR is bigger than expected, had to change a bit how I deal with resolvers. Now I have a similar data model to `FedeartedGraph` for `FieldRequires` & `FieldProvides`, associating them with a subgraph. And added `only_resolvable_in` which restricts the subgraphs in which a field is resolvable if not empty.

The meaning of subgraph in engine-v2 is very similar, but it's not limited to federated subgraphs. Introspection is also considered to be a subgraph. Connectors/resolvers/etc. would also. I've also changed a bit the `engine-v2-schema` `sources::federation::DataSource` to be `sources::graphql::GraphqlEndpoints` to remove the notion of subgraphs in there. They're just Graphql endpoints that happen that support federation entity resolvers from the engine-v2 perspective.

I've updated the planning with those changes but separated the argument handling into the next PR.


engine/crates/engine-v2/schema/src/sources/graphql.rs
```diff
@@ -0,0 +1,156 @@
+use url::Url;
+
+use crate::{HeaderId, HeaderWalker, RequiredFieldSet, RequiredFieldSetId, SchemaWalker, StringId, SubgraphId, UrlId};
+
+#[derive(Default)]
```
Finistere, line 5: Pretty much the same a `sources/federation.rs` that was removed, just with renames

jakubadamw (approved):
I don't see any explosives.

PR #1504, opened 2024-03-25
cli: validate the registry in grafbase dev

tomhoule:
Since we use graphql_schema_validation on the generated schemas in the API, we should also validate in `grafbase dev` to tighten the feedback loop.

The engine tends to produce empty types, so we also have to start pruning these when finalizing the registry.

closes GB-6335

### DIFF
--- a/engine/crates/engine/src/registry/mod.rs
+++ b/engine/crates/engine/src/registry/mod.rs
@@ -1914,6 +1914,22 @@ impl Registry {
         }
     }
 
+    pub fn remove_empty_types(&mut self) {
+        let mut to_be_removed = Vec::new();
+
+        for r#type in self.types.values() {
+            match r#type.fields() {
+                None => to_be_removed.push(r#type.name().to_owned()),
+                Some(fields) if fields.is_empty() => to_be_removed.push(r#type.name().to_owned()),
+                Some(_) => (),
+            }
+        }
+
+        for type_name in to_be_removed {
+            self.types.remove(&type_name);
+        }
+    }
+
     pub fn find_visible_types(&self, ctx: &ContextField<'_>) -> HashSet<&str> {
         let mut visible_types = HashSet::new();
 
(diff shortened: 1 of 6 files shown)


engine/crates/engine/src/registry/mod.rs
```diff
@@ -1914,6 +1914,22 @@ impl Registry {
...
+                None => to_be_removed.push(r#type.name().to_owned()),
+                Some(fields) if fields.is_empty() => to_be_removed.push(r#type.name().to_owned()),
+                Some(_) => (),
+            }
+        }
+
+        for type_name in to_be_removed {
+            self.types.remove(&type_name);
+        }
+    }
```
obmarg, line 1931: I wonder (though I'm not sure) whether this also needs to consider places that the removed types are used....

engine/crates/parser-sdl/src/rules/visitor/context.rs
```diff
@@ -184,6 +185,7 @@ impl<'a> VisitorContext<'a> {
         }
 
         registry.remove_unused_types();
+        registry.remove_empty_types();
```
obmarg, line 188: I wonder about the position of this call as well actually.  This is going to be called whenever _any_ registry is finalised, so if a user gives us `type Blah` without any fields we'll just silently remove `Blah` and never complain.

I wonder if this call would be better off called against `src_registry` here: https://github.com/grafbase/grafbase/blob/main/engine/crates/parser-sdl/src/connector_parsers.rs#L63 so it only acts on connector generated registries and not on anything a user gives us....

tomhoule: Good point, that indeed sounds like a better solution.

engine/crates/engine/src/registry/mod.rs
tomhoule: I wondered about that. Currently, that would lead to a validation error right after the registry is generated. It's hard to say if it's going to be a problem in practice.

obmarg: I suppose this change is unlikely to make anything _more_ broken, so maybe we can go for it as is and improve if it's not good enough.

engine/crates/parser-sdl/src/rules/visitor/context.rs
tomhoule: done

PR #1509, opened 2024-03-26
Implement trusted documents configuration in SDK and registry

tomhoule:
This is part of the work to enable trusted documents for managed standalone graphs.

The config looks like this:

```typescript
config({
      graph: g,
      trustedDocuments: {
        enabled: true
      }
    })
```

closes GB-6346


yoav-lavi (approved):
Nice!

PR #1510, opened 2024-03-26
composition: fix bug on entity interfaces, support more directives

tomhoule:
Fields of entity interfaces were duplicated in entities defined in the
same subgraph as the entity interface, because we were adding the fields
from the interface _again_, while they were already guaranteed to be
there by the rules of GraphQL validation.

Also, we were not taking into account `@requires`, `@overrides` and
`@provides` on fields of objects with `@interfaceObject` in other subgraphs.# Description


Finistere (approved):
Thanks!

PR #1513, opened 2024-03-26
chore(engine-v2): Add authenticated & requiresScopes support

Finistere:
Reworking directives to be friendlier to use and move cache logic into them. Adding `@authenticated` & `@requiresScopes` support from federation 2.5


engine/crates/engine-v2/schema/src/walkers/definition.rs
```diff
@@ -106,6 +107,17 @@ impl<'a> DefinitionWalker<'a> {
...
+
+    pub fn directives(&self) -> TypeSystemDirectivesWalker<'a> {
+        match self.item {
+            Definition::Scalar(s) => self.walk(s).directives(),
+            Definition::Object(o) => self.walk(o).directives(),
+            Definition::Interface(i) => self.walk(i).directives(),
+            Definition::Union(u) => self.walk(u).directives(),
+            Definition::Enum(e) => self.walk(e).directives(),
+            Definition::InputObject(io) => self.walk(io).directives(),
+        }
```
tomhoule, line 119: Sometimes I think we are missing a `Definition` table in FederatedGraph.

engine/crates/graphql-mocks/src/secure.rs
```diff
@@ -0,0 +1,66 @@
+use async_graphql::{EmptyMutation, EmptySubscription, Object, TypeDirective};
+
+/// A schema that only uses String types.
+///
+/// This is used to make sure that we're not pruning built in scalars that aren't used
```
tomhoule, line 5: I don't understand this doc comment

engine/crates/integration-tests/tests/federation/auth/authenticated.rs
```diff
@@ -0,0 +1,97 @@
...
+          "data": {
+            "mustBeAuthenticated": "You are authenticated"
+          }
+        }
+        "###);
+    });
+}
+
+#[test]
+fn not_authenticated() {
```
tomhoule, line 69: I'm reading the tests first so maybe it becomes later, but if only part of the query is on `@authenticated` fields and the rest is not, are we still returning data with nulls for the authenticated fields? This test is correct because the whole query would be null'ed.

engine/crates/integration-tests/tests/federation/auth/requires_scopes.rs
```diff
@@ -0,0 +1,264 @@
...
+
+#[test]
+fn anonymous_does_not_any_scope() {
+    with_secure_schema(|engine| async move {
+        let response = engine.execute("query { mustHaveReadScope }").await;
+        insta::assert_json_snapshot!(response, @r###"
+        {
+          "errors": [
+            {
+              "message": "Invalid OAuth2 scope",
```
tomhoule, line 12: Nitpick, but a "not authenticated" error here would be closer to the root cause.

engine/crates/engine-v2/src/sources/introspection/writer.rs
```diff
@@ -272,14 +272,8 @@ impl<'a> IntrospectionWriter<'a> {
...
-            __Field::IsDeprecated => target.is_deprecated().into(),
-            __Field::DeprecationReason => target
-                .directives()
-                .find_map(|directive| match directive {
-                    Directive::Deprecated { reason } => *reason,
-                    _ => None,
-                })
-                .into(),
+            __Field::IsDeprecated => target.directives().has_deprecated().into(),
+            __Field::DeprecationReason => target.directives().deprecated().map(|d| d.reason).into(),
```
tomhoule, line 276: much nicer

tomhoule (approved):
:100:

engine/crates/graphql-mocks/src/secure.rs
Finistere: ah copy pasted and didn't pay attention to the comment, it's irrelevant.

engine/crates/integration-tests/tests/federation/auth/authenticated.rs
Finistere: hum, excellent point... Excellent catch. Definitely should be.

engine/crates/integration-tests/tests/federation/auth/requires_scopes.rs
Finistere: I wouldn't say closer to root scope. The user has an access token. It's missing the right permissions to access this field. We could replace it with a `Not allowed` message though

tomhoule: Ah I didn't get that the user is authenticated. All good then :+1:

engine/crates/engine-v2/schema/src/walkers/definition.rs
Finistere: don't think it would be useful, you have a definition abstraction in the composition step. FederatedGraph is a storage format from my perspective.

PR #1521, opened 2024-03-27
feat: join support for mongo db connector

obmarg:
A potential customer was trying to use our join directive to hook up two mongo db models up, like so:

```typescript
g.extend("Customer", (extend) => {
  extend.addField(
    "customerContacts",
    g.ref("CustomerConnection").optional().join(
      "mongoDB { contactCollection(first: 10, filter:{businessid:{eq: $customerid}})"
    )
  )
})
```

But this was giving them the following error:

> Field customerContacts got an undefined type: CustomerConnection

This particular error is because the mongo connector doesn't put its generated types into `ctx.types` which is where the code responsible for the above error looks for types.  I've fixed that by updating the undefined type code to also look in `registry.types` instead of just `ctx.types`.

After fixing that I noticed that the connector doesn't parse join directives, so I added support for that.

Then my test started complaining that the `first` argument wasn't provided, which turned out to be because some code involved in joins was turning every number into a float, and that argument is an integer.  So some serde code that's involved was throwing an error which was being silently turned into `None`.  So that was fun.

I also updated a lot of the type generation code to filter out resolver & join fields from the generated input types, because users probably shouldn't be able to write fields that are computed.

Finally, I also discovered some ordering issues that are far too big a job for me to fix today.  For now users probably just need to be careful to define types before they're used.

### DIFF
--- a/engine/crates/engine/src/validation/utils.rs
+++ b/engine/crates/engine/src/validation/utils.rs
@@ -43,7 +43,10 @@ pub fn is_valid_input_value(
 ) -> Option<String> {
     match registry::MetaTypeName::create(type_name) {
         registry::MetaTypeName::NonNull(type_name) => match value {
-            ConstValue::Null => Some(valid_error(&path, format!("expected type \"{type_name}\""))),
+            ConstValue::Null => Some(valid_error(
+                &path,
+                format!("expected type \"{type_name}\" but found null"),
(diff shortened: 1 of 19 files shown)


pimeys (approved):
OMG, this looks really painful SDL-wise. Immediately approved! :D

PR #1522, opened 2024-03-27
federated-graph: new export to generate final API schema from FederatedGraph

tomhoule:
The purpose is for us to stop relying on introspection to generate these schemas. The implementation reuses bits of the existing `render_sdl()` (which is renamed to `render_federated_sdl()` for clarity), but mostly, it is new. In this commit, we modify the composition test setup to reuse the test schemas we already have there.

closes GB-5666


engine/crates/composition/tests/composition/tag_directive_basic/api.graphql
```diff
@@ -0,0 +1,31 @@
...
+    HONEYCRISP @tag(name: "appleEnumValue")
+}
+
+type Apple @tag(name: "appleType") {
+    id: ID!
+    texture: Texture
+    variety: AppleVariety
+}
+
+type Orange @tag(name: "orangeType") {
```
tomhoule, line 13: Hmm I think `@tag` shouldn't be in the public API. right?

engine/crates/composition/tests/composition/authenticated_basic/api.graphql
```diff
@@ -0,0 +1,10 @@
+type Car {
```
Finistere, line 1: why isn't `@authenticated` part of the api schema  if we show `@deprecated` and `@tag`?

engine/crates/composition/tests/composition/requiresScopes_basic/api.graphql
```diff
@@ -0,0 +1,19 @@
+type B {
```
Finistere, line 1: I think I would leave `@requiresScopes`

engine/crates/composition/tests/composition/requires_basic/api.graphql
```diff
@@ -0,0 +1,21 @@
+type Farm {
```
Finistere, line 1: I'd consider leaving `@requires`. Contrary to `@provides`, it's not an optimization, it's more related to business logic.

engine/crates/composition/tests/composition/tag_directive_basic/api.graphql
Finistere: No idea for `@tag` to honest. I don't think we should treat it as a public API. The public API is introspection. This is the schema we show to the maintainer/developers of the project. So they need to be aware of the business logic that affects runtime behavior IMHO. We should only strip out the federation aspects (keys, provides, inaccessible, override, etc.)

engine/crates/federated-graph/src/render_sdl/render_api_sdl.rs
```diff
@@ -0,0 +1,246 @@
...
+        .any(|directive| matches!(directive, Directive::Inaccessible))
+}
+
+fn write_public_directives(
+    f: &mut fmt::Formatter<'_>,
+    directives: Directives,
+    graph: &FederatedGraphV3,
+) -> fmt::Result {
+    for directive in graph[directives].iter().filter(|directive| match directive {
+        Directive::Authenticated | Directive::Inaccessible | Directive::Policy(_) | Directive::RequiresScopes(_) => {
```
Finistere, line 228: as mentioned, would include `@authenticated`, `@requiresScopes` and `@requires`. Probably wouldn't bother with `@policy` though, we can't use it...

Finistere (approved):

engine/crates/composition/tests/composition/authenticated_basic/api.graphql
tomhoule: I erred towards removing anything that is federation-specific / is only meant for the engine, not for clients. But there’s really no deep rationale, let’s go with your suggestions.

PR #1540, opened 2024-04-01
feat:cache span

hfgbarrigas:
# Description

Add a new cache span to capture any cache read status

# Type of change

- [ ] 💔 Breaking
- [x] 🚀 Feature
- [ ] 🐛 Fix
- [ ] 🛠️ Tooling
- [ ] 🔨 Refactoring
- [ ] 🧪 Test
- [ ] 📦 Dependency


engine/crates/integration-tests/tests/tracing/cache.rs
```diff
@@ -0,0 +1,106 @@
...
+    fn metadata(&self) -> CacheMetadata {
+        CacheMetadata::default()
+    }
+}
+
+#[test]
+pub fn cache_status() {
+    runtime().block_on(async {
+        // prepare
+        let cache = runtime::cache::Cache::new(
```
Finistere, line 75: It would be really better if you tested the engine rather than `Cache`, it's an implementation detail and you rely way too much on the internal API here IMHO. This test shouldn't change if the internal cache API is changed.

Finistere (approved):

engine/crates/integration-tests/tests/tracing/cache.rs
hfgbarrigas: I tend to disagree, I'm adding a specific span to our `runtime::Cache` and that's what I'm testing. It's supposed to be agnostic to what uses it, testing the cache through something that uses adds an unnecessary layer. 
I'm not sure what you consider internal API, everything I'm use has `pub` visibility on `runtime::cache`.

Finistere: >  I'm not sure what you consider internal API, everything I'm use has pub visibility on runtime::cache.

Our public API that our users use is the SDK & GraphQL API exposed by the gateway. Anything else is an implementation detail. `pub` in this context only allows us to define our internal APIs, such as `runtime::cache::Cache` which is an abstraction of the cache for the engine as it depends on the environment. I typically did change the cache interface during my gateway refactor for example. I had also removed the gateway. Changing it now would require me to rewrite this entire test.

Effectively this means you're not testing that Grafbase gateway configured with cache will generate a cache span. You're testing that the `cached_execution` method produces a cache span. While it's equivalent today, it's not *inherently* the same thing. Hence why any change to any of the traits you're implementing would require changing this test, even if the behavior the user expects never changed.

Testing through the engine is pretty easy, all the integrations are about this. Not harder than the 50 loc you added to implement traits at least. 🤷

But well, won't insist more than that

PR #1561, opened 2024-04-05
chore: only use `codegen-units=1` for lambda releases

obmarg:
We recently added `codegen-units=1` to the release build config.  This was aimed at reducing the size of the binaries that we ship for lambdas, and it was successful in that regard.  It slows release builds down quite significantly though, which has had an impact on CI times, as you can see on this graph.

CleanShot 2024-04-05 at 12 01 13@2x

That graph shows a day or two between the merge & a noticeable increase in CI times, but the merge was on friday and I assume it wasn't until the monday that every other piece of work had rebased onto `main` and pulled in the change.

This PR makes two changes:

- Creates a new cargo profile `lambda` that with `codegen-units=1` and updates CI to use this profile for building the lambda.  This shaved 5-6 minutes off.
- I then updated the lambda build to only use this profile when doing a tagged release build, which shaved another ~2:30 off.  Although I'm open to discussions if we'd rather not do this bit.

### DIFF
--- a/.github/workflows/binaries.yml
+++ b/.github/workflows/binaries.yml
@@ -288,7 +288,7 @@ jobs:
       - name: Build gateway lambda release
         shell: bash
         run: |
-          RUSTFLAGS="-C opt-level=s" cargo build --release -p grafbase-gateway --target ${{ matrix.archs.target }} --features lambda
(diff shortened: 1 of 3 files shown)


tomhoule (approved):
:100:

PR #1563, opened 2024-04-05
fix(mongo): normalize elemMatch correctly in AND/OR blocks

pimeys:
Part of: GB-6423


tomhoule (approved):
I think I get it, lgtm :+1:

PR #1566, opened 2024-04-05
feat: export otel metrics with same config as traces

Finistere:
I guess at some point we might want to have a different config, but we'll see then.
Re-using the same structure as traces with `tracing` integration.

The request metric is simple and basic, the goal is only to have a working integration test right now.


engine/crates/tracing/src/otel/metrics.rs
```diff
@@ -0,0 +1,90 @@
...
+            .map_err(|e| TracingError::MetricsExporterSetup(e.to_string()))?;
+        let reader = PeriodicReader::builder(exporter, runtime.clone())
+            .with_interval(Duration::from_secs(10))
+            .with_timeout(Duration::from_secs(10))
+            .build();
+
+        provider = provider.with_reader(reader);
+    }
+
+    opentelemetry::global::set_meter_provider(provider.build());
```
hfgbarrigas, line 59: we shouldn't be doing this here. We should return the provider back to the user of this crate, and that's where it should decided if its global or not.

gateway/crates/federated-server/src/server/engine.rs
```diff
@@ -36,6 +37,7 @@ pub(super) async fn post(
     headers: HeaderMap,
     Json(request): Json<engine::BatchRequest>,
 ) -> impl IntoResponse {
+    state.request_counter().add(1, &[]);
```
hfgbarrigas, line 40: I'm guessing the `&[]` is for metric tags? would be nice to know the http method and response status.

Finistere: It's just a dummy request counter to have something to test against in the integration test^^ Will be improved later

Finistere: and yeah those are the attributes

engine/crates/tracing/src/otel/metrics.rs
```diff
@@ -0,0 +1,79 @@
...
+            .with_timeout(Duration::from_secs(10))
+            .build();
+
+        provider = provider.with_reader(reader);
+    }
+
+    #[cfg(feature = "otlp")]
+    if let Some(config) = config.exporters.otlp.as_ref().filter(|cfg| cfg.enabled) {
+        use opentelemetry_otlp::MetricsExporterBuilder;
+        let exporter = crate::otel::provider::build_otlp_exporter::<MetricsExporterBuilder>(config)?
```
hfgbarrigas, line 39: I'd move the `build_otlp_exporter` to a separate mod (`exporter`?) and rename the existing `provider` to `tracing`.

hfgbarrigas (approved):
Nice, left some nits

engine/crates/tracing/src/error.rs
```diff
@@ -1,10 +1,28 @@
...
     FileReadError(std::io::Error),
     /// Error configuring an exporter
     #[error("unable to configure span exporter: {0}")]
     SpanExporterSetup(String),
+    /// Error configuring a metric exporter
+    #[error("unable to configure metrics exporter: {0}")]
+    MetricsExporterSetup(String),
+}
+
+impl From<String> for TracingError {
```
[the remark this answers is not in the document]
Finistere: Conflicts with `impl<T> From<T> for T` IIRC because the error has `to_string()`

engine/crates/tracing/src/otel/metrics.rs
Finistere: done, renaming to `traces` rather than `tracing` though

engine/crates/tracing/src/error.rs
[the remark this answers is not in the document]
hfgbarrigas: really? just replaced with ```impl<T> From<T> for TracingError where T: Into<String>{
    fn from(s: T) -> Self {
        TracingError::Internal(s.into())
    }
}
``` and cargo check --all-features didn't complain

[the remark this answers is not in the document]
Finistere: hum, confused with `std::fmt::Display` then, my bad

PR #1571, opened 2024-04-08
composition: implement composition of field argument defaults

tomhoule:
The rule is that they have to match exactly between subgraphs, otherwise it's an error.

closes GB-6159

### DIFF
--- a/engine/crates/composition/src/compose/object.rs
+++ b/engine/crates/composition/src/compose/object.rs
@@ -99,6 +99,14 @@ pub(super) fn merge_field_arguments<'a>(
     ids.unwrap_or(federated::NO_INPUT_VALUE_DEFINITION)
 }
 
+/// Default values on arguments (e.g. `field(arg: String! = "N.A")`) are _not_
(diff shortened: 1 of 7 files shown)


engine/crates/composition/src/compose/object.rs
```diff
@@ -97,6 +99,32 @@ pub(super) fn merge_field_arguments<'a>(
...
 }
 
+fn compose_field_argument_defaults(
+    arguments: &[(StringId, subgraphs::FieldArgumentWalker<'_>)],
+    ctx: &mut Context<'_>,
+) {
+    let mut default: Option<(&subgraphs::Value, subgraphs::FieldArgumentWalker<'_>)> = None;
+
+    for (_, argument) in arguments {
+        let Some(value) = argument.default() else { continue };
```
obmarg, line 109: If subgraph A doesn't have a default and subgraph B does, should that be an error?  This looks like it'd allow it (though correct me if I'm misunderstanding)

obmarg (approved):

PR #1576, opened 2024-04-10
composition: validate that keys match in entity interfaces

tomhoule:
Before this commit, we were only validating that the implementers of the interface — in the subgraph where the interface is defined — have the same key, but the error message mentioned @interfaceObject, which is not about implementers, but about types in other subgraphs.

This commit

- resolves the confusion: the error message now mentions the right problem
- implements validation of matching key for interface objects in other subgraphs

closes gb-5825


Finistere (approved):
nice

PR #1578, opened 2024-04-10
feat: predefined subgraphs in federated dev config

obmarg:
I've been tasked with writing a guide for setting up federation with gqlgen (a go library).  We noticed when looking at existing guides that it's nice to be able to just show users a config file for their federated development server.  That config can define all the development subgraphs, saving the guide from explaining how to register 3 separate subgraphs.

This brings that predefined subgraph functionality into `grafbase dev`.  We'll watch the federated config for changes, and any new or changed subgraphs will be introspected and then composed.  If a subgraph isn't up/working at the time of introspection  we'll log that and then retry in the background - so if the subgraph later comes up we'll add it into the federated dev setup.

Part of GB-6127

I still have the SDK part of this to do, that'll be my next PR.


engine/crates/parser-sdl/src/rules/subgraph_directive.rs
```diff
@@ -34,6 +39,13 @@ impl Directive for SubgraphDirective {
           "The name of the subgraph"
           name: String!
 
+          """
+          The URL to use for this API in development
+
+          In deployed environments this is ignored.
+          """
+          websocketUrl: String!
```
tomhoule, line 47: I think you meant `developmentUrl` here

tomhoule (approved):

PR #1583, opened 2024-04-10
feat: support caching for interfaces

hfgbarrigas:
# Description

As per title.

# Type of change

- [ ] 💔 Breaking
- [x] 🚀 Feature
- [ ] 🐛 Fix
- [ ] 🛠️ Tooling
- [ ] 🔨 Refactoring
- [ ] 🧪 Test
- [ ] 📦 Dependency

### DIFF
--- a/engine/crates/tracing/src/otel/exporter.rs
+++ b/engine/crates/tracing/src/otel/exporter.rs
@@ -1,10 +1,8 @@
-#[cfg(feature = "otlp")]
 use crate::{
     config::{TracingOtlpExporterConfig, TracingOtlpExporterProtocol},
     error::TracingError,
 };
 
-#[cfg(feature = "otlp")]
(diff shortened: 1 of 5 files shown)


engine/crates/tracing/src/otel/exporter.rs
```diff
@@ -1,16 +1,16 @@
-use std::{str::FromStr, time::Duration};
-
+#[cfg(feature = "otlp")]
```
obmarg, line 1: This intentional?  Might have been worth making it a module attr if so `#![cfg(feature = "otlp")]`

obmarg (approved):

engine/crates/tracing/src/otel/exporter.rs
hfgbarrigas: typo! good catch

hfgbarrigas: for a moment there I thought I had a typo in the #[cfg]. Moved it to `otel.rs` instead and removed it from here.

PR #1596, opened 2024-04-12
feat: a gqlgen & federation example

obmarg:


examples/gqlgen-federation/readme.md
```diff
@@ -0,0 +1 @@
+TODO
```
obmarg, line 1: Need to do this

examples/gqlgen-federation/gateway/package.json
```diff
@@ -0,0 +1,15 @@
...
+  "description": "",
+  "main": "index.js",
+  "scripts": {
+    "test": "echo \"Error: no test specified\" && exit 1"
+  },
+  "keywords": [],
+  "author": "",
+  "license": "ISC",
+  "devDependencies": {
+    "@grafbase/sdk": "~0.20.0"
```
fbjork, line 13: add grafbase as a dependency here and start/dev scripts?

fbjork: another idea is to use the grafbase federation gateway and pass the federated schema SDL via path?

obmarg: Yeah, could do.  Though I was just using `npx grafbase@latest` in `start.sh`, which seemed less annoying (having to update versions etc.)

obmarg: > another idea is to use the grafbase federation gateway and pass the federated schema SDL via path?

Maybe, but then I'd have to hardcode/script/instruct people how to build the federated schema SDL.  I can do that if you think we should though

fbjork: True, better to keep the dev server then.

fbjork (approved):

PR #1603, opened 2024-04-15
cli: return early with an error on missing schema in grafbase check

tomhoule:
If you run grafbase check without the --schema argument, the CLI will wait for a schema from stdin, giving the impression it is hanging. Similar to what we already did for `grafbase publish`, this commit makes the CLI exit early in this situation by detecting if stdin is a tty.

Also replaces the `atty` crate in the CLI with the new `IsTerminal` trait from the standard library (thanks @obmarg for the tip).

closes GH 6494

### DIFF
--- a/cli/crates/cli/src/check.rs
+++ b/cli/crates/cli/src/check.rs
@@ -1,6 +1,10 @@
 use crate::{cli_input::CheckCommand, errors::CliError, report};
 use backend::api::check;
-use std::{fs, io::Read, process::Command};
+use std::{
+    fs,
+    io::{IsTerminal, Read},
+    process::Command,
+};
 
 const FAILED_CHECK_EXIT_STATUS: i32 = 1;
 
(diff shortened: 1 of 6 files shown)


fbjork (approved):

cli/crates/cli/src/check.rs
```diff
@@ -16,6 +16,9 @@ pub(crate) async fn check(command: CheckCommand) -> Result<(), CliError> {
 
     let schema = match schema {
         Some(schema) => fs::read_to_string(schema).map_err(CliError::SchemaReadError)?,
+        None if atty::is(atty::Stream::Stdin) => {
```
obmarg, line 19: [`std::io::IsTerminal`](https://doc.rust-lang.org/std/io/trait.IsTerminal.html) was stabilised in 1.70 iirc - wonder if we could just start using that instead of the 3rd party crate.

tomhoule: It seems to be working in my local testing, so let's do this.

tomhoule: done

PR #1606, opened 2024-04-15
graphql-schema-diff: replace async-graphql-parser with cynic-parser

tomhoule:
The main reason is that we are dealing with large GraphQL schemas where async-graphql-parser blows up in size and memory usage in production. cynic-parser doesn't.

The other main reason is that cynic-parser has a much nicer API and sounder foundations to build on.

closes GB-6465

### DIFF
--- a/engine/crates/graphql-schema-diff/src/state.rs
+++ b/engine/crates/graphql-schema-diff/src/state.rs
@@ -4,11 +4,11 @@ pub(crate) type DiffMap<K, V> = HashMap<K, (Option<V>, Option<V>)>;
 
 #[derive(Default)]
 pub(crate) struct DiffState<'a> {
-    pub(crate) schema_definition_map: [Option<&'a ast::SchemaDefinition>; 2],
+    pub(crate) schema_definition_map: [Option<ast::SchemaDefinition<'a>>; 2],
     pub(crate) types_map: DiffMap<&'a str, DefinitionKind>,
-    pub(crate) fields_map: DiffMap<[&'a str; 2], Option<&'a ast::Type>>,
-    pub(crate) interface_impls: DiffMap<&'a str, &'a [Positioned<async_graphql_value::Name>]>,
-    pub(crate) arguments_map: DiffMap<[&'a str; 3], (&'a ast::Type, Option<&'a ConstValue>)>,
+    pub(crate) fields_map: DiffMap<[&'a str; 2], Option<ast::Type<'a>>>,
+    pub(crate) interface_impls: DiffMap<&'a str, Vec<&'a str>>,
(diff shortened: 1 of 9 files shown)


engine/crates/graphql-schema-diff/src/state.rs
```diff
@@ -266,3 +262,56 @@ fn push_schema_definition_changes(
...
+                    let Some((_, b_value)) = b.iter().find(|(b_name, _)| b_name == name) else {
+                        return false;
+                    };
+
+                    value_eq(value, b_value)
+                })
+        }
+        _ => false,
+    }
+}
```
tomhoule, line 317: Should these last two be upstreamed in some way?

engine/crates/graphql-schema-diff/tests/diff/field_wrapping_type_changes.graphql
```diff
@@ -0,0 +1,13 @@
+scalar Date
```
tomhoule, line 1: This wasn't tested directly in this crate, but in operation-checks. Added this test for tighter feedback loop when debugging.

engine/crates/graphql-schema-diff/src/state.rs
obmarg: Yes, these seem like they'd be useful.

engine/crates/graphql-schema-diff/src/state.rs
```diff
@@ -266,3 +262,45 @@ fn push_schema_definition_changes(
...
+        (ast::Value::BlockString(a), ast::Value::BlockString(b)) => a == b,
+        (ast::Value::Boolean(a), ast::Value::Boolean(b)) => a == b,
+        (ast::Value::Null, ast::Value::Null) => true,
+        (ast::Value::Enum(a), ast::Value::Enum(b)) => a == b,
+        (ast::Value::List(a), ast::Value::List(b)) => {
+            a.len() == b.len() && a.iter().zip(b.iter()).all(|(a, b)| value_eq(a, b))
+        }
+        (ast::Value::Object(a), ast::Value::Object(b)) => {
+            a.len() == b.len()
+                && a.iter().all(|(name, value)| {
```
tomhoule, line 296: this is wrong: b could have keys a doesn't have and this would still return true

obmarg: Doesn't the equal length prevent that?

This looks fairly similar to the HashMap impl: https://doc.rust-lang.org/src/std/collections/hash/map.rs.html#1289

tomhoule: ah true, should have self-reviewed more carefully

obmarg (approved):

PR #1608, opened 2024-04-16
feat: cors directive for managed graphs

pimeys:
first part of GB-6507

### DIFF
--- a/engine/crates/engine/src/registry/mod.rs
+++ b/engine/crates/engine/src/registry/mod.rs
@@ -1392,6 +1392,17 @@ pub struct TrustedDocuments {
     pub bypass_header_value: Option<String>,
 }
 
+#[derive(Clone, Debug, Default, serde::Deserialize, serde::Serialize, PartialEq)]
+#[serde(rename_all = "camelCase")]
+pub struct CorsConfig {
+    pub max_age: Option<u32>,
+    pub allow_credentials: bool,
+    pub allowed_headers: Option<Vec<String>>,
+    pub allowed_methods: Option<Vec<String>>,
+    pub allowed_origins: Option<Vec<String>>,
+    pub exposed_headers: Option<Vec<String>>,
+}
+
 // TODO(@miaxos): Remove this to a separate create as we'll need to use it outside engine
 // for a LogicalQuery
 #[derive(Clone, Debug, serde::Serialize, serde::Deserialize)]
@@ -1428,6 +1439,8 @@ pub struct Registry {
     pub operation_limits: OperationLimits,
     #[serde(default)]
     pub trusted_documents: Option<TrustedDocuments>,
+    #[serde(default)]
+    pub cors_config: Option<CorsConfig>,
 }
 
 impl Default for Registry {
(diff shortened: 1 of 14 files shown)


engine/crates/engine/src/registry/mod.rs
```diff
@@ -1428,6 +1439,8 @@ pub struct Registry {
     pub operation_limits: OperationLimits,
     #[serde(default)]
     pub trusted_documents: Option<TrustedDocuments>,
+    #[serde(default)]
+    pub cors_config: Option<CorsConfig>,
```
obmarg, line 1443: Are we planning on loading the whole registry to service CORS requests?  It can be moderately slow to deser if it's large (which I imagine the CC one is), worth bearing in mind

pimeys: It kind of is deserialized when we load the config in every request. We have the config there already, so why not?

pimeys: here https://github.com/grafbase/api/blob/main/workers/gateway/src/lib.rs#L52

engine/crates/parser-sdl/src/rules/cors_directive/mod.rs
```diff
@@ -0,0 +1,210 @@
...
+const CORS_DIRECTIVE_NAME: &str = "cors";
+
+#[derive(Debug, Clone, serde::Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct CorsDirective {
+    max_age: Option<u32>,
+    #[serde(default)]
+    allow_credentials: bool,
+    allowed_headers: Option<AnyOrStringArray>,
+    allowed_methods: Option<AnyOrStringArray>,
```
obmarg, line 20: Not going to fight against it, but since we're in control of the valid HTTP methods it seems a bit weird to let users say "no, you can't do a POST actually" (or to allow PUT & DELETE, which we won't accept either).

pimeys: I agree, it was in the "RFC", but I don't think we need these...

pimeys: I would expose only `allowedOrigins`, but that was not in the issue...

engine/crates/engine/src/registry/mod.rs
Finistere: the registry isn't sent to the gateway. We might be able to store it somewhere else don't remember, but it's the most convenient that comes to mind to get the CORS config there https://github.com/grafbase/api/blob/main/lambda/deployment/gateway-deploy/src/gateway_config.rs#L114
so we can generate the `CustomerDeploymentConfig` with it

engine/crates/parser-sdl/src/rules/cors_directive/mod.rs
pimeys: Discussed with Fredrik, only having `allowedOrigins` and `maxAge` in the settings now.

obmarg (approved):

engine/crates/engine/src/registry/mod.rs
Finistere: the registry for v1 is loaded in the executor: https://github.com/grafbase/api/blob/main/workers/executor/src/api.rs#L66

PR #1619, opened 2024-04-18
fix: add missing http methods to parser-openapi

obmarg:
Bit weird to see these in a spec, but ok I suppose

### DIFF
--- a/engine/crates/parser-openapi/src/graph/mod.rs
+++ b/engine/crates/parser-openapi/src/graph/mod.rs
@@ -299,6 +299,10 @@ pub enum HttpMethod {
     Put,
     Delete,
     Patch,
+    Head,
+    Trace,
+    Options,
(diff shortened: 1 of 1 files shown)


engine/crates/parser-openapi/src/graph/mod.rs
```diff
@@ -299,6 +299,10 @@ pub enum HttpMethod {
     Put,
     Delete,
     Patch,
+    Head,
+    Trace,
```
Finistere, line 303: didn't know it even existed

Finistere (approved):

engine/crates/parser-openapi/src/graph/mod.rs
obmarg: Nor did I.  Still not sure what it's used for

PR #1625, opened 2024-04-20
chore: give the registry a haircut

obmarg:
I've been profiling with CCs actual queries & schema today, and the number one consumer of memory with these tiny queries is the registry.  As always when I go on one of these missions, there's a ton of low hanging fruit we can get rid of.  MetaField in particular was _huge_ - I forget the numbers, but 700 bytes or so per field.

In my profiling this reduces the total memory use from ~16.5mb to ~11.5.  This is all on arm rather than wasm so actual numbers probably vary, but it's an improvement either way.


engine/crates/engine/src/types/merged_object.rs
```diff
@@ -50,26 +50,26 @@ where
     fn create_type_info(registry: &mut Registry) -> crate::registry::MetaFieldType {
         registry.create_output_type::<Self, _>(|registry| {
             let mut fields = IndexMap::new();
-            let mut cc = CacheControl::default();
+            let _cc = CacheControl::default();
 
             if let MetaType::Object(ObjectType {
                 fields: b_fields,
-                cache_control: b_cc,
+                // cache_control: b_cc,
```
tomhoule, line 57: Are changes in this file as intended? Are there consequences to not merging the cache control there?

tomhoule (approved):
:100:

engine/crates/engine/src/types/merged_object.rs
obmarg: Yes, as intended.  This is part of the original async_graphql codebase, not in the codepath for production, I'm not wasting my saturday keeping it right 😆

tomhoule: Ah good, if it's async_graphql stuff we don't use that makes complete sense.

PR #1626, opened 2024-04-20
fix: add more info to traces for http requests

pimeys:
We should have more info where time is spent when triggering an http request:

- is it connecting
- or loading data from the backend
- or sending logs data

### DIFF
--- a/engine/crates/engine/src/registry/resolvers/logged_fetch.rs
+++ b/engine/crates/engine/src/registry/resolvers/logged_fetch.rs
@@ -35,7 +35,7 @@ pub async fn send_logged_request(
                     .headers_mut()
                     .unwrap()
                     .extend(response.headers_mut().drain());
-                let bytes = response.bytes().await?;
+                let bytes = response.bytes().instrument(info_span!("response_data_fetch")).await?;
                 let body = String::from_utf8(bytes.to_vec()).ok();
                 let response_to_return = response_to_return_builder.body(bytes).expect("must be valid").into();
                 (response_to_return, body)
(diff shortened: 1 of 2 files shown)


engine/crates/engine/src/registry/resolvers/logged_fetch.rs
```diff
@@ -35,7 +35,7 @@ pub async fn send_logged_request(
                     .headers_mut()
                     .unwrap()
                     .extend(response.headers_mut().drain());
-                let bytes = response.bytes().await?;
+                let bytes = response.bytes().instrument(info_span!("respose_data_fetch")).await?;
```
tomhoule, line 38:
```suggestion
                let bytes = response.bytes().instrument(info_span!("response_data_fetch")).await?;
```

tomhoule (approved):

tomhoule (approved):

engine/crates/engine/src/registry/resolvers/logged_fetch.rs
```diff
@@ -35,7 +35,7 @@ pub async fn send_logged_request(
                     .headers_mut()
                     .unwrap()
                     .extend(response.headers_mut().drain());
-                let bytes = response.bytes().await?;
+                let bytes = response.bytes().instrument(info_span!("response_data_fetch")).await?;
```
obmarg, line 38: Not against this change, but I will be (pleasantly) surprised if it gives us any more timing info

pimeys: Yeah, I also hope it's the backend server we are calling that is slow. But it's good to understand a bit more where the time goes if it's some fucked up WASM/Cloudflare thing we're witnessing here.

PR #1633, opened 2024-04-22
chore: put registry into an arc

obmarg:
I've hacked together an in memory cache in the executor to try and share registries between concurrent requests.  This means we need to use an `Arc<Engine>` most places we currently use an `Engine`.

### DIFF
--- a/engine/crates/engine/src/schema.rs
+++ b/engine/crates/engine/src/schema.rs
@@ -32,37 +32,12 @@ use crate::{new_futures_spawner, QuerySpawnedFuturesWaiter};
 /// Schema builder
 pub struct SchemaBuilder {
     validation_mode: ValidationMode,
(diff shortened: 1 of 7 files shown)


engine/crates/engine/src/schema.rs
```diff
@@ -103,43 +78,8 @@ impl SchemaBuilder {
...
-        self.registry.set_description(&T::type_name(), desc);
-        self
-    }
-
-    /// Override the name of the specified output type.
-    #[must_use]
-    pub fn override_output_type_description<T: LegacyOutputType>(mut self, desc: &'static str) -> Self {
-        self.registry.set_description(&T::type_name(), desc);
-        self
-    }
```
tomhoule, line 134: nice

tomhoule (approved):

PR #1636, opened 2024-04-22
feat: allow deleting and listing branches, allow deploying without linking

pimeys:
Videos tell more than words here:

Creating a new graph interactively, we can now add env vars from this dialog. The deployment errors, because we can't introspect the postgres with the given connection string:
[Screencast_20240425_185634.webm](https://github.com/grafbase/grafbase/assets/34967/63f4f0c8-39b8-42bc-848c-ae6404a0b95a)

Creating a new graph with env vars directly using CLI arguments. Now it works.
[Screencast_20240425_190139.webm](https://github.com/grafbase/grafbase/assets/34967/dd395759-e1ff-410d-91f9-978a5a6267e6)

Deploying from a linked project with branch, then listing branches, deleting the `.grafbase` directory and deploying with the graph ref that does not need a linked project.

[Screencast_20240425_190251.webm](https://github.com/grafbase/grafbase/assets/34967/1e6d6868-d32b-468e-8987-cbec0b03f6f0)

closes: GB-6546

### DIFF
--- a/cli/crates/backend/src/api/deploy.rs
+++ b/cli/crates/backend/src/api/deploy.rs
@@ -5,11 +5,12 @@ use super::graphql::mutations::{
     ArchiveFileSizeLimitExceededError, DailyDeploymentCountLimitExceededError, DeploymentCreate,
     DeploymentCreateArguments, DeploymentCreateInput, DeploymentCreatePayload,
 };
+use super::graphql::queries::deployment_poll::{Deployment, DeploymentLogs, DeploymentLogsArguments, Node};
 use super::types::ProjectMetadata;
 use common::consts::{PROJECT_METADATA_FILE, USER_AGENT};
 use common::environment::Project;
 use cynic::http::ReqwestExt;
-use cynic::{Id, MutationBuilder};
+use cynic::{Id, MutationBuilder, QueryBuilder};
 use ignore::DirEntry;
 use reqwest::{header, Body, Client};
 use std::ffi::OsStr;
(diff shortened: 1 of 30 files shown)


cli/crates/backend/src/api/deploy.rs
```diff
@@ -19,26 +21,118 @@ use tokio_util::codec::{BytesCodec, FramedRead};
...
+    };
 
-    match project_metadata_file_path.try_exists() {
-        Ok(true) => {}
-        Ok(false) => return Err(ApiError::UnlinkedProject),
-        Err(error) => return Err(ApiError::ReadProjectMetadataFile(error)),
+    let tar_file = create_tar_file().await?;
+
+    #[allow(clippy::cast_possible_truncation)] // must fit or will be over allowed limit
+    let archive_file_size = tar_file.metadata().await.map_err(ApiError::ReadArchiveMetadata)?.len() as i32;
```
obmarg, line 76: Probably not important but should this be a `i32::try_from().unwrap()` rather than an as?  A truncation seems like it might be a lot more subtle (although it's very unlikely so 🤷)

obmarg (approved):
Nice

cli/crates/backend/src/api/deploy.rs
pimeys: Yeah maybe, I just didn't really read how this part was done, copied it out. Maybe I should expect and crash if it's too large?

obmarg: > Maybe I should expect and crash if it's too large?

Yeah, that seems more correct.  Although now we're speaking about it I also wonder if we should have a limit far below i32::MAX.  If someone's trying to upload gigabytes of stuff then something must have gone wrong...

PR #1638, opened 2024-04-22
graphql-schema-validation: fix stack overflow on input type cycles

tomhoule:
Whenever we had a graph of input objects with multiple cycles in graphql-schema-validation, we would go into an infinite loop and stack overflow. This commit changes the function at fault to do a proper BFS.

closes GB-6570

### DIFF
--- a/engine/crates/validation/src/validate/input_object_cycles.rs
+++ b/engine/crates/validation/src/validate/input_object_cycles.rs
@@ -1,5 +1,6 @@
 use crate::context::Context;
 use async_graphql_parser::{types as ast, Positioned};
+use std::collections::HashSet;
 
 /// http://spec.graphql.org/draft/#sec-Input-Objects.Type-Validation
 pub(crate) fn input_object_cycles<'a>(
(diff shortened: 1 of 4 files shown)


engine/crates/validation/src/validate/input_object_cycles.rs
```diff
@@ -26,14 +30,20 @@ fn references_input_object_rec<'a>(
         }
 
         let field_type_name = super::extract_type_name(&field.ty.node.base);
+
         if field_type_name == name {
             return Some(vec![field.name.node.as_str()]);
         }
 
+        if visited.contains(field_type_name) {
```
Finistere, line 38: So the error was already prepared but you had an infinite loop? funny^^

Finistere (approved):

engine/crates/validation/src/validate/input_object_cycles.rs
tomhoule: yes :/

PR #1642, opened 2024-04-23
openapi: request fields that are not required should be nullable in GraphQL

tomhoule:
closes GB-6574


engine/crates/parser-openapi/src/tests/snapshots/parser_openapi__tests__flat_output.snap
```diff
@@ -1,13 +1,13 @@
...
 expression: registry.export_sdl(false)
 ---
 type Mutation {
 	updatePet(input: PetstorePetInput!): PetstorePet
 	addPet(input: PetstorePetInput!): PetstorePet
 	uploadFile(petId: Int!, additionalMetadata: String): PetstoreApiResponse
-	placeOrder(input: PetstoreOrderInput!): PetstoreOrder
-	createUsersWithListInput(input: [PetstoreUserInput!]!): PetstoreUser
+	placeOrder(input: PetstoreOrderInput): PetstoreOrder
+	createUsersWithListInput(input: [PetstoreUserInput!]): PetstoreUser
```
tomhoule, line 10: I was puzzled by these two lines but I think the change is correct. Just like for fields, we used to add a `NonNull` wrapper for request bodies if the referenced type was not declared as nullable. But the source of truth for the optionality of the body is the `required` key in `requestBody` in the spec, defaulting to `false` (source for that last point: https://swagger.io/docs/specification/describing-request-body/).

So I think this is correct.

engine/crates/parser-openapi/src/graph/input_value.rs
```diff
@@ -39,9 +39,12 @@ impl InputValue {
                     unreachable!()
                 };
 
-                // The HasType edge can introduce more wrapping so we need to make sure to account
+                // The HasType edge can introduce list wrapping so we need to make sure to account
                 // for that.
-                let wrapping = wrapping.wrap_with(edge_wrapping.clone());
+                let wrapping = match edge_wrapping {
+                    WrappingType::List(_) => wrapping.wrap_required().wrap_with(edge_wrapping.clone()),
+                    WrappingType::NonNull(_) | WrappingType::Named => wrapping,
```
tomhoule, line 46: I think there is a blind spot in this logic: if the referenced type is explicitly declared as `nullable: true` or as a union of the type and null, we should mutate `wrapping` to remove the outermost `NonNull` if it exists. Can we assume that `WrappingType::Named` in this match would always correspond to the referenced type being declared as nullable?

tomhoule: > we should mutate wrapping to remove the outermost NonNull if it exists

to elaborate on this case with an example: the field is part of `required` on the parent, but the referenced type is nullable. So omitting the field would not be allowed, but passing the field with a `null` value would be. This is hard to model properly in GraphQL...

engine/crates/parser-openapi/src/tests/snapshots/parser_openapi__tests__flat_output.snap
obmarg: Hmm, yeah.  Doesn't make sense as an API, but it does seem to be correct.

engine/crates/parser-openapi/src/graph/input_value.rs
obmarg: > Can we assume that WrappingType::Named in this match would always correspond to the referenced type being declared as nullable?

I think that's a safe assumption for the HasType edge at least.  Others I am less sure of.

tomhoule: Let me check what kind of diff that produces in snapshots and if that makes sense given the api definitions.

obmarg (approved):
Nice.  I'm never certain we've done the precisely right thing with nullability in parser-openapi, but comparing the snapshot changes to the spec it looks right to me.

engine/crates/parser-openapi/src/graph/input_value.rs
tomhoule: This:

```
                    WrappingType::Named => wrapping.set_required(false),
                    WrappingType::NonNull(_) => wrapping,
```

produces no difference in snapshots. I guess we don't test references to schema items that are explicitly nullable.

PR #1645, opened 2024-04-24
sdk: add ability to pass GraphQL SDL type definitions to standalone graphs

tomhoule:
This lets users mix GraphQL SDL and SDK to their preference when defining their configuration. The idea and implementation are both very basic.

The API takes the shape of a new argument to `graph.Standalone()`:

```typescript
const g = graph.Standalone({
  typeDefs: /* GraphQL */ `
    type Invoice {
      id: ID!
      invoiceNumber: String!
      dueDate: Date!
      totalAmount: Int!
    }

    extend type Query {
      invoiceByNumber(invoiceNumber: String!): Invoice @resolver(name: "invoice/byNumber")
    }
  `
})

```

The implementation is equally straightforward: we append the GraphQL that was passed in at the end of the generated SDL configuration.

Potential issues:

- Ordering: we have a set ordering for schema items in the SDK. Appending the string as-is will not respect it.
- Error message: it will not be clear where the error originated, but this is no worse than if you defined your whole config in SDL in the first place.

The issues seem to be minor enough to overlook, since it's a small change that enables new use cases.

closes GB-6547

### DIFF
--- a/packages/grafbase-sdk/tests/unit/typeDef.test.ts
+++ b/packages/grafbase-sdk/tests/unit/typeDef.test.ts
@@ -1,0 +1,41 @@
+import { config, graph } from '../../src/index'
+import { describe, expect, it } from '@jest/globals'
+import { renderGraphQL } from '../utils'
+
+const g = graph.Standalone({ subgraph: true })
+
+describe('graph.Standalone() typeDef argument', () => {
+  it('renders the content of typeDefs at the end', () => {
+    const g = graph.Standalone({
+      typeDefs: /* GraphQL */ `
+        type Invoice {
+          id: ID!
+          invoiceNumber: String!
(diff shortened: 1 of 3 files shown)


packages/grafbase-sdk/tests/unit/typeDef.test.ts
```diff
@@ -0,0 +1,41 @@
...
+                dueDate: Date!
+                totalAmount: Int!
+              }
+
+              extend type Query {
+                invoiceByNumber(invoiceNumber: String!): Invoice @resolver(name: "invoice/byNumber")
+              }
+            "
+    `)
+  })
```
pimeys, line 40: Is this the indentation?

obmarg (approved):

PR #1648, opened 2024-04-25
typed-resolvers: implement correct codegen for interfaces

tomhoule:
Before this commit, codegen would generate a TS interface with the fields of the GraphQL interface. But we are generating code for resolvers, and the resolver should not return only the interface fields, but a whole object that implements the interface, including `__typename`. This commit changes the generated code to a TypeScript union of all implementers of the interface.

closes GB-5171

### DIFF
--- a/cli/crates/typed-resolvers/src/analyze.rs
+++ b/cli/crates/typed-resolvers/src/analyze.rs
@@ -60,7 +60,7 @@ fn analyze_top_level<'doc>(graphql_document: &'doc ast::ServiceDocument, schema:
                             schema.push_enum_variant(id, &variant.node.value.node);
                         }
                     }
-                    // Interfaces are only interesting insofar as they are implemented, so they are handled in the Object branch.
+                    // Interfaces are only relevant insofar as they are implemented, so they are handled in the Object branch.
                     ast::TypeKind::Interface(_) => {}
                 }
             }
--- a/cli/crates/typed-resolvers/src/analyze.rs
+++ b/cli/crates/typed-resolvers/src/analyze.rs
@@ -1,6 +1,6 @@
 use engine_parser::types as ast;
 use engine_value::ConstValue;
-use std::{collections::HashMap, ops, str::FromStr};
+use std::{collections::HashMap, iter, ops, str::FromStr};
 
 #[must_use]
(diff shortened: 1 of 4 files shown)


cli/crates/typed-resolvers/src/analyze.rs
```diff
@@ -60,6 +60,8 @@ fn analyze_top_level<'doc>(graphql_document: &'doc ast::ServiceDocument, schema:
                             schema.push_enum_variant(id, &variant.node.value.node);
                         }
                     }
+                    // Interfaces are only interesting insofar as they are implemented, so they are handled in the Object branch.
```
tomhoule, line 63:
```suggestion
                    // Interfaces are only relevant insofar as they are implemented, so they are handled in the Object branch.
```

cli/crates/typed-resolvers/tests/schema_types/keyword_names.expected.ts
```diff
@@ -12,7 +12,7 @@
 //  }
 
 export type Schema = {
-  'holycow': | Schema['type'] | Schema['interface'] | Schema['object'] | Schema['union'];
+  'holycow': | Schema['type'] | Schema['interface'] | Schema['object'];
```
tomhoule, line 15: This line is not obvious because of the extremely confusing naming in this test, but the change is correct: `union` here is an interface (I know). The [Good Spec says interfaces can't be union members](http://spec.graphql.org/October2021/#sec-Unions.Type-Validation) so not taking them into account here does not matter because it won't happen in a valid GraphQL schema.

cli/crates/typed-resolvers/tests/schema_types/interfaces.expected.ts
```diff
@@ -23,11 +23,6 @@ export type Schema = {
...
     name: string;
   };
-  'LivingThing': {
-    metabolicRate: number;
-    age: number;
-  };
-  'Animal': {
-    pettable: boolean;
-  };
+  'Animal': | Schema['Dog'];
```
obmarg, line 26: Didn't realise TS allowed leading pipes.  How considerate of it.

obmarg (approved):

PR #1651, opened 2024-04-26
feat: add registry-v2 crate

obmarg:
As I mentioned in the meeting this morning, I've been working on refactoring the registry to a more compact & performant representation (hopefully also easier to use, but time will tell).

I've done this in a new crate `registry-v2`.  Separated from the `engine` because having the registry inside engine always limited how modular we could be: if the engine needed to call code that needed the registry that code usually to live inside the `engine` crate: leading to a severely bloated crate that takes an age to compile (waiting 30 seconds for feedback on my changes while working on this was not fun)

The structure of this crate follows the same pattern I've been using in cynic-parser (which is itself similar to patterns used in engine-v2 & elsewhere).  Each data type:

- Is stored as a "record" in an array in the Regsitry struct.
- Has an ID that is just a thin wrapper around index into the array.
- References other records by id where it makes sense.
- Has reader structs built on top of IDs that provide a reasonably ergonomic API for reading the data (known as walkers in some of our other implementations of this pattern).  Since these are the primary way to interact with the data, they just take the name of the data, no reader/walker postfix.
- Functions on a shared Writer struct that can be used to populate the Registry.

Writing code for all of the above is quite laborious (but not that complicated) so I've borrowed a generator that I wrote for cynic-parser to do a lot of it. All the code in the `generated` folder was produced by this.  The generator will be in a separate PR.  Note that not everything is worth generating - any special cases that are easier to write by hand should be.

Some of the code in this crate is just copied & pasted from the old registry. We could update that code to use the above pattern sometime, but this change is already massive.

Any questions, comments, criticisms - give them to me.

### DIFF
--- a/engine/crates/engine/registry-v2/src/common/iter.rs
+++ b/engine/crates/engine/registry-v2/src/common/iter.rs
@@ -14,8 +14,7 @@ where
     T: IdReader,
 {
     range: IdRange<T::Id>,
-    current: Option<T::Id>,
-    document: &'a crate::Registry,
+    registry: &'a crate::Registry,
 }
 
 impl<'a, T> Iter<'a, T>
@@ -23,12 +22,8 @@ where
     T: IdReader,
     T::Id: IdOperations,
 {
-    pub(crate) fn new(range: IdRange<T::Id>, document: &'a Registry) -> Self {
-        Iter {
-            current: (IdOperations::distance(range.start, range.end) > 0).then_some(range.start),
-            range,
-            document,
-        }
+    pub(crate) fn new(range: IdRange<T::Id>, registry: &'a Registry) -> Self {
+        Iter { range, registry }
     }
 }
 
@@ -44,18 +39,14 @@ where
     type Item = <T::Id as RegistryId>::Reader<'a>;
 
     fn next(&mut self) -> Option<Self::Item> {
-        let current = self.current?;
-        let next = self.range.next(current);
-        self.current = next;
+        let next = self.range.next()?;
+        self.range.start = next;
 
-        Some(self.document.read(current))
+        Some(self.registry.read(next))
     }
 
     fn size_hint(&self) -> (usize, Option<usize>) {
-        let Some(current) = self.current else {
-            return (0, Some(0));
-        };
-        let remaining = IdOperations::distance(current, self.range.end);
+        let remaining = IdOperations::distance(self.range.start, self.range.end);
         (remaining, Some(remaining))
     }
 }
--- a/engine/crates/engine/registry-v2/src/common/iter.rs
+++ b/engine/crates/engine/registry-v2/src/common/iter.rs
@@ -1,0 +1,75 @@
+use std::iter::FusedIterator;
+
+use crate::{
+    common::{IdOperations, IdRange},
+    Registry, RegistryId,
+};
+
+/// Iterator for readers
+///
+/// T indicates the type that will be yielded by the Iterator
+#[derive(Clone, Copy)]
+pub struct Iter<'a, T>
+where
+    T: IdReader,
+{
+    range: IdRange<T::Id>,
+    current: Option<T::Id>,
+    document: &'a crate::Registry,
+}
+
+impl<'a, T> Iter<'a, T>
+where
+    T: IdReader,
+    T::Id: IdOperations,
+{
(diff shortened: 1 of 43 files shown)


engine/crates/engine/registry-v2/src/common/types.rs
```diff
@@ -0,0 +1,207 @@
...
+                .collect::<Vec<_>>(),
+            vec![
+                WrappingType::List,
+                WrappingType::NonNull,
+                WrappingType::List,
+                WrappingType::NonNull
+            ]
+        );
+    }
+}
```
tomhoule, line 207: Is there a reason to not use the `wrapping` crate we extracted? (https://github.com/grafbase/grafbase/tree/main/engine/crates/wrapping). If there's missing features we could add them there.

engine/crates/engine/registry-v2/src/resolvers/mongo.rs
```diff
@@ -0,0 +1 @@
+
```
tomhoule, line 1: Accidentally committed?

engine/crates/engine/registry-v2/src/resolvers/postgres.rs
```diff
@@ -0,0 +1,43 @@
+// #[serde_with::minify_field_names(serialize = "minified", deserialize = "minified")]
```
tomhoule, line 1: I see a bunch of these annotations. Was the idea abandoned? I don't know what it does but the generated code with numbers for field names looks pretty compact already. In JSON I think it could only get more compact by serializing to an array instead of objects, but it would probably be more fragile.

tomhoule (approved):
I like where this is going

engine/crates/engine/registry-v2/src/common/types.rs
obmarg: This module was just copied over wholesale from cynic-parser.  Could almost certainly use that crate instead, though I might hold off that change till I've got everything else in place

engine/crates/engine/registry-v2/src/resolvers/mongo.rs
obmarg: Ah yes, oops

engine/crates/engine/registry-v2/src/resolvers/postgres.rs
obmarg: Oh, this being commented out is an oversight.  I copied & pasted all the resolvers over from engine, but then commented out the serde_with stuff initially so I could get it working.

I've ended up adding serde_with back in for the types that aren't yet otherwise compact, but I obviously missed this one.

I _think_ I want to eventually move the resolver heirarchy into the generated code, which would remove the need for these annotations.  But not planning on doing that immediately, it's a medium sized job.

engine/crates/engine/registry-v2/src/common/iter.rs
```diff
@@ -0,0 +1,75 @@
...
+/// Iterator for readers
+///
+/// T indicates the type that will be yielded by the Iterator
+#[derive(Clone, Copy)]
+pub struct Iter<'a, T>
+where
+    T: IdReader,
+{
+    range: IdRange<T::Id>,
+    current: Option<T::Id>,
```
Finistere, line 17: any reason to have `current`? sounds only useful to implement `peek()`

Finistere (approved):

engine/crates/engine/registry-v2/src/common/iter.rs
obmarg: Probably not.  First pass of this had a(n incorrect) DoubleEndedIterator impl that needed current.  It is pretty useless now that I've realised how you're actually meant to do DoubleEndedIterator.  

But it's also harmless so I might hold off removing it till I'm done with the mountain of other things

engine/crates/engine/registry-v2/src/common/iter.rs
```diff
@@ -0,0 +1,75 @@
...
+    document: &'a crate::Registry,
```
obmarg, line 18: I need to rename this field

engine/crates/engine/registry-v2/src/common/iter.rs
obmarg: This was trivial to do, so done

PR #1653, opened 2024-04-26
feat: add registry-upgrade crate

obmarg:
This is a crate that takes a `registry-v1::Registry` and spits out an equivalent `registry-v2::Registry`.  This allows us to hold off updating the parsers, that all use `registry-v1` while allowing the engine to work with `registry-v2`.  It's pretty straightforward other than that.

### DIFF
--- a/engine/crates/engine/registry-upgrade/src/lib.rs
+++ b/engine/crates/engine/registry-upgrade/src/lib.rs
@@ -1,0 +1,495 @@
+use std::collections::HashMap;
+
+use indexmap::IndexMap;
+use registry_v2::{
+    ids::{InputValidatorId, MetaFieldId},
+    storage::*,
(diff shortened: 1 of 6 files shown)


pimeys (approved):
nice and simple

engine/crates/engine/registry-upgrade/src/lib.rs
```diff
@@ -0,0 +1,495 @@
...
+        subscription_type,
+        disable_introspection,
+        enable_federation,
+        federation_subscription,
+        auth,
+        mongodb_configurations,
+        http_headers,
+        postgres_databases,
+        enable_caching,
+        enable_kv,
```
obmarg, line 29: I should ditch this kv flag, as it's not a feature anymore

PR #1654, opened 2024-04-26
feat: code generator for registry-v2

obmarg:
As outlined in [the PR that adds registy-v2][1] a lot of that crate is generated. This PR contains the generator.  It:

- Reads the data structure from a GraphQL file (SDL for life).
- Generates the records, readers and neccesary plumbing for everything to work.
- Manually implements some things that would usually be derived (i.e. `Debug` & `PartialEq`), but can't because of the patterns being used.
- Adds some serde annotations on things to make our JSON smaller.  We've used serde_as for this in the past, but why have a macro do work on every compile when you can have a generator do it once (with the bonus that it's more obvious what transforms are taking place)

This crate started as a c&p from cynic-parser and is very quick & dirty. it does what it needs to do and nothing more, it probably has bugs or strange behaviour if you try to do new things with it.  But it's worked so far, and is far easier than writing all this by hand (and makes changing things en masse relatively easy).

[1]: https://github.com/grafbase/grafbase/pull/1651

### DIFF
--- a/engine/crates/engine/registry-v2-generator/domain/registry.graphql
+++ b/engine/crates/engine/registry-v2-generator/domain/registry.graphql
@@ -1,0 +1,234 @@
+# type SchemaDefinition @file(name: "schemas") {
+#   description: StringLiteral
+#   directives: [Directive!]!
+#   root_operations: [RootOperationTypeDefinition!]!
+# }
+
+# type RootOperationTypeDefinition @file(name: "schemas") {
+#   operation_type: OperationType!
+#   named_type: String!
+# }
+
+# type ScalarDefinition @file(name: "scalars") {
+#   name: String!
+#   description: StringLiteral
+#   directives: [Directive]
+#   span: Span!
+# }
+
+union MetaType
+  @distinct
+  @file(name: "metatype")
+  @variant(names: ["Object", "Interface", "Union", "Enum", "InputObject", "Scalar"]) =
+    ObjectType
+  | InterfaceType
+  | UnionType
+  | EnumType
+  | InputObjectType
+  | ScalarType
+
+type ObjectType @file(name: "objects") @distinct {
+  name: String!
+  description: String
+  fields: [MetaField!]!
+  cache_control: CacheControl
+  external: bool!
+  shareable: bool!
+  # directives: [Directive]
+  # implements_interfaces: [String]
+  # span: Span!
+}
+
+type MetaField @file(name: "field") {
+  name: String!
+  mapped_name: String
+  description: String
+  args: [MetaInputValue!]!
+  ty: MetaFieldTypeRecord! @noreader
+  deprecation: Deprecation
+  cache_control: CacheControl
+  requires: FieldSet
+  federation: FederationProperties # inline Box this maybe?
+  resolver: Resolver!
+  required_operation: Operations
+  auth: AuthConfig # inline box this also?
+}
+
+# TODO: Think about whether these should be inline or not...
+# TODO: implement @box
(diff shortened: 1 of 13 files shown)


engine/crates/engine/registry-v2-generator/domain/registry.graphql
```diff
@@ -0,0 +1,234 @@
...
+#   operation_type: OperationType!
+#   named_type: String!
+# }
+
+# type ScalarDefinition @file(name: "scalars") {
+#   name: String!
+#   description: StringLiteral
+#   directives: [Directive]
+#   span: Span!
+# }
```
obmarg, line 17:
```suggestion
```

engine/crates/engine/registry-v2-generator/domain/registry.graphql
```diff
@@ -0,0 +1,234 @@
...
+  cache_control: CacheControl
+  requires: FieldSet
+  federation: FederationProperties # inline Box this maybe?
+  resolver: Resolver!
+  required_operation: Operations
+  auth: AuthConfig # inline box this also?
+}
+
+# TODO: Think about whether these should be inline or not...
+# TODO: implement @box
```
obmarg, line 58:
```suggestion
```

engine/crates/engine/registry-v2-generator/domain/registry.graphql
```diff
@@ -0,0 +1,234 @@
...
+# }
+
+# scalar Span @inline
+# scalar DirectiveLocation @inline
+# scalar OperationType @inline
+
+# Type & Value are kind of special cases that aren't worth automating
+# so we make them scalars and implement them by hand
+# scalar Type @file(name: "types")
+# scalar Value @file(name: "value")
```
obmarg, line 227:
```suggestion
```

Finistere (approved):
really nice work

engine/crates/engine/registry-v2-generator/src/exts.rs
```diff
@@ -0,0 +1,98 @@
...
+
+pub trait DistinctExt {
+    fn is_distinct(&self) -> bool;
+}
+
+impl DistinctExt for ObjectDefinition<'_> {
+    fn is_distinct(&self) -> bool {
+        // Indicates that a type is "distinct" i.e. no other instances of this type can
+        // have the same values.  We can use this to generate an optimised PartialEq
+        self.directives().any(|directive| directive.name() == "distinct")
```
tomhoule, line 12: This is a cool idea.

engine/crates/engine/registry-v2-generator/src/main.rs
```diff
@@ -0,0 +1,110 @@
...
+    use xshell::{cmd, Shell};
+    let sh = Shell::new()?;
+
+    let stdout = cmd!(sh, "rustfmt").stdin(&text.to_string()).read()?;
+
+    Ok(stdout)
+}
+
+fn shared_imports() -> proc_macro2::TokenStream {
+    quote::quote! {
```
tomhoule, line 107: Not a criticism, but wondering: do we gain something by using TokenStream's instead of treating the generated code as just a string, in this crate's use case?

obmarg: It's a mixed bag tbh.

Main advantages of quote for this are the fancy interpolation features (particularly the stuff it can do with lists etc.), and some degree of syntax highlighting on the code fragments.

It'd be a lot easier to control formatting (particularly blank lines) if we used strings, but I've managed to hack around that.

PR #1655, opened 2024-04-26
Stabilize resolver type code generation (codegen experimental feature)

tomhoule:
This stabilization PR does two things:

- Introduce the internal `@codegen` directive to configure codegen, replacing `@experimental(codegen: true)`. The experimental flag continues working for now but it is marked as deprecated in the SDK, so it will be a visible warning in users' IDE.
- Make the path where the generated code is written configurable.

We will need a SDK release after this is merged to expose the new configuration. In practice, it looks like this:

```ts
export default config({
  graph: g,
  codegen: {
     enabled: true,
     path: 'cat/meow'
  }
});
```

closes GB-6468
closes GB-5246
closes GB-5245
closes GB-5244

### DIFF
--- a/cli/crates/cli/tests/dev_watch.rs
+++ b/cli/crates/cli/tests/dev_watch.rs
@@ -81,3 +81,27 @@ async fn dev_watch() {
         assert!(generated_types_path.is_file());
     }
 }
+
+#[tokio::test(flavor = "multi_thread", worker_threads = 2)]
+async fn dev_watch_with_custom_codegen_path() {
+    let mut env = Environment::init();
+
+    env.grafbase_init(GraphType::Single);
+
+    env.write_schema(
+        r#"
+        extend schema @codegen(enabled: true, path: "custom-generated/directory/")
+
+        extend type Query {
+            hello: String! @resolver(name: "hello")
+        }
+        "#,
+    );
+    env.write_resolver("hello.js", "export default function Resolver() { return 'hello'; }");
+
+    env.grafbase_dev_watch();
+
+    // Check that the TS resolver types are being generated.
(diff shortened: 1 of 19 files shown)


cli/crates/cli/tests/dev_watch.rs
```diff
@@ -81,3 +80,29 @@ async fn dev_watch() {
         assert!(generated_types_path.is_file());
     }
 }
+
+#[tokio::test(flavor = "multi_thread", worker_threads = 2)]
+async fn dev_watch_with_custom_codegen_path() {
+    let mut env = Environment::init();
+
+    env.grafbase_init(GraphType::Single);
```
yoav-lavi, line 88: Shouldn't this be `Standalone`? Weird that we didn't change it

engine/crates/engine/src/registry/tests.rs
```diff
@@ -4,7 +4,7 @@ use sha2::{Digest, Sha256};
 
 use super::*;
 
-const EXPECTED_SHA: &str = "e6b38585432832f53a25f44a549ad6a4bdfd969983992b5bb7274aeaf85ddaa8";
+const EXPECTED_SHA: &str = "7e7bd6644c51d7b6ffcb1aad5b6cdb37660d0d05cb835d417d8e79d2a08bbd74";
```
yoav-lavi, line 7: We really need to rethink this test

yoav-lavi (approved):
Nice work!

cli/crates/cli/tests/dev_watch.rs
tomhoule: Hmm true, let me check that out in another branch.

tomhoule: PR coming

PR #1657, opened 2024-04-26
chore: extract some more crates from `engine`

obmarg:
The validation in engine is currently used for two things: 

1. Validating requests in the executor
2. Some caching stuff in the gateway.

I've updated the validation code to use registry-v2, which is fine.  But the cachng code in the gateway does not use registry-v2, but still needs to use the validation code.  Oh dear.

I'm probably going to make the code generic over registry types or something, but I figured I might as well split it out into it's own crate before doing that.  I did this but then discovered a few bits of engine it still needed, so I split those out as well.

I was probably going to need `scalars` in it's own crate anyway, as registry-v1 needs to know about the custom scalars in order to feed registry-v2 valid schemas when these custom scalars are  being used.


yoav-lavi (approved):

yoav-lavi:
Nice!

PR #1663, opened 2024-04-29
fix: order columns must be in selection, implicit if needed

pimeys:
Closes: GB-6619

### DIFF
--- a/engine/crates/engine/src/registry/resolvers/postgres/context/selection.rs
+++ b/engine/crates/engine/src/registry/resolvers/postgres/context/selection.rs
@@ -47,47 +47,48 @@ impl<'a> SelectionIterator<'a> {
 
         let mut extra_columns = Vec::new();
 
-        match selection_field
+        if let Some(order_by) = selection_field
             .field
             .get_argument("orderBy")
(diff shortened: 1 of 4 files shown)


engine/crates/engine/src/registry/resolvers/postgres/context/selection.rs
```diff
@@ -47,47 +47,48 @@ impl<'a> SelectionIterator<'a> {
...
+
+                    extra_columns.push(column);
                 }
             }
+        };
+
+        for column in table.implicit_ordering_key().unwrap().columns() {
+            if selection
+                .iter()
+                .any(|select| select.field.name.as_str() == column.table_column().client_name())
```
tomhoule, line 82: Is client_name the one from the GraphQL API? If so, shouldn't it be the one from the database?

tomhoule (approved):

engine/crates/engine/src/registry/resolvers/postgres/context/selection.rs
tomhoule: Ah ok, we select `database_column AS client_name`. Makes sense :+1:

PR #1666, opened 2024-04-29
chore: port CodegenConfig to new registry code

obmarg:
This was added to the original Registry on friday, so porting it over to the new registry setup while I remember.


tomhoule (approved):
looks good!

PR #1671, opened 2024-04-30
feat: use partial cache registry

obmarg:
#1675 introduces a new partial cache registry, this PR updates the code to make use of it.

Fairly simple, mostly just sprinkling generic parameters around in the validation code.


engine/crates/engine/registry-upgrade/src/lib.rs
```diff
@@ -69,6 +69,8 @@ pub fn convert_v1_to_v2(v1: registry_v1::Registry) -> registry_v2::Registry {
 
     // Build a map of type name -> the ID it'll have when we insert it.
     let preallocated_ids = writer.preallocate_type_ids(types.len()).collect::<Vec<_>>();
+    assert_eq!(preallocated_ids.len(), types.len());
```
Finistere, line 72: weird assert to have in production code, any reason this can fail at runtime?

engine/crates/engine/registry-upgrade/src/lib.rs
```diff
@@ -444,24 +452,30 @@ fn insert_validators(
...
+    type_ids: &HashMap<String, MetaTypeId>,
 ) -> MetaInputValueTypeRecord {
     MetaInputValueTypeRecord {
         wrappers: wrappers_from_string(ty.as_str()),
-        target: type_ids[ty.base_type_name()],
+        target: lookup_type_id(type_ids, ty.base_type_name()),
     }
 }
 
+fn lookup_type_id(type_ids: &HashMap<String, MetaTypeId>, name: &str) -> MetaTypeId {
```
Finistere, line 473: for a better error?

engine/crates/engine/validation/src/registries/mod.rs
```diff
@@ -0,0 +1,67 @@
+mod partial_caching_registry;
+mod v2;
+
+/// A trait for registries that provides enough functionalty for the validation visitors
+pub trait ValidationRegistry {
```
Finistere, line 5: Would have used `Any` prefix instead of `Validation` I think

Finistere (approved):

engine/crates/engine/registry-upgrade/src/lib.rs
obmarg: Yeah, a bug was causing the lookup to fail and "no thing at index" is the most useless error message.

engine/crates/engine/registry-upgrade/src/lib.rs
obmarg: I put this in to debug a fuckup I'd made, and I'm leaving it in incase that fuckup re-appears.

I'd only expect this to fail if someone messes with the `ID`/`IdRange`/`Iter` implementations in such a way that they're wrong.

Finistere: Sounds like you should have unit tests for whatever you rely on IMHO, but 🤷

PR #1678, opened 2024-05-01
chore: fix a ton of warnings

obmarg:


engine/crates/engine/src/registry/cache_control.rs
```diff
@@ -50,6 +46,10 @@ impl CachePartialRegistry {
     }
 }
 
+fn build_caching_view(_registry_caching_view: Registry) -> registry_v2::Registry {
+    todo!("implement me")
```
Finistere, line 50: if only Rust had an `unimplemented!()` /troll

Finistere (approved):

PR #1679, opened 2024-05-01
fix: the rest of the tests for new registry

obmarg:
- Some snapshot updates (because new registry does things alphabetically)
- Added SimpleJSON scalar, which postgres uses but didn't exist.
- Re-implemented validation for default values
- Fixed some incorrect binary searches
- Added a way to create an empty PartialCacheRegistry
- Some other misc things the api repo needed


tomhoule (approved):
:+1:

PR #1680, opened 2024-05-02
feat: remove kv support

hfgbarrigas:


cli/wrappers/package.json
```diff
@@ -8,7 +8,6 @@
     "typescript": "^5.3.3"
   },
   "dependencies": {
-    "@miniflare/kv": "^2.14.1",
     "@miniflare/storage-file": "^2.14.2"
```
yoav-lavi, line 11: This can also be removed

cli/wrappers/udf.ts
```diff
@@ -26,10 +26,6 @@ interface UdfRequestPayload {
   args: unknown
```
yoav-lavi, line 26: Lines 3-4 can also be removed

yoav-lavi: Line 25 should probably be modified since `kv` won't be passed anymore. We'd probably want it to be an empty object for now

cli/wrappers/package.json
```diff
@@ -7,10 +7,6 @@
     "bun": "1.1.6",
     "typescript": "^5.3.3"
   },
-  "dependencies": {
-    "@miniflare/kv": "^2.14.1",
-    "@miniflare/storage-file": "^2.14.2"
-  },
   "resolutions": {
```
yoav-lavi, line 10: This actually can also be removed since it was an optimization for the miniflare imports. (along with the file it references)

yoav-lavi (approved):
Nice work!

PR #1681, opened 2024-05-02
fix: wait for GDN reload to happen before retrieving the meter prodiver for OTEL metrics

Finistere:
Currently when the gateway is bootstrapped as follows:
1. Immediately setup OTEL metrics & tracing
2. Set up a watcher for the engine and a background task that loads GDN data to update it at regular intervals
3. Start the HTTP server

However, as we do need metadata from the GDN (branch_id typically), there is an async reload mechanism that is used to update the OTEL/tracing setup. Unfortunately, this doesn't work well for the HTTP request metrics as we need a `MeterProvider` at step 3., because it's used to setup a `tower::Layer`, even though neither the GDN call nor the OTEL reload may have happened yet.

So we need to wait for the reload mechanism to happen before starting the HTTP server. We could be lazy with the `MeterProvider`, a bit like the `tracing` integration of metrics is, but it only replaces the problem. If a HTTP request comes in before the OTEL reload we wouldn't have the right resource attributes either.


gateway/crates/gateway-binary/tests/.integration_tests.rs.pending-snap
```diff
@@ -1,6 +0,0 @@
-{"run_id":"2bab5e53-b736-4e31-ab0b-c6c6ef41dfbf","line":621,"new":{"module_name":"integration_tests","snapshot_name":"hybrid_graph","metadata":{"source":"gateway/crates/gateway-binary/tests/integration_tests.rs","assertion_line":621,"expression":"&result"},"snapshot":"{\n  \"errors\": [\n    {\n      \"message\": \"there are no subgraphs registered currently\"\n    }\n  ]\n}"},"old":{"module_name":"integration_tests","metadata":{},"snapshot":"{\n  \"data\": {},\n  \"errors\": [\n    {\n      \"message\": \"error sending request for url (http://127.0.0.1:46697/)\"\n    }\n  ]\n}"}}
```
pimeys, line 1: we probably don't need this commited?

Finistere: exactly why I'm removing the file^^

PR #1682, opened 2024-05-02
GB-6642 - Rust 1.78

yoav-lavi:
# Description

## Tooling

- Updates Rust to 1.78.0

# Type of change

- [ ] 💔 Breaking
- [ ] 🚀 Feature
- [ ] 🐛 Fix
- [x] 🛠️ Tooling
- [ ] 🔨 Refactoring
- [ ] 🧪 Test
- [ ] 📦 Dependency

### DIFF
--- a/cli/crates/cli/src/output/report.rs
+++ b/cli/crates/cli/src/output/report.rs
@@ -280,7 +280,7 @@ fn log_nested_events(nested_events: Vec<NestedRequestScopedMessage>, log_level_f
                 let mut url_string = url.to_string();
 
                 if url.path() == "/" && url.query().is_none() {
-                    url_string = url_string.trim_end_matches('/').to_owned();
+                    url_string = url_string.trim_end_matches('/').to_string();
                 }
 
                 let formatted_duration = format_duration(duration);
--- a/cli/crates/gateway/src/serving.rs
+++ b/cli/crates/gateway/src/serving.rs
@@ -101,7 +101,7 @@ async fn get_graphql(
     let ctx = crate::Context::new(headers, sender);
 
     let mut request: engine::Request = params.into();
-    request.ray_id = ctx.ray_id.clone();
+    request.ray_id.clone_from(&ctx.ray_id);
 
     let result = match streaming_format {
         None => gateway
--- a/engine/crates/engine/src/context/ext.rs
+++ b/engine/crates/engine/src/context/ext.rs
@@ -2,10 +2,6 @@ use std::any::{Any, TypeId};
 
 use engine_value::{ConstValue as Value, Value as InputValue}; // WHY?
 use graph_entities::QueryResponse;
-use http::{
-    header::{AsHeaderName, IntoHeaderName},
-    HeaderValue,
-};
 use postgres_connector_types::database_definition::DatabaseDefinition;
 
 use crate::{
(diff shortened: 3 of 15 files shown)


cli/crates/gateway/src/serving.rs
```diff
@@ -101,7 +101,7 @@ async fn get_graphql(
     let ctx = crate::Context::new(headers, sender);
 
     let mut request: engine::Request = params.into();
-    request.ray_id = ctx.ray_id.clone();
+    request.ray_id.clone_from(&ctx.ray_id);
```
tomhoule, line 104: I find the `.clone_from()` and `.clone_into()` less readable than assignments. Does clippy give good reasons for enforcing that?

yoav-lavi:
```
Performs copy-assignment from source.

a.clone_from(&b) is equivalent to a = b.clone() in functionality, but can be overridden to reuse the resources of a to avoid unnecessary allocations.
```
I'm assuming they added a check for this since it can be more performant

yoav-lavi: These are mostly auto-fixed by the way

obmarg: I thought the same tbh

cli/crates/cli/src/output/report.rs
```diff
@@ -280,7 +280,7 @@ fn log_nested_events(nested_events: Vec<NestedRequestScopedMessage>, log_level_f
                 let mut url_string = url.to_string();
 
                 if url.path() == "/" && url.query().is_none() {
-                    url_string = url_string.trim_end_matches('/').to_owned();
+                    url_string = url_string.trim_end_matches('/').to_string();
```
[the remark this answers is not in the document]
yoav-lavi: It tries to do `clone_from` and then gets an error for using both mutable and immutable borrows, it's an issue with that auto-fix

cli/crates/gateway/src/serving.rs
yoav-lavi: We could bypass this check everywhere, although I'm not sure these few lines warrant it

yoav-lavi: @obmarg @tomhoule any preference?

tomhoule: Hmm yes so it's about specialization. I still find it less readable but I get the point now. No strong opinion.

cli/crates/cli/src/output/report.rs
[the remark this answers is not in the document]
tomhoule: Ah ok, we can just ignore it then.

engine/crates/engine/src/context/ext.rs
```diff
@@ -146,6 +148,8 @@ pub trait ContextExt<'a>: Context<'a> {
     ///     }
     /// }
     /// ```
+
+    #[allow(unused)]
```
yoav-lavi, line 152:
```suggestion
    #[allow(unused)]
```

cli/crates/gateway/src/serving.rs
obmarg: I suppose I can live with it.

PR #1683, opened 2024-05-02
composition: validate that required arguments are provided in @requires selections

tomhoule:
Also an improvement in other `@requires` validations error messages (we weren't telling _which_ require was at fault).

closes GB-6319

### DIFF
--- a/engine/crates/composition/src/validate.rs
+++ b/engine/crates/composition/src/validate.rs
@@ -50,7 +50,14 @@ fn validate_override_labels(ctx: &mut ValidateContext<'_>, field: subgraphs::Fie
 
 fn validate_selections(ctx: &mut ValidateContext<'_>, field: subgraphs::FieldWalker<'_>) {
     for selection in field.directives().requires().into_iter().flatten() {
-        validate_selection(ctx, selection, field.parent_definition())
+        let requires_path = || {
+            format!(
+                "{}.{}",
+                field.parent_definition().name().as_str(),
+                field.name().as_str()
+            )
+        };
+        validate_selection(ctx, selection, field.parent_definition(), &requires_path);
     }
 }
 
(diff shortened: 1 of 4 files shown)


engine/crates/composition/src/validate.rs
```diff
@@ -68,23 +76,37 @@ fn validate_selection(
         ));
     };
 
+    for required_argument in field.arguments().filter(|arg| arg.r#type().is_required()) {
+        let arg_name = required_argument.name();
+        if selection.arguments.iter().all(|(name, _)| *name != arg_name.id) {
+            ctx.diagnostics.push_fatal(format!(
+                "Error in @requires on {requires_path}: the {field_name}.{arg_name} is required but not provided.",
```
obmarg, line 83: should this have the word argument in there?

```suggestion
                "Error in @requires on {requires_path}: the {field_name}.{arg_name} argument is required but not provided.",
```

obmarg (approved):

engine/crates/composition/src/validate.rs
tomhoule: indeed, good catch

engine/crates/composition/src/validate.rs
```diff
@@ -68,23 +76,37 @@ fn validate_selection(
         ));
     };
 
+    for required_argument in field.arguments().filter(|arg| arg.r#type().is_required()) {
```
obmarg, line 79: Oh, one last thing - should you also be checking whether the argument has a default here?  If it's required but has a default then that probably shouldn't be an error

obmarg: (This comment brought to you by [fixing a similar bug in relay](https://github.com/facebook/relay/pull/4647) very recently)

tomhoule: Good point! Will do that before merging.

PR #1685, opened 2024-05-02
composition: validate subgraph names

tomhoule:
The rule we implement here is that the first character must be alphabetic, and the rest of the name must consist of alphanumeric characters and hyphens (`-`).

We also enforce that subgraphs whose names only differ by letter casing cannot coexist.

The main reason for these rules is that we need to come up with an enum value name based on each subgraph name when rendering federated SDL.

closes GB-5648


engine/crates/composition/CHANGELOG.md
```diff
@@ -1,5 +1,12 @@
 # Changelog
 
+## Unreleased
```
pimeys, line 3: This is the first thing I see in this PR and I'm already into the whole thing. Very nice.

pimeys (approved):

PR #1687, opened 2024-05-03
feat: hardcode grafbase otel endpoint

Finistere:
I'm hardcoding our OTEL endpoint whenever a graph reference is used. We don't fail if the token isn't present because the gateway will fail later with a better error message anyway. Using native certs ended up being as easy as just adding the right feature flag in `tonic`...

### DIFF
--- a/gateway/crates/federated-server/src/server/graph_fetch_method.rs
+++ b/gateway/crates/federated-server/src/server/graph_fetch_method.rs
@@ -8,7 +8,6 @@ use tokio::sync::{oneshot, watch};
 /// The method of running the gateway.
 pub enum GraphFetchMethod {
     /// The schema is fetched in regular intervals from the Grafbase API.
-    #[cfg(not(feature = "lambda"))]
     FromApi {
         /// The access token for accessing the the API.
         access_token: ascii::AsciiString,
@@ -37,12 +36,12 @@ impl GraphFetchMethod {
         sender: watch::Sender<Option<Arc<Engine>>>,
     ) -> crate::Result<()> {
         match self {
-            #[cfg(not(feature = "lambda"))]
             GraphFetchMethod::FromApi {
                 access_token,
                 graph_name,
                 branch,
             } => {
+                #[cfg(not(feature = "lambda"))]
                 tokio::spawn(async move {
                     use super::graph_updater::GraphUpdater;
 
--- a/gateway/crates/gateway-binary/src/args.rs
+++ b/gateway/crates/gateway-binary/src/args.rs
@@ -1,13 +1,32 @@
-#[cfg(feature = "lambda")]
 mod lambda;
 mod log;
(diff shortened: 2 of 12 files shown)


gateway/crates/federated-server/src/server/graph_fetch_method.rs
```diff
@@ -37,12 +36,12 @@ impl GraphFetchMethod {
         sender: watch::Sender<Option<Arc<Engine>>>,
     ) -> crate::Result<()> {
         match self {
-            #[cfg(not(feature = "lambda"))]
             GraphFetchMethod::FromApi {
                 access_token,
                 graph_name,
                 branch,
             } => {
+                #[cfg(not(feature = "lambda"))]
```
pimeys, line 44: ?

gateway/crates/gateway-binary/src/args.rs
```diff
@@ -1,13 +1,32 @@
...
+    fn fetch_method(&self) -> anyhow::Result<GraphFetchMethod>;
+    fn config(&self) -> anyhow::Result<Config>;
+    fn log_format<S>(&self) -> BoxedLayer<S>
+    where
+        S: Subscriber + for<'span> LookupSpan<'span> + Send + Sync;
+}
 
-#[cfg(not(feature = "lambda"))]
-pub(crate) use self::std::Args;
+pub(crate) fn parse() -> impl Args {
```
pimeys, line 24: :3

gateway/crates/gateway-binary/src/args/std.rs
```diff
@@ -75,17 +78,38 @@ impl Args {
...
+        let mut config = match self.config.as_ref() {
             Some(path) => {
                 let config = fs::read_to_string(path).context("could not read config file")?;
-                Ok(toml::from_str(&config)?)
+                toml::from_str(&config)?
             }
-            None => Ok(Config::default()),
+            None => Config::default(),
+        };
+        if let Some(ref token) = self.grafbase_access_token {
```
pimeys, line 89: So we still need some way of doing this in AWS Lambda?

pimeys (approved):

gateway/crates/federated-server/src/server/graph_fetch_method.rs
Finistere: The code is still behind the `not(lambda)` flag

gateway/crates/gateway-binary/src/args/std.rs
Finistere: Well yes, not sure how.

gateway/crates/gateway-binary/src/args.rs
Finistere: didn't want to write anymore code hidden behind a `not(feature = "lambda")` flag 😅

PR #1695, opened 2024-05-06
engine: add JWT claims to custom resolver Context

tomhoule:
Under `ctx.request.jwtClaims`.

closes GB-6639


engine/crates/runtime/src/udf.rs
```diff
@@ -4,8 +4,10 @@ use common_types::UdfKind;
 use serde::Serialize;
 
 #[derive(Debug, serde::Serialize)]
+#[serde(rename_all = "camelCase")]
 pub struct UdfRequestContextRequest {
     pub headers: serde_json::Value,
+    pub jwt_claims: BTreeMap<String, serde_json::Value>,
```
tomhoule, line 10: Should we make the field `null` when the request isn't authenticated with a JWT? On the one hand, more context for the custom resolver. On the other hand, it's more work to handle.

yoav-lavi: As opposed to an empty object?

tomhoule: yes

Finistere: IMHO it's better, you can't rely on specific claims to be there even if authenticated. if the user needs to know whether he was authenticated through some method or not, we should just add a `is_authenticated` boolean or something like that imho.

obmarg (approved):
Looks good, assuming you decided not to make the field null on no auth?

tomhoule:
> Looks good, assuming you decided not to make the field null on no auth?

Yes, Benjamin's argument convinced me.

PR #1701, opened 2024-05-07
fix: support `application/vnd.whatever+json` in OpenAPI.

obmarg:
The OpenAPI connector only supports JSON content-types, so it looks for responses w/ `application/json` as the content-type.  But some APIs use an application specific content type such as `application/vnd.amadeus+json`.

### DIFF
--- a/engine/crates/parser-openapi/src/graph/operations.rs
+++ b/engine/crates/parser-openapi/src/graph/operations.rs
@@ -238,7 +238,8 @@ impl DebugNode for Operation {
 }
 
 fn is_json(content_type: &str) -> bool {
-    content_type == "application/json"
+    // This covers `application/json` and the jsonAPI style `application/vnd.whatever+json`
+    content_type.ends_with("json")
 }
 
 #[derive(Debug)]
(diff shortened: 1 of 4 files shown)


fbjork (approved):

engine/crates/parser-openapi/src/graph/operations.rs
```diff
@@ -238,7 +238,8 @@ impl DebugNode for Operation {
 }
 
 fn is_json(content_type: &str) -> bool {
-    content_type == "application/json"
+    // This covers `application/json` and the jsonAPI style `application/vnd.whatever+json`
+    content_type.ends_with("json")
```
yoav-lavi, line 242: Do we want to check for `/json` or `+json` specifically to reduce edge cases?

yoav-lavi (approved):

tomhoule (approved):

engine/crates/parser-openapi/src/graph/operations.rs
obmarg: I'd rather be lax than strict on this tbh.

Unless you have specific examples of content types that end in `json` but are actually not json?

PR #1704, opened 2024-05-07
fix: refs in v2 openapi response map

obmarg:
OpenAPI lets you put a `{"$ref": "#/components/whatever"}` in a lot of places to refer to another bit of the JSON document.  The library we were using for v2 parsing didn't support these in the response map though, which a spec we're trying to use was doing.

I updated our fork to support this [here][1] and pulled it into this PR.  I also had to add support for schemas that aren't explicitly labelled with a type, which this schema is also doing.  I think is valid OpenAPI but wasn't supported by the existing code.

[1]: https://github.com/grafbase/openapi/commit/f1f45282c17d07881bb76a8ba50d1960fdf3c810


yoav-lavi (approved):
Nice work!

PR #1708, opened 2024-05-08
registry: do not produce invalid schema on empty Query type

tomhoule:
...in export_sdl().

We produced a schema with `type Query {\n}` in it, and async_graphql_parser choked on that. Instead, export to an empty string.

closes GB-6675


tomhoule:
Nothing really, I think the engine is fine with an empty schema (only async-graphql-parser isn't), but `grafbase dev` at least doesn't fail with the ugly async-graphql-parser error.

obmarg (approved):

PR #1714, opened 2024-05-13
feat(engine-v1): Add operation metrics

Finistere:
# Description

# Type of change

- [ ] 💔 Breaking
- [ ] 🚀 Feature
- [ ] 🐛 Fix
- [ ] 🛠️ Tooling
- [ ] 🔨 Refactoring
- [ ] 🧪 Test
- [ ] 📦 Dependency


engine/crates/engine/src/schema.rs
```diff
@@ -466,10 +474,12 @@ impl Schema {
         request: impl Into<Request> + Send,
         session_data: Arc<Data>,
     ) -> impl Stream<Item = StreamingPayload> + Send + Unpin {
+        let start = web_time::Instant::now();
         let schema = self.clone();
-        let request = request.into();
+        let request: Request = request.into();
         let extensions = self.create_extensions(session_data.clone());
         let gql_span = GqlRequestSpan::new().into_span();
+        let normalized_query = operation_normalizer::normalize(request.query(), request.operation_name()).ok();
```
obmarg, line 482: Not much we can do to avoid it I supppose but wonder what impact adding this is going to have on the wasm size...

engine/crates/gateway-core/src/lib.rs
```diff
@@ -124,53 +141,28 @@ where
...
+                    },
+                    name: operation.name,
+                    normalized_query_hash: blake3::hash(normalized_query.as_bytes()).into(),
+                    normalized_query,
+                    has_errors: !response.errors.is_empty(),
+                    cache_status: headers
+                        .get(X_GRAFBASE_CACHE)
+                        .and_then(|v| v.to_str().ok().map(|s| s.to_string())),
+                },
+                start.elapsed(),
```
obmarg, line 162: I assume I'm missing something: why are we doing this here _and_ in the engine?

engine/crates/engine/src/schema.rs
Finistere: we do it already in the worker anyway already, so won't change anything

engine/crates/gateway-core/src/lib.rs
Finistere: because gateway-core deals with the caching so need to be here, but the stream interface doesn't let me retrieve the necessary information. So need to do in the engine for stream

Finistere: The alternative would be imposing constraints on `Executor::StreamingResponse` to be able to retrieve headers and send all the metadata through it

obmarg: > So need to do in the engine for stream

Ah, that'll be what I'm missing - didn't notice it was stream only

engine/crates/engine/src/schema.rs
obmarg: We do it in the gateway currently, but this is going to run in the executor, no?  Hence wondering

obmarg (approved):

engine/crates/engine/src/schema.rs
Finistere: hum good point, you're right

PR #1715, opened 2024-05-14
fix: make registry-upgrade more tolerant of problems

obmarg:
registry-upgrade currently panics if you pass it a registry that references a type that dopesn't exist.  The thinking here was that _ideally_ we validate schemas before we even parse them, so that we can give the users a good error message.  And then hopefully our code doesn't do anything to make the registry invalid between validation and passing it to registry-upgrade.

This was wishful thinking - I'm not sure we _do_ validate the schema, and parser-sdl (and probably the connectors) are also perfectly capable of generating bad registries.

This updates registry-upgrade to be more tolerant of these kinds of problems:

1. Where possible it'll just skip any fields with missing types.  We could raise these as errors to the user, but I think they're _usually_ a sign of a bug elsewhere - either we're generating invalid registries, or we're not catching invalid schemas.  So I kind of think a best effort approach to getting _something_ deployed is appropriate.
2. If the user doesn't provide a root `Query` type there's no graceful way to handle that - so for that case I've updated the crate to return an error.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6251,10 +6251,12 @@ dependencies = [
 name = "registry-upgrade"
 version = "0.1.0"
 dependencies = [
+ "anyhow",
  "indexmap 2.2.6",
  "registry-for-cache",
  "registry-v1",
  "registry-v2",
+ "tracing",
  "wrapping",
 ]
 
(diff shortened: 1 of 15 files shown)


tomhoule (approved):
Looks good!

PR #1717, opened 2024-05-14
fix: allow users to extend federation types

obmarg:
I was helping Fredrik setup for a sales demo last week, and we ran into some problems with the fairly common pattern of extending types from another subgraph:

```graphql
extend type Whatever @key(fields: "id") {
  id: String,
  someField: String @join(...)
}
```

1. It would complain that `id` didn't have a join or custom resolver on it.
2. It'd complain that `Whatever` didn't exist for extension.

It's quite easy to work around this by removing the `extend` keyword, but we should probably be friendlier to users and just let them do this.

Fixes GB-6682
Fixes GB-6678

### DIFF
--- a/engine/crates/parser-sdl/src/validations.rs
+++ b/engine/crates/parser-sdl/src/validations.rs
@@ -2,14 +2,18 @@ use std::sync::OnceLock;
 
 use engine::{
     registry::{field_set::FieldSetDisplay, resolvers::Resolver, MetaField, MetaFieldType},
-    Registry,
+    Pos, Registry,
 };
 use engine_parser::types::{BaseType, Type};
 use indexmap::IndexMap;
 use regex::Regex;
+use registry_v1::ObjectType;
 use registry_v2::{resolvers::join::JoinResolver, FederationResolver, Selection};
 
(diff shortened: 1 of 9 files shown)


engine/crates/parser-sdl/src/validations.rs
```diff
@@ -315,3 +319,25 @@ fn validate_federation_joins(registry: &Registry) -> Vec<RuleError> {
...
+            errors.push(RuleError::new(
+                vec![pos],
+                format!(
+                    "The object {} has a key that requires the field {} but that field isn't present",
+                    object.name, &field.field,
+                ),
+            ));
+        }
+        // In an ideal world we'd also validate any nested keys, but we don't really have
+        // access to any of the other types involved at this point :(
```
tomhoule, line 339: Nested keys seem super rare anyway

tomhoule (approved):
Neat!

engine/crates/parser-sdl/src/validations.rs
obmarg: Oh, this comment isn't even true now that I've moved the validation...

Although I don't really want to write nested key validation just now, so maybe I'll just fix the comment 😅

PR #1718, opened 2024-05-15
fix: remove to_plural from postgres connector runtime

obmarg:
I was digging into the binary sizes of the gateway & executor last week.  One of the things I noticed was that the regex crate (a known space waster) was being pulled into the executor, so I dug into that.  Turns out this was from a single innocent looking call to `Inflector::to_plural` in the postgres connector.

So, this works the connector to do that call at deploy time in the postgres-parser instead of at runtime, which should reduce the executor payload by a fair bit.

I've approached this by just storing the singular & plural relation field names against every table and storing that.  I'm not 100% happy with this implementation: these are technically only required for tables with relations, and even then a given relation might only need one of the forms.  But I'm not familiar enough with the code to figure out the best place to do _exactly_ what we need.  Particularly since tables names can change later in the process if their name clashes, so we'd then need to propagate that into our memoised relation names.  If anyone knows of a better way to do this (or wants to revisit) feel free.

### DIFF
--- a/engine/crates/postgres-connector-types/src/database_definition.rs
+++ b/engine/crates/postgres-connector-types/src/database_definition.rs
@@ -174,6 +174,8 @@ impl DatabaseDefinition {
             schema_id: table.schema_id(),
             database_name: self.names.intern_string(table.database_name()),
             client_name: self.names.intern_string(table.client_name()),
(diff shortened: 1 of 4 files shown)


tomhoule (approved):
:100:

yoav-lavi (approved):
Nice!

PR #1720, opened 2024-05-15
federated-graph: do not render empty types, empty enums in SDL

tomhoule:
They are not valid GraphQL. It's better to render an empty string, so let's do that.

closes GB-6715

### DIFF
--- a/engine/crates/federated-graph/src/render_sdl/render_federated_sdl.rs
+++ b/engine/crates/federated-graph/src/render_sdl/render_federated_sdl.rs
@@ -28,62 +28,57 @@ pub fn render_federated_sdl(graph: &FederatedGraphV3) -> Result<String, fmt::Err
         sdl.push('\n');
     }
 
-    let query_type_exists = graph.objects.iter().any(|object| &graph[object.name] == "Query");
-    if !query_type_exists {
-        writeln!(sdl, "type Query\n")?;
-    }
-
     for (idx, object) in graph.objects.iter().enumerate() {
         let object_name = &graph[object.name];
         let is_query_root = graph.root_operation_types.query == ObjectId(idx);
 
-        if let Some(description) = object.description {
-            write!(sdl, "{}", Description(&graph[description], ""))?;
-        }
-
-        sdl.push_str("type ");
-        sdl.push_str(object_name);
-
-        if !object.implements_interfaces.is_empty() {
-            sdl.push_str(" implements ");
-
-            for (idx, interface) in object.implements_interfaces.iter().enumerate() {
-                let interface_name = &graph[graph[*interface].name];
-                sdl.push_str(interface_name);
-
-                if idx < object.implements_interfaces.len() - 1 {
-                    sdl.push_str(" & ");
-                }
-            }
-        }
-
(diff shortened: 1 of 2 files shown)


engine/crates/federated-graph/src/render_sdl/render_federated_sdl.rs
```diff
@@ -28,62 +28,57 @@ pub fn render_federated_sdl(graph: &FederatedGraphV3) -> Result<String, fmt::Err
...
-    let query_type_exists = graph.objects.iter().any(|object| &graph[object.name] == "Query");
-    if !query_type_exists {
-        writeln!(sdl, "type Query\n")?;
-    }
-
     for (idx, object) in graph.objects.iter().enumerate() {
         let object_name = &graph[object.name];
         let is_query_root = graph.root_operation_types.query == ObjectId(idx);
 
-        if let Some(description) = object.description {
```
tomhoule, line 40: The diff is not very readable from here, but it's just a bunch of code moving inside an if block.

engine/crates/federated-graph/src/render_sdl/render_federated_sdl.rs
```diff
@@ -28,62 +28,57 @@ pub fn render_federated_sdl(graph: &FederatedGraphV3) -> Result<String, fmt::Err
...
-            write!(sdl, "{}", Description(&graph[description], ""))?;
-        }
+        let mut fields = graph[object.fields.clone()]
+            .iter()
+            .filter(|field| !(is_query_root && ["__type", "__schema"].contains(&graph[field.name].as_str())))
```
obmarg, line 37: The engine just filters out any fields starting w/ a `__`.  Not a big deal but do you think it's worth doing that here as well?

obmarg (approved):

engine/crates/federated-graph/src/render_sdl/render_federated_sdl.rs
obmarg: This was meant to have a link: https://github.com/grafbase/grafbase/blob/main/engine/crates/engine/src/registry/export_sdl.rs#L66

tomhoule: Since the rendered SDL is visible in the dashboard, I think it's better to leave these out (same as __type and __schema), but it's not super important.

obmarg: Yeah, I was just suggesting an alternative method for leaving them out.  But doesn't matter.

tomhoule: Ah, you mean feed the schema to engine-v2, then use the introspection machinery to render the api schema as sdl?

obmarg: Apparently I chose the worlds least clear way to suggest this:

```suggestion
            .filter(|field| !field.starts_with("__"))
```

tomhoule: aah ok, yes, let's do that :+1:

PR #1721, opened 2024-05-16
GB-6598 - Format dev introspection

yoav-lavi:
# Description

## Features

- Better formatting for `grafbase introspect --dev`

# Type of change

- [ ] 💔 Breaking
- [ ] 🚀 Feature
- [x] 🐛 Fix
- [ ] 🛠️ Tooling
- [ ] 🔨 Refactoring
- [ ] 🧪 Test
- [ ] 📦 Dependency

### DIFF
--- a/cli/crates/server/src/introspect_local.rs
+++ b/cli/crates/server/src/introspect_local.rs
@@ -26,6 +26,13 @@ pub async fn introspect_local() -> Result<IntrospectLocalOutput, ServerError> {
     if federated_graph_config.is_some() && rendered_sdl.is_empty() {
         Ok(IntrospectLocalOutput::EmptyFederated)
     } else {
-        Ok(IntrospectLocalOutput::Sdl(rendered_sdl))
+        Ok(IntrospectLocalOutput::Sdl(prettify(rendered_sdl)))
+    }
+}
+
+fn prettify(graphql: String) -> String {
+    match cynic_parser::parse_type_system_document(&graphql) {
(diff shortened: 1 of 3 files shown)


cli/crates/server/src/introspect_local.rs
```diff
@@ -23,9 +23,18 @@ pub async fn introspect_local() -> Result<IntrospectLocalOutput, ServerError> {
...
     let rendered_sdl = registry.export_sdl(registry.enable_federation);
 
+    let rendered_sdl = prettify(rendered_sdl);
+
     if federated_graph_config.is_some() && rendered_sdl.is_empty() {
         Ok(IntrospectLocalOutput::EmptyFederated)
     } else {
         Ok(IntrospectLocalOutput::Sdl(rendered_sdl))
     }
 }
```
obmarg, line 33: Maybe safer to not try and prettify an empty schema.  I assume it would always fail to parse.  Assume it would always fail to parse and just return the empty string again, but why even try

```suggestion
    if federated_graph_config.is_some() && rendered_sdl.is_empty() {
        Ok(IntrospectLocalOutput::EmptyFederated)
    } else {
        Ok(IntrospectLocalOutput::Sdl(prettify(rendered_sdl);))
    }
}
```

obmarg (approved):

cli/crates/server/src/introspect_local.rs
yoav-lavi: Agreed, updated

PR #1724, opened 2024-05-17
feat: add client name/version

Finistere:
# Description

# Type of change

- [ ] 💔 Breaking
- [ ] 🚀 Feature
- [ ] 🐛 Fix
- [ ] 🛠️ Tooling
- [ ] 🔨 Refactoring
- [ ] 🧪 Test
- [ ] 📦 Dependency


engine/crates/tracing/src/tower.rs
```diff
@@ -1,67 +1,175 @@
...
+        #[pin]
+        inner: F,
+        metrics: RequestMetrics,
+        span: Span,
+        start: Instant,
+        client: Option<Client>,
+    }
+}
+
+impl<F, ResBody, E> Future for ResponseFuture<F>
```
yoav-lavi, line 131: Why are we implementing this directly?

Finistere: hum fair point, could have just returned a `BoxFuture` actually. Thought I could specify the future type in the trait initially but couldn't without creating a struct like `tower-http` did and went that way instead of falling back to a box

PR #1727, opened 2024-05-22
cli/federated-dev: print composition errors for subgraphs from config

tomhoule:
closes GB-6768

### DIFF
--- a/rust-toolchain.toml
+++ b/rust-toolchain.toml
@@ -1,6 +1,6 @@
 [toolchain]
 profile = "default"
-channel = "1.78.0"
+channel = "1.77.2"
 targets = [
     "aarch64-apple-darwin",
     "x86_64-apple-darwin",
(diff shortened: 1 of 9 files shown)


obmarg (approved):

rust-toolchain.toml
```diff
@@ -1,6 +1,6 @@
 [toolchain]
 profile = "default"
-channel = "1.78.0"
```
Finistere, line 3: why?

tomhoule: Since I don't have a windows machine to debug with, and the cause of the failure isn't clear (although I agree, it looks like it's trying to dynamically load something and failing), I'm trying to see if changing runners / rustc versions or removing the cached artifacts is helping. It's not, so far. Next step will be trying to see if I can get my hands on a windows VM to reproduce the issue.

Finistere: ah ok makes sense with the commit names 😅

PR #1728, opened 2024-05-22
feat: first pass of partial caching query splitting

obmarg:
For partial caching, we're going to partition an incoming query based on the cache config that applies to each field in the query.  This PR contains the initial implementation of that.  It's not complete: in particular it does not handle variables at the moment, only looks at a fields cache config (not the config on the type), doesn't propagate configs in anyway, and possibly other shortcomings.  But this PR is already larger than I'd intended, so want to get it in and iterate.

I've implemented this as a separate crate, with the hopes that it can be plugged in where-ever it's needed.  

I've chosen to implement this with cynic-parser, rather than one of the many other options, because it assigns an `Id` to all of the parsed elements, which makes implementing this kind of thing easier than it otherwise would be:

- Each cache partition is basically just a set of fragment & selection IDs that indicate which part of the document to include for that partition.
- This will make merging partitions that miss back into the main query trivial - it's just merging some sets.
- We build these partitions up by walking the operation with some visitors, and then recursing over all the necessary fragments doing the same.
- We then build a graph of fragment relations, and traverse that to determine the full set of fragments each partition might need.  This is necessary because a fragment might not directly contain fields for a partition, but might just include a fragment that does.  The graph makes this fairly easy to calculate.


engine/crates/partial-caching/src/planning/mod.rs
```diff
@@ -0,0 +1,215 @@
...
+        cache_queries: cache_groups
+            .into_iter()
+            .map(|(control, group)| (control, QuerySubset::new(operation, group, &document)))
+            .collect(),
+        executor_query: QuerySubset::new(operation, uncached_group, &document),
+        document,
+    }))
+}
+
+// TODO: This name is awful
```
obmarg, line 74: deal with this

engine/crates/partial-caching/src/planning/mod.rs
```diff
@@ -0,0 +1,215 @@
...
+        operation,
+        registry,
+        &mut VisitorContext::new(&mut [&mut cache_group_visitor, &mut fragment_tracker]),
+    );
+
+    let (cache_groups, uncached_group) = visit_fragments(&document, registry, fragment_tracker, cache_group_visitor)?;
+
+    let operation = operation.id();
+
+    // TODO: probably want to return none if there are no cache groups
```
obmarg, line 61: do something about this

tomhoule (approved):
I can't say I get all the nuances with fragments from a first read, but the structure of the crate and the planning is clear, all the modules have a well defined purpose and I didn't see anything that looks wrong. Really nice!

PR #1735, opened 2024-05-29
feat: partial cache fetching logic

obmarg:
This implements the logic for building cache keys out of the various cachable query partitions, along with fetching them from the cache and then determining the final query to send along to the executor.

I've introduce the notion of "phases" to the partial-caching crate - we enter a "CacheFetchPhase", represented by a struct, that exposes functions to handle all the logic of this phase.  When done you can call ".finish()" which will either move to an execution phase or some as yet unnamed final phase.

I've also started integrating this with the gateway_core crate.  This is a very broken implementation, but one that will only be called if you enable partial caching in your registry (don't do that) and you have the cargo feature flag enabled (which it is in the CLI, so that I can work on this easier, but won't be in production till this is closer to ready).

The cache key calculations in this differ in a couple of ways from the existing setup:

1. I'm using blake3 for the hashing instead of SipHash.  I didn't feel comfortable using a non-cryptographic hash function - from a security standpoint, and also because collisions will mess this partial caching thing up horrendously, and I do not fancy debugging that down the line.
2. I'm hashing the JSON serialised form of the data, instead of doing order agnostic things with Objects like we do in the existing hash functions.  This is a bit annoying, although I'm not sure how much of an issue this is in practice. It was way faster to implement though. I've raised a linear issue to revisit this if I get time/we decide that it matters.

Fixes GB-6741

### DIFF
--- a/engine/crates/partial-caching/src/fetching/keys.rs
+++ b/engine/crates/partial-caching/src/fetching/keys.rs
@@ -1,5 +1,3 @@
-use std::collections::BTreeSet;
-
 use common_types::auth::ExecutionAuth;
 use engine_value::{ConstValue, Variables};
 use registry_for_cache::{CacheAccessScope, CacheControl};
@@ -48,14 +46,14 @@ fn cache_scopes<'a>(
 
     let actual_scopes = scopes
         .iter()
-        .filter_map(|scope| match scope {
+        .map(|scope| match scope {
             CacheAccessScope::Public | CacheAccessScope::ApiKey => Some(auth.global_ops().to_string()),
             CacheAccessScope::Jwt { claim } => auth.as_token().and_then(|token| token.get_claim(claim)),
             CacheAccessScope::Header { header: header_name } => headers
                 .get(header_name)
                 .and_then(|header| Some(header.to_str().ok()?.to_string())),
         })
-        .collect::<BTreeSet<_>>();
+        .collect::<Vec<_>>();
 
     if actual_scopes.is_empty() {
         tracing::warn!("Could not find any scopes for cache_control {cache_control:?}");
@@ -67,7 +65,7 @@ fn cache_scopes<'a>(
 
 #[derive(Debug, Hash, serde::Serialize)]
 enum CacheAccess<'a> {
-    Scoped(BTreeSet<String>),
+    Scoped(Vec<Option<String>>),
     Default(&'a ExecutionAuth),
 }
 
--- a/engine/crates/partial-caching/src/fetching/keys.rs
+++ b/engine/crates/partial-caching/src/fetching/keys.rs
@@ -1,0 +1,115 @@
+use std::collections::BTreeSet;
+
+use common_types::auth::ExecutionAuth;
+use engine_value::{ConstValue, Variables};
+use registry_for_cache::{CacheAccessScope, CacheControl};
+
+use super::CachingPlan;
+use crate::query_subset::QuerySubsetDisplay;
+
+pub fn build_cache_keys(
+    plan: &CachingPlan,
+    auth: &ExecutionAuth,
+    headers: &http::HeaderMap,
+    variables: &Variables,
+) -> Vec<Option<String>> {
+    plan.cache_partitions
+        .iter()
+        .map(|(cache_control, query_subset)| {
+            // If we can't figure out the scope for a cache_control entry we just skip it.
+            // This will force it to always be fetched from the cache.
+            let scopes = cache_scopes(cache_control, auth, headers)?;
+
+            let serializer = CacheKeySerializer {
+                query: query_subset.as_display(&plan.document),
+                scopes,
+                variables: variables_required(query_subset, &plan.document, variables),
+            };
+
+            match serializer.build_key_string() {
+                Ok(key) => Some(key),
+                Err(err) => {
+                    tracing::error!("Could not build key string for {cache_control:?}: {err:?}");
+                    None
+                }
+            }
+        })
+        .collect()
+}
+
+fn cache_scopes<'a>(
+    cache_control: &CacheControl,
+    auth: &'a ExecutionAuth,
+    headers: &http::HeaderMap,
+) -> Option<CacheAccess<'a>> {
+    let Some(scopes) = &cache_control.access_scopes else {
+        return Some(CacheAccess::Default(auth));
+    };
+
+    let actual_scopes = scopes
+        .iter()
+        .filter_map(|scope| match scope {
+            CacheAccessScope::Public | CacheAccessScope::ApiKey => Some(auth.global_ops().to_string()),
+            CacheAccessScope::Jwt { claim } => auth.as_token().and_then(|token| token.get_claim(claim)),
+            CacheAccessScope::Header { header: header_name } => headers
+                .get(header_name)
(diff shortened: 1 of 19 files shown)


engine/crates/partial-caching/src/fetching/mod.rs
```diff
@@ -0,0 +1,98 @@
...
+        self,
+        auth: &ExecutionAuth,
+        headers: &http::HeaderMap,
+        variables: &Variables,
+    ) -> CacheFetchPhase {
+        let cache_headers = headers
+            .typed_get::<headers::CacheControl>()
+            .unwrap_or_else(headers::CacheControl::new);
+
+        // TODO: don't forget to use CacheControlHeaders
```
obmarg, line 28: I'm going to do this later, have put an issue in linear to track it

engine/crates/partial-caching/src/fetching/keys.rs
```diff
@@ -0,0 +1,115 @@
...
+        })
+        .collect()
+}
+
+fn cache_scopes<'a>(
+    cache_control: &CacheControl,
+    auth: &'a ExecutionAuth,
+    headers: &http::HeaderMap,
+) -> Option<CacheAccess<'a>> {
+    let Some(scopes) = &cache_control.access_scopes else {
```
Finistere, line 45: IMHO this behavior seems useless to me. If the whole token is the scope, I understand cache control is private and should only be cached by the browse, but well that's what we did before. So might be worth a comment, or just ignore me 🤷

engine/crates/partial-caching/src/fetching/keys.rs
```diff
@@ -0,0 +1,115 @@
...
+    let actual_scopes = scopes
+        .iter()
+        .filter_map(|scope| match scope {
+            CacheAccessScope::Public | CacheAccessScope::ApiKey => Some(auth.global_ops().to_string()),
+            CacheAccessScope::Jwt { claim } => auth.as_token().and_then(|token| token.get_claim(claim)),
+            CacheAccessScope::Header { header: header_name } => headers
+                .get(header_name)
+                .and_then(|header| Some(header.to_str().ok()?.to_string())),
+        })
+        .collect::<BTreeSet<_>>();
```
Finistere, line 58: Why`BtreeSet`? Cache scope ordering matters afaik

engine/crates/partial-caching/src/fetching/mod.rs
```diff
@@ -0,0 +1,98 @@
...
+
+    /// The keys for each cache_query in the plan.
+    ///
+    /// Will be None if we couldn't determine a key for whatever reason, in
+    /// which case we'll always query the executor for those fields
+    pub(crate) cache_keys: Vec<Option<String>>,
+
+    /// The cache control headers a user has provided
+    #[allow(unused)]
+    cache_headers: CacheControlHeaders,
```
Finistere, line 53: would have used `request_cache_control: RequestCacheControl` and `RequestCacheControl` could even be a new type to avoid mixups.

Finistere (approved):

engine/crates/partial-caching/src/fetching/keys.rs
obmarg: Not sure I follow - in what way does it matter?

engine/crates/partial-caching/src/fetching/keys.rs
obmarg: I don't think this is useless.  

If there's some really expensive field, and it appears in a whole range of queries - why wouldn't we want to cache that, even if the cache is user specific?  Sure, it could be cached by the client, but as soon as the user opens a tab or hits refresh that cache is gone.  Or maybe it's a CLI that's running the call which doesn't even have a cache.

Or if the auth token is shared amongst many users or many devices.

engine/crates/partial-caching/src/fetching/mod.rs
obmarg: Yeah, maybe.  I was kind of trying to get away from the phrase `cache_control` altogether.  It's a reasonable name, I was just finding the clash very confusing at points.

I think ideally I'd keep cache_control for this and have something different for the registry type, but urgh: that is a lot more work.

I'll give this some thought though, may change it.

engine/crates/partial-caching/src/fetching/keys.rs
Finistere: if you have cache scopes: `Jwt(key1), Header(key2)` resulting in the values `Some(X), None` you could also send `None, Some(X)` and have the same cache scope. That's not what I would expect at least

engine/crates/partial-caching/src/fetching/keys.rs
Finistere: Well for partial caching it's indeed a bit different.

> Sure, it could be cached by the client, but as soon as the user opens a tab or hits refresh that cache is gone.

Would still be there, in the browser cache. Unless I don't understand at all how it works. But on another device no indeed. I doubt the same token would be used across devices though. 

The other problem with caching potentially private stuff is we don't know what's in there and how it'll be stored by CF. With `Cache-Control: private` header a server should not cache the response at all and only let the client should do it, scoping by auth feels like it but we don't behave in the same way, so well. Not convinced it's a good default behavior.

engine/crates/partial-caching/src/fetching/mod.rs
Finistere: Was using `request_cache_control` for this in my refactor and `reponse_cache_control` for the response header IIRC. Not necessarily the best, but at least `request_cache_control` doesn't need a comment explaining what it is^^

engine/crates/partial-caching/src/fetching/keys.rs
obmarg: > Would still be there, in the browser cache

I don't know that it would.  Any headers we return to the client need to use the minimum cache age of any individual field in the response.  So if this theoretical field is cacheable for 60 days, but everything else in the response is only cacheable for 30 seconds, then after 30 seconds the browser has no cache.

> I doubt the same token would be used across devices though.

In a lot of cases sure, but it's not impossible.  

Also bear in mind that ExecutionAuth isn't just tokens - if it's public auth then that goes in there too.  So we get a separate cache for public users, which seems faintly sensible.

> The other problem with caching potentially private stuff is we don't know what's in there and how it'll be stored by CF. With Cache-Control: private header a server should not cache the response at all and only let the client should do it, scoping by auth feels like it but we don't behave in the same way, so well. Not convinced it's a good default behavior.

So what would you have the default behaviour be?  

Not sure I agree that no cache scopes means private thing that should never be cached - if the user turned on caching, then presumably they do want caching.  We're just making sure that all cached data doesn't become 100% public if caching is turned on and they're using authentication.  It seems reasonable to me...

Finistere: >Also bear in mind that ExecutionAuth isn't just tokens - if it's public auth then that goes in there too.

Ah right, missed that.

> Not sure I agree that no cache scopes means private thing that should never be cached - if the user turned on caching, then presumably they do want caching. We're just making sure that all cached data doesn't become 100% public if caching is turned on and they're using authentication. It seems reasonable to me...

Hum fair point, never mind

PR #1738, opened 2024-05-30
feat: write partial cache updates

obmarg:
This is the final PR to get the basics of partial caching working - writing back to the cache after we get results from the executor.

I do this by splitting off a copy of the response, and running a custom `Serialize` impl that only writes the fields that the particular cache partition cares about.  Otherwise I think this is quite straightforward.


engine/crates/partial-caching/src/execution.rs
```diff
@@ -48,25 +51,44 @@ impl ExecutionPhase {
...
+                    keys_to_write.push((key.unwrap(), PartitionIndex(index)));
+                }
                 Entry::Miss => {}
             }
         }
-        (response, CacheUpdates)
+
+        let mut update_phase = None;
+        if !keys_to_write.is_empty() {
+            // TODO: Also check for errors before creating this...
```
obmarg, line 79: I need to do this

tomhoule (approved):
Nice!

engine/crates/gateway-core/src/cache/partial.rs
```diff
@@ -67,6 +81,31 @@ where
     }
 }
 
+async fn run_update_phase(update_phase: CacheUpdatePhase, cache: Cache) {
+    // Unsure whether I should run these in parallel or series.
```
Finistere, line 85: It's still the same CPU usage though, so ideally for CF I guess you should create all future serially and wait on them in parallel? Though for memory it's kind of fucked... So 🤷

obmarg: Yeah, it's the same usage overall, it's the blocking for longer that I'm concerned about.  This approach will at least yield the CPU for a bit while we write to the cache.  But I dunno, hard to know the impact without properly measuring it.

Finistere: ah, I see what you mean, CPU for other requests. Fair point

PR #1740, opened 2024-05-31
fix: support variables in partial caching

obmarg:
For the sake of time I skipped figuring out which variables apply to which query partitions earlier in the week.  This revisits and fixes those.

Fixes GB-6776

### DIFF
--- a/engine/crates/partial-caching/src/planning/variables.rs
+++ b/engine/crates/partial-caching/src/planning/variables.rs
@@ -1,0 +1,59 @@
+use cynic_parser::{
+    executable::{ids::VariableDefinitionId, Argument, Directive, OperationDefinition},
+    ExecutableDocument,
+};
+use indexmap::IndexSet;
+
+use crate::query_subset::CacheGroup;
+
+pub fn variables_required(
+    partition: &CacheGroup,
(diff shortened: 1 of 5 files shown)


engine/crates/partial-caching/src/planning/variables.rs
```diff
@@ -0,0 +1,59 @@
...
+    operation: OperationDefinition<'_>,
+) -> IndexSet<VariableDefinitionId> {
+    let mut variables = partition
+        .selections
+        .iter()
+        .flat_map(|id| -> Box<dyn Iterator<Item = &str>> {
+            match document.read(*id) {
+                cynic_parser::executable::Selection::Field(field) => Box::new(
+                    variables_used_in_argument(field.arguments())
+                        .chain(variables_used_in_directive(field.directives())),
```
tomhoule, line 21: :bulb: I would 100% have forgotten to check for variable usage in directives.

tomhoule (approved):

engine/crates/partial-caching/src/planning/variables.rs
obmarg: Yeah, I nearly did as well

PR #1743, opened 2024-06-03
composition: ignore subgraph federation mandated fields and types

tomhoule:
Subgraphs must expose the _entities and _service fields (and the corresponding _Entity and _Service types) in order to function as subgraphs behind the federated gateway. Depending on how you introspect them, these fields and types may end up in the subgraph SDL sent for composition. Since these fields are not part of the domain API, and they do not compose cleanly (duplicated fields and types across subgraphs), composition can and should ignore them.

This is what is implemented in this commit.

closes GB-6830

### DIFF
--- a/engine/crates/composition/src/ingest_subgraph.rs
+++ b/engine/crates/composition/src/ingest_subgraph.rs
@@ -15,6 +15,12 @@ use crate::{
 use async_graphql_parser::{types as ast, Positioned};
 use async_graphql_value::ConstValue;
 
+/// _Service is a special type exposed by subgraphs. It should not be composed.
+const SERVICE_TYPE_NAME: &str = "_Service";
+
+/// _Entity is a special union type exposed by subgraphs. It should not be composed.
+const ENTITY_UNION_NAME: &str = "_Entity";
+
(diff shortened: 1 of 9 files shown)


engine/crates/composition/src/ingest_subgraph.rs
```diff
@@ -197,10 +213,13 @@ fn ingest_definition_bodies(
                     subgraphs.push_interface_impl(definition_name, implemented_interface);
                 }
 
+                let is_query = root_type_matcher.is_query(&definition.node.name.node);
```
Finistere, line 216: would name it `is_query_type` / `is_query_root_type`

engine/crates/composition/src/ingest_subgraph/fields.rs
```diff
@@ -63,10 +64,17 @@ pub(super) fn ingest_fields(
     definition_id: DefinitionId,
     fields: &[Positioned<ast::FieldDefinition>],
     directive_matcher: &DirectiveMatcher<'_>,
+    parent_is_query_root: bool,
```
Finistere, line 67: would name it `parent_is_query_type` / `parent_is_query_root_type`

Finistere (approved):

tomhoule:
Good suggestions, applying them before merging.

PR #1746, opened 2024-06-03
feat: partial cache settings propagation

obmarg:
This updates the partial caching implementation to handle propagate settings from types & settings onto their descendants.  So if a type `User` as a cache control, then all the fields inside `User` will inherit those settings.

If a field _and_ the type that field contains both have a cache control then we'll give preference to the field setting, as it's more specific.

This complicated the fragment logic a bit, as we have to propagate through a fragment spread as well - which means we potentially need to process fragments multiple times if they are spread multiple times with different inherited cache settings. So were we were before just keying fragments by their id, we're now using a `FragmentKey` - which takes the inhereted cache control value into account as well. 

Fixes GB-6777

### DIFF
--- a/engine/crates/partial-caching/src/planning/fragments/tracker.rs
+++ b/engine/crates/partial-caching/src/planning/fragments/tracker.rs
@@ -27,10 +27,10 @@ impl FragmentTracker {
     }
 
     fn key_for(&self, spread: FragmentSpread<'_>) -> Option<FragmentKey> {
-        Some(FragmentKey {
-            id: spread.fragment()?.id(),
-            spread_cache_control: self.cache_control_stack.last().cloned(),
-        })
+        Some(FragmentKey::new(
+            spread.fragment()?.id(),
+            self.cache_control_stack.last().cloned(),
+        ))
     }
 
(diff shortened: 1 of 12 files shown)


engine/crates/partial-caching/src/planning/fragments/tracker.rs
```diff
@@ -26,10 +60,24 @@ impl super::super::visitor::Visitor for FragmentTracker {
...
+            self.cache_control_stack.pop();
+        }
+    }
+
     fn fragment_spread(&mut self, spread: FragmentSpread<'_>) {
-        self.used_fragments
-            .entry(spread.fragment_name().to_string())
-            .or_default()
-            .extend(self.selection_stack.iter().copied())
+        let Some(key) = self.key_for(spread) else {
```
Finistere, line 76: why not raise an error here? instead of doing it later?

Finistere (approved):

engine/crates/partial-caching/src/planning/fragments/tracker.rs
obmarg: Because this is the implementation of a trait and the trait function doesn't have a return type to raise an error through.

I could add one, but I think that would make the trait & visitor code worse.

PR #1747, opened 2024-06-04
fix: some issues with join

obmarg:
After #1744 yesterday we support schemas joins on fields of the root query, but there were a couple of further issues:

1. We were generating a `@requires` on any join that uses variables, even if those variables actually all refer to arguments.
2. The argument resolution code for joins did an early return if there's no parent resolve data (as is the case in root fields) - this meant that the arguments weren't being resolvd.

This commit fixes both of those, as well as adding an integration-test for this scenario.


tomhoule (approved):
:100:

PR #1750, opened 2024-06-04
fix: non-partial caching behaviour

obmarg:
When demoing partial caching last week, I first showed an example of how things work with non-partial caching.  While doing this we came to the conclusion that the current behaviour is confusing: if you have 2 fields in a query `cacheMe` and `noCache` - then the query will always be cached, even though `noCache` is not marked to be cached.

This was harder to fix than you'd think.  I needed to make enough changes that I suspect caching has always behaved like this.  Or if this was a regression, it's not clear when it regressed.

To fix this:

1. If a field doesn't have a `CacheControl` we need to treat it as having the default cache control (i.e. one with `maxAge: 0`).
2. I needed to update the `CacheControl::merge` function to not ignore `maxAge: 0`.  Previously if one of the cache controls had `maxAge: 0` then it would always prefer the other.  But now if we see `maxAge: 0` then the merged cache control will also have `maxAge: 0`
3. We now only do a merge on the cache controls when we reach a leaf field, rather than doing it for every type & field we encounter on the traversal.  This was necessary, as otherwise every type & field in the hierarchy needs to be marked as cachable.  So if you only wanted to cache the `email` field in `query { user { email } }` you would also have to mark `type Query` and `type User` as cacheable, which would in turn make the whole schema cacheable.  Not what we want.
4. I had to add a cache control stack into the CacheControlVisitor, so that we can propagate cache controls down from parent types & fields.

Fixes GB-6822

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -455,8 +455,9 @@ dependencies = [
 
 [[package]]
 name = "async-tungstenite"
-version = "0.25.1"
-source = "git+https://github.com/grafbase/async-tungstenite?rev=02222f222532b5cf01bba77e6d1cc8a142fe5593#02222f222532b5cf01bba77e6d1cc8a142fe5593"
+version = "0.26.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "7cf3e0ba34d3cf04731e376a1cf6c16c357396ab66da227bd283e74d49890b2a"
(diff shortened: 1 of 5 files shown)


tomhoule (approved):

jakubadamw (approved):
Honey to my eyes!

PR #1751, opened 2024-06-06
fix: handle headers in partial caching

obmarg:
This updates partial caching to respect an incoming `Cache-Control` header and also provide outgoing `x-grafbase-cache` & `Cache-Control` headers.

The maxAge of a returned `Cache-Control` header is set to whatever the lowest `maxAge` of the individual partitions is. As part of this I had to update the `Entry` interface to return the current TTLof a hit, as otherwise we'd have no way to know whether that TTL is actually lower than any of the maxAge values of the freshly fetched partitions.

Fixes GB-6806
Fixes GB-6819

### DIFF
--- a/engine/crates/partial-caching/src/headers.rs
+++ b/engine/crates/partial-caching/src/headers.rs
@@ -1,0 +1,25 @@
+pub struct RequestCacheControl {
+    /// Whether the cache code should use the cache to provide responses
+    ///
+    /// Controlled by the `no-cache` directive in headers
+    ///
+    /// The MDN description of `no-cache` seems to say that you can fetch
+    /// the cache but need to "revalidate" when this directive is present.
+    /// But we currently have no revalidation mechanism so this seems the
+    /// closest option
+    pub should_read_from_cache: bool,
+
+    /// Whether the cache code should store responses in the cache
+    ///
+    /// Controlled by the `no-store` directive in headers
+    pub should_write_to_cache: bool,
+}
+
+impl From<headers::CacheControl> for RequestCacheControl {
+    fn from(value: headers::CacheControl) -> Self {
+        RequestCacheControl {
+            should_read_from_cache: !value.no_cache(),
+            should_write_to_cache: !value.no_store(),
+        }
+    }
+}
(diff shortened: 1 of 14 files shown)


engine/crates/partial-caching/src/headers.rs
```diff
@@ -0,0 +1,25 @@
+pub struct RequestCacheControl {
+    /// Whether the cache code should use the cache to provide responses
+    ///
+    /// Controlled by the `no-cache` directive in headers
+    ///
+    /// The MDN description of `no-cache` seems to say that you can fetch
+    /// the cache but need to "revalidate" when this directive is present.
+    /// But we currently have no revalidation mechanism so this seems the
+    /// closest option
+    pub should_read_from_cache: bool,
```
Finistere, line 10: I would just keep the http header naming convention imho.

engine/crates/runtime/src/cache/mod.rs
```diff
@@ -38,15 +38,16 @@ pub enum EntryState {
 /// Wraps an entry from cache when getting it from there
 #[derive(Debug, PartialEq, Eq)]
 pub enum Entry<T> {
-    Hit(T),
+    /// A hit, and the current time till stale of that hit
```
Finistere, line 41: would say current `age`, and rename `max_age` underneath to `age` to follow the example of the `Age` header.

Finistere (approved):

engine/crates/runtime/src/cache/mod.rs
obmarg: It's not the age though?  It's the time till it's dead

Finistere: ah, would just say that in the comment then :D
And use `time_left` or something like that rather than `max_age` underneath

obmarg: > ah, would just say that in the comment then :D

You find "time till dead" clearer than "time till stale"?  Happy to change it if so, but stale seemed the better word to me.

Will fix the `max_age` bit though, yes

Finistere: no looks good actually, guess would have worded it this way, but it's clear.

engine/crates/partial-caching/src/headers.rs
obmarg: I dunno - `no_cache` sort of implies "do absolutely no caching" to me.  Which is not what it actually means.  That and what we are treating the headers as meaning and what they originally mean are not _exactly_ the same thing.  Given both of those, using clearer more accurate names seemed good to me?

Finistere: hum, ok. fair point

PR #1757, opened 2024-06-10
fix(engine-v1): bad request error & proper metrics

Finistere:
Currently, we don't track operations that end up as request errors even if we could parse the operation. We also don't track the number of errors nor whether those are request or field errors and whether data ended up `null` or not.

So adding a `GraphqlResponseStatus` for this and adding engine-v1 support. As engine-v1 still added the `data` field even on request errors, I'm fixing it. engine-v2 ended up being more work as error handling wasn't entirely done in a few cases, so ended up fixing that in another PR.

I'll mostly test all of this with the clickhouse integration tests in api. Had to also update the flake to have the latest pnpm to build the cli-app.

### DIFF
--- a/engine/crates/engine-v2/src/engine.rs
+++ b/engine/crates/engine-v2/src/engine.rs
@@ -9,6 +9,7 @@ use futures_util::{SinkExt, Stream};
 use gateway_core::StreamingFormat;
 use gateway_v2_auth::AuthService;
 use grafbase_tracing::{
+    gql_response_status::GraphqlResponseStatus,
     grafbase_client::Client,
     metrics::{GraphqlOperationMetrics, GraphqlOperationMetricsAttributes},
     span::{gql::GqlRequestSpan, GqlRecorderSpanExt, GqlRequestAttributes},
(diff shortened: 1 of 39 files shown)


engine/crates/engine-v2/src/engine.rs
```diff
@@ -169,28 +170,27 @@ impl<'ctx> ExecutionContext<'ctx> {
...
-        let mut metrics_attributes = operation_normalizer::normalize(request.query(), request.operation_name())
+        let metrics_attributes = operation_normalizer::normalize(request.query(), request.operation_name())
             .ok()
             .map(|normalized_query| GraphqlOperationMetricsAttributes {
                 normalized_query_hash: blake3::hash(normalized_query.as_bytes()).into(),
                 name: operation.name.clone(),
                 ty: operation.ty.as_str(),
                 normalized_query,
-                has_errors: false,
+                status: GraphqlResponseStatus::Success,
```
Finistere, line 187: Wrong but fixing it in another PR for engine-v2

engine/crates/integration-tests/src/engine_v1/gateway.rs
```diff
@@ -161,7 +172,21 @@ impl IntoFuture for GatewayTesterExecutionRequest {
 
             // TODO: Do something with headers...
```
obmarg, line 173: Can probably get rid of this TODO now

obmarg (approved):

PR #1758, opened 2024-06-10
chore: Add cli docker image

Finistere:
A user had trouble setting up a docker image with Grafbase, so I thought I would just revive an older docker PR and add a simple sanity check test. I didn't realize how much time I would spend on it 😅 

It took me an absurd amount of time to have a working docker image because of permissions issues, missing assets, and just incredibly slow build times. What I expect to be the linking stage still stakes forever in Docker in release, but I improved as much as I reasonably could the Dockerfile by using a separate step for building dependencies and using https://github.com/LukeMathWalker/cargo-chef. Unfortunately, limiting the preparation step to a single binary didn't work,  but it still avoids dependency download and some of the compilations. A bunch of those are recompiled likely because of feature flags. Did the same to the gateway image.

I considered not adding any tests given the amount of time I had already spent, but we have no tests that check that resolvers work with TS config. So inspired by clickhouse integration tests I've added a simple one that uses the docker image.

### DIFF
--- a/.github/workflows/binaries.yml
+++ b/.github/workflows/binaries.yml
@@ -153,7 +153,7 @@ jobs:
           set -euo pipefail
           cargo check --package ${{ matrix.package }}
 
-  docker:
+  docker-grafbase:
     needs: [test]
     if: ${{ needs.detect-change-type.outputs.build == 'true' || startsWith(github.ref, 'refs/tags/') }}
     runs-on: ubicloud-standard-8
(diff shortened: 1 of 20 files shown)


.github/workflows/binaries.yml
```diff
@@ -191,6 +192,7 @@ jobs:
       - name: Push docker image
         run: |
           docker push --all-tags ghcr.io/grafbase/gateway
+          docker push --all-tags ghcr.io/grafbase/cli
```
fbjork, line 195: or perhaps `ghcr.io/grafbase/grafbase`?

obmarg (approved):

.github/workflows/binaries.yml
Finistere: done

tomhoule (approved):
:100:

PR #1759, opened 2024-06-10
chore(engine-v2): fix error management & add metrics for bad requests

Finistere:
So this PR is pretty big because errors were not entirely well handled in engine-v2. 

The original goal was to handle correctly bad requests for operation metrics. Even if the operation was parsed correctly, we wouldn't generate any metrics if validation failed or an unknown field was used. So had to tweak stuff in both engines. This PR focuses on engine-v2.

While adding this feature I realized that we don't handle a few error cases, like execution failure propagating up to the root field ending up `null`. We would always return `{}` in those cases. So from there went down the rabbit hole of fixing various stuf...

v2 partitions the query into different plans, for each of those a response part is created which contains the data. So for errors, there are three cases:

- errors propagating within a response part, that was well handled
- errors that need to propagate across response parts, that worked more or less
- plan execution failure which results in an error for fields we were supposed to retrieve. That was poorly handled.

So this PR adds the latest and reworks some of the response part writing to simplify the actual plan code with fewer edge cases. Error paths from subgraphs are now also handled and only add an `upstream_path` error extension if they don't match anything on my side.

I also ended up tweaking a bit the subscription execution to not have an error special case and parallelize a bit more work.

And I'm also fixing operation metrics for bad requests. So now as long as we manage to parse an operation, we'll return operation metrics

### DIFF
--- a/engine/crates/integration-tests/tests/tracing/v2.rs
+++ b/engine/crates/integration-tests/tests/tracing/v2.rs
@@ -7,7 +7,6 @@ use graphql_mocks::{FakeFederationProductsSchema, FakeGithubSchema, MockGraphQlS
 use integration_tests::{federation::GatewayV2Ext, runtime};
 
 #[test]
-#[ignore] // Maybe fixing it in next PR.
 fn query_bad_request() {
     runtime().block_on(async {
         // prepare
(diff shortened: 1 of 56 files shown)


engine/crates/integration-tests/tests/tracing/v2.rs
```diff
@@ -90,6 +91,7 @@ fn query_named() {
 }
 
 #[test]
+#[ignore]
```
pimeys, line 94: ?

pimeys (approved):
Lots of code here, but also with good test coverage. LGTM.

engine/crates/integration-tests/tests/tracing/v2.rs
Finistere: ah thanks, should either remove the test or add it again. those tracing tests are quite sensible to changes and are pretty much useless for anything that involves I/O since enter&exit of spans for future concurrency are also recorded...

PR #1760, opened 2024-06-11
feat: support unions in SDL

obmarg:
For a while grafbase didn't support unions, because our execution was based around our old database feature and we hadn't built support for unions into that.  That changed when we added our OpenAPI & GraphQL connectors though - both of those connectors make use of union types, so execution needed to support union types.

However, we didn't come back and allow users to add unions into their schemas when doing this - they're only supported via one of the connectors.

A user on discord is trying to use unions now and is running into this limitation.

This PR updates parser-sdl to pull in unions, solving this problem.  It does not add support for them to the SDK for now - if we want that it'll need to be a separate PR.


tomhoule (approved):
Nice that most of the infrastructure was already there. LGTM!

PR #1764, opened 2024-06-11
postgres connector: implement "like" filter

tomhoule:
It produces a `LIKE` clause (https://www.postgresql.org/docs/current/functions-matching.html).

The feature request was for a `contains` filter, but:

1. The `contains` name is already used for the array membership filter
2. "like" better matches database terminology
3. LIKE is more powerful. It can also serve as `startsWith`, `endsWith`, ...

closes GB-6869

### DIFF
--- a/engine/crates/integration-tests/tests/postgres/introspection.rs
+++ b/engine/crates/integration-tests/tests/postgres/introspection.rs
@@ -1681,6 +1681,14 @@ fn table_with_serial_primary_key_string_unique() {
             The value is not in the given array of values
           """ nin: [String]
           not: StringSearchFilterInput
+          """
+            	The string matches the given pattern.
+
+            Example: "%ear%" would match strings containing the substring "ear".
+
+            See the reference at https://www.postgresql.org/docs/current/functions-matching.html#FUNCTIONS-LIKE
+
+          """ like: String
         }
 
         """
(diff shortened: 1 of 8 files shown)


engine/crates/integration-tests/tests/postgres/introspection.rs
```diff
@@ -1681,6 +1681,14 @@ fn table_with_serial_primary_key_string_unique() {
             The value is not in the given array of values
           """ nin: [String]
           not: StringSearchFilterInput
+          """
+            	The string matches the given pattern.
```
tomhoule, line 1685: I don't understand what's happening with the leading whitespace here.

engine/crates/parser-postgres/src/registry/types/scalar.rs
```diff
@@ -90,6 +90,26 @@ pub(super) fn register(input_ctx: &InputContext<'_>, output_ctx: &mut OutputCont
         create_scalar_update_type(input_ctx, TypeKind::Scalar(scalar), output_ctx);
         create_array_update_type(input_ctx, TypeKind::Scalar(scalar), output_ctx);
     }
+
+    registrer_string_filters(input_ctx, output_ctx);
```
tomhoule, line 94: s/registrer/register

engine/crates/integration-tests/tests/postgres/introspection.rs
obmarg: Wonder if the graphql-parser roundtrip is messing it up. Looks like graphql-parser didn't handle block strings properly prior to this PR: https://github.com/graphql-rust/graphql-parser/pull/75 (which is unreleased as far as I can see)

obmarg (approved):

PR #1766, opened 2024-06-11
fix(engine-v2): Mixed up enum values

Finistere:
# Description

# Type of change

- [ ] 💔 Breaking
- [ ] 🚀 Feature
- [ ] 🐛 Fix
- [ ] 🛠️ Tooling
- [ ] 🔨 Refactoring
- [ ] 🧪 Test
- [ ] 📦 Dependency


engine/crates/engine-v2/schema/src/builder/coerce/mod.rs
```diff
@@ -173,7 +173,7 @@ impl<'a> InputValueCoercer<'a> {
         let value_ids = r#enum.value_ids;
         match self.graph[value_ids].binary_search_by(|enum_value| self.ctx.strings[enum_value.name].as_str().cmp(name))
         {
-            Ok(id) => Ok(SchemaInputValue::EnumValue(EnumValueId::from(id))),
+            Ok(id) => Ok(SchemaInputValue::EnumValue(r#enum.value_ids.get(id).unwrap())),
```
tomhoule, line 176: So we were getting the value with id(x), say, 3, instead of the third value _inside the range_, right?

tomhoule (approved):
Nice fix!

engine/crates/engine-v2/schema/src/builder/coerce/mod.rs
Finistere: yeah

Finistere: so easy to miss with simple tests

PR #1770, opened 2024-06-12
postgres connector: multiple relations between same tables

tomhoule:
Given a table with two foreign keys referencing the same table, for example:

```sql
CREATE TABLE public.colors (
    id INT PRIMARY KEY GENERATED ALWAYS AS IDENTITY,
    name VARCHAR(255) NOT NULL UNIQUE,
    rgb INT NOT NULL
);

CREATE TABLE public.users (
    id INT PRIMARY KEY GENERATED ALWAYS AS IDENTITY,
    name VARCHAR(255) NOT NULL UNIQUE,
    favorite_color_id INT NOT NULL REFERENCES colors(id),
    least_favorite_color_id INT REFERENCES colors(id)
);
```

Previously, we would have only a `colors` field on the `users` type, using one of the two relations at random.

This commit changes the connector so we have two fields: `colorsByFavoriteColorId` and `colorsByLeastFavoriteColorId`. The fields on `colors` that walk the relation in the other direction are also adapted.

closes GB-6878


engine/crates/postgres-connector-types/src/database_definition/walkers/relation.rs
```diff
@@ -41,8 +41,8 @@ impl<'a> RelationWalker<'a> {
         }
     }
 
-    /// True, if the referenced row is unique, this means there can only be at most one related row.
-    pub fn is_referenced_row_unique(self) -> bool {
+    /// True, if the referenced column(s) is (are) unique, this means there can only be at most one row on the other side of the relation.
+    pub fn is_other_side_one(self) -> bool {
```
tomhoule, line 45: Renaming and rewriting the docstring because this is really not about the foreign key. Maybe `other_side_is_one()` would be a better name?

engine/crates/integration-tests/tests/postgres/find_one/joins.rs
```diff
@@ -1430,3 +1430,65 @@ fn one_to_one_with_one_to_many_joins_parent_side() {
...
+              ('Flageolet', 3, 2),
+              ('Canezou', 3, 4);
+             "#,
+        ];
+
+        for statement in setup {
+            api.execute_sql(statement).await;
+        }
+
+        std::thread::sleep(std::time::Duration::from_millis(1000));
```
pimeys, line 1466: ?

engine/crates/integration-tests/tests/postgres/find_one/joins.rs
```diff
@@ -1430,3 +1430,65 @@ fn one_to_one_with_one_to_many_joins_parent_side() {
...
+                  usersByFavoriteColorId(first: 10) { edges { node { id name favoriteColorId leastFavoriteColorId } } }
+                  usersByLeastFavoriteColorId(first: 10) { edges { node { id name favoriteColorId leastFavoriteColorId } } }
+              }
+            }
+        "#};
+
+        api.execute(query).await
+    });
+
+    let expected = expect![[""]];
```
pimeys, line 1491: Should we have a test that returns something too?

tomhoule: PR is not submitted for review yet, I haven't updated the snapshot

pimeys (approved):

PR #1773, opened 2024-06-13
cli: release 0.75.0

tomhoule:
Features

- You can now filter on text columns with the LIKE operator on the postgres connector (https://www.postgresql.org/docs/current/functions-matching.html). #1764
- The postgres connector now uses a connection pool instead of a single connection in `grafbase start`. #1771
- The postgres connector now supports multiple relations between two tables. See the pull request (https://github.com/grafbase/grafbase/pull/1770) for an example.
- Fixed a bug with enum values sent to subgraphs in `grafbase dev` for federated graphs. #1766
- Expand the scope of files watched for changes in `grafbase dev`. #1774

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2140,7 +2140,7 @@ dependencies = [
 
 [[package]]
 name = "engine-config-builder"
-version = "0.74.3"
+version = "0.75.0"
 dependencies = [
  "engine-v2-config",
  "graphql-federated-graph",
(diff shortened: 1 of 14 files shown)


tmp/.grafbase/generated/schemas/schema.graphql
```diff
@@ -0,0 +1,15 @@
+extend schema
```
pimeys, line 1: Are these files supposed to be in the PR?

tomhoule: nope, good catch

PR #1774, opened 2024-06-13
cli: expand what files we watch in grafbase dev

tomhoule:
The previous logic looks like this:

- The file is schema.graphql, grafbase.config.ts or .env at the root, or inside a grafbase/ subdirectory

OR

- The file is under grafbase/, resolvers/ or auth/
- AND it isn't inside node_modules/, or generate/, or a lockfile
- AND has one of the following extensions: "js", "ts", "jsx", "tsx", "mjs", "mts", "wasm", "cjs", "json", "yaml", "yml",

This commit expands the second branch of the conditional to any directory in the project (we still watch the same project root). There are new ignores for dist/ and target/ folders, since these tend to contain build artifacts rather than source files.

closes GB-6881

### DIFF
--- a/cli/crates/server/src/file_watcher.rs
+++ b/cli/crates/server/src/file_watcher.rs
@@ -165,14 +165,24 @@ const ROOT_FILE_WHITELIST: &[&str] = &[
 const EXTENSION_WHITELIST: &[&str] = &[
     "js", "ts", "jsx", "tsx", "mjs", "mts", "wasm", "cjs", "json", "yaml", "yml",
 ];
-const DIRECTORY_BLACKLIST: &[&str] = &[DOT_GRAFBASE_DIRECTORY_NAME, "node_modules", "generated"];
-const ROOT_WHITELIST: &[&str] = &[GRAFBASE_DIRECTORY_NAME, "resolvers", "auth"];
+const DIRECTORY_BLACKLIST: &[&str] = &[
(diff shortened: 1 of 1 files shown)


cli/crates/server/src/file_watcher.rs
```diff
@@ -157,22 +157,32 @@ impl Clone for ChangeStream {
...
-    is_whitelisted_root_file(path, root)
-        || in_whitelisted_root(path, root)
-            && (!(is_likely_a_directory(path) || in_blacklisted_directory(path, root) || is_lock_file_path(path, root))
-                && has_whitelisted_extension(path))
+    is_allowlisted_root_file(path, root)
+        || !(is_likely_a_directory(path) || in_denylisted_directory(path, root) || is_lock_file_path(path, root))
+            && has_allowlisted_extension(path)
+}
+
+fn is_allowlisted_root_file(path: &Path, root: &Path) -> bool {
```
yoav-lavi, line 182: Do we really need this if we're going from the root anyhow?

yoav-lavi: Nevermind, there are

tomhoule: Sorry, I missed this comment. We need this because .env and schema.graphql don't have one of the allowed extensions.

PR #1775, opened 2024-06-14
1.79

yoav-lavi:
# Description

## Tooling

- Updates Rust to 1.79

# Type of change

- [ ] 💔 Breaking
- [ ] 🚀 Feature
- [ ] 🐛 Fix
- [x] 🛠️ Tooling
- [ ] 🔨 Refactoring
- [ ] 🧪 Test
- [ ] 📦 Dependency

### DIFF
--- a/cli/crates/backend/src/api/graphql/types/mutations/environment_variable_delete.rs
+++ b/cli/crates/backend/src/api/graphql/types/mutations/environment_variable_delete.rs
@@ -47,10 +47,6 @@ pub struct EnvironmentVariableDeleteByValuesSuccess {
 
 #[derive(cynic::InlineFragments, Debug)]
 pub enum EnvironmentVariableDeleteByValuesPayload {
-    #[allow(dead_code)]
-    EnvironmentVariableDeleteByValuesSuccess(EnvironmentVariableDeleteByValuesSuccess),
-    #[allow(dead_code)]
-    EnvironmentVariableDoesNotExistError(EnvironmentVariableDoesNotExistError),
     #[cynic(fallback)]
     Unknown(String),
 }
--- a/cli/crates/backend/src/api/graphql/types/mutations/environment_variable_delete.rs
+++ b/cli/crates/backend/src/api/graphql/types/mutations/environment_variable_delete.rs
@@ -47,7 +47,9 @@ pub struct EnvironmentVariableDeleteByValuesSuccess {
 
 #[derive(cynic::InlineFragments, Debug)]
 pub enum EnvironmentVariableDeleteByValuesPayload {
+    #[allow(dead_code)]
     EnvironmentVariableDeleteByValuesSuccess(EnvironmentVariableDeleteByValuesSuccess),
+    #[allow(dead_code)]
     EnvironmentVariableDoesNotExistError(EnvironmentVariableDoesNotExistError),
     #[cynic(fallback)]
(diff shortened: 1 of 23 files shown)


fbjork (approved):

cli/crates/backend/src/api/graphql/types/mutations/environment_variable_delete.rs
```diff
@@ -47,7 +47,9 @@ pub struct EnvironmentVariableDeleteByValuesSuccess {
 
 #[derive(cynic::InlineFragments, Debug)]
 pub enum EnvironmentVariableDeleteByValuesPayload {
+    #[allow(dead_code)]
     EnvironmentVariableDeleteByValuesSuccess(EnvironmentVariableDeleteByValuesSuccess),
+    #[allow(dead_code)]
```
obmarg, line 52: These seem like they _should_ be used, no?  Or if we don't need them, should they be removed?

engine/crates/parser-sdl/src/rules/visitor/context.rs
```diff
@@ -45,18 +45,21 @@ pub struct VisitorContext<'a> {
...
     /// schema (such as global headers), but *MUST NOT* be used for anything that requires a stable
-    /// identifier across schema generations.
+    /// identifier across schema generations.    
+    #[allow(dead_code)]
     pub connector_id_generator: ConnectorIdGenerator,
 
     /// Each schema to build should contains a SchemaID -> MetaType String to be
     /// able to construct the whole SchemaRegistry at the end of the parsing.
+    #[allow(dead_code)]
     pub schema_to_build: RwLock<HashMap<SchemaID, String>>,
```
obmarg, line 63: If these aren't used maybe we should delete them

obmarg (commented):
Should we should be taking advantage of the improved dead_code detection to remove/fix unused things instead of just marking them all as allowed?

I think some of them make sense to mark as allow, but others look quite suspect to me

yoav-lavi:
@obmarg normally I'd go over these individually and remove them, but I won't have the time to today with the migration and I don't want to stall the upgrade. We should likely remove them in the future or check why they're not used

cli/crates/backend/src/api/graphql/types/mutations/environment_variable_delete.rs
fbjork: Yeah let's delete all dead code

cli/crates/backend/src/api/graphql/types/mutations/environment_variable_upsert_by_slugs.rs
```diff
@@ -28,13 +28,15 @@ pub struct NameContainsInvalidCharactersError {
 #[derive(cynic::QueryFragment, Debug)]
 #[cynic(graphql_type = "Mutation", variables = "EnvironmentVariableUpsertBySlugsVariables")]
 pub struct EnvironmentVariableUpsertBySlugs {
+    #[allow(dead_code)]
```
fbjork, line 31: Can't we delete the dead code?

pimeys: Another PR. There's lots of it, so just removed the things Yoav added in this PR.

PR #1781, opened 2024-06-17
feat: WASI component loader

pimeys:
This adds the crate to load and call WASI components and the callback for gateway requests. The subgraph stuff comes later when the types are more clear what is required/needed.

This is _not_ yet added to the gateway. Just the library with the tests. Do not be afraid for the large line count, the example guests have generated bindings which can be skipped.

Closes: GB-6899


gateway/crates/wasi-component-loader/src/callbacks/gateway.rs
```diff
@@ -0,0 +1,90 @@
...
+    pub(crate) async fn new(
+        loader: &ComponentLoader,
+        headers: HeaderMap,
+        request: engine::Request,
+    ) -> crate::Result<Self> {
+        let mut store = super::initialize_store(loader.config(), loader.engine())?;
+
+        // adds the data to the shared memory
+        let headers = store.data_mut().push_resource(headers)?;
+        let request = store.data_mut().push_resource(request)?;
```
pimeys, line 42: maybe we don't need this here _yet_

PR #1789, opened 2024-06-20
feat: plug a simple gateway callback with headers to the engine

pimeys:
Closes: GB-6901

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7518,6 +7518,7 @@ dependencies = [
  "tracing",
  "tungstenite",
  "ulid",
+ "wasi-component-loader",
 ]
 
 [[package]]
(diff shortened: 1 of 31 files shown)


engine/crates/runtime/src/user_callbacks.rs
```diff
@@ -0,0 +1,46 @@
...
+#[derive(Clone)]
+pub struct UserCallbacks(Arc<dyn UserCallbacksImpl>);
+
+impl UserCallbacks {
+    pub fn new(inner: impl UserCallbacksImpl + 'static) -> Self {
+        Self(Arc::new(inner))
+    }
+
+    /// Called on federated gateway, before authentication
+    pub async fn on_gateway_request(&self, headers: HeaderMap) -> Result<HeaderMap, UserCallbackError> {
```
Finistere, line 38: could just add a `std::ops::Deref` isntead

Finistere (approved):
I think I would have called it `UserExtension` or `Hooks` rather than callbacks. For me, it has the meaning of a function being called at the end of some work, kind of javascript callbacks used before promises were a thing.
But those functions are called at specific event to modify existing behavior before or after some work is done, hence extension or hooks.

PR #1790, opened 2024-06-20
feat(engine): generate operation name from first field if none was provided

Finistere:
# Description

# Type of change

- [ ] 💔 Breaking
- [ ] 🚀 Feature
- [ ] 🐛 Fix
- [ ] 🛠️ Tooling
- [ ] 🔨 Refactoring
- [ ] 🧪 Test
- [ ] 📦 Dependency

### DIFF
--- a/engine/crates/engine/src/context.rs
+++ b/engine/crates/engine/src/context.rs
@@ -34,7 +34,7 @@ use crate::{
     registry::type_kinds::SelectionSetTarget,
     request::IntrospectionState,
     schema::SchemaEnv,
-    GraphqlOperationMetadata, Name, Positioned, Result, ServerError, ServerResult, UploadValue,
+    GraphqlOperationAnalyticsAttributes, Name, Positioned, Result, ServerError, ServerResult, UploadValue,
 };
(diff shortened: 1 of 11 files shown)


engine/crates/engine/src/context.rs
```diff
@@ -127,6 +126,12 @@ pub struct QueryEnvInner {
     pub cache_control: CacheControl,
 }
 
+impl QueryEnvInner {
+    pub fn operation_name(&self) -> &Option<String> {
+        &self.operation_metadata.name
+    }
```
obmarg, line 132: Is there a good reason to always use this fake operation_name?  

I can live with it as a hack for analytics but outside of that pretending that the operation actually had a name when it didn't seems like it could cause problems...

Finistere: hum, seems like the opposite could also be true if we're not consistent. The operation name is only necessary when deciding which operation to use in the document afaik. Don't have a strong opinion, I just wasn't a fan of the inconsistency I guess 🤷 Want me to revert that part?

obmarg (approved):

PR #1791, opened 2024-06-20
feat: add OutputStore to partial-caching

obmarg:
Adding defer support to partial caching is going to mean keeping response values around in memory for a while.  We can't write to the cache until the full request is finished, so we need to keep the initial response and all it's deferred chunks in memory until the request is finished, when we can write the whole thing into the cache.

Because of this, I want to keep the memory footprint of this data small.  This PR contains an `OutputStore` type that I'll use to store the output, code for writing a QueryResponse into this structure, and a quick sanity test to make sure it works.

There's a lot of todos in this PR, which I'll come back to later.

Part of GB-6940

### DIFF
--- a/engine/crates/graph-entities/src/response/mod.rs
+++ b/engine/crates/graph-entities/src/response/mod.rs
@@ -350,6 +350,14 @@ impl ResponseList {
         self.0.iter().copied()
     }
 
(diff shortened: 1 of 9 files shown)


tomhoule (approved):
Makes sense

PR #1793, opened 2024-06-21
fix: a single shape hierarchy per query in partial-caching

obmarg:
In #1786 I started calculating the shape of output for partial-caching, which we can use to tell whether or not we're inside a defer when traversing some JSON.

In #1791 I added an "OutputStore" that ties into this, which stores objects something like this:

```
struct Object {
  shape_id: ObjectShapeId,
  fields: Vec<Value>
}
```

This saves having to store the keys of objects alongside the data, as you can just grab them from the associated shape.

The problem is, in #1786 I was building up multiple shape heirarchies - one for each partition of the query.  So when it comes time to merge those partitions together there's no single object shape that has all the fields we need an object to have.  Oops.

This PR fixes that by generating a single heirarchy for the entire query.  This is also simpler, so that's nice.


tomhoule (approved):
It does seem simpler

PR #1795, opened 2024-06-23
chore(engine-v2): Refactor required fields

Finistere:
In a next PR I'll need required field to be deduplicated, before I only
deduplicated arguments, not the full field itself. This makes it a lot
easier to track requirements later.

I'm also adding a `Unavailable` case for variables, planning will be
done without variables to be cache-able. This can lead us to add extra
fields which aren't actually required for some variables, but that's
something to optimize for later.

I've fixed what would definitely be bugs in field comparison between
required ones and operation ones. I was assuming a specific ordering
of input values at various places that wasn't guaranteed. I really
need to add tests for this...


obmarg (approved):

Finistere:
Fixes: GB-6975

PR #1798, opened 2024-06-24
feat: passable context object for hooks

pimeys:
We need a context object for the full request lifecycle which can be used for storing and retrieving data in the engine hooks. The issue is that we need to pass it by-value, and the request metadata is either passed by reference or as an Arc. So, to plan for that, we store the context data in a mutex. Yes, I know, but this data is always per-request, meaning the locks will not block other requests anyhow. In the upcoming hooks, we need to lock the mutex, get a mutable reference to the context object, use some `std::mem` tricks to get a value out of it and again return the possible changes to the request metadata.

Or, all the calls need to have a mutable reference to the metadata struct (which will be annoying in the websocket code)...

Closes: GB-6958

### DIFF
--- a/engine/crates/engine-v2/src/engine.rs
+++ b/engine/crates/engine-v2/src/engine.rs
@@ -1,10 +1,10 @@
-use std::sync::Arc;
+use std::{borrow::Cow, collections::HashMap, sync::Arc};
 use web_time::Instant;
 
 use async_runtime::stream::StreamExt as _;
 use engine::{BatchRequest, Request};
 use engine_parser::types::OperationType;
-use futures::{channel::mpsc, StreamExt};
+use futures::{channel::mpsc, lock::Mutex, StreamExt};
 use futures_util::{SinkExt, Stream};
 use gateway_core::StreamingFormat;
 use gateway_v2_auth::AuthService;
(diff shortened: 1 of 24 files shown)


engine/crates/engine-v2/src/engine.rs
```diff
@@ -63,30 +63,38 @@ impl Engine {
...
         } else {
             HttpGraphqlResponse::request_error("Unauthorized")
         }
     }
 
-    pub async fn create_session(self: &Arc<Self>, headers: http::HeaderMap) -> Option<Session> {
+    pub async fn create_session(
+        self: &Arc<Self>,
+        headers: http::HeaderMap,
+        context: HashMap<String, String>,
```
Finistere, line 93: `create_session` should call `on_gateway_request`. `Context` is an implementation detail. `create_session` is used in websocket connection to accept multiple incoming requests.

engine/crates/engine-v2/src/engine.rs
```diff
@@ -63,30 +63,38 @@ impl Engine {
         }
     }
 
+    pub fn user_hooks(&self) -> &UserHooks {
```
Finistere, line 66: why? `UserHooks` is an implementation detail for the engine caller perspective

Finistere (approved):

PR #1799, opened 2024-06-24
feat: merge partial cache entries into initial response

obmarg:
#1791 added an `OutputStore` and some code to write the initial response from the engine into that.  This PR handles merge cache response into the OutputStore.  This is a bit more involved than the initial response, because we need to be careful not to add any deferred fields into the response, and instead save them for when the deferred chunks come in.

Still a bunch of TODOs in here.  Will revisit these later though.

Fixes GB-6940

### DIFF
--- a/engine/crates/partial-caching/src/output/cache_merging.rs
+++ b/engine/crates/partial-caching/src/output/cache_merging.rs
@@ -1,0 +1,157 @@
+//! Handles merging cache responses into the OutputStore
+
+use super::{
+    engine_response::InitialOutput,
+    shapes::{ConcreteShape, Field, ObjectShape, OutputShapes},
+    store::{ObjectId, ValueId, ValueRecord},
+};
+
+impl<'a> InitialOutput<'a> {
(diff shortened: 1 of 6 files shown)


engine/crates/partial-caching/src/output/cache_merging.rs
```diff
@@ -0,0 +1,157 @@
...
+impl<'a> InitialOutput<'a> {
+    pub fn merge_cache_entry(&mut self, json: serde_json::Value, shapes: &'a OutputShapes) {
+        let Some(root_object_id) = self.store.root_object() else {
+            // Presumably an error bubbled up to the root, so not much we can do here.
+            return;
+        };
+
+        let root_object_shape = shapes.root();
+        let serde_json::Value::Object(mut object) = json else {
+            todo!("something");
```
yoav-lavi, line 18: Intentional?

obmarg: yeah, all the todos in here are intentional - just trying to get the basics working before I figure out error handling and such.

yoav-lavi (approved):
I'm not super up to date with partial caching but looks good!

PR #1800, opened 2024-06-24
feat: authorization wasi hook

pimeys:
Just adds the wasi side of this hook. We can change the types and rename things if needed, but the machinery is in this PR.

Closes: GB-6960


engine/crates/runtime/src/user_hooks.rs
```diff
@@ -51,4 +51,10 @@ pub trait UserHooksImpl: Send + Sync {
     type Context;
 
     async fn on_gateway_request(&self, headers: HeaderMap) -> Result<(Self::Context, HeaderMap), UserHookError>;
+
+    async fn on_authorization(
```
Finistere, line 55: IMHO the function should be:
```
async fn authorized&self, context: &mut Self::Context, input: impl ExactSizeIterator<Item = impl serde::Serialize>) -> ...
```
Might as well use the directive name, engine will provide a &mut Self::Context. The need to use `std::mem::take` is an implementation details of wasm IMHO. No need either to specify the types of of input either. How it's serialized is also up to the implementation IMO.

Why differentiate `UserError` from `UserHookError`? In all cases it'll end up as a GraphQL error 🤷

Finistere: IMHO would just `UserError` as `HookError` and that's it.

engine/crates/wasi-component-loader/src/hooks/authorization.rs
```diff
@@ -0,0 +1,62 @@
...
+use wasmtime::{
+    component::{Resource, TypedFunc},
+    Store,
+};
+
+use crate::{names::AUTHORIZATION_HOOK_FUNCTION, state::WasiState, ComponentLoader, ContextMap, ErrorResponse};
+
+pub(crate) type HookParameters = (Resource<ContextMap>, Vec<String>);
+
+pub(crate) type HookResponse = (Result<Vec<Option<ErrorResponse>>, ErrorResponse>,);
```
Finistere, line 12: If those are scoped to `hooks::authorization` anyway, might as well drop the `Hook` prefix^^

Finistere (approved):

engine/crates/runtime/src/user_hooks.rs
pimeys: The `UserError` gives user control to define the message and the extensions.

Finistere: Well yes, but in the end it doesn't make a difference to the engine whether it's wasm error or a hook error. It'll be treated the same and transformed into a GraphqlError with message & extensions. so as it'll be done in all cases we could simplify the types here IMHO. 🤷

PR #1802, opened 2024-06-24
chore(engine-v2): Split query plan solver from output shape computation

Finistere:
In this PR I'm separating the query plan "solver", which ensures we have a working plan with all requirements fulfilled, from the "output shape" computation.
So now I have `Plan` for the first step, which is kind of akin to partial cache query partitions with an associated resolver, and `ExecutionPlan` which defines the actual execution that will happen.

The solving step traverses the operation and splits it into different plans, adding extra fields as necessary with eventually a new plan. We keep track of which operation field matches which `RequiredField` for a given boundary (object resolved by multiple plans)

The second step re-traverses the query and requirements again and with the help of the previous step's metadata defines the actual plans and their dependencies. We know which resolver to use for which field, which operation field to use for which requirement etc.

Currently, ExecutionPlan and Plan are equivalent, but this allows us to modify the execution selection with directives like `@requiresScopes`, etc. after creating the plans but before the execution plans. The tricky part for this will be to ensure we modify the selection set in a consistent way.

### DIFF
--- a/engine/crates/engine-v2/schema/src/builder/mod.rs
+++ b/engine/crates/engine-v2/schema/src/builder/mod.rs
@@ -100,6 +100,7 @@ impl BuildContext {
                     directives: Default::default(),
                 },
             ],
+            field_to_parent_entity: vec![EntityId::Object(0.into()); 2],
             enum_definitions: Vec::new(),
             union_definitions: Vec::new(),
             scalar_definitions: Vec::new(),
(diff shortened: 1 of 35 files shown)


engine/crates/engine-v2/schema/src/builder/mod.rs
```diff
@@ -100,6 +100,7 @@ impl BuildContext {
                     directives: Default::default(),
                 },
             ],
+            field_to_parent_entity: vec![EntityId::Object(0.into()); 2],
```
obmarg, line 103: Why 2 entries of 0 here?  And is it worth a comment explaining

obmarg (approved):

engine/crates/engine-v2/schema/src/builder/mod.rs
Finistere: ah not obvious in the PR, but it's just because I'm adding two fields `__schema` and `__type` before, so adding `Query` as their parent entity

PR #1803, opened 2024-06-25
composition: implement @authorized

tomhoule:
This is a first version to unblock progress on further implementation.

Missing:

- Some validation on selections (nested selections on arguments and fields)
- Validation of metadata
- https://github.com/Authorized on objects, interfaces and fields of interfaces

That should already be captured in issues if I haven't forgotten anything.

closes gb-6961

### DIFF
--- a/engine/crates/composition/src/emit_federated_graph.rs
+++ b/engine/crates/composition/src/emit_federated_graph.rs
@@ -1,3 +1,4 @@
+mod attach_argument_selection;
 mod context;
 mod emit_fields;
 mod field_types_map;
@@ -26,15 +27,18 @@ pub(crate) fn emit_federated_graph(mut ir: CompositionIr, subgraphs: &Subgraphs)
         scalars: mem::take(&mut ir.scalars),
         input_objects: mem::take(&mut ir.input_objects),
         directives: mem::take(&mut ir.directives),
-        input_value_definitions: vec![],
-        strings: Vec::new(),
-        subgraphs: vec![],
         root_operation_types: RootOperationTypes {
             query: ir.query_type.unwrap(),
             mutation: ir.mutation_type,
             subscription: ir.subscription_type,
         },
+
+        subgraphs: vec![],
         fields: vec![],
+        input_value_definitions: vec![],
+        strings: vec![],
+        authorized_directives: vec![],
+        field_authorized_directives: vec![],
     };
 
     let mut ctx = Context::new(&mut ir, subgraphs, &mut out);
@@ -106,7 +110,7 @@ fn emit_fields<'a>(
     __type: federated::StringId,
     ctx: &mut Context<'a>,
 ) {
-    // We have to accumulate the `@provides` and `@requires` and delay emitting them because
+    // We have to accumulate the `@provides`, `@requires` and `@authorized` and delay emitting them because
     // attach_selection() depends on all fields having been populated first.
     let mut field_provides: Vec<(
         federated::FieldId,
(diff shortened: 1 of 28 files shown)


engine/crates/composition/src/emit_federated_graph.rs
```diff
@@ -263,6 +274,36 @@ fn emit_fields<'a>(
             .requires
             .push(federated::FieldRequires { subgraph_id, fields });
     }
+
+    for (field_id, definition, authorized) in field_authorized {
+        dbg!(ctx.subgraphs.walk(authorized.rule).as_str());
```
tomhoule, line 279: dbg to be removed

engine/crates/composition/src/emit_federated_graph/attach_argument_selection.rs
```diff
@@ -0,0 +1,35 @@
...
+                field_arguments.0 .0
+                    + ctx.out[field_arguments]
+                        .iter()
+                        .position(|arg| arg.name == selection_field)
+                        .unwrap(),
+            );
+
+            let subselection: federated::InputValueDefinitionSet =
+                if let federated::Definition::InputObject(_) = ctx.out[argument_id].r#type.definition {
+                    todo!("Nested selections not implement yet for arguments in @authorized: GB-6965")
```
tomhoule, line 24: limitation, but should not block merging imo

engine/crates/composition/src/subgraphs/directives.rs
```diff
@@ -248,6 +257,15 @@ pub(crate) struct OverrideDirective {
     pub(crate) label: Option<StringId>,
 }
 
+#[derive(Debug)]
+pub(crate) struct AuthorizedDirective {
+    pub(crate) rule: StringId,
+    pub(crate) arguments: Option<Vec<Selection>>,
+    pub(crate) fields: Option<Vec<Selection>>,
+    /// We expect JSON here, but since this will be passed as-is, we don't need to parse.
```
Finistere, line 265: well by `JSON` I meant the JSON scalar where anything works, so should be of type `Value` like:
```
@authorized(metadata: { a: 1 })
```
We also can't rely here on the format we might use to send data to hooks.

Finistere: The policy example uses this for example:
```
type User {
  id: ID!
  contracts: Contract @authorized(rule: "policy", metadata: [["role:admin"]] )
}
```

Finistere (approved):
Nice work

Finistere:
Not sure why you didn't add it inside `Directive` enum though instead of the new `field_authorized_directives`

Finistere:
`@authorized` directive isn't present on objects or interfaces or I don't know where to look for them :D

engine/crates/composition/src/subgraphs/directives.rs
tomhoule: Aah ok, yeah, that needs to be changed then. I assumed a custom scalar with JSON in a strin.

tomhoule:
> Not sure why you didn't add it inside Directive enum though instead of the new field_authorized_directives
EDIT: not worth changing though 🤷

The directives are already produced as-is during composition, but we need the already composed graph to construct the selection sets for Authorized. So the alternative would have been to split the directive type into two, then transform the intermediate directive enum type into the final directive enum type after we attach the selections, which was more work and a larger refactoring, and I was in a bit of a rush.

tomhoule:
>@authorized directive isn't present on objects or interfaces or I don't know where to look for them :D

Yes it's part of the missing points, I have an issue for it.

PR #1805, opened 2024-06-25
chore(engine-v2): Simplify FlatSelectionSet + fix requires within plan

Finistere:
The current implementation wasn't handling properly requires within a
plan, it would have tried to create a new plan with resolvers that don't
exist if the fields were only providable by the current subgraph. This
will only happen with directives like `@authorized` which isn't tied to
subgraph boundaries.

The initial goal of this PR was to simplify the flatten of selection
sets I'm relying on during planning. It had complex logic to deal with
chains of type conditions, but it's actually irrelevant. Type conditions
are only useful to construct an appropriate query. Once we've identified
the field definitions, they will present if the response object matches
the field parent entity (object/interface).

Fixes: GB-6977


tomhoule (approved):
:+1:

PR #1807, opened 2024-06-25
chore(engine-v2): Replace plan boundaries by entity locations

Finistere:
To read the necessary input for children plan we keep track of the
root response objects for those. This avoids the need to traverse the
response graph with complex logic. As we would only track those at plan
boundaries the term "plan boundary" objects, refs, items etc. were used.

Now with features like `@authenticated` we'll need to track response
objects at various places. The immediate need was to be able to identify
which extra fields were added for a `@authenticated` requirements to
avoid retrieving those if access was denied. Those extra field would be
retrieved through the tuple `(PlanBoundaryId, RequiredFieldId)`, now
it's `(EntityLocation, RequiredFieldId)` making it work even outside
plan boundaries.

Fixes: GB-6978

### DIFF
--- a/engine/crates/engine-v2/src/execution/coordinator.rs
+++ b/engine/crates/engine-v2/src/execution/coordinator.rs
@@ -1,4 +1,4 @@
-use std::{borrow::Cow, sync::Arc};
+use std::sync::Arc;
 
 use async_runtime::make_send_on_wasm;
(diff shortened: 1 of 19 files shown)


tomhoule (approved):
Nice to see the abstraction working

PR #1810, opened 2024-06-26
feat: hook up partial caching defer work

obmarg:
I've spent the past couple of weeks working on defer partial caching without actually hooking it up to anything.  I'd written most of the code I'd needed, but until it was half way functional I didn't want to call it from the gateway.

This PR (mostly) changes that:

- Added a new `execute_stream_v2` entrypoint to the Gateway.  This is very similar to the old `execute_stream` function, but it returns a `Stream<StreamingPayload>` rather than an http response.  This allows partial caching to work against this stream instead of having to deal with the streaming transports that were contained within the http response.
- This function calls into a new `StreamingExecutionPhase` struct, which is similar to the `ExecutionPhase` used by non-streaming transports.
- The `gateway_core` interface also now has an `execute_stream_v2` function.

Once all of this is implemented and hooked up I'll retire the `execute_stream` functions in favor of v2, but that won't be until next week at the earliest.

I've tested this with some integration tests that'll be in my next PR.  There are issues with it, but some of the basics work at least.  It's not yet usable outside of integration tests - I need to implement transports in the CLI & cloud gateway first.

Fixes GB-6746


engine/crates/gateway-core/src/cache/partial.rs
```diff
@@ -84,6 +72,138 @@ where
...
+}
+
+async fn run_execution_phase_stream(
+    mut execution_phase: StreamingExecutionPhase,
+    mut engine_stream: BoxStream<'static, engine::StreamingPayload>,
+    mut response_sender: mpsc::Sender<engine::StreamingPayload>,
+    cache: Cache,
+) {
+    let Some(StreamingPayload::InitialResponse(mut payload)) = engine_stream.next().await else {
+        todo!("GB-BLAH")
```
obmarg, line 138: Fix me

engine/crates/gateway-core/src/cache/partial.rs
```diff
@@ -84,6 +72,138 @@ where
...
+}
+
+async fn run_execution_phase_stream(
+    mut execution_phase: StreamingExecutionPhase,
+    mut engine_stream: BoxStream<'static, engine::StreamingPayload>,
+    mut response_sender: mpsc::Sender<engine::StreamingPayload>,
+    cache: Cache,
+) {
+    let Some(StreamingPayload::InitialResponse(mut payload)) = engine_stream.next().await else {
+        todo!("GB-6966");
```
yoav-lavi, line 138: Is this meant to still be here?

engine/crates/partial-caching/src/execution/defer.rs
```diff
@@ -0,0 +1,174 @@
...
+        path: &[&QueryPathSegment],
+        data: QueryResponse,
+        errors: bool,
+    ) -> QueryResponse {
+        let Some(output) = &mut self.output else {
+            todo!("GB-6966");
+        };
+
+        if !self.execution_phase.cache_entries.is_empty() {
+            // If we still have cache entries, we should merge them into the
```
yoav-lavi, line 111: Into the what?

obmarg: haha, good question.

yoav-lavi (approved):
Nice!

engine/crates/partial-caching/src/execution/defer.rs
obmarg: Clearly I am not doing that, oops - will fix

PR #1816, opened 2024-06-26
feat: cli in memory rate limiting

hfgbarrigas:
# Description

Adds in memory rate limiting for local dev through the CLI.

# Type of change

- [ ] 💔 Breaking
- [x] 🚀 Feature
- [ ] 🐛 Fix
- [ ] 🛠️ Tooling
- [ ] 🔨 Refactoring
- [ ] 🧪 Test
- [ ] 📦 Dependency

### DIFF
--- a/engine/crates/integration-tests/src/engine_v1/gateway.rs
+++ b/engine/crates/integration-tests/src/engine_v1/gateway.rs
@@ -82,12 +82,14 @@ impl GatewayBuilder {
         };
 
         // This AuthService is used to authenticate "user" requests
-        let auth = self.auth_service.unwrap_or_else(|| AuthService::new_v1(
-            self.auth_config,
-            KvStore::new(NoopKvStore),
-            UdfInvoker::new(self.authorizers.unwrap_or_default()),
-            "my-identity-is-ray".into(),
-        ));
+        let auth = self.auth_service.unwrap_or_else(|| {
+            AuthService::new_v1(
+                self.auth_config,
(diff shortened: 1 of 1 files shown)


engine/crates/integration-tests/src/engine_v1/gateway.rs
```diff
@@ -64,19 +82,21 @@ impl GatewayBuilder {
         };
 
         // This AuthService is used to authenticate "user" requests
-        let auth = AuthService::new_v1(
+        let auth = self.auth_service.unwrap_or(AuthService::new_v1(
```
obmarg, line 85: Not super important, but should this be `unwrap_or_else` to avoid always creating the AuthService?

engine/crates/runtime-local/src/rate_limiting.rs
```diff
@@ -0,0 +1,174 @@
...
+                .map(|rule| {
+                    let quota = rule
+                        .limit
+                        .checked_div(rule.duration.as_secs() as u32)
+                        .expect("rate limiter with invalid per second quota");
+
+                    (
+                        rule.condition.clone(),
+                        governor::RateLimiter::keyed(Quota::per_second(
+                            NonZeroU32::new(quota).expect("rate limit duration cannot be 0"),
```
obmarg, line 32: Should the parser be enforcing this non-zero constraint as well?  Took a quick look and couldn't see it doing that, correct me if I just missed it

obmarg (approved):

PR #1829, opened 2024-06-28
composition: implement validation of nested selection sets 

tomhoule:
In addition, fix a test that is broken on main, probably because of a bad
rebase. We didn't handle `@authorized` on objects in from_sdl.

closes gb-6965

### DIFF
--- a/engine/crates/composition/src/validate/selection.rs
+++ b/engine/crates/composition/src/validate/selection.rs
@@ -1,0 +1,142 @@
+use super::*;
+
+pub(super) fn validate_selections(ctx: &mut ValidateContext<'_>, field: subgraphs::FieldWalker<'_>) {
+    let directives = field.directives();
+    for (selection, directive_name) in directives
+        .requires()
+        .into_iter()
+        .flatten()
+        .map(|selection| (selection, "requires"))
+        .chain(
+            directives
+                .authorized()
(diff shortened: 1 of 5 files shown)


engine/crates/composition/src/validate/selection.rs
```diff
@@ -0,0 +1,142 @@
...
+            ctx,
+            selection,
+            field
+                .r#type()
+                .definition(on_definition.subgraph_id())
+                .expect("type is defined in subgraph"),
+            directive_path,
+            directive_name,
+        );
+    }
```
tomhoule, line 102: This block is the important part.

engine/crates/federated-graph/src/from_sdl.rs
```diff
@@ -315,9 +315,70 @@ fn ingest_object_interfaces(
...
+        let rule = state.insert_string(
+            authorized
+                .node
+                .get_argument("rule")
+                .and_then(|rule| match &rule.node {
+                    async_graphql_value::ConstValue::String(s) => Some(s.as_str()),
+                    _ => None,
+                })
+                .unwrap_or_default(),
+        );
```
pimeys, line 358: Not needed anymore...

pimeys (approved):

PR #1830, opened 2024-06-28
fix: remove rule from @authorized

pimeys:
For the demo. These hooks will change more next week...


Finistere (approved):

tomhoule:
The composition tests need to be adjusted, but I can take care of it later.

pimeys:
> The composition tests need to be adjusted, but I can take care of it later.

done in this pr

PR #1831, opened 2024-06-30
chore(engine-v2): Engine generic over the runtime & hooks trait update

Finistere:
Was a bit tedious to set up and encountered a bug with Rust compiler but
now engine is generic over its runtime. This allow us to have
considerably more flexibility in what we can do in our interfaces.

I've also updated the hook trait to be closer to the next RFC, but only
added the `pre-execution` hooks for now. I've simplified the error
handling also, by having a single `runtime::error::GraphqlError` instead
of `HookError` and `UserError`. This `GraphqlError` only has message and
extension, the rest will be added by the engine. The distinction between
a HookError & UserError isn't necessary for the engine, in all cases
it's treated as an error that must be shown.

To avoid re-compiling the engine for various runtimes in tests, I've
added a `TestRuntime` struct and added a `DynHooks` trait with default
implementation which allows providing dynamically dispatched hooks.


engine/crates/runtime-local/src/hooks.rs
```diff
@@ -1,68 +1,102 @@
...
-        let results = self
-            .0
-            .authorized(context, input)
+        context: &Self::Context,
+        definition: EdgeDefinition<'a>,
+        arguments: impl serde::Serialize + serde::de::Deserializer<'a> + Send,
+        _metadata: impl serde::Serialize + serde::de::Deserializer<'a> + Send,
+    ) -> Result<(), GraphqlError> {
+        let Some(ref loader) = self.0 else {
+            unreachable!("authorize_edge_pre_execution called without Wasm");
```
pimeys, line 50: We should fix this in the future. It will be a very common thing where the user adds an `@authorized` directive but fails to add the wasi side. Now we get a really ugly panic, and for sure this is not unreachable.

engine/crates/runtime-noop/src/hooks.rs
```diff
@@ -1,19 +1,34 @@
...
-    async fn on_gateway_request(&self, headers: HeaderMap) -> Result<(Self::Context, HeaderMap), HookError> {
-        Ok((HashMap::new(), headers))
+    async fn authorize_edge_pre_execution<'a>(
+        &self,
+        _: &Self::Context,
+        _: runtime::hooks::EdgeDefinition<'a>,
+        _: impl serde::Serialize + serde::de::Deserializer<'a> + Send,
+        _: impl serde::Serialize + serde::de::Deserializer<'a> + Send,
+    ) -> Result<(), runtime::error::GraphqlError> {
+        unreachable!("@authorization directive not available outside of local context")
```
pimeys, line 23: Could this be called if somebody deploys a schema with `@authorization` to a managed runtime on Cloudflare and we crash to unreachable. So this would not really be unreachable, it can be reached very easily.

pimeys (approved):
Awesome, removes a ton of vtables which is nice. The error handling for certain things needs more work, but can be done in subsequent PRs.

engine/crates/runtime-noop/src/hooks.rs
Finistere: yeah you're right, changing that quickly

PR #1836, opened 2024-07-01
cli: error early when we fail to find bun on NixOS

tomhoule:
We already had the error ("Could not find a `bun` executable in PATH. Bun is required in order to evaluate your TypeScript configuration. Please install `bun` or run `nix shell nixpkgs#bun`."), so this is a short fix.

closes GB-7005


yoav-lavi (approved):
Nice!

PR #1838, opened 2024-07-01
Pooling and auth refactor

pimeys:
The complexity here is for a situation if the hook traps, we can't call it again. We check the return value and reinitialize it only if it succeeds. If the hook traps, deadpool will remove the instance from the pool.

This will cut the hook execution time to half, about 100us off.

Closes: GB-7017
Closes: GB-7016


engine/crates/runtime-local/src/hooks/pool.rs
```diff
@@ -0,0 +1,54 @@
...
+impl managed::Manager for GatewayHookManager {
+    type Type = GatewayHookInstance;
+    type Error = wasi_component_loader::Error;
+
+    async fn create(&self) -> Result<Self::Type, Self::Error> {
+        GatewayHookInstance::new(&self.component_loader).await
+    }
+
+    async fn recycle(&self, instance: &mut Self::Type, _: &managed::Metrics) -> managed::RecycleResult<Self::Error> {
+        instance.cleanup()?;
```
pimeys, line 25: This must fail if the instance is in a broken state, which means the pool will drop the instance and create a new one.

Finistere (approved):

engine/crates/wasi-component-loader/gateway-hooks.wit
```diff
@@ -72,21 +86,30 @@ interface gateway-request {
...
-    use types.{error-response, shared-context};
+    use types.{error-response, shared-context, edge-definition, node-definition};
 
-    // The hook is called in the request cycle if the schema defines an authorization directive.
-    // The context is read only, and provides data written in the gateway request hook.
-    //
-    // The input is combined from the directive arguments and fields, taking the data from the 
-    // provided arguments and the field values as defined in the directive.
+    // The hook is called in the request cycle if the schema defines an authorization directive on
+    // an edge, providing the arguments of the edge and the metadata of the directive to the hook.
```
tomhoule, line 92: Should we clarify that the arguments are not all arguments but only the ones selected in the directive?

tomhoule (approved):
Nice!

PR #1844, opened 2024-07-02
cli: make grafbase schema return federated SDL

tomhoule:
The `grafbase schema` command returns the API schema for federated graphs, but that can already be fetched with `grafbase introspect`.

This commit:

- Makes it return the federated SDL for federated graphs
- Makes it work with graph refs that don't contain a branch (it then defaults to the graph's production branch)

Alternatively, we could add a flag to return either api schema or federated sdl.

closes GB-7015

### DIFF
--- a/cli/crates/backend/src/api/schema.rs
+++ b/cli/crates/backend/src/api/schema.rs
@@ -1,5 +1,8 @@
 use crate::api::graphql::queries::{
-    fetch_federated_graph_schema::{FetchFederatedGraphSchemaArguments, FetchFederatedGraphSchemaQuery},
+    fetch_federated_graph_schema::{
+        FetchFederatedGraphSchemaArguments, FetchFederatedGraphSchemaProductionBranchArguments,
+        FetchFederatedGraphSchemaProductionBranchQuery, FetchFederatedGraphSchemaQuery,
+    },
     fetch_subgraph_schema::{FetchSubgraphSchemaArguments, FetchSubgraphSchemaQuery},
 };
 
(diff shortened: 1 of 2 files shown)


cli/crates/backend/src/api/schema.rs
```diff
@@ -47,16 +45,35 @@ async fn subgraph_schema(
...
 
-    response
-        .data
-        .as_ref()
-        .and_then(|data| data.branch.as_ref())
-        .ok_or_else(|| ApiError::SubgraphsError(format!("{response:#?}")))
-        .map(|branch| branch.schema.clone())
+    if let Some(branch) = branch {
+        let operation =
+            FetchFederatedGraphSchemaQuery::build(FetchFederatedGraphSchemaArguments { account, graph, branch });
```
obmarg, line 53: I may be missing some reason why it doesn't work, but seems like you could have made `FetchFederatedGraphSchemaArguments` take a `branch: Option<&str>` and reused it for both of these cases.

Not a big deal though

obmarg (approved):

cli/crates/backend/src/api/schema.rs
tomhoule: I'm not sure what it would look like, since the branch is required on `Query.branch`. We would need to branch on whether the branch is provided or not. Maybe there's another field in the API that would work?

obmarg: Ah yeah - I was indeed missing a reason why it wouldn't work.  Ignore this comment

PR #1846, opened 2024-07-02
feat: support nested defers in partial caching

obmarg:
Prior to this change partial caching didn't work properly with nested defers. This fixes that: mostly by tracking active defers in incremental payloads (similar to what we do for the initial response) and then using that in the cache merging for deferred payloads.

Also unified the merging of deferred payloads & the initial payload as part of this, since the logic is almost identical but the implementations were not.

Fixes GB-6963


tomhoule (approved):
nice!

PR #1852, opened 2024-07-03
feat(nix): build script for grafbase-gateway

pimeys:
Allows building grafbase-gateway through nix.

```bash
nix build .#grafbase-gateway
```

### DIFF
--- a/gateway/nix/grafbase-gateway.nix
+++ b/gateway/nix/grafbase-gateway.nix
@@ -1,0 +1,73 @@
+{
+  pkgs,
+  crane,
+  lib,
+  ...
+}: let
+  rustToolchain = pkgs.rust-bin.stable.latest.default;
+  craneLib = (crane.mkLib pkgs).overrideToolchain rustToolchain;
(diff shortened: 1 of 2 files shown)


gateway/nix/grafbase-gateway.nix
```diff
@@ -0,0 +1,78 @@
...
+    package.json
+    assets.tar.gz
+
+    # We can't ignore /packages wholesale because we need to include
+    # grafbase-sdk/package.json later.
+    /packages/**/*.json
+    /packages/**/*.js
+    /packages/**/*.ts
+    /packages/**/*.html
+
```
tomhoule, line 29: Would an extraIgnore on `cli/` work here?

tomhoule (approved):
:100:

gateway/nix/grafbase-gateway.nix
pimeys: no because the workspace cargo lock references crates there and the build would fail

PR #1857, opened 2024-07-03
chore: add supertype info to partial caching registry

obmarg:
Partial caching needs to know the relationships between types in order to handle type conditions correctly.  This is essential for us to support APIs that use unions or interfaces.

This PR is pretty simple: it just adds those details into the PartialCacheRegistry (which I have realised is named somewhat ambiguously - it's a partial registry for caching, not a registry for partial caching.  May need to fix that sometime).

I am very aware that this could bump up the size of the caching registry a bit, and therefore slow down cache responses.  I have tried to minimise that as much as possible though.  It's structured as a map of `StringId -> IdRange<SupertypeId>`, where `SupertypeId` is an index into a separate Vec.  Using this separate Vec allows us to deduplicate things a bit - if we have an interface like `Node` that has 1000s of subtypes we don't necessarily need to store `Node` 1000s of times.

Fixes GB-6942

### DIFF
--- a/engine/crates/engine/registry-for-cache/src/ids.rs
+++ b/engine/crates/engine/registry-for-cache/src/ids.rs
@@ -13,4 +13,6 @@ make_id!(InterfaceTypeId, InterfaceTypeRecord, interfaces, PartialCacheRegistry)
 
 make_id!(OtherTypeId, OtherTypeRecord, others, PartialCacheRegistry);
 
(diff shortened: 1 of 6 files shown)


tomhoule (approved):
:+1:

Finistere (approved):

pimeys (approved):

PR #1859, opened 2024-07-04
refactor polymorphic shape in partial caching

obmarg:
Partial caching precalculates the shape of all the objects that can be returned from a given query.  When it encounters a selection set with a type condition in it, we mark that selection set as having a polymorphic shape, with a different concrete shape for each type condition that appears in the selection set.

Prior to attempting to work with these polymorphic shape I'd assumed I could just put them into a list, and then when we get a `__typename` we just iterate over that list until we find a match.  This was a bit naive though, as more than one typecondition can match a given typename.  So with this approach for every new object we see, we need to check all the conditions, and then if > 1 match we need to figure out which is the most specific.  

So instead, I'm now storing these in a graph: if a condition matches then you need to check only the conditions under that match, and not any of the other branches of the graph.  This was a bit more work to pre-calculate, but should be more efficient when handling responses, so I think the tradeoff is worth it.

Part of GB-6949


pimeys (approved):

engine/crates/partial-caching/src/output/shapes/shape_builder.rs
```diff
@@ -0,0 +1,249 @@
...
+        nodes: impl ExactSizeIterator<Item = Option<TypeConditionId>>,
+    ) -> Box<[TypeConditionId]> {
+        nodes
+            .map(|option| option.expect("the node to be present because we did a topsort"))
+            .collect::<Vec<_>>()
+            .into_boxed_slice()
+    }
+}
+
+fn topsort(adjacency_list: &Vec<Vec<usize>>) -> Result<Vec<usize>, ()> {
```
Finistere, line 120: would have named it `topological_sort`

PR #1862, opened 2024-07-04
feat: support type conditions in partial caching

obmarg:
Prior to this commit we'd just panic if we found a polymorphic shape when we were handling a partial caching response.  This commit fixes that: we now use walk the possible type conditions looking for the most specific match, and then use the corresponding concrete type.

I've also added a bunch of integration tests, which are the bulk of this PR.

One thing I haven't done yet is update the cache writing code to write type names into the cache entry.  I'll do that in my next PR, as this one is already quite large.

Fixes GB-6949


engine/crates/partial-caching/src/output/shapes/mod.rs
```diff
@@ -154,12 +151,61 @@ impl<'a> ConcreteShape<'a> {
...
+        ) -> ConcreteShapeId {
+            let condition_match = conditions.iter().find(|id| {
+                type_relationships
+                    .type_condition_matches(shapes.type_conditions[id.0 as usize].type_condition.as_str(), typename)
+            });
+
+            match condition_match {
+                Some(id) => {
+                    let node = &shapes.type_conditions[id.0 as usize];
+                    search(
```
Finistere, line 188: A bit weird to have a search within a search when you already did a `find()` imho

Finistere (approved):

PR #1866, opened 2024-07-05
chore(engine-v2): Early validation of parsed operation

Finistere:
In a further commit checking for fragment cycles is pretty hard, so
doing it just after parsing in a separate step now. I've also moved
all operation limits check that could be verified at this step (all but
height).

### DIFF
--- a/engine/crates/engine-v2/src/operation/bind/validation.rs
+++ b/engine/crates/engine-v2/src/operation/bind/validation.rs
@@ -1,0 +1,130 @@
+use config::latest::OperationLimits;
+use engine::Positioned;
+
+use crate::operation::parse::ParsedOperation;
+
+use super::{BindError, BindResult};
+
+pub(super) fn validate_parsed_operation(operation: &ParsedOperation, limits: &OperationLimits) -> BindResult<()> {
+    Visitor {
+        operation,
+        current_fragments_stack: Vec::new(),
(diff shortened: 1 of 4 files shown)


engine/crates/engine-v2/src/operation/bind/validation.rs
```diff
@@ -0,0 +1,133 @@
...
+impl<'p> Visitor<'p> {
+    fn visit_selection_set(
+        &mut self,
+        root: bool,
+        selection_set: &'p Positioned<engine_parser::types::SelectionSet>,
+    ) -> BindResult<()> {
+        for item in &selection_set.items {
+            match &item.node {
+                engine_parser::types::Selection::Field(field) => {
+                    self.root_fields += root as usize;
```
tomhoule, line 46: Isn't root fields only the top-most selection set?

tomhoule: (if so, this can be wrapped in an `if current_depth == 0`)

tomhoule (approved):

obmarg (approved):

engine/crates/engine-v2/src/operation/bind/validation.rs
Finistere: you're right

PR #1868, opened 2024-07-05
chore(engine-v2): flatten selection sets immediately in the binding

Finistere:
Initially, I tried to keep the query structure with fragments intact believing it would be useful for directives and such. But after dealing with it and better understanding of GraphQL it's pretty much useless server-side. Fragments are just a tool to _write_ queries, nothing more or less. Executable directives like `@include`, `@defer`, etc. all require custom support from the engine and they can be associated with fields directly.

During the different steps after the binding, I would never rely on those:
- the planning step flattens selection sets* and only needs to process them by their parent entity (object/interface) as that is the scope of `@requires`, `@provides` etc.
- the output shape computation would also need to group them
- the query generation for the subgraphs would also ignore fragments because only some of their fields might be requested.

*By flattening I mean grouping all fields, including fragment ones, into a single selection set.

Keeping this structure led to a bunch of complications though and a lot of noise.

So this PR focuses on flattening the selection sets immediately in the binding step so that later ones wouldn't have to do it. This considerably simplifies later steps and allowed me to start doing some optimizations on the selection sets such as deduplicating fields in a trivial way. Later the flattening process would ideally be a bit smarter and deduplicate selection sets as we lose all fragment information, and maybe other optimizations.

As the binding step got more complex with it, I also split it up into multiple files.


engine/crates/engine-v2/src/lib.rs
```diff
@@ -1,4 +1,5 @@
 #![deny(clippy::future_not_send)]
+#![allow(dead_code)]
```
obmarg, line 2: This seems a slightly bad thing to allow at a crate level.

engine/crates/engine-v2/src/plan/planning/boundary.rs
```diff
@@ -1,279 +1,265 @@
...
-    response::{ResponseKey, SafeResponseKey, UnpackedResponseEdge},
+    response::{SafeResponseKey, UnpackedResponseEdge},
 };
 
 /// The Planner traverses the selection sets to plan all the fields, but it doesn't define the
 /// plans directly. That's the job of the BoundaryPlanner which will attribute a plan for each
 /// field for a given selection set and satisfy any requirements.
-pub(super) struct BoundarySelectionSetPlanner<'schema, 'a> {
-    planner: &'a mut OperationSolver<'schema>,
+pub(super) struct SelectionSetSolver<'schema, 'a> {
```
obmarg, line 25: Docstring needs updated for the rename I think?

obmarg (approved):

PR #1870, opened 2024-07-08
Add health checks to self-hosted federated gateway

tomhoule:
Please react if you see additional health checks we could do in the handler.

closes GB-7048

### DIFF
--- a/gateway/crates/gateway-binary/tests/integration_tests.rs
+++ b/gateway/crates/gateway-binary/tests/integration_tests.rs
@@ -167,6 +167,18 @@ impl Client {
             bearer: None,
         }
     }
+
+    pub async fn get<Response>(&self, path: &str) -> (http::StatusCode, Response)
+    where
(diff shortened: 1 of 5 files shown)


gateway/crates/gateway-binary/tests/integration_tests.rs
```diff
@@ -654,3 +666,20 @@ fn hybrid_graph() {
...
+    with_static_server(config, &schema, None, None, |client| async move {
+        let (_, result): (_, serde_json::Value) = client.get("/health").await;
+
+        insta::assert_json_snapshot!(&result, @r###"
+        {
+          "status": "healthy"
+        }
+        "###);
+    });
+}
```
tomhoule, line 685: I'd have liked to have a test for the unhealthy status, but I couldn't make it happen — seems we fail early before there's a chance to become unhealthy, which is a good thing.

pimeys (approved):

PR #1871, opened 2024-07-08
feat(engine-v2): Add error code for all GraphQL errors

Finistere:
Adding a mandatory error code in engine-v2 errors.
I'm not sure how we should handle custom error codes from hooks though if we want to support that later.

I have renamed the `runtime::error:{GraphqlError, ErrorCode}` to `runtime::error::{PartialGraphqlError, PartialErrorCode}` as they're extended by the engine on top. Not a great name, but avoids having identical names at least.

### DIFF
--- a/engine/crates/engine-v2/src/engine.rs
+++ b/engine/crates/engine-v2/src/engine.rs
@@ -28,7 +28,7 @@ use crate::{
     http_response::{HttpGraphqlResponse, HttpGraphqlResponseExtraMetadata},
     operation::{Operation, OperationMetadata, Variables},
     plan::OperationPlan,
-    response::{GraphqlError, Response},
+    response::{ErrorCode, GraphqlError, Response},
     websocket,
 };
 
@@ -100,7 +100,7 @@ impl<R: Runtime> Engine<R> {
     ) -> HttpGraphqlResponse {
         let (hooks_context, headers) = match self.runtime.hooks().on_gateway_request(headers).await {
             Ok(result) => result,
-            Err(error) => return Response::execution_error(error).into(),
+            Err(error) => return Response::pre_execution_error(error).into(),
         };
 
         if let Some(access_token) = self.auth.authorize(&headers).await {
@@ -283,13 +283,14 @@ impl<'ctx, R: Runtime> PreExecutionContext<'ctx, R> {
 
         let metadata = Some(operation.metadata.clone());
         let response = if matches!(operation.ty(), OperationType::Subscription) {
-            Response::bad_request(GraphqlError::new(
+            Response::pre_execution_error(GraphqlError::new(
                 "Subscriptions are only suported on streaming transports. Try making a request with SSE or WebSockets",
+                ErrorCode::BadRequest,
             ))
         } else {
             match self.finalize_operation(operation, request.variables).await {
                 Ok(operation_plan) => self.execute_query_or_mutation(operation_plan).await,
-                Err(errors) => Response::bad_request_from_errors(errors),
+                Err(errors) => Response::pre_execution_errors(errors),
             }
         };
 
@@ -315,7 +316,7 @@ impl<'ctx, R: Runtime> PreExecutionContext<'ctx, R> {
         let operation_plan = match self.finalize_operation(operation, request.variables).await {
             Ok(operation_plan) => operation_plan,
             Err(errors) => {
-                let response = Response::bad_request_from_errors(errors);
+                let response = Response::pre_execution_errors(errors);
                 let status = response.status();
                 sender.send(response).await.ok();
                 return (metadata, status);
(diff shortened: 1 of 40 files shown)


engine/crates/engine/src/error.rs
```diff
@@ -8,18 +8,47 @@ use std::{
...
+    SubgraphError,
+    SubgraphInvalidResponseError,
+    SubgraphRequestError,
+    // Auth
+    Unauthenticated,
+    Unauthorized,
+    // Operation preparation phases
+    OperationParsingError,
+    OperationValidationError,
+    OperationPlanningError,
```
obmarg, line 38: Why do engine-v2 errors live inside engine-v1?

Finistere: because the `ErrorCode` was initially created for `PersistedQueryNotFound`. Can create one specific to engine-v2 though if you prefer

obmarg: Not going to insist, but seems like it'd make more sense from an organisational perspective.

obmarg: (Bearing in mind I have thought about this for all of 30 seconds - feel free to disagree)

engine/crates/engine-v2/src/engine.rs
```diff
@@ -285,6 +285,7 @@ impl<'ctx, R: Runtime> PreExecutionContext<'ctx, R> {
         let response = if matches!(operation.ty(), OperationType::Subscription) {
             Response::bad_request(GraphqlError::new(
                 "Subscriptions are only suported on streaming transports. Try making a request with SSE or WebSockets",
+                ErrorCode::BadRequest,
```
pimeys, line 288: If we already call `Response::bad_request`, could this automatically generate a `GraphqlError` with the corresponding error code?

engine/crates/engine-v2/src/response/error.rs
```diff
@@ -1,44 +1,68 @@
...
-    pub locations: Vec<crate::operation::Location>,
+    pub message: Cow<'static, str>,
+    pub code: ErrorCode,
+    pub locations: Vec<Location>,
     pub path: Option<ResponsePath>,
-    // ensures consistent ordering for tests
-    pub extensions: BTreeMap<String, serde_json::Value>,
+    // Serialized as a map, but kept as a Vec for efficiency.
+    // "code" must not be present
+    pub extensions: Vec<(Cow<'static, str>, serde_json::Value)>,
```
pimeys, line 17: This is kind of breaking, because the btree is ordered by the key and this is not. It's probably fine, just hilighting this here.

engine/crates/engine-v2/src/response/error.rs
```diff
@@ -1,44 +1,68 @@
...
+            path: None,
+            extensions: Vec::new(),
         }
     }
 
-    pub fn with_error_code(mut self, code: ErrorCode) -> Self {
-        self.extensions
-            .insert("code".to_string(), serde_json::Value::String(code.to_string()));
+    #[must_use]
+    pub fn with_location(mut self, location: Location) -> Self {
```
pimeys, line 32: Builder stuff is awesome here.

engine/crates/engine-v2/src/response/read/ser.rs
```diff
@@ -105,9 +108,32 @@ impl<'a> serde::Serialize for SerializableError<'a> {
...
+impl<'a> serde::Serialize for SerializableExtension<'a> {
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: serde::Serializer,
+    {
+        let mut map = serializer.serialize_map(Some(self.extensions.len() + 1))?;
+        for (key, value) in self.extensions {
+            map.serialize_entry(key, value)?;
         }
+        map.serialize_entry("code", &self.code)?;
```
pimeys, line 136: What happens if a hook adds a code to the error extensions. Do we get two codes in the response?

engine/crates/engine/src/error.rs
pimeys: :+1:

pimeys (approved):
Approving with caveats to check out before merging.

engine/crates/engine-v2/src/response/error.rs
Finistere: it's serialized as a map, no ordering guarantees in GraphQL, but yeah fair point

engine/crates/engine-v2/src/response/read/ser.rs
Finistere: gets ignored today, I should probably just let it through though

engine/crates/engine-v2/src/engine.rs
Finistere: fair point

Finistere: Ok not the same thing, bad request here means it's a request error in the GraphQL-spec sense. So no `data` field. Error could also be validation one or persisted query not found.

Finistere: renamed to `pre_execution_error` which is closer to the spec meaning

PR #1873, opened 2024-07-10
fix(engine-v2): Add missing (sanitized) query in traces

Finistere:
Adding missing (sanitized) GraphQL query in traces and cleaning up a few things. Renamed `gql.document` to `gql.operation.query` as it's more accurate. Removed some of the tracing v2 tests, the tracing mock ones are super hard to get right and understand. We should just test against the output in ClickHouse as we do for metrics.

### DIFF
--- a/engine/crates/engine-v2/src/sources/graphql/federation.rs
+++ b/engine/crates/engine-v2/src/sources/graphql/federation.rs
@@ -101,12 +101,15 @@ pub(crate) struct FederationEntityExecutor<'ctx, R: Runtime> {
 impl<'ctx, R: Runtime> FederationEntityExecutor<'ctx, R> {
     #[tracing::instrument(skip_all)]
     pub async fn execute(self, mut response_part: ResponsePart) -> ExecutionResult<ResponsePart> {
-        let subgraph_gql_request_span = SubgraphRequestSpan::new(self.subgraph.name())
-            .with_url(self.subgraph.url())
-            .with_operation_type(OperationType::Query.as_ref())
-            // The query string contains no input values, only variables. So it's safe to log.
-            .with_document(&self.operation.query)
-            .into_span();
+        let span = SubgraphRequestSpan {
+            name: self.subgraph.name(),
+            operation_type: OperationType::Query.as_str(),
+            // The generated query does not contain any data, everything are in the variables, so
+            // it's safe to use.
+            sanitized_query: &self.operation.query,
+            url: self.subgraph.url(),
+        }
+        .into_span();
 
         async {
             let bytes = self
@@ -150,7 +153,7 @@ impl<'ctx, R: Runtime> FederationEntityExecutor<'ctx, R> {
             .deserialize(&mut serde_json::Deserializer::from_slice(&bytes))?;
             Ok(response_part)
         }
-        .instrument(subgraph_gql_request_span.clone())
+        .instrument(span.clone())
         .await
     }
 }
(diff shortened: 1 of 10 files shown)


obmarg (approved):

engine/crates/engine-v2/src/sources/graphql/federation.rs
```diff
@@ -101,12 +101,15 @@ pub(crate) struct FederationEntityExecutor<'ctx, R: Runtime> {
...
-            .with_operation_type(OperationType::Query.as_ref())
-            // The query string contains no input values, only variables. So it's safe to log.
-            .with_document(&self.operation.query)
-            .into_span();
+        let span = SubgraphRequestSpan {
+            name: self.subgraph.name(),
+            operation_type: OperationType::Query.as_str(),
+            // The generated query does not contain any data, everything are in the variables, so
+            // it's safe to use.
+            sanitized_query: &self.operation.query,
```
pimeys, line 109: Is this still sanitized though, should we consider a different name?

engine/crates/engine-v2/src/sources/graphql/mod.rs
```diff
@@ -94,12 +94,15 @@ pub(crate) struct GraphqlExecutor<'ctx, R: Runtime> {
...
-            .with_operation_type(self.operation.ty.as_ref())
-            // The query string contains no input values, only variables. So it's safe to log.
-            .with_document(&self.operation.query)
-            .into_span();
+        let span = SubgraphRequestSpan {
+            name: self.subgraph.name(),
+            operation_type: self.operation.ty.as_str(),
+            // The generated query does not contain any data, everything are in the variables, so
+            // it's safe to use.
+            sanitized_query: &self.operation.query,
```
pimeys, line 102: Sam here, should we maybe consider a different name here because the query is not really sanitized?

engine/crates/integration-tests/tests/tracing/v1.rs
```diff
@@ -106,20 +124,29 @@ async fn query_named() {
         .with_filter(|meta| meta.is_span() && meta.target() == "grafbase" && *meta.level() >= Level::INFO)
         .new_span(graphql_span.clone())
         .enter(graphql_span.clone())
+        .new_span(
+            resolver_span
+                .clone()
+                .with_field(expect::field("resolver.name").with_value(&"test")),
+        )
+        // Who knows why we're doing that
```
pimeys, line 132: git blame?

pimeys (approved):
Awesome. Just a few questions, still approving...

engine/crates/engine-v2/src/sources/graphql/federation.rs
Finistere: yes it is, the query generated for the subgraph has no query variables except enums. Everything else are passed on as variables

engine/crates/integration-tests/tests/tracing/v1.rs
Finistere: I meant I don't know why we're entering & exiting the resolver span, but tracing mock is tricky. It tracks everything that is happening to a span, meaning also when tokio decides to switch to a different future.... so who knows 🤷

engine/crates/engine-v2/src/sources/graphql/mod.rs
Finistere: same thing

engine/crates/engine-v2/src/sources/graphql/federation.rs
pimeys: It's still a bit different, no? The sanitized query e.g. changes all strings to empty strings, all numbers to 0 etc.

Finistere: that's the normalized query you're talking about. Hence why I chose "sanitized" here. I'm not super convinced normalized query is that much useful to be honest, but it's also sanitized so well works for now.

Finistere: the only goal here is to prevent any sensitive data leaks in the logs. Query normalization is used for analytics currently, which is different and where I have my doubts. But as said it's sanitized so also works for the logs.

pimeys: Fair. Sashay away!

PR #1875, opened 2024-07-10
fix: defer served entirely from partial cache

obmarg:
When testing defer on production, I saw some weird behaviour: sometimes a cached defer would not return some of the fields.

This turned out to be because I'd forgotten to account for the case where a defer is served entirely from the cache and doesn't make it as far as the query we send to the executor.

Easy to fix though: just need to check which defers aren't in the executor partition and then count those amongst the active defers when doing the initial cache merge.


tomhoule (approved):
Makes sense

PR #1881, opened 2024-07-10
chore(engine-v2): Add shape partitioning

Finistere:
Add the logic to partition the output of a field into distinct object ID subsets with the same object fields (shape)

### DIFF
--- a/engine/crates/engine-v2/src/execution/planner/partition.rs
+++ b/engine/crates/engine-v2/src/execution/planner/partition.rs
@@ -1,0 +1,706 @@
+use std::borrow::Cow;
+
+#[derive(Debug, PartialEq, Eq, PartialOrd, Ord)]
+pub(super) enum Partition<Id> {
+    One(Id),
+    Many(Vec<Id>),
+}
+
+impl<Id> From<Id> for Partition<Id> {
+    fn from(id: Id) -> Self {
+        Partition::One(id)
+    }
+}
+
+impl<Id: Copy> From<Vec<Id>> for Partition<Id> {
+    fn from(ids: Vec<Id>) -> Self {
+        match ids[..] {
+            [] => unreachable!(),
+            [id] => Partition::One(id),
+            _ => Partition::Many(ids),
+        }
+    }
+}
+
+/// For selection sets with type conditions, we must know the concrete object id being
+/// resolved to apply correctly the field collection as defined in the specification:
+/// https://spec.graphql.org/October2021/#CollectFields()
+///
+/// However, we're building the expect shape of the subgraph response in advance to drive the
+/// de-serialization, so we don't have it! So we need to pre-compute all possible shapes and
+/// use the right one later.
+///
+/// This function will create a partition set of the selection set possible types that guarantees
+/// the following:
+/// - Each partition is disjoint.
+/// - All types in a partition have the same fields selected.
+///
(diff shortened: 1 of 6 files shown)


obmarg (approved):
Approved, but with the proviso that I only sort of understand what I'm approving

engine/crates/engine-v2/src/execution/planner/partition.rs
```diff
@@ -0,0 +1,775 @@
...
+/// https://spec.graphql.org/October2021/#CollectFields()
+///
+/// However, we're building the expect shape of the subgraph response in advance to drive the
+/// de-serialization, so we don't have it! So we need to pre-compute all possible shapes and
+/// use the right one later.
+///
+/// This function will create a partition set of the output object ids that guarantees
+/// the following:
+/// - Each partition is disjoint.
+/// - All object ids in a partition will have the same fields.
```
obmarg, line 36: Wasn't clear from our slack chat whether this line was actually true?  Worth rewording/removing/something?

engine/crates/engine-v2/src/execution/planner/partition.rs
```diff
@@ -0,0 +1,775 @@
...
+///
+/// To do so it's given:
+/// - output_ids: the list of possible object ids from that the parent field may return
```
obmarg, line 39: is it worth framing this in terms of the current selection set rather than the "parent field" - I can see what you mean after we've chatted about this for a while, but wonder if "the list of possible object ids for the current selection set" or similar would be clearer.

engine/crates/engine-v2/src/execution/planner/partition.rs
```diff
@@ -0,0 +1,775 @@
...
+/// - type_conditions: All possible type conditions, each treated as a list of possible object ids.
```
obmarg, line 40: Again, maybe could be framed in terms of the current selection set?

"Each type condition in the selection set" rather than "all possible"?  Maybe also worth clarifying that "possible object ids" means the object ids that would match the given type condition?  (Kind of implied by what you've got, but not sure how obvious it was).

engine/crates/engine-v2/src/execution/planner/partition.rs
```diff
@@ -0,0 +1,775 @@
...
+/// Generics are only used for easier testing.
+///
+#[allow(dead_code)]
+pub(super) fn partition_shapes<Id: Copy + Ord + std::fmt::Debug>(
+    // Must be sorted and unique
+    output_ids: &[Id],
+    // Must be sorted and unique
+    mut single_object_ids: Vec<Id>,
+    // Individual arrays must be sorted and unique
+    mut type_conditions: Vec<&[Id]>,
```
obmarg, line 62: `output_ids` in particular seems like a confusing name?  `all_possible_types` or something?  Not sure.

engine/crates/engine-v2/src/execution/planner/partition.rs
```diff
@@ -0,0 +1,775 @@
...
+        }
+    }
+
+    // All possible cases are defined as single object ids
+    if single_object_ids.len() == output_ids.len() {
+        return Some(single_object_ids.into_iter().map(Into::into).collect());
+    }
+
+    // Everything relies on sorted arrays.
+    single_object_ids.sort_unstable();
```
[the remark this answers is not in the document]
Finistere: ah good point, wasn't sorted initially, leftover

PR #1886, opened 2024-07-11
chore(hooks): add missing node_post_execution method, more generics and split trait

Finistere:
Add the missing method and split the authorization part of the `Hooks` into a different one. It started to be overcrowded.
Using generics for all inputs for more flexibility.


pimeys (approved):
supah!

PR #1902, opened 2024-07-16
chore(engine-v2/id-newtypes): few improvements, bitset and idtomany

Finistere:
Adding two new structs: BitSet and IdToMany. The first is obvious, the second is used to store a list of `(Id, Value)` to provide all `Value` associated with an `Id`. It's not great yet, might never be. It's an attempt to avoid using `usize::from(id)` too much to access data for an id, as we lose type safety, and have one correct implementation. I made the mistake initially of using `binary_search` rather than `partition_point` here for exampe.

### DIFF
--- a/engine/crates/engine-v2/id-newtypes/src/many.rs
+++ b/engine/crates/engine-v2/id-newtypes/src/many.rs
@@ -1,0 +1,71 @@
+#[derive(serde::Serialize, serde::Deserialize)]
+pub struct IdToMany<Id, V>(Vec<(Id, V)>);
+
+impl<Id, V> Default for IdToMany<Id, V> {
+    fn default() -> Self {
+        Self(Vec::new())
+    }
+}
+
+impl<Id: Ord + Copy, V> FromIterator<(Id, V)> for IdToMany<Id, V> {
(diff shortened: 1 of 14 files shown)


engine/crates/engine-v2/id-newtypes/src/many.rs
```diff
@@ -0,0 +1,71 @@
...
+    pub fn get(&self, id: Id) -> impl Iterator<Item = &V> + '_
+    where
+        Id: Copy + Ord,
+    {
+        ValueIterator {
+            relations: self,
+            id,
+            idx: self.0.partition_point(|probe| probe.0 < id),
+        }
+    }
```
obmarg, line 50: Not super important but `get` seems an odd name for an operation that's not a straight lookup _and_ returns an iterator.

`find_all` or something maybe?  Not sure

obmarg (approved):

engine/crates/engine-v2/id-newtypes/src/many.rs
Finistere: oh yeah better. Wanted to use `Index` originally but it can only return a reference so couldn't use it

PR #1905, opened 2024-07-17
feat: support self-hosted federation in-memory rate limiting

hfgbarrigas:


engine/crates/engine-v2/src/engine/rate_limit.rs
```diff
@@ -0,0 +1,92 @@
...
+        None
+    }
+
+    fn key(&self) -> Option<&str> {
+        Some(self.0)
+    }
+}
+
+#[derive(Default)]
+pub struct InMemoryRateLimiter {
```
Finistere, line 39: Should be in `runtime-local`

Finistere: the interface probably needs to be adjusted, but the implementation (`governor`) can't be in engine-v2

Finistere (approved):

engine/crates/engine-v2/src/engine.rs
```diff
@@ -103,6 +105,14 @@ impl<R: Runtime> Engine<R> {
             Err(response) => return HttpGraphqlResponse::build(response, format, Default::default()),
         };
 
+        if let Err(err) = self.runtime.rate_limiter().limit(&RateLimitContext::Global).await {
```
Finistere, line 108: should also be called in `execute_websocket` underneath. Maybe create_session, but would need some tweaking

PR #1906, opened 2024-07-17
chore(engine-v2): Create ResponseModifier

Finistere:
`ResponseModifier` are built in the same way QueryModifiers are during the
operation preparation. As their name suggests, they will modify the
response instead of modifying the query.

I've implemented most of the code for the response modifier in this PR:

- struct themselves
- authorized directives requirements should be taken into account during planning
- being able to forward a view of the response objects to the
  hooks
- As a response modifier may be used on various selection sets but
  should be called at most once per subgraph response I
  created a `FilteredResponseObjectSet` which can iterate over multiple
  `ResponseObjectSet`
- I've renamed `ResponsePart` to `SubgraphResponse` for hopefully a
  bit more clarity, not sure though... I've tweaked a bit how the executor receives it in this PR, but I'm changing it back next PR so well 🤷 

I've stopped there because defining the "ReadSelectionSet", and what and how
we read from the response is a bit broken outside of plan root fields.
I'm not entirely sure what I'll do here, the best solution would be to
add the necessary metadata in the shapes.


engine/crates/engine-v2/src/response/object_set.rs
```diff
@@ -0,0 +1,142 @@
...
+
+const SET_INDEX_OFFSET: usize = 24;
+const OBJECT_INDEX_MASK: u32 = (1 << SET_INDEX_OFFSET) - 1;
+
+#[derive(Default, Clone)]
+pub(crate) struct FilteredResponseObjectSet {
+    sets: Vec<Arc<ResponseObjectSet>>,
+    // Upper 8 bits in the set index, the 24 lower is the object index.
+    indices: Vec<u32>,
+}
```
obmarg, line 48: What are all of these things?

engine/crates/engine-v2/src/response/read/view/de.rs
```diff
@@ -0,0 +1,175 @@
...
+            ResponseValue::Null => visitor.visit_none(),
+            ResponseValue::Boolean { value, .. } => visitor.visit_bool(*value),
+            ResponseValue::Int { value, .. } => visitor.visit_i32(*value),
+            ResponseValue::BigInt { value, .. } => visitor.visit_i64(*value),
+            ResponseValue::Float { value, .. } => visitor.visit_f64(*value),
+            ResponseValue::String { value, .. } => visitor.visit_borrowed_str(value),
+            ResponseValue::StringId { id, .. } => visitor.visit_borrowed_str(&self.schema[*id]),
+            ResponseValue::Json { value, .. } => value
+                .clone()
+                .deserialize_any(visitor)
```
obmarg, line 117: Can you not avoid this clone?

obmarg (approved):

engine/crates/engine-v2/src/execution/hooks/authorized.rs
```diff
@@ -32,6 +35,34 @@ impl<'ctx, H: Hooks> super::RequestHooks<'ctx, H> {
...
+                parents,
+                metadata,
+            )
+            // FIXME: Unfortunately, boxing seems to be the only solution for the bug explained here:
+            //        https://github.com/rust-lang/rust/issues/110338#issuecomment-1513761297
+            //        Otherwise is not correctly evaluated to be Send due to the impl IntoIterator
+            .boxed()
+            .await;
+
+        todo!()
```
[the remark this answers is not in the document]
Finistere: well not really, but not used yet

engine/crates/engine-v2/src/response/read/view/de.rs
Finistere: I could handle a few special cases like String/Number as JSON is used as fallback for any scalar we don't understand. Eventually, I would like to get rid of it by making `ResponseValue` a superset of it, but well. not there yet. Can't avoid the copy here otherwise

obmarg: Is this not a call to `Deserializer::deserialize_any` then?  Because it seems [like `&Value` implements that](https://docs.rs/serde/1.0.203/x86_64-unknown-linux-gnu/serde/de/trait.Deserializer.html#tymethod.deserialize_any)? Or am I missing something?

Finistere: wow ok you're right, good catch. Didn't work because it's `&Box<Value>` so had to do `as_ref()`

engine/crates/engine-v2/src/response/object_set.rs
Finistere: adding doc for this tomorrow

engine/crates/engine-v2/src/execution/hooks/authorized.rs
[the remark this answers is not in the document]
Finistere: added docstrings, will merge the PR after the release to avoid any conflicts I think

PR #1908, opened 2024-07-18
gateway: implement timeouts

tomhoule:
Additions to the configuration:

```toml
timeout = "10s" # defaults to "30s"

[subgraphs.products]
timeout = "5s" # defaults to "30s"
```

closes GB-7115


engine/crates/runtime-local/src/fetch.rs
```diff
@@ -10,28 +10,49 @@ use self::websockets::StreamingRequest;
 
 pub struct NativeFetcher {
     client: reqwest::Client,
+    /// Timeout configuration per subgraph
+    timeouts: HashMap<String, Duration>,
```
tomhoule, line 14: I am wondering: does this belong here, or one level down in the engine v2 config?

gateway/crates/gateway-binary/tests/integration_tests.rs
```diff
@@ -785,3 +785,207 @@ fn health_custom_listener() {
...
+          "errors": [
+            {
+              "message": "Subgraph reviews timed out",
+              "path": [
+                "topProducts",
+                0,
+                "reviews"
+              ],
+              "extensions": {
+                "code": "SUBGRAPH_REQUEST_ERROR"
```
tomhoule, line 975: Separate error code for subgraph timeouts?

engine/crates/runtime-local/src/fetch.rs
```diff
@@ -10,28 +10,49 @@ use self::websockets::StreamingRequest;
...
             .post(request.url.clone())
             .body(request.json_body)
             .header("Content-Type", "application/json")
-            .headers(request.headers.clone())
-            .send()
-            .await
-            .map_err(|e| FetchError::AnyError(e.to_string()))?;
+            .headers(request.headers.clone());
+
+        if let Some(timeout) = self.timeouts.get(subgraph_name) {
```
tomhoule, line 45: That fallback behaviour is wrong. There should be a default (from the config), not fall back to no timeout.

tomhoule: That would be solved if we moved timeouts inside the engine's subgraph configuration instead. Is that the right move?

engine/crates/runtime-local/src/fetch.rs
```diff
@@ -10,28 +10,49 @@ use self::websockets::StreamingRequest;
...
 
 impl NativeFetcher {
     pub fn runtime_fetcher() -> Fetcher {
         Fetcher::new(Self {
             client: reqwest::Client::new(),
+            timeouts: HashMap::new(),
+        })
+    }
+
+    pub fn runtime_fetcher_with_timeouts(timeouts: HashMap<String, Duration>) -> Fetcher {
```
Finistere, line 25: instead, you could have the `runtime::fetch::Fetcher` implement the retry & timeout logic on top of all fetcher implementation. So have something like

```
async fn post(&self, request: FetchRequest<'_>, retry_config: RetryConfig) -> FetchResult<FetchResponse> {
   ... do stuff
  self.post(request)
}
```

engine/crates/runtime-local/src/fetch.rs
Finistere: the timeout config should be in `engine_v2::Config` IMO and intuitively would pass it to the fetcher when requesting data. With defaultss being baked into the fetcher at creation.

This would be also easier to test with integration-tests and ensure this also works with cloudflare and other platforms.

tomhoule: done

engine/crates/runtime-local/src/fetch.rs
tomhoule: Good idea. Done and pushed.

engine/crates/runtime/src/fetch.rs
```diff
@@ -23,6 +23,8 @@ pub struct FetchRequest<'a> {
     pub url: &'a url::Url,
     pub headers: http::HeaderMap,
     pub json_body: String,
+    pub subgraph_name: &'a str,
```
Finistere, line 26: Wouldn't add the subgraph_name IMHO

engine/crates/engine-v2/Cargo.toml
```diff
@@ -38,6 +38,7 @@ thiserror.workspace = true
 tracing.workspace = true
 http.workspace = true
 headers.workspace = true
+tokio.workspace = true
```
Finistere, line 41: Cannot depend on tokio directly

Finistere (approved):

engine/crates/runtime/src/fetch.rs
tomhoule: Yeah we can handle that in the caller and avoid passing the subgraph name, if that feels better. Just need to add an error variant.

Finistere: nah just leave it as is for now, we'll see later 🤷

PR #1910, opened 2024-07-18
feat(engine-v2): Add on_subgraph_request hook

Finistere:
- Hooked up the `on_subgraph_request` hook.
- removed the `Result` wrapping of `headers.get` and `headers.delete`. 
- added `headers.entries` to list all headers


engine/crates/engine-v2/src/sources/graphql/request.rs
```diff
@@ -0,0 +1,65 @@
...
+use tracing::Span;
+
+use crate::{
+    engine::RateLimitContext,
+    execution::{ExecutionContext, ExecutionError, ExecutionResult},
+    response::ResponsePart,
+    Runtime,
+};
+
+pub(super) async fn execute_subgraph_request<'ctx, 'a, R: Runtime>(
```
pimeys, line 16: This function is really cool and beautiful.

engine/crates/integration-tests/tests/federation/hooks/on_subgraph_request.rs
```diff
@@ -0,0 +1,245 @@
...
+use integration_tests::{federation::EngineV2Ext, runtime};
+use runtime::{
+    error::{PartialErrorCode, PartialGraphqlError},
+    hooks::{DynHookContext, DynHooks},
+};
+use runtime_local::HooksWasiConfig;
+use url::Url;
+
+#[test]
+fn wasi() {
```
pimeys, line 13: We should leave a note for people to compile the hooks before running the tests. I have a hunch this might be annoying. If you just `cargo test` here this will fail if you don't compile the hooks examples first.

pimeys (approved):
super

engine/crates/integration-tests/tests/federation/hooks/on_subgraph_request.rs
Finistere: oh right, good catch. Was thinking of CI, not devs

PR #1917, opened 2024-07-19
grafbase-gateway: implement subgraph retries

tomhoule:
Using the Budget logic from tower.

closes GB-7116


cli/crates/federated-dev/src/dev/subgraph_config_watcher.rs
```diff
@@ -126,19 +126,21 @@ fn determine_changes<'a>(
     let deleted_names = previous_names.difference(&next_names).collect::<BTreeSet<_>>();
 
     let changed_subgraphs = {
-        let previous_subgraphs = previous_subgraphs
+        let previous_subgraph_names = previous_subgraphs
```
tomhoule, line 129: This change is here because the config now has a float in it, so it can't implement `Ord` and `Eq`.

engine/crates/engine-v2/src/engine.rs
```diff
@@ -70,6 +72,20 @@ impl<R: Runtime> Engine<R> {
...
 
+        let retry_budgets = schema
+            .iter_graphql_endpoints()
+            .map(|endpoint| {
+                let retry_config = endpoint.retry_config()?;
+
+                // Defaults: https://docs.rs/tower/0.4.13/src/tower/retry/budget.rs.html#137-139
+                let ttl = retry_config.ttl.unwrap_or(std::time::Duration::from_secs(10));
+                let min_per_second = retry_config.min_per_second.unwrap_or(10);
+                let retry_percent = retry_config.retry_percent.unwrap_or(0.2);
```
tomhoule, line 83: Since the config is `None` by default we don't retry at all, but if the retry section is present, we'll retry. I think we could add an `enabled` option for clarity.

engine/crates/engine-v2/src/engine.rs
```diff
@@ -122,7 +139,7 @@ impl<R: Runtime> Engine<R> {
             } else {
                 self.schema.settings.timeout
             };
-            self.runtime.sleep(timeout).await;
+            self.runtime.sleep(dbg!(timeout)).await;
```
tomhoule, line 142: stray dbg

engine/crates/engine-v2/src/engine.rs
```diff
@@ -57,6 +58,7 @@ pub struct Engine<R: Runtime> {
     pub(crate) runtime: R,
     operation_metrics: GraphqlOperationMetrics,
     auth: AuthService,
+    retry_budgets: Vec<Option<RetryBudget>>,
```
tomhoule, line 61: We need this here and not in config since it's state, and it can't be serialized.

engine/crates/graphql-mocks/src/state_mutation.rs
```diff
@@ -49,6 +49,18 @@ impl Query {
...
+    async fn increment_and_fail_if_less_than(&self, ctx: &Context<'_>, n: usize) -> FieldResult<usize> {
+        let state = ctx.data_unchecked::<Arc<AtomicUsize>>();
+        let new = state.fetch_add(1, Ordering::Relaxed);
+        if new < n {
+            // Trigger a 500
+            panic!("State value is {new} < {n}")
+        } else {
+            Ok(new)
+        }
+    }
```
tomhoule, line 63: Could be in a separate mock if that sounds better

engine/crates/integration-tests/src/federation/builder/bench.rs
```diff
@@ -152,7 +152,7 @@ impl DummyFetcher {
 
 #[async_trait::async_trait]
 impl runtime::fetch::FetcherInner for DummyFetcher {
-    async fn post(&self, _request: FetchRequest<'_>) -> FetchResult<FetchResponse> {
+    async fn post(&self, _request: &FetchRequest<'_>) -> FetchResult<FetchResponse> {
```
tomhoule, line 155: We have to take the request by reference for retries, otherwise we'd have to recreate it in the caller.

engine/crates/integration-tests/tests/federation/subgraph_retries.rs
```diff
@@ -0,0 +1,87 @@
...
+            "#,
+            )
+            .build()
+            .await;
+
+        let response = engine.execute("mutation { incrementAndFailIfLessThan(n: 5) }").await;
+
+        insta::assert_json_snapshot!(response, {
+            ".errors[0].message" => "REDACTED".to_owned(),
+        });
```
tomhoule, line 85: We don't need the redaction here, that could be inline.

engine/crates/runtime/src/fetch.rs
```diff
@@ -60,6 +61,39 @@ impl Fetcher {
...
+
+    #[allow(unused_variables)]
+    pub async fn post<SleepFut>(
+        &self,
+        request: &FetchRequest<'_>,
+        sleep: impl Fn(Duration) -> SleepFut,
+    ) -> FetchResult<FetchResponse>
+    where
+        SleepFut: std::future::Future<Output = ()>,
+    {
```
tomhoule, line 73: This is where the retries happen. We're missing backoff, preferably with jitter.

engine/crates/runtime/src/fetch.rs
```diff
@@ -60,6 +61,38 @@ impl Fetcher {
             inner: Arc::new(fetcher),
         }
     }
+
+    #[allow(unused_variables)]
+    pub async fn post<SleepFut>(
+        &self,
+        request: &FetchRequest<'_>,
+        sleep: impl Fn(Duration) -> SleepFut + Send + Sync,
```
tomhoule, line 69: This argument and the `#[allow(unused_variables)]` are remainders of playing with backoff, please remove if not necessary anymore.

engine/crates/runtime-local/src/fetch.rs
```diff
@@ -22,15 +22,16 @@ impl NativeFetcher {
...
         let subgraph_name = request.subgraph_name;
 
         let n = request.json_body.len();
+
         let response = self
             .client
             .post(request.url.clone())
-            .body(request.json_body)
-            .headers(request.headers)
+            .body(request.json_body.clone())
```
Finistere, line 33: Can't you pass the body as a `Bytes` instead?

engine/crates/runtime-local/src/rate_limiting/redis.rs
```diff
@@ -207,25 +207,21 @@ impl RedisRateLimiter {
...
                 let previous_count = previous_count.unwrap_or_default().min(config.limit as u64);
-                let current_count = current_count.min(config.limit as u64);
+                let current_count = current_count.unwrap_or_default().min(config.limit as u64);
 
                 // Sum is a percentage what is left from the previous window, and the count of the
                 // current window.
                 let average = previous_count as f64 * (1.0 - bucket_percentage) + current_count as f64;
 
-                if average <= config.limit as f64 {
+                if average < config.limit as f64 {
```
Finistere, line 222: It's not obvious that doing two requests is better than everything at once. Have you measured it? 
Cloudflare doesn't make *any* requests to their Memcache if the user is rate-limited. They keep track of those in-memory.

engine/crates/runtime/src/fetch.rs
```diff
@@ -60,6 +72,58 @@ impl Fetcher {
...
```
Finistere, line 80: `sleep` shouldn't be part of the trait, the implementation can do sleep by itself without any problem. It's aware of Cloudflare/Tokio. I's only the engine that can't rely on tokio. `runtime-local` can

engine/crates/engine-v2/schema/src/lib.rs
```diff
@@ -135,6 +135,11 @@ impl Schema {
             .map(|pos| FieldDefinitionId::from(usize::from(fields.start) + pos))
     }
 
+    pub fn iter_graphql_endpoints(&self) -> impl ExactSizeIterator<Item = sources::graphql::GraphqlEndpointWalker<'_>> {
```
Finistere, line 138: The method would have been better placed in the `SchemaWalker`:
https://github.com/grafbase/grafbase/blob/main/engine/crates/engine-v2/schema/src/walkers/mod.rs#L119

engine/crates/runtime/src/fetch.rs
```diff
@@ -60,6 +72,58 @@ impl Fetcher {
...
+                    if retry_budget.withdraw().is_ok() {
+                        let jitter = rand::random::<f64>() * 2.0;
+                        let exp_backoff = (100 * 2u64.pow(counter)) as f64;
+                        let backoff_ms = (exp_backoff * jitter).round() as u64;
+
+                        sleep(Duration::from_millis(backoff_ms)).await;
+
+                        counter += 1;
+
+                        let rate_limit = request
```
Finistere, line 114: nitpick, but IMHO `rate_limit` variable isn't a great name. It's more an `am_i_rate_limited`. So at that point you could just do:
```rust
request..rate_limiter.limit(&request.subgraph_name).await?
self.inner.post(request).await
```
would be clearer imho

engine/crates/runtime/src/fetch.rs
```diff
@@ -60,6 +72,58 @@ impl Fetcher {
...
+                            .rate_limiter
+                            .limit(&request.subgraph_name)
```
Finistere, line 116: Just provide the `RateLimitContext` to `post`, There might be other stuff in it and `impl RateLimitContext for &str` is very ugly.

pimeys: This was surprisingly hard. The graphql/federation.rs went totally crazy with lifetime errors by passing a RateLimitContext in the FetchRequest.

engine/crates/runtime/src/fetch.rs
pimeys: but we need to get that one result from both futures.

engine/crates/runtime/src/fetch.rs
pimeys: Yeah but the implementation of retries needs to move from runtime to runtime local.

engine/crates/runtime/src/fetch.rs
Finistere: what do you mean? the rate limiter raises an error or not, you're using the ok value.

engine/crates/runtime/src/fetch.rs
Finistere: What was hard? the `subgraph_name` is a `&'ctx str` which is guaranteed to exist during the future lifetime, so seems like you just need to pass a `Box<dyn RateLimiterContext>` to the future

engine/crates/runtime/src/fetch.rs
Finistere: ah it's loop, ok nevermind

engine/crates/runtime/src/fetch.rs
Finistere: Honestly I would move all of this to `engine-v2`, would make more sense as you rely on retry budget, rate limiter and sleep

Finistere: will remove `subgraph_name` from the `FetchRequest` also, shouldn't be there, it was an error. `FetchError` should just have a timeout case

engine/crates/runtime-local/src/rate_limiting/redis.rs
pimeys: We cannot increment this value in a hot loop in retries. And if we don't increment, the <= check would lead to a wrong result. Now if we do it like this, we load the values first, retry as many times as needed and _when_ we have a successful result, we increment the value asynchronously in a task.

Finistere (approved):

engine/crates/parser-sdl/src/rules/subgraph_directive.rs
```diff
@@ -73,13 +99,28 @@ impl Directive for SubgraphDirective {
           Timeout for requests to that subgraph
           """
           entityCacheTtl: String
+
+          Retry configuration for that subgraph
```
obmarg, line 103:
```suggestion

          """
          Retry configuration for that subgraph
```

PR #1922, opened 2024-07-20
fix(engine-v2): Do not forward headers that impact our subgraph request processing

Finistere:
Do not forward hop-by-hop headers and other headers like `Accept` and `Content-Length` which would impact the subgraph request


engine/crates/engine-v2/src/execution/header_rule.rs
```diff
@@ -0,0 +1,123 @@
...
+    header::ACCEPT_CHARSET,
+    header::ACCEPT_LANGUAGE,
+    header::ACCEPT_RANGES,
+    header::CONNECTION,
+    header::HeaderName::from_static("keep-alive"),
+    header::CONTENT_LENGTH,
+    header::CONTENT_TYPE,
+    header::HOST,
+    header::USER_AGENT,
+];
```
obmarg, line 21: There's a few other hop-by-hop headers in the spec that we should probably omit: https://datatracker.ietf.org/doc/html/rfc2616#section-13.5.1

Might also dispute user-agent.  Can see reasons to not include it, but could also see reasons to include it?  I'm not sure...

engine/crates/engine-v2/src/execution/header_rule.rs
```diff
@@ -0,0 +1,123 @@
...
+use schema::{HeaderRuleWalker, NameOrPatternRef};
+
+use crate::engine::RequestContext;
+
+static FORWARD_BLACKLIST: [header::HeaderName; 12] = [
+    header::ACCEPT,
+    header::ACCEPT_CHARSET,
+    header::ACCEPT_ENCODING,
+    header::ACCEPT_CHARSET,
+    header::ACCEPT_LANGUAGE,
```
obmarg, line 13: I can see why we'd want to get rid of most of the accepts, but language doesn't seem like something we'd care about?  Ofc not sure who would use it, just a thought

Finistere: hum you're right, ended up just copy pasting all ACCEPT

engine/crates/engine-v2/src/execution/header_rule.rs
Finistere: they can always forward it with a different name, so sounded fair to me 🤷 
Not critical, but felt more "sincere" for a service to use a proper user agent

Finistere: good catch on the hop-by-hop headers!

engine/crates/engine-v2/src/execution/header_rule.rs
```diff
@@ -0,0 +1,129 @@
...
+use http::header;
+use schema::{HeaderRuleWalker, NameOrPatternRef};
+
+use crate::engine::RequestContext;
+
+static FORWARD_BLACKLIST: [header::HeaderName; 17] = [
+    header::ACCEPT,
+    header::ACCEPT_CHARSET,
+    header::ACCEPT_ENCODING,
+    header::ACCEPT_CHARSET,
```
obmarg, line 12: You’ve got charset twice btw

engine/crates/engine-v2/src/execution/header_rule.rs
```diff
@@ -0,0 +1,129 @@
...
+) -> http::HeaderMap {
+    let mut headers = default;
+    for header in rules {
+        match header.rule() {
+            schema::HeaderRuleRef::Forward { name, default, rename } => match name {
+                NameOrPatternRef::Pattern(regex) => {
+                    let filtered = request_context
+                        .headers
+                        .iter()
+                        .filter(|(name, _)| !FORWARD_BLACKLIST.contains(name))
```
obmarg, line 42: Might be better using a set or sorting & doing a binary search?

engine/crates/engine-v2/src/execution/header_rule.rs
```diff
@@ -0,0 +1,143 @@
...
+                }
+            }
+        }
+    }
+
+    headers
+}
+
+fn is_blacklisted_header(name: &HeaderName) -> bool {
+    static BLACKLIST: OnceLock<[&str; 14]> = OnceLock::new();
```
fbjork, line 120: Nice. Also add an issue to document which headers are not forwarded.

engine/crates/engine-v2/src/execution/header_rule.rs
```diff
@@ -0,0 +1,143 @@
...
+                    headers.insert(rename, value.into_owned());
+                }
+            }
+        }
+    }
+
+    headers
+}
+
+fn is_blacklisted_header(name: &HeaderName) -> bool {
```
fbjork, line 119: better to use the term denylist

pimeys (approved):

engine/crates/integration-tests/tests/federation/basic/headers.rs
```diff
@@ -141,38 +139,112 @@ fn test_subgraph_specific_header_forwarding() {
...
+
+        let response = engine
+            .execute("query { headers { name value }}")
+            .header("User-Agent", "Rusty")
+            .header("Accept", "application/json")
+            .header("Accept-Encoding", "gzip")
+            .header("Accept-Charset", "utf-8")
+            .header("Accept-Ranges", "bytes")
+            .header("Content-Length", "728")
+            .header("Content-Type", "application/jpeg")
```
obmarg, line 204: Kind of seems like a bug that we respond to requests with `Content-Type: application/jpeg` tbh.  Possibly the same for some of these other ones.

obmarg: (Not saying we should fix it right now, just an observation)

Finistere: what do you mean? the response shows the headers the subgraph received, so `application/json`

Finistere: ahhh ok got what you mean, `Content-Type` for the request is handled by `axum` since it deserializes the JSON.

Finistere: so yeah wouldn't be able to test that we don't forward this header with the gateway

obmarg: Oh right of course - I forgot we're skipping the whole HTTP layer in these tests.  Cool, never mind then.

PR #1928, opened 2024-07-22
fix(engine-v2): forward header rule should not duplicate

pimeys:
A test case:

```toml
[[headers]]
rule = "forward"
pattern = "^.*$"

[[headers]]
rule = "remove"
pattern = "^x-amzn-mtls-.*"

[[headers]]
rule = "forward"
name = "x-client-auth-ctx"
rename = "x-graph-client-auth-ctx"
```

This would first forward all headers, then forward again `x-client-auth-ctx` with a new name. Not good.

What we do here is that for every forward rule, we _first_ remove the header from the header map and right after that we add the forwarded header.


obmarg (approved):

jakubadamw (approved):
Great!

PR #1929, opened 2024-07-23
chore(engine-v2): Create response modifier executor with proper dependencies

Finistere:
So adjusting the `ResponseModifier` creation to be the same as the `QueryModiifer`, used a different data layout initially, but was a mistake. So a modifier is roughly just a rule (which `@authorized` directive basically) and a list of impacted fields.

Now I'm defining whenever a SelectionSet object needs to be tracked in the "logical planning" phase, the "shape" phase then create an appropriate `ResponseObjectSetId` whenever necessary. Had to do this because now a ResponseModifier points to its impacted fields (where we need to track objects), rather than the opposite.

In the execution planning phase, I'm building `ResponseModifierExecutor` which will do the actual modification if the field is still queried after the query modifications. This is also where I define all the dependencies between execution plans and response modifiers. I define all of those by relying on the input & output fields of each of them.

### DIFF
--- a/engine/crates/engine-v2/src/execution/planner/builder.rs
+++ b/engine/crates/engine-v2/src/execution/planner/builder.rs
@@ -1,49 +1,51 @@
-use std::{borrow::Cow, collections::HashMap};
+use std::{
+    borrow::{Borrow, Cow},
+    collections::HashMap,
+};
 
+use itertools::Itertools;
 use schema::{RequiredFieldSet, ResolverWalker};
 
 use crate::{
-    execution::{ExecutableOperation, ExecutionPlan, ExecutionPlanId, PlanWalker, PreExecutionContext},
-    operation::{FieldId, LogicalPlanId, OperationWalker, SelectionSetId},
-    response::{ResponseViewSelection, ResponseViewSelectionSet, ResponseViews},
+    execution::{
+        ExecutableOperation, ExecutionPlan, ExecutionPlanId, PlanWalker, PreExecutionContext, ResponseModifierExecutor,
+    },
+    operation::{FieldId, LogicalPlanId, OperationWalker, ResponseModifierRule, SelectionSetId, SelectionSetType},
+    response::{ResponseObjectSetId, ResponseViewSelection, ResponseViewSelectionSet},
     sources::PreparedExecutor,
-    utils::BufferPool,
     Runtime,
 };
 
-use super::PlanningResult;
+use super::{BuildContext, PlanningResult};
 
-pub(super) struct RequirementsBuildContext<'ctx, 'op, 'parent, R: Runtime> {
-    ctx: &'op PreExecutionContext<'ctx, R>,
-    operation: &'op ExecutableOperation,
-    response_views: &'parent mut ResponseViews,
-    response_view_selection_buffer_pool: &'parent mut BufferPool<ResponseViewSelection>,
+pub(super) struct ExecutionBuilder<'ctx, 'op, R: Runtime> {
+    pub(super) ctx: &'op PreExecutionContext<'ctx, R>,
+    pub(super) operation: &'op ExecutableOperation,
+    pub(super) build_context: &'op mut BuildContext,
 }
 
-impl<'ctx, 'op, 'parent, R: Runtime> RequirementsBuildContext<'ctx, 'op, 'parent, R>
+impl<'ctx, 'op, R: Runtime> std::ops::Deref for ExecutionBuilder<'ctx, 'op, R> {
+    type Target = BuildContext;
+    fn deref(&self) -> &Self::Target {
+        self.build_context
+    }
+}
+
+impl<'ctx, 'op, R: Runtime> std::ops::DerefMut for ExecutionBuilder<'ctx, 'op, R> {
+    fn deref_mut(&mut self) -> &mut Self::Target {
+        self.build_context
+    }
+}
+
+impl<'ctx, 'op, R: Runtime> ExecutionBuilder<'ctx, 'op, R>
 where
     'ctx: 'op,
-    'op: 'parent,
 {
-    pub(super) fn new(
-        ctx: &'op PreExecutionContext<'ctx, R>,
-        operation: &'op ExecutableOperation,
-        response_views: &'parent mut ResponseViews,
-        response_view_selection_buffer_pool: &'parent mut BufferPool<ResponseViewSelection>,
-    ) -> Self {
-        Self {
-            ctx,
-            operation,
-            response_views,
-            response_view_selection_buffer_pool,
-        }
-    }
-
-    pub(super) fn build_execution_plan(
+    pub(super) fn insert_execution_plan(
         mut self,
         execution_plan_id: ExecutionPlanId,
         logical_plan_id: LogicalPlanId,
-    ) -> PlanningResult<(ExecutionPlan, Vec<FieldId>)> {
+    ) -> PlanningResult<()> {
         let logical_plan = &self.operation[logical_plan_id];
         let resolver = self
             .ctx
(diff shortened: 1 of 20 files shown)


engine/crates/engine-v2/src/execution/planner/mod.rs
```diff
@@ -1,131 +1,212 @@
...
+                {
+                    self.operation[parent_id].dependent_response_modifiers.push(child_id);
+                    self.operation[child_id].parent_count += 1;
+                }
+            }
+        }
 
-        operation.response_views = response_views;
+        // All execution plans that must be executed *after* a response modifier
+        for (i, output_fields) in take(&mut self.response_modifier_executors_output_fields)
```
Finistere, line 155: the logic here isn't really accurate, fixed it in later PR

engine/crates/engine-v2/src/execution/planner/mod.rs
```diff
@@ -1,131 +1,217 @@
...
     response::{ResponseViewSelection, ResponseViews},
     sources::PreparedExecutor,
     utils::BufferPool,
     Runtime,
 };
 
-use super::{header_rule::create_subgraph_headers_with_rules, ExecutableOperation};
+use super::{header_rule::create_subgraph_headers_with_rules, ExecutableOperation, ResponseModifierExecutor};
 
 pub(super) struct ExecutionPlanner<'ctx, 'op, R: Runtime> {
```
obmarg, line 21: Does this even need to be public?  Looks like it's only used inside the plan function

engine/crates/engine-v2/src/execution/planner/mod.rs
```diff
@@ -1,131 +1,217 @@
...
     ctx: &'op PreExecutionContext<'ctx, R>,
     operation: ExecutableOperation,
+    build_context: BuildContext,
+}
+
```
obmarg, line 26: What's the point/difference between this and the ExecutionBuilder I just finished reading?  I can only see 3 functions on this thing - could they just be free standing functions that take 1-3 parameters?  There's a ton of ambiguous structs in this code, it'd be nice if we could ditch some of them...

engine/crates/engine-v2/src/execution/planner/mod.rs
```diff
@@ -1,131 +1,217 @@
...
+    let operation = ExecutionPlanner {
+        ctx,
+        build_context: BuildContext {
+            logical_plan_to_execution_plan_id: vec![None; operation.plan.logical_plans.len()],
+            io_fields: Vec::with_capacity(operation.fields.len()),
+            ..Default::default()
+        },
+        operation,
+    }
+    .build()?;
```
obmarg, line 103: `ExecutionPlanner::build` returns an operation?  I'd kind of assume that it returns a builder for or an instance of ExecutionPlanner.  Might be better named plan?  Although as I said above - if we could just get rid of this that'd be even better.

engine/crates/engine-v2/src/execution/planner/mod.rs
```diff
@@ -1,131 +1,217 @@
...
         if let Some(mut prev) = mutation_fields_plan_order.next() {
             for next in mutation_fields_plan_order {
-                if !operation[prev].children.contains(&next) {
-                    operation[prev].children.push(next);
-                    operation[next].parent_count += 1;
+                if !self.operation[prev].children.contains(&next) {
+                    self.operation[prev].children.push(next);
+                    self.operation[next].parent_count += 1;
                 }
                 prev = next;
```
obmarg, line 207: This kind of looks like [tuple_windows](https://docs.rs/itertools/latest/itertools/trait.Itertools.html#method.tuple_windows), might be easier to just use that (if i'm reading right)

obmarg (approved):

engine/crates/engine-v2/src/execution/planner/builder.rs
```diff
@@ -1,63 +1,65 @@
...
 
-pub(super) struct RequirementsBuildContext<'ctx, 'op, 'parent, R: Runtime> {
-    ctx: &'op PreExecutionContext<'ctx, R>,
-    operation: &'op ExecutableOperation,
-    response_views: &'parent mut ResponseViews,
-    response_view_selection_buffer_pool: &'parent mut BufferPool<ResponseViewSelection>,
+impl<'ctx, 'op, R: Runtime> std::ops::DerefMut for ExecutionBuilder<'ctx, 'op, R> {
+    fn deref_mut(&mut self) -> &mut Self::Target {
+        self.build_context
+    }
```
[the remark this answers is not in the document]
Finistere: Well have to admit I'm probably abusing this deref pattern and that's not great. I'm not entirely sure how to solve it though. I often end up with basically 3 levels of context:
- `ctx`: PreExecutionContext / ExecutionContext which holds the engine, schema typically
- A root struct for the current phase: binding, solving requirements, computing shapes, building executors/execution plans
- A nested struct with additional local context: the current plan being built, the current selection set being solved, etc..

I'm always using `Deref` so that the 3rd level has access to the 2nd one and I add utility methods for schema/walker at the 2nd one usually. Without `Deref` I would need to add an extra field for like 3/4 of the `self.` usage.

But maybe I should replace the 2nd level with a Context, have its function be free-standing functions and pass that around in the 3rd level with local context

engine/crates/engine-v2/src/execution/planner/mod.rs
Finistere: The `ResponseModifierExecutor` could be `pub(in crate::execution)` currently it's `pub(crate)`

engine/crates/engine-v2/src/execution/planner/builder.rs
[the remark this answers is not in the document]
obmarg: > Without `Deref` I would need to add an extra field for like 3/4 of the `self.` usage.

Yeah, I am not sure either - one thing that occurred to me is literally just doing `let context = &self.build_context;` at the start of the function (or similarly `let Self { ctx, build_context } = &self`).  Although I don't know if lifetimes/mutability restrictions or whatever are going to fuck that up. 

> But maybe I should replace the 2nd level with a Context, have its function be free-standing functions and pass that around in the 3rd level

Yeah a few extra parameters (potentially on free standing functions) are another possibility.

Without actually trying things I don't really know the solution (and also don't think it _needs_ to be solved right now) - just wanted to raise it as a concern.  Deal with it as you like (or don't)

engine/crates/engine-v2/src/execution/planner/mod.rs
obmarg: I'm talking about ExecutionPlanner - it's `pub(super)` but appears to be both constructed and dropped inside this module.  Seemed like an implementation detail of this module.  Unless I'm missing something

Finistere: ah right, leftover you're right

engine/crates/engine-v2/src/execution/planner/builder.rs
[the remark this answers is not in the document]
Finistere: I think you're right and needs to be fixed, what I'm doing is akin to the cake pattern in Scala which is an anti-pattern. And here it's particularly bad with the mutability issue of the operation leading to even more complexity. Since I'm re-using this same anti-pattern all the time, need to figure out how to structure it properly without it. I keep coming back to it being the lowest-effort solution for me right now. I'll do it next week after field analytics

engine/crates/engine-v2/src/execution/planner/mod.rs
Finistere: Well yeah, my problem is there I currently need to modify the `ExecutableOperation` at some point, and then I'm relying on it being immutable and I'm alternating those phases... So that's something I want to fix, and seems like I fixed the core issue that I had at some point. Those could be free-standing functions. I will take a stab on this next week

PR #1931, opened 2024-07-24
feat: entity caching for top level fields

obmarg:
This adds entity caching for requests to top level fields.  The name entity caching isn't _really_ accurate for these, since they're not neccesarily entities, but whatever.

The implementation is very barebones at the moment: we just hash the query, and look that up in the cache.  We don't yet bother with auth scopes, cache invalidation, or even much of the configuration concerns for this feature (beyond what I needed for tests).

Fixes GB-7148

### DIFF
--- a/engine/crates/engine-v2/src/sources/graphql/mod.rs
+++ b/engine/crates/engine-v2/src/sources/graphql/mod.rs
@@ -1,8 +1,13 @@
-use grafbase_telemetry::span::subgraph::SubgraphRequestSpan;
-use request::execute_subgraph_request;
+use std::{borrow::Cow, time::Duration};
+
+use bytes::Bytes;
+use grafbase_telemetry::{gql_response_status::GraphqlResponseStatus, span::subgraph::SubgraphRequestSpan};
+use hex::ToHex;
+use request::{execute_subgraph_request, ResponseIngester};
 use runtime::fetch::FetchRequest;
 use schema::sources::graphql::{GraphqlEndpointId, RootFieldResolverWalker};
 use serde::de::DeserializeSeed;
+use sha2::Digest;
 use tracing::Instrument;
 
 use self::query::PreparedGraphqlOperation;
@@ -59,6 +64,7 @@ impl GraphqlPreparedExecutor {
             variables: &self.operation.variables,
             inputs: Vec::new(),
         };
+
         tracing::debug!(
             "Query {}\n{}\n{}",
             subgraph.name(),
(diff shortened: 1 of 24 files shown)


engine/crates/engine-v2/src/sources/graphql/mod.rs
```diff
@@ -92,20 +133,65 @@ impl GraphqlPreparedExecutor {
...
+impl<'ctx, R> ResponseIngester for GraphqlIngester<'ctx, R>
+where
+    R: Runtime,
+{
+    async fn ingest(
+        mut self,
+        bytes: Bytes,
+    ) -> Result<(GraphqlResponseStatus, SubgraphResponse), crate::execution::ExecutionError> {
+        if let Some(cache_key) = self.cache_key {
+            // We could probably put this call into the background at some point, but for
```
Finistere, line 173: hum indeed, there is a FutureUnordered executing background tasks here
https://github.com/grafbase/grafbase/blob/main/engine/crates/engine-v2/src/execution/coordinator.rs#L44

but no way to write in it, I guess we could have a MPSC channel with the sender inside the `ExecutionContext`.

engine/crates/runtime/src/kv.rs
```diff
@@ -60,5 +60,5 @@ impl std::ops::Deref for KvStore {
 #[async_trait::async_trait]
 pub trait KvStoreInner: Send + Sync {
     async fn get(&self, name: &str, cache_ttl: Option<Duration>) -> KvResult<Option<Vec<u8>>>;
-    async fn put(&self, name: &str, bytes: Vec<u8>, expiration_ttl: Option<Duration>) -> KvResult<()>;
+    async fn put(&self, name: &str, bytes: &[u8], expiration_ttl: Option<Duration>) -> KvResult<()>;
```
Finistere, line 63: `Cow<'_, [u8]>` would probably be the most flexible.

it's not great to use KV for caching. In Cloudflare KV behavior is different from cache. KV is more meant as a distributed KV than pure cache, so it'll try to fetch data from single colocation. Hence the `cache_ttl` in the `get()` to cache the result inside the current colocation. Entity cache shouldn't rely on that

Finistere (approved):

engine/crates/engine-v2/src/sources/graphql/mod.rs
obmarg: Yeah, was going to add a channel when/if we do this (not necessarily to that FutureUnordered, but maybe).  But out of scope for now

Finistere: yeah just wanted to mention that this existed for pretty much the same purpose, just at a different step of the execution. so could likely be re-used.

engine/crates/runtime/src/kv.rs
obmarg: > Cow<'_, [u8]> would probably be the most flexible.

I guess, though I'm struggling to see a situation where it'd be useful?  An owned thing can always be made into a slice and a slice doesn't need converted?  Although let me know if I'm missing some advantage.

> it's not great to use KV for caching. In Cloudflare KV behavior is different from cache. KV is more meant as a distributed KV than pure cache, so it'll try to fetch data from single colocation. Hence the cache_ttl in the get() to cache the result inside the current colocation. Entity cache shouldn't rely on that

Raised #1938 to make this a bit more obvious.  I'll do something to fix this in a future PR: likely pulling in some kind of cache impl

Finistere: Ok flexible wasn't the right word for `Cow`, wanted to say it can avoid copies

PR #1933, opened 2024-07-24
feat(engine-v2): authorized parent edge post execution

Finistere:
So this applies the response modifiers, once all the requirements are present, and adds some basic tests for them for the case:
```graphql
type User {
   name @authorized(fields: "id")
}
```

### DIFF
--- a/engine/crates/engine-v2/src/execution/mod.rs
+++ b/engine/crates/engine-v2/src/execution/mod.rs
@@ -5,6 +5,7 @@ mod header_rule;
 pub(crate) mod hooks;
 mod ids;
 mod planner;
+mod response_modifier;
 mod state;
 mod walkers;
 
@@ -89,10 +90,10 @@ pub(crate) struct QueryModifications {
 }
 
 // Modifies the response based on a given rule
-#[allow(unused)]
 pub(crate) struct ResponseModifierExecutor {
     pub rule: ResponseModifierRule,
     /// Which object & fields are impacted
+    /// ordered
     pub on: Vec<(ResponseObjectSetId, Option<EntityId>, ResponseKey)>,
     /// What fields the hook requires
     pub requires: ResponseViewSelectionSet,
--- a/engine/crates/engine-v2/src/execution/planner/mod.rs
+++ b/engine/crates/engine-v2/src/execution/planner/mod.rs
@@ -163,12 +163,14 @@ where
         {
             let parent_id = ResponseModifierExecutorId::from(i);
             for &field_id in &self.build_context[output_fields] {
-                let child_logical_plan_id = self.operation.plan.field_to_logical_plan_id[usize::from(field_id)];
-                let child_id = self.logical_plan_to_execution_plan_id[usize::from(child_logical_plan_id)]
-                    .expect("Depend on unfinished plan");
-                if !self.operation[parent_id].children.contains(&child_id) {
-                    self.operation[parent_id].children.push(child_id);
-                    self.operation[child_id].parent_count += 1;
+                for (i, input_fields) in self.build_context.execution_plans_input_fields.iter().enumerate() {
+                    let child_id = ExecutionPlanId::from(i);
+                    if self.build_context[*input_fields].contains(&field_id)
+                        && !self.operation[parent_id].children.contains(&child_id)
+                    {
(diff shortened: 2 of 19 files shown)


engine/crates/engine-v2/src/execution/mod.rs
```diff
@@ -89,10 +90,10 @@ pub(crate) struct QueryModifications {
 }
 
 // Modifies the response based on a given rule
-#[allow(unused)]
 pub(crate) struct ResponseModifierExecutor {
     pub rule: ResponseModifierRule,
     /// Which object & fields are impacted
+    /// ordered
```
obmarg, line 96: ordered by what?

engine/crates/engine-v2/src/execution/planner/mod.rs
```diff
@@ -163,12 +163,14 @@ where
...
-                    self.operation[parent_id].children.push(child_id);
-                    self.operation[child_id].parent_count += 1;
+                for (i, input_fields) in self.build_context.execution_plans_input_fields.iter().enumerate() {
+                    let child_id = ExecutionPlanId::from(i);
+                    if self.build_context[*input_fields].contains(&field_id)
+                        && !self.operation[parent_id].children.contains(&child_id)
+                    {
+                        self.operation[parent_id].children.push(child_id);
+                        self.operation[child_id].parent_count += 1;
+                    }
```
obmarg, line 173: Did you consider just putting this change in 1929 rather than putting it in this one and leaving a comment in 1929?  

Not that big a deal, but would have been nice to not just skip over it in the last one and then try to remember the context to review it here

engine/crates/engine-v2/src/execution/response_modifier.rs
```diff
@@ -0,0 +1,109 @@
...
+use super::{state::OperationExecutionState, ExecutionContext, ResponseModifierExecutorId};
+
+impl<'ctx, R: Runtime> ExecutionContext<'ctx, R> {
+    pub(super) async fn execute_response_modifier(
+        &self,
+        state: &mut OperationExecutionState<'ctx>,
+        response: &mut ResponseBuilder,
+        response_modifier_executor_id: ResponseModifierExecutorId,
+    ) {
+        let executor = &self.operation[response_modifier_executor_id];
```
obmarg, line 20: This is very confusing.  Why is a thing named context doing the execution, and the thing named executor not doing the one thing that it's name says it does.

I'm not even sure what to suggest here - is the executor named wrong?  Would this make more sense as a function on the executor, or just a freestanding function that _takes_ a context?

obmarg (approved):

engine/crates/engine-v2/src/execution/response_modifier.rs
Finistere: Oh, good catch. Yeah, this is more of a freestanding function. You're right that it should be an executor method for clarity, I mostly started with it not entirely how it would end up and left it as is

engine/crates/engine-v2/src/execution/mod.rs
Finistere: should have said `sorted`, it's just sorted

engine/crates/engine-v2/src/execution/planner/mod.rs
Finistere: yeah sorry about that

engine/crates/engine-v2/src/execution/mod.rs
obmarg: By what though?  Or just the "natural" order that a tuple of ints has?

Finistere: yeah just the natural order

Finistere: just `sort_unstable()`

obmarg: Maybe worth clarifying that in the comment?  It wasn't immediately obvious for some reason (guess maybe it should have been, but I dunno)

Finistere: `sorted by natural order`?

obmarg: works for me

PR #1938, opened 2024-07-25
fix: the runtime-local implementation of KvStoreInner

obmarg:
In #1931 it was pointed out that I was passing a None ttl to KvStoreInner::get, which on CF will cause the KV store to always look up in a central location instead of in a local datastore cache.

I did not realise that this was what the `cache_ttl` function was for, because I based my understanding on the runtime-local implementation, which used this parameter to decide whether an entry had expired.  although even that was a bit weird - it added the cache_ttl to the orignially requested expiry time, and ignored the original expiry time if you didn't provide a cache_ttl

This PR fixes the KV impl to be closer to what the parameters actually mean, and adds some docstrings.  So nobody else falls into this trap in the future.

### DIFF
--- a/engine/crates/runtime-local/src/kv.rs
+++ b/engine/crates/runtime-local/src/kv.rs
@@ -1,12 +1,17 @@
 use runtime::kv::{KvResult, KvStore, KvStoreInner};
 use std::{
-    collections::HashMap,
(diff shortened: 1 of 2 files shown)


jakubadamw (approved):
Nice!

PR #1941, opened 2024-07-26
chore: refine redis keys a bit nicer

pimeys:
If subgraph has a name "global" we would be in trouble.


Finistere (approved):
The `RateLimitContext` makes even less sense now IMHO. The actual Redis key looks good though.

PR #1942, opened 2024-07-26
chore: replace index! with proc macro

obmarg:
`engine-v2` makes extensive use of [a few proc macros][1] to define `Id` newtypes and to simplify adding `Index` impls that use those newtypes.  This was quite a good a idea, but as they've needed to support more features and more structs with lifetimes, generic parameters, where bounds etc. it's become harder to understand the definition of these macros ([this line in particular][2])

This PR contains a prototype of some derive based alternatives: 

1. An `IndexImpls` derive for generating the `Index` impl.  This goes on the struct being indexed rather than being associated with the Id.  I think this makes a bit more sense, as you can now tell from the struct definition what it's various entries are indexed by.
2. An `Id` derive that can go on any newtype struct that contains a `NonZero<T>`.  This makes defining the individual Ids a bit more verbose, but means there's a bit more freedom to add arbitrary derives without having to add them to all the Ids.  Maybe worth the tradeoff?  I'm open to disagreement here though, wouldn't be against going back to macro_rules for these.

I haven't finished changing everything over to use this, but have done enough to get a feel for it.  Can do the rest if people think this is a good idea.

[1]: https://github.com/grafbase/grafbase/blob/1b6f688/engine/crates/engine-v2/id-newtypes/src/lib.rs#L4-L180
[2]: https://github.com/grafbase/grafbase/blob/1b6f688/engine/crates/engine-v2/id-newtypes/src/lib.rs#L39


engine/crates/engine-v2/id-derives/src/id.rs
```diff
@@ -0,0 +1,98 @@
...
+
+    let name = proc_macro2::Literal::string(&ident.to_string());
+
+    Some(quote!(
+        assert!(value <= #max, "{} id {} exceeds maximum {}", #name, value, stringify!(#max));
+    ))
+}
+
+/// Finds the u8 in NonZero<u8>
+fn find_non_zero_kind(fields: &syn::FieldsUnnamed) -> Option<&syn::Type> {
```
Finistere, line 84: Does that still work with `pub struct SomeId(u8)`?

engine/crates/engine-v2/id-derives/src/lib.rs
```diff
@@ -0,0 +1,12 @@
+mod id;
+mod indexes;
+
+#[proc_macro_derive(Id, attributes(max))]
+pub fn derive_id(input: proc_macro::TokenStream) -> proc_macro::TokenStream {
+    id::derive_id(input)
+}
+
+#[proc_macro_derive(IndexImpls, attributes(indexed_by))]
```
Finistere, line 9: maybe `IndexedFields`?

engine/crates/engine-v2/id-derives/Cargo.toml
```diff
@@ -0,0 +1,12 @@
+[package]
```
Finistere, line 1: Don't think I would have created a new crate for it, but 🤷

Finistere (approved):

Finistere:
nice work!

engine/crates/engine-v2/id-derives/Cargo.toml
obmarg: I don't think you would have had a choice, proc macros live in proc macro crates.

Finistere: wow ok interesting, thanks for pointing it out

engine/crates/engine-v2/id-derives/src/lib.rs
obmarg: Tbh the more I dug into how your `proxy` stuff worked (which this was intended to replace) the less I thought it was a good idea...  Far too magic

So not sure this is either

engine/crates/engine-v2/id-derives/Cargo.toml
obmarg: It's because they have to be compiled into a shared library and then the rust compiler does some dynamic linking at runtime to actually pull the code in.  

Way simpler to have them isolated in their own crates than have users (and the compiler) deal with the burden of figuring out which bits are macro-time code and which bits are runtime code.

engine/crates/engine-v2/id-derives/src/id.rs
obmarg: Don't think so - I was kind of assuming that everything would be `NonZero<T>`.  Is that not the case currently?

Finistere: well, no, but doesn't matter today. the only use case I see needs to be changed anyway

engine/crates/engine-v2/src/operation/mod.rs
```diff
@@ -42,6 +42,16 @@ impl std::ops::Deref for PreparedOperation {
...
+impl std::ops::Index<LogicalPlanId> for PreparedOperation {
+    type Output = LogicalPlan;
+
+    fn index(&self, index: LogicalPlanId) -> &Self::Output {
+        &self.plan[index]
+    }
+}
+
+// Is this even a good idea?
+// It's very difficult to actually tell where data lives when we're going through three shadow Index impls....
```
obmarg, line 54: Oops, personal note

```suggestion
```

PR #1948, opened 2024-07-29
fix(engine-v2): Panic could happen with both pre & post execution auth

Finistere:
The post-execution mechanism, `ResponseModifierExecutor`, was expecting
its input, a `ResponseObjectSet`, to exist once its parent plans were
executed. However, with pre-execution auth we modify the subgraph query
to not retrieve any unauthorized fields. This could lead to the input
never being created. Added a test for it.

I'm also changing a bit how I'm testing type system directives. Before I
was adding them manually in the "secure" graphql-mock, but it's already
a mess today. So instead I created a simple TeaShop schema and I'm
overriding the SDL to add directives for a given test. Making it a lot
easier to follow and more reusable. Will try to move the other hook
stuff to this model over time.

### DIFF
--- a/engine/crates/graphql-mocks/src/tea_shop.rs
+++ b/engine/crates/graphql-mocks/src/tea_shop.rs
@@ -1,0 +1,218 @@
+use async_graphql::{Context, EmptyMutation, EmptySubscription, Object, SimpleObject, Union};
+
+use crate::Schema;
+
+/// A schema that only uses String types.
+///
+/// This is used to make sure that we're not pruning built in scalars that aren't used
+#[derive(Default)]
+pub struct TeaShop {
+    sdl: Option<String>,
+}
+
+impl TeaShop {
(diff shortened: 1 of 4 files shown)


engine/crates/graphql-mocks/src/tea_shop.rs
```diff
@@ -0,0 +1,218 @@
+use async_graphql::{Context, EmptyMutation, EmptySubscription, Object, SimpleObject, Union};
+
+use crate::Schema;
+
+/// A schema that only uses String types.
```
obmarg, line 5: I see at least one usize so this comment seems slightly inaccurate

obmarg (approved):

engine/crates/graphql-mocks/src/tea_shop.rs
Finistere: oh yeah missed that in the copy-paste

PR #1953, opened 2024-07-29
feat: cache entity requests in engine-v2

obmarg:
Following on from #1931 - this adds caching on the entity requests to subgraphs.  These are a bit more complicated because one request will usually map to more than one cache entry.  We need to filter out any representations that have a cache hit from the request to the subgraph, and then make sure we use the cache entry for those during deserialisation rather than trying to read them from the response.

I've made use of `serde_json::RawValue` in quite a few places to try to make things more efficient - so we don't have to go through a bunch of extra serde roundtrips.

Fixes GB-7159

### DIFF
--- a/engine/crates/engine-v2/src/sources/graphql/deserialize/entities.rs
+++ b/engine/crates/engine-v2/src/sources/graphql/deserialize/entities.rs
@@ -8,7 +8,7 @@ use serde::{
 use crate::{
     execution::PlanWalker,
     response::{ErrorCode, GraphqlError, ResponseKeys, ResponsePath, SubgraphResponseRefMut, UnpackedResponseEdge},
-    sources::graphql::CachedEntity,
+    sources::graphql::CacheEntry,
 };
 
 use super::errors::GraphqlErrorsSeed;
@@ -16,7 +16,7 @@ use super::errors::GraphqlErrorsSeed;
 pub(in crate::sources::graphql) struct EntitiesDataSeed<'resp> {
     pub response: SubgraphResponseRefMut<'resp>,
     pub plan: PlanWalker<'resp>,
-    pub cache_entries: Option<&'resp [CachedEntity]>,
+    pub cache_entries: Option<&'resp [CacheEntry]>,
 }
 
 impl<'resp, 'de> DeserializeSeed<'de> for EntitiesDataSeed<'resp>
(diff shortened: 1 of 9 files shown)


engine/crates/engine-v2/src/sources/graphql/deserialize/entities.rs
```diff
@@ -73,6 +76,7 @@ enum EntitiesKey {
 struct EntitiesSeed<'resp, 'parent> {
     response_part: &'parent SubgraphResponseRefMut<'resp>,
     plan: PlanWalker<'resp>,
+    cache_entries: Option<std::slice::Iter<'parent, CachedEntity>>,
```
Finistere, line 79: If you end up with an iterator in all cases, I would have made it generic over the iterator and provided an empty iterator in case nothing was cached, rather than an operation of an iterator.

Finistere (approved):

engine/crates/engine-v2/src/sources/graphql/deserialize/entities.rs
```diff
@@ -99,12 +103,23 @@ where
...
     }
 
-    fn visit_seq<A>(self, mut seq: A) -> Result<Self::Value, A::Error>
+    fn visit_seq<A>(mut self, mut seq: A) -> Result<Self::Value, A::Error>
     where
         A: SeqAccess<'de>,
     {
         while let Some(seed) = self.response_part.next_seed(self.plan) {
-            match seq.next_element_seed(seed) {
+            let next_cache_entry = self.cache_entries.as_mut().and_then(Iterator::next);
```
Finistere, line 111: this could be actually just expressed as:
```rust
while let Some(cache_entry) = ... { 
  // deserialize
}
while let Some(seed) = ... {
   // process as before
,}
```

engine/crates/engine-v2/src/sources/graphql/federation.rs
```diff
@@ -91,34 +77,214 @@ impl FederationEntityPreparedExecutor {
...
+
+        if let Some(cache_entries) = cache_entries.filter(|_| status.is_success()) {
+            update_cache(ctx, cache_ttl, bytes, cache_entries).await
+        }
+
+        Ok((status, subgraph_response))
+    }
+}
+
+async fn update_cache<R: Runtime>(
```
Finistere, line 205: this could be a bit simpler if you kept only the cache_entries that need to be updated, as you do for the representations

engine/crates/engine-v2/src/sources/graphql/federation.rs
```diff
@@ -91,34 +77,214 @@ impl FederationEntityPreparedExecutor {
...
+        .await
+        .inspect_err(|err| tracing::warn!("Failed to read the cache key {key}: {err}"))
+        .ok()
+        .flatten();
+
+    CachedEntity { key, data }
+}
+
+fn build_cache_key(subgraph_name: &str, repr: &RawValue) -> String {
+    let mut hasher = sha2::Sha256::new();
```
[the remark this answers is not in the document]
obmarg: I was just going off what we had in the Cargo.toml already.

PR #1956, opened 2024-07-30
feat(engine): Expose used fields in OTEL metrics

Finistere:
For a schema:
```graphql
type Query {
   user(id: ID!): User
}

type User {
  id: ID!
  name: String
}
```
and query:
```graphql
query {
  user(id: "0x1") {
    id
    name
  }
}
```
We produce the following string
```
"Query.user,User.id+name"
```

I'm also renaming `gql.operation.normalized_query` to `gql.operation.query` in metrics for consistency with tracing. That's the only `query` we're sending and likely will ever send anyway.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -53,4 +53,4 @@ WORKDIR /data
 ENTRYPOINT ["/bin/grafbase-gateway"]
 CMD ["--config", "/etc/grafbase.toml", "--listen-address", "0.0.0.0:5000"]
 
-EXPOSE 4000
+EXPOSE 5000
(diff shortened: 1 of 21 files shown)


Dockerfile
```diff
@@ -53,4 +53,4 @@ WORKDIR /data
 ENTRYPOINT ["/bin/grafbase-gateway"]
 CMD ["--config", "/etc/grafbase.toml", "--listen-address", "0.0.0.0:5000"]
 
-EXPOSE 4000
```
pimeys, line 56: this should be in the changelog when we release

pimeys (approved):

PR #1958, opened 2024-07-30
feat: gateway config for entity caching

obmarg:
Adds entity caching config to the gateway toml.  Allowing users to set defaults and also enable/disable and set the TTL per subgraph.

Config should look something like:

```toml
[entity_caching]
enabled = true
ttl = "60s"

[subgraphs.products.entity_caching]
ttl = "30s"

[subgraphs.reviews.entity_caching]
enabled = false
```

This PR is just the config for enabling & setting the TTL: configuring redis etc. will come when I get to adding redis support.

Fixes GB-7160
Fixes GB-7051


pimeys (approved):
I'd add at least one test to parse the toml. Otherwise sashay away.

PR #1959, opened 2024-07-30
chore(engine-v2): Remove unnecessary subgraph name from runtime fetch

Finistere:


engine/crates/engine-v2/src/sources/graphql/request.rs
```diff
@@ -23,6 +23,15 @@ pub trait ResponseIngester: Send {
...
 }
 
+impl<T> ResponseIngester for T
+where
+    T: FnOnce(Bytes) -> Result<(GraphqlResponseStatus, SubgraphResponse), ExecutionError> + Send,
+{
+    async fn ingest(self, bytes: Bytes) -> Result<(GraphqlResponseStatus, SubgraphResponse), ExecutionError> {
+        self(bytes)
+    }
+}
```
obmarg, line 33: Is this even used anymore?  Think I forgot to remove it.

obmarg (approved):

pimeys (approved):

engine/crates/engine-v2/src/sources/graphql/request.rs
Finistere: not used, but not a big deal to keep it though imho 🤷

PR #1963, opened 2024-07-31
feat: add support for dynamic rate limit configurations with hot reload

pimeys:
- Introduce dynamic rate limiting configurations
- Incorporate hot reload functionality to update rate limits without a server restart
- Allow configuration of hot reload of rate limits in `gateway` server setup


engine/crates/runtime-local/src/rate_limiting/in_memory/key_based.rs
```diff
@@ -34,48 +37,73 @@ impl RateLimiterContext for RateLimitingContext {
...
-    inner: HashMap<String, governor::DefaultKeyedRateLimiter<usize>>,
+    limiters: Arc<RwLock<HashMap<String, governor::DefaultKeyedRateLimiter<usize>>>>,
 }
 
 impl InMemoryRateLimiter {
-    pub fn runtime(config: KeyedRateLimitConfig<'_>) -> RateLimiter {
-        let mut limiter = Self::default();
+    pub fn runtime(
+        config: KeyedRateLimitConfig,
+        mut updates: mpsc::Receiver<HashMap<String, GraphRateLimit>>,
```
Finistere, line 47: Have you considered using a Watch channel like Redis implementation? You would just need to use https://docs.rs/tokio/latest/tokio/sync/watch/struct.Receiver.html#method.changed AFAIK to only handle the last event. Intermediate events have no value after all

engine/crates/runtime-local/src/rate_limiting/in_memory/key_based.rs
```diff
@@ -34,48 +37,73 @@ impl RateLimiterContext for RateLimitingContext {
...
+                continue;
+            };
+
+            limiters.insert(name.to_string(), limiter);
         }
 
-        RateLimiter::new(limiter)
-    }
+        let limiters = Arc::new(RwLock::new(limiters));
+        let limiters_copy = limiters.clone();
```
Finistere, line 61: Did you consider using a weak clone instead? Would avoid leaking futures if multiple RateLimiter are created (schema reload in cli/gateway, etc.)

gateway/crates/federated-server/src/config.rs
```diff
@@ -67,6 +73,25 @@ pub struct Config {
     pub entity_caching: EntityCachingConfig,
 }
 
+impl Config {
+    /// Load the rate limit configuration for global and subgraph level settings.
+    pub fn as_keyed_rate_limit_config(&self) -> HashMap<&str, GraphRateLimit> {
+        let mut key_based_config = HashMap::new();
+
+        if let Some(global_config) = self.gateway.rate_limit.as_ref().and_then(|c| c.global) {
+            key_based_config.insert("global", global_config);
```
Finistere, line 82: this will break with a subgraph named `"global"`

gateway/crates/federated-server/src/config/hot_reload.rs
```diff
@@ -0,0 +1,124 @@
...
+use grafbase_telemetry::span::GRAFBASE_TARGET;
+use notify::{EventHandler, EventKind, PollWatcher, Watcher};
+use runtime::rate_limiting::GraphRateLimit;
+use tokio::sync::{mpsc, watch};
+
+use crate::Config;
+
+type RateLimitData = HashMap<String, GraphRateLimit>;
+
+pub(crate) enum RateLimitSender {
```
Finistere, line 12: Would also simplify this code to re-use a watcher in the in-memory case

engine/crates/runtime-local/src/rate_limiting/in_memory/key_based.rs
pimeys: mm, yeah I don't like weaks that much but here that kind of makes sense.

Finistere (approved):

PR #1964, opened 2024-07-31
chore: decouple redis pools from rate limiting

obmarg:
I'm in the process of adding redis support to entity caching.  Rate limiting already uses redis, and ideally we want to be able to share the connection pool between caching & rate limiting, assuming the user is using the same redis settings for both.

This is a bit of prep work for that: extracting the generic redis connection code into it's own module with its own structs, where they were part of the rate limiting config/code before.

I've also added a `RedisPoolFactory` that is used to construct pools - it'll return the same pool if you give it the same details twice.  This should make it easy to share connection pools without having to explicity write code to compare a rate limiting config with an entity caching config.

Part of GB-6894

### DIFF
--- a/engine/crates/runtime-local/src/lib.rs
+++ b/engine/crates/runtime-local/src/lib.rs
@@ -8,6 +8,7 @@ mod kv;
 mod log;
 mod pg;
 pub mod rate_limiting;
+pub mod redis;
 mod ufd_invoker;
 
 pub use bridge::Bridge;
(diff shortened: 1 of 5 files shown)


Finistere (approved):

pimeys (approved):
Nice. I wonder will our config have multiple redis configurations or should we aim to have a global config somewhere and then things like rate limit or cache can do their own if needed?

obmarg:
> Nice. I wonder will our config have multiple redis configurations or should we aim to have a global config somewhere and then things like rate limit or cache can do their own if needed?

My initial plan was to do the easy thing and just duplicate the config in both places.  Not against consolidating, kinda just wanting to avoid breaking what we have already (and spending more time than necessary on config)

PR #1965, opened 2024-07-31
chore(gateway): Separate gateway config & use TOML in integration-tests

Finistere:
Instead of adding custom methods to imitate the TOML configuration, integration-tests
can now just use it directly.

Moving the gateway config into a separate crate also allows
`runtime-local` to depend on it directly if relevant, looking at you
Redis.

Had to do a bunch of clone with the config, we should ideally separate it into multiple struct with `#[serde(flatten)]` if TOML supports it


engine/crates/engine-config-builder/src/from_sdl_config.rs
```diff
@@ -0,0 +1,299 @@
+//! Glue crate between parser-sdl & engine-v2-config
```
Finistere, line 1: copy-pasted what was in `lib` before

engine/crates/engine-config-builder/src/from_toml_config.rs
```diff
@@ -0,0 +1,66 @@
+use engine_v2_config::VersionedConfig;
```
Finistere, line 1: moved the logc from `federated-server`

engine/crates/integration-tests/Cargo.toml
```diff
@@ -23,9 +23,10 @@ engine-config-builder = { path = "../engine-config-builder" }
 expect-test = "1.5"
 futures = "0.3.30"
 gateway-core = { workspace = true, features = ["partial-caching"] }
+gateway-config.workspace = true
 grafbase-graphql-introspection.workspace = true
 graphql-composition.workspace = true
-graphql-federated-graph.path = "../federated-graph"
+graphql-federated-graph = { path = "../federated-graph" }
```
pimeys, line 29: why not .workspace for this too?

gateway/crates/config/Cargo.toml
```diff
@@ -0,0 +1,32 @@
...
+version = "0.6.1"
+
+[lints]
+workspace = true
+
+[dependencies]
+ascii = { version = "1.1.0", features = ["serde"] }
+duration-str = "0.11.0"
+engine.workspace = true
+grafbase-telemetry = { workspace = true, features = ["tower", "otlp"] }
```
pimeys, line 17: I'm not really sure about this. Should we actually just have the configuration options for telemetry in the config crate, so we don't get this kind of crazy dependency chain here...

gateway/crates/config/Cargo.toml
```diff
@@ -0,0 +1,32 @@
...
+repository.workspace = true
+version = "0.6.1"
+
+[lints]
+workspace = true
+
+[dependencies]
+ascii = { version = "1.1.0", features = ["serde"] }
+duration-str = "0.11.0"
+engine.workspace = true
```
pimeys, line 16: It feels kind of weird to depend on engine in a config crate that configures engine.

gateway/crates/config/Cargo.toml
```diff
@@ -0,0 +1,32 @@
...
+grafbase-telemetry = { workspace = true, features = ["tower", "otlp"] }
+http.workspace = true
+parser-sdl = { version = "0.1.0", path = "../../../engine/crates/parser-sdl" }
```
pimeys, line 19: Should whatever needs this crate to actually get moved to parser-sdl? Would be great if the whole workspace depends on the config, not the other way around.

pimeys: Also workspace would be nicer here than path.

gateway/crates/federated-server/src/server.rs
```diff
@@ -61,24 +59,7 @@ pub async fn serve(
...
-                header_rules: config.headers,
-                rate_limit: config.gateway.rate_limit,
-                timeout: config.gateway.timeout,
-                entity_caching: config.entity_caching,
-            },
-            otel_reload,
-            sender,
-        )
-        .await?;
+    fetch_method.start(&config, otel_reload, sender).await?;
```
pimeys, line 62: This one here I need to change again, and is a big merge conflict for this PR and mine: the gateway config has the config file location and boolean if hot reloading is enabled on my side. I really also don't want to introduce two new parameters to this function, especially if one of them is boolean.

What I think I'll do is I'll add a new `GatewayConfig` with the config, otel reload, sender, config path and hot reload.

pimeys (approved):
good stuff.

some paper cuts ahead with merge conflicts.

gateway/crates/config/Cargo.toml
Finistere: fixed

gateway/crates/config/Cargo.toml
Finistere: did the workspace, but parser-sdl is kind of required anyway today. We're passing through SDL config to build the engine configuration today. And some tests for entity caching directly rely on stuff from `parser-sdl`. So well not fixing it in this PR

PR #1970, opened 2024-08-01
chore: stop using KV trait for entity caching

obmarg:
The KV trait was designed to wrap cloudflares KV implementation - where entries can have a TTL but don't have to, and there's also a local cache that has its own TTL you specfiy when performing a get.  

When writing entity caching I used the KV trait because 1) I didn't spot the differences and 2) it was there.

This PR fixes that mistake.  I've added a new trait for this, because the existing cache trait is more complicated than what entity caching requires, and I don't want to implement all of the original cache trait on top of redis.  

We may want to consolidate this new trait with the existing Cache trait at some point, but I've got a limited time to finish all this entity caching stuff before going on holiday so not going to do that today.

Fixes GB-7178


engine/crates/runtime-local/src/entity_cache/memory.rs
```diff
@@ -0,0 +1,67 @@
...
+    collections::{hash_map::Entry, HashMap},
+    sync::Mutex,
+    time::Instant,
+};
+
+use futures_util::future::BoxFuture;
+
+#[derive(Default)]
+pub struct InMemoryEntityCache {
+    inner: Mutex<HashMap<String, CacheValue>>,
```
Finistere, line 11: You could use `mini_moka::sync::Cache` like I do for operations

PR #1973, opened 2024-08-01
feat: implement and hook up RedisEntityCache

obmarg:
As per the title this finishes adding redis entity caching support.  

I haven't written any tests of this yet because the existing tests of the gateway binary don't seem to call subgraphs, so there's a bit of setup I'd need to do for that.  Have added GB-7210 to handle that.

Fixes GB-6894

### DIFF
--- a/engine/crates/integration-tests/tests/federation/trusted_documents.rs
+++ b/engine/crates/integration-tests/tests/federation/trusted_documents.rs
@@ -77,7 +77,7 @@ fn apollo_client_style_happy_path() {
             engine
                 .execute("")
                 .extensions(
-                    &json!({"persistedQuery": { "version": 1, "sha256Hash": &TRUSTED_DOCUMENTS[0].document_id }}),
+                    json!({"persistedQuery": { "version": 1, "sha256Hash": &TRUSTED_DOCUMENTS[0].document_id }}),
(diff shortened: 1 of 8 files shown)


obmarg:
Note to self: add some tests and do a rebase after #1970 is merged

engine/crates/runtime-local/src/entity_cache/memory.rs
```diff
@@ -0,0 +1,67 @@
...
+}
+
+struct CacheValue {
+    data: Vec<u8>,
+    expires_at: Instant,
+}
+
+impl InMemoryEntityCache {
+    async fn get(&self, name: &str) -> anyhow::Result<Option<Vec<u8>>> {
+        let mut lock = self.inner.lock().unwrap();
```
yoav-lavi, line 21: Should this be expect?

yoav-lavi (approved):
Nice work, see comment

engine/crates/runtime-local/src/entity_cache/memory.rs
obmarg: It shouldn't even be in this PR, accidentally hit ready for review before I'd done a rebase, please ignore.

PR #1976, opened 2024-08-02
feat(config): Global retry settings

pimeys:
A global setting to define retries for all subgraphs.


engine/crates/engine-v2/config/src/v2/mod.rs
```diff
@@ -87,7 +87,7 @@ pub struct RetryConfig {
     /// The fraction of the successful requests budget that can be used for retries.
     pub retry_percent: Option<f32>,
     /// Whether mutations should be retried at all. False by default.
-    pub retry_mutations: Option<bool>,
+    pub retry_mutations: bool,
```
Finistere, line 90: config needs to be backwards compatible, it may be stored

Finistere (approved):

PR #1982, opened 2024-08-04
chore(engine-v2): Use Definition suffix in schema structs

Finistere:
Not having this suffix makes a lot of things more confusing than
necessary. I've also renamed `Executor` to simply `Resolver` now with
this change.


pimeys (approved):
remember to update engines to API soon :)

PR #1987, opened 2024-08-05
chore: switch to mainline otel

pimeys:
Let's got some of the forks we need. The otel setup is now in http 1.0 and all that. Seems like we even get otel logs even when removing that provider set methods. Looks good.

### DIFF
--- a/gateway/crates/gateway-binary/src/main.rs
+++ b/gateway/crates/gateway-binary/src/main.rs
@@ -137,12 +137,10 @@ fn init_global_tracing(args: &impl Args, config: Option<TelemetryConfig>) -> any
         Some(logger) => {
             tracing_subscriber::registry()
                 .with(tracer.layer.boxed())
-                .with(logger.layer.boxed())
+                .with(logger.boxed())
                 .with(args.log_format())
                 .with(env_filter)
                 .init();
-
-            grafbase_telemetry::otel::opentelemetry::global::set_logger_provider(logger.provider.clone());
         }
         None => {
             tracing_subscriber::registry()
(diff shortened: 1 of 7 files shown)


gateway/crates/gateway-binary/src/main.rs
```diff
@@ -181,6 +179,9 @@ fn otel_layer_reload<S>(
         let ReloadableOtelLayers {
             tracer,
             meter_provider,
+            // I swear I don't really want to know what the compiler is doing here.
+            // Set this to `logger: _`, and the OTEL reload will explode.
```
tomhoule, line 183: `logger: _` drops the logger instantly, which is probably the problem. But `logger: _something` wouldn't drop immediately, so that should work without the annotation.

gateway/crates/gateway-binary/src/main.rs
```diff
@@ -212,10 +213,6 @@ fn otel_layer_reload<S>(
         grafbase_telemetry::otel::opentelemetry::global::set_meter_provider(meter_provider);
         grafbase_telemetry::otel::opentelemetry::global::set_tracer_provider(tracer.provider.clone());
 
-        if let Some(logger) = logger {
-            grafbase_telemetry::otel::opentelemetry::global::set_logger_provider(logger.provider.clone());
```
tomhoule, line 216: No longer needed?

tomhoule:
LGTM :+1:

gateway/crates/gateway-binary/src/main.rs
Finistere: that's suspicious, why is logging removed?

Finistere (approved):
Be sure to test that this still works with API before merging

gateway/crates/gateway-binary/src/main.rs
pimeys: Yeah they removed this from the API completely. Seems it works without.

pimeys:
work ongoing in https://github.com/grafbase/api/pull/3888

PR #1988, opened 2024-08-05
fix: include headers in entity caching keys

obmarg:
One of the things we want to avoid in entity caching is sharing cache entries that shouldn't be shared: one way that could happen is if some kind of authentication header was included in a subgraph request.  This handles that case by adding subgraph headers into the entity caching key.

We probably still need to handle adding auth scopes and such into the caching key, but that's a bigger job - this should handle a lot of the obvious cases.

### DIFF
--- a/engine/crates/engine-v2/src/sources/graphql/federation.rs
+++ b/engine/crates/engine-v2/src/sources/graphql/federation.rs
@@ -87,27 +87,21 @@ impl FederationEntityResolver {
                     cache_ttl,
                 };
 
+                let headers = ctx.subgraph_headers_with_rules(endpoint.header_rules());
+
                 if cache_ttl.is_some() {
-                    let fetches = representations
-                        .iter()
-                        .map(|repr| cache_fetch(ctx, endpoint.subgraph_name(), repr));
+                    match cache_fetches(ctx, endpoint, &headers, representations, &mut ingester).await {
+                        CacheFetchOutcome::FullyCached => {
+                            let (_, response) = ingester
+                                .ingest(Bytes::from_static(br#"{"data": {"_entities": []}}"#))
+                                .await?;
 
-                    let cache_entries = join_all(fetches).await;
-                    let fully_cached = !cache_entries.iter().any(CacheEntry::is_miss);
-                    ingester.cache_entries = Some(cache_entries);
-                    if fully_cached {
-                        let (_, response) = ingester
-                            .ingest(Bytes::from_static(br#"{"data": {"_entities": []}}"#))
-                            .await?;
-
-                        return Ok(response);
+                            return Ok(response);
+                        }
+                        CacheFetchOutcome::Other {
+                            filtered_representations,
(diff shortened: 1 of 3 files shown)


engine/crates/engine-v2/src/sources/graphql/federation.rs
```diff
@@ -151,6 +145,59 @@ impl FederationEntityResolver {
     }
 }
 
+async fn cache_fetches<'ctx, R: Runtime>(
+    ctx: ExecutionContext<'ctx, R>,
+    endpoint: schema::SchemaWalker<'_, GraphqlEndpointId>,
```
Finistere, line 150: just FYI, there is a `GraphqlEndpointWalker` alias

engine/crates/engine-v2/src/sources/graphql/federation.rs
```diff
@@ -151,6 +145,59 @@ impl FederationEntityResolver {
...
+    headers: &http::HeaderMap,
+    representations: Vec<Box<RawValue>>,
+    ingester: &mut EntityIngester<'ctx, R>,
+) -> CacheFetchOutcome {
+    let header_result = serde_json::to_vec(
```
Finistere, line 155: Nitpick: would have avoided the JSON by hashing directly the name & value (with headers length, name/value length or null separator for name).

engine/crates/engine-v2/src/sources/graphql/federation.rs
```diff
@@ -151,6 +145,59 @@ impl FederationEntityResolver {
...
+    };
+
+    let fetches = representations
+        .iter()
+        .map(|repr| cache_fetch(ctx, endpoint.subgraph_name(), &headers_json, repr));
+
+    let cache_entries = join_all(fetches).await;
+    let fully_cached = !cache_entries.iter().any(CacheEntry::is_miss);
+
+    ingester.cache_entries = Some(cache_entries);
```
Finistere, line 176: Wouldn't mutate the ingester here but return the `cache_entires` in the `CacheFetchOutcome`.

engine/crates/engine-v2/src/sources/graphql/root_fields.rs
```diff
@@ -129,10 +134,21 @@ impl GraphqlResolver {
     }
 }
 
-fn build_cache_key(subgraph_request_body: &[u8]) -> String {
+fn build_cache_key(subgraph_name: &str, subgraph_request_body: &[u8], headers: &http::HeaderMap) -> Option<String> {
+    let header_json = serde_json::to_vec(
```
Finistere, line 138: same comment on directly hashing the headers without an intermediate allocation

Finistere (approved):

engine/crates/engine-v2/src/sources/graphql/federation.rs
obmarg: rust analyzer doesn't do those though

PR #1992, opened 2024-08-06
chore(engine-v2): Remove now useless response metadata & simplify caching

Finistere:
Now that operation checks rely on ClickHouse, no need for response
metadata anymore. Also removed the trusted document cache, the operation
cache already fills that role and having two in-memory caches means
twice the monitoring.


tomhoule (approved):
Nice simplification!

PR #2001, opened 2024-08-08
feat(engine-v2): Add support for SSE subscriptions to subgraphs.

Finistere:
If the websocket URL is not defined we'll send a request with `Accept:
text/event-stream,application/json;q=0.9` to start a SSE subscriptions.
We may want to allow a different URL for this at some point, not sure.

I'm using the quick start of the JS `graphql-sse` library as a SSE
subgraph in the integration-tests.

I've changed the `Fetcher` trait to remove a bunch of allocations we're
doing for websockets, couldn't resist not removing those.


engine/crates/engine-v2/src/engine/retry_budget.rs
```diff
@@ -33,7 +33,7 @@ impl RetryBudgets {
 }
 
 impl<R: Runtime> super::Engine<R> {
-    pub(crate) fn get_retry_budget_for_query(&self, endpoint_id: GraphqlEndpointId) -> Option<&Budget> {
+    pub(crate) fn get_retry_budget_for_non_mutation(&self, endpoint_id: GraphqlEndpointId) -> Option<&Budget> {
```
tomhoule, line 36: :+1:

engine/crates/integration-tests/data/sse-subgraph/tmp.html
```diff
@@ -0,0 +1,113 @@
...
+   *
+   * This source code is licensed under the MIT license found in the
+   * LICENSE file in the root directory of this source tree.
+   *)
+*/
+;
+      ;globalThis[Symbol.for("Bun__renderFallbackError")](globalThis.__BUN_DATA__);
+    </script>
+  </body>
+</html>
```
tomhoule, line 113: Is that committed intentionally?

engine/crates/integration-tests/data/tmp.html
```diff
@@ -0,0 +1 @@
+{"data":{"hello":"world"}}
```
tomhoule, line 1: (same here)

engine/crates/runtime/src/fetch.rs
```diff
@@ -20,53 +21,135 @@ impl FetchError {
 
 pub type FetchResult<T> = Result<T, FetchError>;
 
-// very minimal for now, but will be expanded as we need it.
-pub struct FetchRequest<'a> {
-    pub url: &'a url::Url,
+/// reqwest uses Url instead of Uri, so as long as it's the actual implementation underneath it's a
+/// bit of a waste to use http::Request
```
tomhoule, line 25: I don't understand this comment — what's URI in this context?

tomhoule (approved):

engine/crates/integration-tests/data/sse-subgraph/tmp.html
Finistere: oops, thanks

engine/crates/runtime/src/fetch.rs
Finistere: meant this: https://docs.rs/http/latest/http/uri/struct.Uri.html
instead of `url::Url`

tomhoule: Aah ok, makes sense

PR #2003, opened 2024-08-08
gateway: error with helpful message when account name provided in graph ref

tomhoule:
They should not be passed. Currently, you only get a warning that the schema cannot be found. With the change in this commit, you get a (hopefully) helpful error message.

closes GB-7251


Finistere (approved):
nice

PR #2011, opened 2024-08-13
chore(metrics): Separate our internal operation metrics attributes

Finistere:
Prefixing a few custom metrics attributes to `__grafbase` and separating our custom operation name from the real one.

I've also fixed the operation name handling. We're generating a custom one if there isn't one already but we were sending it as `gql.operation.name` directly as if it was the real one which is too ambiguous for "public" use. So separated it into `__grafbase.gql.operation.inferred.name`

### DIFF
--- a/engine/crates/gateway-core/src/cache/partial.rs
+++ b/engine/crates/gateway-core/src/cache/partial.rs
@@ -72,6 +72,7 @@ where
                     body,
                     engine::GraphqlOperationAnalyticsAttributes {
                         name: request.operation_name().map(str::to_string),
+                        name_or_generated_one: request.operation_name().map(str::to_string).unwrap_or_default(),
                         r#type: operation_type,
                         used_fields: String::new(),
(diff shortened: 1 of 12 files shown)


engine/crates/gateway-core/src/cache/partial.rs
```diff
@@ -72,6 +72,7 @@ where
                     body,
                     engine::GraphqlOperationAnalyticsAttributes {
                         name: request.operation_name().map(str::to_string),
+                        name_or_generated_one: request.operation_name().map(str::to_string).unwrap_or_default(),
```
tomhoule, line 75: Do we have a generated name at this stage, or is it ok to default to an empty string?

tomhoule (approved):

fbjork:
@Finistere this functionality should happen server side, not in the gateway. We should not have any custom or internal metrics IMO.

PR #2012, opened 2024-08-13
feat: subgraph metrics

pimeys:
Implements subgraph request related metrics:

- latencies
- retries
- request body size
- response body size
- inflight requests

### DIFF
--- a/engine/crates/engine-v2/src/engine.rs
+++ b/engine/crates/engine-v2/src/engine.rs
@@ -60,7 +60,7 @@ pub struct Engine<R: Runtime> {
     pub(crate) schema: Arc<Schema>,
     pub(crate) schema_version: SchemaVersion,
     pub(crate) runtime: R,
-    operation_metrics: GraphqlOperationMetrics,
+    pub(crate) operation_metrics: GraphqlOperationMetrics,
     auth: AuthService,
     retry_budgets: RetryBudgets,
     operation_cache: <R::OperationCacheFactory as OperationCacheFactory>::Cache<Arc<PreparedOperation>>,
@@ -294,7 +294,7 @@ impl<R: Runtime> Engine<R> {
             if let Some(operation_metrics_attributes) = operation_metrics_attributes {
                 span.record_gql_request((&operation_metrics_attributes).into());
 
-                self.operation_metrics.record(
+                self.operation_metrics.record_operation(
                     GraphqlRequestMetricsAttributes {
                         operation: operation_metrics_attributes,
                         status,
@@ -345,7 +345,7 @@ impl<R: Runtime> Engine<R> {
                 if let Some(operation_metrics_attributes) = operation_metrics_attributes {
                     tracing::Span::current().record_gql_request((&operation_metrics_attributes).into());
 
-                    engine.operation_metrics.record(
+                    engine.operation_metrics.record_operation(
                         GraphqlRequestMetricsAttributes {
                             operation: operation_metrics_attributes,
                             status,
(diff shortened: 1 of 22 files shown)


gateway/crates/gateway-binary/tests/telemetry/metrics/operation/subgraph.rs
```diff
@@ -0,0 +1,242 @@
...
+                .bind(&service_name)
+                .bind(start_time_unix)
+                .fetch_all::<SumRow>()
+                .await
+                .unwrap();
+
+            insta::assert_json_snapshot!(rows, @r###"
+            [
+              {
+                "Value": 1.0,
```
tomhoule, line 169: Weird, do we expect to retry only once with that config? I guess so, since it's a minimum of 1 per second, and we haven't put any successful request in the budget yet at this point. Kind of makes sense, but I am still having trouble wrapping my head around how the budget works at the very beginning sometimes.

tomhoule (approved):
:100:

gateway/crates/gateway-binary/tests/telemetry/metrics/operation/subgraph.rs
pimeys: So what happens here is that it retries once (counter +1), then sleeps, then tries to retry the second time but has no balance in the bucket and aborts. Two rows to the database.

engine/crates/engine-v2/src/sources/graphql/request/execute.rs
```diff
@@ -70,20 +72,44 @@ pub(crate) async fn execute_subgraph_request<'ctx, 'a, R: Runtime>(
...
+    .await;
+
+    let duration = SystemTime::now()
+        .duration_since(start)
+        .unwrap_or(Duration::from_secs(0));
+
+    let response = match response {
+        Ok(response) => response,
+        Err(e) => {
+            let status = SubgraphResponseStatus::HttpError;
```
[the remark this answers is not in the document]
pimeys: These are the error codes. I wanted to have the error message too to the metrics, but after discussing with Benjamin we already have these in the logs and so forth we don't add that to the metrics.

gateway/crates/gateway-binary/tests/telemetry/metrics/operation/subgraph.rs
pimeys: The retry counter has that aborted flag, which you can either filter out or have as a separate message, which then tells you that this retry stopped and will not continue.

engine/crates/engine-v2/src/engine.rs
```diff
@@ -60,7 +60,7 @@ pub struct Engine<R: Runtime> {
     pub(crate) schema: Arc<Schema>,
     pub(crate) schema_version: SchemaVersion,
     pub(crate) runtime: R,
-    operation_metrics: GraphqlOperationMetrics,
```
Finistere, line 63: could just be `metrics` as you added all of them inside the structs

engine/crates/engine-v2/src/sources/graphql/request/execute.rs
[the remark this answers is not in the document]
Finistere: Not sure we understood each other I guess, sounds definitely like a good thing to have the `http.response.status_code` inside the metric for the subgraph response.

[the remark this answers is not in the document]
Finistere: might have said non-sense at some point 🤷

PR #2015, opened 2024-08-14
feat: implement caching metrics

pimeys:
### subgraph entity cache
- hits
- misses
### operation cache:
- hits
- misses
- size

Also enables in-memory operation cache for gateway.

### DIFF
--- a/engine/crates/runtime-local/src/operation_cache.rs
+++ b/engine/crates/runtime-local/src/operation_cache.rs
@@ -1,3 +1,4 @@
+use mini_moka::sync::ConcurrentCacheExt;
 use runtime::operation_cache::{OperationCache, OperationCacheFactory};
 
 pub struct InMemoryOperationCacheConfig {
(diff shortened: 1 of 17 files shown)


engine/crates/runtime-local/src/operation_cache.rs
```diff
@@ -49,8 +50,10 @@ impl<V> OperationCache<V> for InMemoryOperationCache<V>
 where
     V: Clone + Send + Sync + 'static + serde::Serialize + serde::de::DeserializeOwned,
 {
-    async fn insert(&self, key: String, value: V) {
+    async fn insert(&self, key: String, value: V) -> u64 {
         self.inner.insert(key, value);
+        self.inner.sync();
```
tomhoule, line 55: I wonder if that's expensive. I guess we'll see.

gateway/crates/federated-server/src/server/gateway.rs
```diff
@@ -146,30 +148,36 @@ pub struct GatewayRuntime {
...
         &self.meter
     }
+
     fn hooks(&self) -> &HooksWasi {
         &self.hooks
     }
+
     fn operation_cache_factory(&self) -> &Self::OperationCacheFactory {
-        &()
+        &self.operation_cache_factory
```
tomhoule, line 180: Did you manage to clarify why that was a no-op?

tomhoule (approved):

engine/crates/runtime-local/src/operation_cache.rs
pimeys: we run this outside from the main thread anyhow. this is the only way of getting the real counter value right after adding a thing to the cache.

gateway/crates/federated-server/src/server/gateway.rs
pimeys: No. There's a PR

https://github.com/grafbase/grafbase/pull/1854

This PR is about adding back the op cache. But it did not do it eventually...

PR #2018, opened 2024-08-15
cli: make creating federated graphs possible in grafbase create

tomhoule:
Both interactively (we prompt the user) and through CLI args with the `--mode` argument. In both cases, we default to managed graphs.

```
  Set up and deploy a new graph

Usage: grafbase create [OPTIONS]

Options:
  -n, --name <NAME>
          The name to use for the new graph
  -a, --account <SLUG>
          The slug of the account in which the new graph should be created
  -m,  --mode <MODE>
          Whether the graph is self-hosted or managed. Possible values: self-hosted or managed
          (default)
  -e, --env [<ENVIRONMENT_VARIABLES>...]
          Adds an environment variable to the graph
  -h, --help
          Print help

```

closes GB-7297

### DIFF
--- a/cli/crates/backend/src/api/create.rs
+++ b/cli/crates/backend/src/api/create.rs
@@ -17,6 +17,8 @@ use cynic::{MutationBuilder, QueryBuilder};
 use std::iter;
 use tokio::fs;
 
+pub use super::graphql::mutations::GraphMode;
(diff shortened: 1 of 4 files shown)


yoav-lavi (approved):
Nice!

PR #2025, opened 2024-08-16
feat: various metrics (part 2)

pimeys:
moar

### DIFF
--- a/engine/crates/engine-v2/src/engine.rs
+++ b/engine/crates/engine-v2/src/engine.rs
@@ -8,6 +8,7 @@ use ::runtime::{
 use async_runtime::stream::StreamExt as _;
 use bytes::Bytes;
 use engine_parser::types::OperationType;
+use enumset::EnumSet;
 use futures::{channel::mpsc, FutureExt, StreamExt, TryFutureExt};
 use futures_util::{SinkExt, Stream};
 use gateway_v2_auth::AuthService;
@@ -15,7 +16,7 @@ use grafbase_telemetry::{
     gql_response_status::GraphqlResponseStatus,
     grafbase_client::Client,
     metrics::{
-        GraphqlOperationMetrics, GraphqlRequestMetricsAttributes, OperationMetricsAttributes,
+        GraphqlErrorAttributes, GraphqlOperationMetrics, GraphqlRequestMetricsAttributes, OperationMetricsAttributes,
         QueryPreparationAttributes,
     },
     span::{gql::GqlRequestSpan, GqlRecorderSpanExt, GRAFBASE_TARGET},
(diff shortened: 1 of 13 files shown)


engine/crates/engine-v2/src/engine.rs
```diff
@@ -324,6 +327,16 @@ impl<R: Runtime> Engine<R> {
                 tracing::debug!(target: GRAFBASE_TARGET, "{message}")
             }
 
+            if let Some(attributes) = operation_metrics_attributes {
+                for error in response.errors() {
+                    self.operation_metrics.increment_graphql_errors(GraphqlErrorAttributes {
+                        code: error.code.to_string(),
```
Finistere, line 333: Change this to a `&'static str`

engine/crates/engine-v2/src/engine.rs
```diff
@@ -324,6 +327,16 @@ impl<R: Runtime> Engine<R> {
                 tracing::debug!(target: GRAFBASE_TARGET, "{message}")
             }
 
+            if let Some(attributes) = operation_metrics_attributes {
+                for error in response.errors() {
```
Finistere, line 331: should be distinct error codes as discussed

engine/crates/engine-v2/src/engine.rs
```diff
@@ -435,28 +455,46 @@ impl<'ctx, R: Runtime> PreExecutionContext<'ctx, R> {
         };
 
         let mut status: GraphqlResponseStatus = GraphqlResponseStatus::Success;
+
         struct Sender<'a> {
             sender: mpsc::Sender<Response>,
             status: &'a mut GraphqlResponseStatus,
+            operation_name: Option<String>,
+            client: Option<Client>,
+            operation_metrics: GraphqlOperationMetrics,
```
Finistere, line 464: why clone? could just be a `&GraphqlOperationMetrics`

engine/crates/telemetry/src/metrics/operation.rs
```diff
@@ -23,6 +24,12 @@ pub struct GraphqlOperationMetrics {
     query_preparation_latency: Histogram<u64>,
     batch_sizes: Histogram<u64>,
     request_body_sizes: Histogram<u64>,
+    graphql_errors: Counter<u64>,
+}
+
+#[derive(Clone)]
```
Finistere, line 30: You can just remove the `Clone`, you don't need it for the response sender stuff in engine-v2, and `Clone` isn't even used in this repo. Don't think it is in api either.

engine/crates/engine-v2/src/engine.rs
pimeys: Yeah, we could, if the async block starting from line 430 would not take the ownership of self...

Finistere: `self` is retrieved back as `ctx`, so you can just do `ctx.engine.operation_metrics`

Finistere (approved):

PR #2028, opened 2024-08-16
feat(metrics): update metrics for request and operation latency

pimeys:
The last pieces in grafbase/grafbase. This PR will break the API and our own analytics, so we need to fix that before merging.


engine/crates/telemetry/src/metrics/request.rs
```diff
@@ -35,23 +42,53 @@ impl RequestMetrics {
...
+            attributes.push(KeyValue::new("server.port", listen_address.port() as i64));
+        }
+
+        if let Some(scheme) = url_scheme {
+            attributes.push(KeyValue::new("url.scheme", scheme.to_string()));
+        }
+
         if let Some(cache_status) = cache_status {
-            attributes.push(KeyValue::new("http.response.headers.cache_status", cache_status));
+            attributes.push(KeyValue::new("http.response.headers.cache.status", cache_status));
```
Finistere, line 77: maybe `x-grafbase-cache`?

engine/crates/telemetry/src/metrics/request.rs
```diff
@@ -17,12 +19,17 @@ pub struct RequestMetricsAttributes {
     pub cache_status: Option<String>,
     pub gql_status: Option<GraphqlResponseStatus>,
     pub client: Option<Client>,
+    pub url_scheme: Option<String>,
+    pub route: Option<String>,
```
Finistere, line 23: Think could be a `&'static str`. Not sure we'll ever need a dynamic string even in the future

Finistere (approved):

engine/crates/telemetry/src/metrics/request.rs
pimeys: Ummm... we do. https://grafbase.com/docs/self-hosted-gateway#graph-configuration

Finistere: hum... ok fair point forgot that was configurable. Could `Box::leak` I guess but with config reloading and all...

PR #2031, opened 2024-08-19
gateway: shut down tracer and logger providers in graceful shutdown

tomhoule:
So the exporters have time to flush during the grace period.

### DIFF
--- a/engine/crates/telemetry/src/graceful_shutdown.rs
+++ b/engine/crates/telemetry/src/graceful_shutdown.rs
@@ -1,0 +1,4 @@
+pub fn graceful_shutdown() {
+    opentelemetry::global::shutdown_tracer_provider();
+    opentelemetry::global::shutdown_logger_provider();
+}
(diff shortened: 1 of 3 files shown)


yoav-lavi (approved):
Xzibit: "so I put a graceful shutdown in your graceful shutdown so you can shutdown while you shutdown"

engine/crates/telemetry/src/graceful_shutdown.rs
```diff
@@ -0,0 +1,4 @@
+pub fn graceful_shutdown() {
+    opentelemetry::global::shutdown_tracer_provider();
```
Finistere, line 2: would need to shut down the meter provider but that's also only available through the "SDK" implementation, not the interface -_-"

Finistere: can't even shutdown simultaneously lovely.

tomhoule: Yes it's challenging to wire that into this function, but I could spend some more time and try harder.

Finistere: Honestly you can probably leave it as is, I'll remove the reload logic, so can probably keep the SdkMeter ref somewhere

Finistere (approved):

PR #2041, opened 2024-08-20
fix(engine-v2): Correct operationName handling

Finistere:


engine/crates/integration-tests/tests/federation/basic/operations.rs
```diff
@@ -0,0 +1,166 @@
...
+
+#[test]
+fn multiple_operations_without_name() {
+    let response = runtime().block_on(async move {
+        let engine = Engine::builder().with_subgraph(FakeGithubSchema).build().await;
+
+        engine
+            .post(
+                r"
+                    query First {
```
tomhoule, line 13: I don't get this test, the operations have names here right?

yoav-lavi (approved):

engine/crates/integration-tests/tests/federation/basic/operations.rs
Finistere: ah without providing an `operationName`

tomhoule (approved):
:100:

PR #2043, opened 2024-08-21
feat: respect subgraph CacheControl headers

obmarg:
Prior to this PR we'd always use the default cache TTL for a subgraph as provided in the configuration.  However, we want subgraphs to be able to override that on a per-response basis by returning a CacheControl header.

This PR implements that:

1. If a subgraph returns `no-store` or `private` directives we avoid writing to the entity cache.
2. If the `Age` header and the `max-age` directive are present we set the TTL to `age - max_age`
3. If the `max-age` directive is present with no `Age` header we take the `max-age` as the TTL

Fixes GB-7164

### DIFF
--- a/engine/crates/engine-v2/src/sources/graphql/mod.rs
+++ b/engine/crates/engine-v2/src/sources/graphql/mod.rs
@@ -5,25 +5,17 @@ mod request;
 mod root_fields;
 mod subscription;
 
-use std::time::Duration;
-
 pub(crate) use federation::*;
-use headers::HeaderMapExt;
 pub(crate) use root_fields::*;
 
 use grafbase_telemetry::gql_response_status::GraphqlResponseStatus;
-use runtime::bytes::OwnedOrSharedBytes;
-
-fn should_update_cache(status: GraphqlResponseStatus, cache_control: Option<&headers::CacheControl>) -> bool {
-    status.is_success()
-        && cache_control
-            .map(|cache_control| !(cache_control.private() || cache_control.no_store()))
-            .unwrap_or(true)
-}
+use headers::HeaderMapExt;
+use http::HeaderMap;
+use std::time::Duration;
 
 fn calculate_cache_ttl(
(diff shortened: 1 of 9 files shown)


engine/crates/engine-v2/src/sources/graphql/mod.rs
```diff
@@ -5,5 +5,45 @@ mod request;
...
 pub(crate) use federation::*;
+use headers::HeaderMapExt;
 pub(crate) use root_fields::*;
+
+use grafbase_telemetry::gql_response_status::GraphqlResponseStatus;
+use runtime::bytes::OwnedOrSharedBytes;
+
+fn should_update_cache(status: GraphqlResponseStatus, cache_control: Option<&headers::CacheControl>) -> bool {
+    status.is_success()
+        && cache_control
```
Finistere, line 19: I would have moved that to the TTL computation, if `no-store` / `private`, would return a `None` for the TTL. Would keep the `Cache-Control` header manipulation in a single place. And would have provided only the headers to the `calculate_cache_ttl`, looked a bit weird to provide the full http response rather than just the headers IMHO.

engine/crates/engine-v2/src/sources/graphql/mod.rs
```diff
@@ -5,5 +5,45 @@ mod request;
...
+            .map(|cache_control| !(cache_control.private() || cache_control.no_store()))
+            .unwrap_or(true)
+}
+
+fn calculate_cache_ttl(
```
Finistere, line 24: would have created a `graphql/entity_cache.rs` or something like this

Finistere (approved):

engine/crates/engine-v2/src/sources/graphql/mod.rs
Finistere: well I mean, you didn't :D If you intended to do later then great. Just saying what comes to mind reading the PR once marked as reviewable

PR #2044, opened 2024-08-21
gateway: tracing context propagation

tomhoule:
The gateway now implements tracing propagation: it can receive tracing context (trace id, parent span id, additional context) from headers in the requests it receives, and pass that context on to subgraphs when it makes requests to them. That enables correlation of requests across network boundaries.

There are different standards for propagation. In this release, we implement [Trace Context](https://www.w3.org/TR/trace-context/) and [Baggage](https://opentelemetry.io/docs/concepts/signals/baggage/) — the standard OpenTelemetry mechanisms for trace parent and trace context propagation respectively —, and AWS X-Ray. Multiple mechanisms can be combined.

This is configurable through the `telemetry.tracing.propagation` key in the configuration. For example:

```toml
[telemetry.tracing.propagation]
trace_context = true
baggage = true
```

In addition, a new option named `parent_based_sampler` appears in `telemetry.tracing`:

```toml
[telemetry.tracing]
sampling = 0.2
parent_based_sampler = true
```

When enabled, the gateway will honor sampling configuration passed in through the `sampled` header (W3C spec), then fall back to its own configuration. This option is disabled by default and should not be enabled if the gateway is exposed directly to the internet, since malicious clients could increase load on the gateway by forcing traces to be recorded.

### DIFF
--- a/engine/crates/telemetry/src/http.rs
+++ b/engine/crates/telemetry/src/http.rs
@@ -1,0 +1,12 @@
+/// FIXME: Can be replaced with tracing-http once we are back to upstream opentelemetry.
+pub struct HeaderInjector<'a>(pub &'a mut http::HeaderMap);
+
+impl<'a> opentelemetry::propagation::Injector for HeaderInjector<'a> {
+    fn set(&mut self, key: &str, value: String) {
+        if let Ok(name) = http::header::HeaderName::from_bytes(key.as_bytes()) {
+            if let Ok(val) = http::header::HeaderValue::from_str(&value) {
+                self.0.insert(name, val);
+            }
+        }
+    }
+}
--- a/engine/crates/telemetry/src/span/request.rs
+++ b/engine/crates/telemetry/src/span/request.rs
@@ -46,13 +46,6 @@ pub struct HttpRequestSpan<'a> {
 }
 
 impl<'a> HttpRequestSpan<'a> {
-    /// Sets the span ray_id
-    pub fn with_ray_id(mut self, ray_id: impl Into<Option<Cow<'a, http::HeaderValue>>>) -> Self {
-        self.header_ray_id = ray_id.into();
-
-        self
-    }
-
     /// Sets the span git_branch
     pub fn with_git_branch(mut self, git_branch: impl Into<Option<Cow<'a, http::HeaderValue>>>) -> Self {
         self.git_branch = git_branch.into();
--- a/engine/crates/telemetry/src/tower.rs
+++ b/engine/crates/telemetry/src/tower.rs
@@ -56,18 +56,11 @@ pub struct TelemetryService<S> {
 }
 
 impl<S> TelemetryService<S> {
-    #[cfg(not(feature = "lambda"))]
-    fn make_span<B: Body>(&mut self, request: &Request<B>) -> Span {
-        HttpRequestSpan::from_http(request).into_span()
-    }
-
-    #[cfg(feature = "lambda")]
     fn make_span<B: Body>(&self, request: &Request<B>) -> Span {
-        use opentelemetry::Context;
         use tracing_opentelemetry::OpenTelemetrySpanExt;
 
         let parent_ctx = opentelemetry::global::get_text_map_propagator(|propagator| {
-            propagator.extract_with_context(&Context::current(), &HeaderExtractor(request.headers()))
+            propagator.extract(&HeaderExtractor(request.headers()))
         });
 
         let span = HttpRequestSpan::from_http(request).into_span();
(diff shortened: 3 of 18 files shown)


engine/crates/telemetry/src/span/request.rs
```diff
@@ -46,13 +46,6 @@ pub struct HttpRequestSpan<'a> {
 }
 
 impl<'a> HttpRequestSpan<'a> {
-    /// Sets the span ray_id
-    pub fn with_ray_id(mut self, ray_id: impl Into<Option<Cow<'a, http::HeaderValue>>>) -> Self {
-        self.header_ray_id = ray_id.into();
-
-        self
```
tomhoule, line 53: Dead code

gateway/crates/federated-server/src/server/graph_fetch_method.rs
```diff
@@ -65,6 +65,8 @@ impl GraphFetchMethod {
                 let gateway = gateway::generate(&federated_schema, None, config, hot_reload_config_path).await?;
 
                 sender.send(Some(Arc::new(gateway)))?;
+
+                std::mem::forget(otel_reload);
```
tomhoule, line 69: The drop causes errors on stderr since the channel is closed on drop and the reload task freaks out. Becomes irrelevant with the removal of the reload layer.

gateway/crates/gateway-binary/src/telemetry.rs
```diff
@@ -15,25 +15,56 @@ pub(crate) struct OpenTelemetryProviders {
...
-                grafbase_telemetry::otel::opentelemetry::global::set_text_map_propagator(XrayPropagator::default());
-
-                XrayIdGenerator::default()
-            } else {
-                use grafbase_telemetry::otel::opentelemetry_sdk::trace::RandomIdGenerator;
-
-                RandomIdGenerator::default()
-            }
+
+    {
```
tomhoule, line 27: separate function for this block? Would help readability I think

engine/crates/telemetry/src/http.rs
```diff
@@ -0,0 +1,12 @@
+/// FIXME: Can be replaced with tracing-http once we are back to upstream opentelemetry.
```
pimeys, line 1: had this shiver when I thought about it... nice.

engine/crates/telemetry/src/tower.rs
```diff
@@ -56,18 +56,11 @@ pub struct TelemetryService<S> {
...
-    }
-
-    #[cfg(feature = "lambda")]
     fn make_span<B: Body>(&self, request: &Request<B>) -> Span {
-        use opentelemetry::Context;
         use tracing_opentelemetry::OpenTelemetrySpanExt;
 
         let parent_ctx = opentelemetry::global::get_text_map_propagator(|propagator| {
-            propagator.extract_with_context(&Context::current(), &HeaderExtractor(request.headers()))
+            propagator.extract(&HeaderExtractor(request.headers()))
```
pimeys, line 63: Why we needed to switch here? Read the docs on `extract_with_context` and `extract` and I still don't get the difference.

gateway/crates/gateway-binary/src/telemetry.rs
```diff
@@ -15,25 +15,56 @@ pub(crate) struct OpenTelemetryProviders {
...
+        }
+
+        if tracing_config.propagation.aws_xray {
+            propagators.push(Box::new(XrayPropagator::default()));
+            xray_propagator_added = true;
         }
-    };
+
+        if cfg!(feature = "lambda") && !xray_propagator_added {
+            propagators.push(Box::new(XrayPropagator::default()))
```
pimeys, line 51: Should we get rid of this and use the config option only?

gateway/crates/gateway-binary/tests/telemetry/tracing.rs
```diff
@@ -0,0 +1,420 @@
+use std::{
```
pimeys, line 1: Fantastic.

pimeys (approved):

gateway/crates/gateway-binary/src/telemetry.rs
tomhoule: That would work, a bit more configuration burden for the user but simpler for us, and less branching.

pimeys: exactly. nobody uses it on lambda yet, so we can break things.

engine/crates/telemetry/src/tower.rs
tomhoule: The context argument is there for fallback in case a context (trace parent + metadata) can't be extracted from the headers. `extract()` uses the current context as fallback so that's strictly equivalent to what we were doing.

gateway/crates/gateway-binary/src/telemetry.rs
tomhoule: Ok doing that then.

PR #2045, opened 2024-08-21
cli: make ts config optional for gb create --mode=self-hosted

tomhoule:
When --mode=self-hosted, no grafbase configuration is required anymore


yoav-lavi (commented):
Wouldn't this be a bit of friction long term if you always have the same mode but need to specify it?

tomhoule:
Yes, we could have heuristics, but I'm wondering how to do that without introducing too much implicit behaviour.

fbjork (approved):

PR #2053, opened 2024-08-23
chore: Remove OTEL reload mechanism

Finistere:
- Remove all the OTEL reload mechanism and sending `grafbase-graph-ref` in gRPC metadata for OTEL
- Simplifying a bit the gateway crate with the `not(feature = "lambda")` feature flag portions to be as small as possible
- Telemetry-related logic is now passed down to the federated-sevrer as a `ServerRuntime` trait which has `graceful_shutdown` and `after_request` (for lambda case where we export traces after each request). Telemetry shutdown is now correctly called _after_ the server has finished serving requests.


gateway/crates/federated-server/src/server/graph_updater.rs
```diff
@@ -118,8 +114,9 @@ impl GraphUpdater {
             sender,
             current_id: None,
             gateway_config,
-            otel_reload,
-            latencies: None,
+            latencies: meter_from_global_provider()
```
pimeys, line 117: hell yes

gateway/crates/gateway-binary/src/main.rs
```diff
@@ -33,15 +23,15 @@ fn main() -> anyhow::Result<()> {
...
+    let config = args.config()?;
 
     let runtime = runtime::Builder::new_multi_thread()
         .enable_all()
         .thread_name(THREAD_NAME)
         .build()?;
 
     runtime.block_on(async move {
-        let otel_tracing = if std::env::var("__GRAFBASE_RUST_LOG").is_ok() {
+        let telemetry = if std::env::var("__GRAFBASE_RUST_LOG").is_ok() {
```
pimeys, line 34: do we still need this?

gateway/crates/gateway-binary/src/server_runtime.rs
```diff
@@ -0,0 +1,61 @@
+use crate::telemetry::OpenTelemetryProviders;
```
pimeys, line 1: nice

gateway/crates/gateway-binary/src/telemetry.rs
```diff
@@ -0,0 +1,102 @@
...
+
+    let config = config.unwrap_or(TelemetryConfig {
+        service_name: "unknown".to_string(),
+        resource_attributes: Default::default(),
+        tracing: Default::default(),
+        exporters: Default::default(),
+        logs: Default::default(),
+        metrics: Default::default(),
+        grafbase: Default::default(),
+    });
```
pimeys, line 46: Can't we do this in the config crate instead? Unknown is also a bit weird default value.

gateway/crates/gateway-binary/src/telemetry.rs
```diff
@@ -0,0 +1,102 @@
...
+        logger,
+    } = grafbase_telemetry::otel::layer::build(config, id_generator, Tokio)?;
+
+    if let Some(ref meter_provider) = meter_provider {
+        grafbase_telemetry::otel::opentelemetry::global::set_meter_provider(meter_provider.clone());
+    }
+
+    if let Some(ref logger) = logger {
+        grafbase_telemetry::otel::opentelemetry::global::set_logger_provider(logger.provider.clone());
+    }
```
pimeys, line 60: I think this is not needed. It's removed in the latest otel crates, and worked fine just without when I tried.

gateway/crates/gateway-binary/tests/telemetry/metrics.rs
```diff
@@ -79,7 +79,7 @@ where
 
     println!("service_name: {}", service_name);
     with_static_server(config, &schema, None, None, |client| async move {
-        const WAIT_SECONDS: u64 = 2;
+        const WAIT_SECONDS: u64 = 3;
```
pimeys, line 82: The game of changing this number. Been there too.

pimeys (approved):
Awesome. Check the few comments, and merge if you think it's ready.

gateway/crates/gateway-binary/src/main.rs
Finistere: Current logs are unreadable IMHO for debugging... :/

PR #2055, opened 2024-08-23
graphql-federated-graph: decouple FederatedGraph from serialized format

tomhoule:
The original rationale for this data structure being serializable was twofold:

1. Easier to deserialize than parse from sdl. Convenience for us.
2. Faster to deserialize than parse from sdl. Performance on engine start up.

Both reasons do not apply anymore.

1. We had to implement from_sdl anyway, so now there are three ways to construct a FederatedGraph (from composition, from sdl and deserialization)
2. In managed contexts, the executor stores engine_v2::Schema in its cache, not the config, so startup times are unaffected.

This change will let us iterate on FederatedGraph without having to provide an upgrade path for the data structure, without fear of breaking changes.

closes GB-7359

### DIFF
--- a/cli/crates/federated-dev/src/dev/gateway_nanny.rs
+++ b/cli/crates/federated-dev/src/dev/gateway_nanny.rs
@@ -51,7 +51,7 @@ impl EngineNanny {
 
 pub(super) async fn new_gateway(config: Option<engine_v2::VersionedConfig>) -> Option<Arc<Engine<CliRuntime>>> {
     let config = config?.into_latest();
-    let graph = graphql_federated_graph::from_sdl(&config.graph).ok()?;
+    let graph = graphql_federated_graph::from_sdl(&config.federated_sdl).ok()?;
 
     let runtime = CliRuntime {
         fetcher: NativeFetcher::default(),
@@ -92,7 +92,7 @@ pub(super) async fn new_gateway(config: Option<engine_v2::VersionedConfig>) -> O
         entity_cache: InMemoryEntityCache::default(),
     };
 
-    let schema = (config, graph).try_into().ok()?;
+    let schema = config.try_into().ok()?;
     let engine = Engine::new(Arc::new(schema), None, runtime).await;
 
     Some(Arc::new(engine))
--- a/cli/crates/federated-dev/src/dev/gateway_nanny.rs
+++ b/cli/crates/federated-dev/src/dev/gateway_nanny.rs
@@ -51,6 +51,7 @@ impl EngineNanny {
 
 pub(super) async fn new_gateway(config: Option<engine_v2::VersionedConfig>) -> Option<Arc<Engine<CliRuntime>>> {
     let config = config?.into_latest();
+    let graph = graphql_federated_graph::from_sdl(&config.graph).ok()?;
 
     let runtime = CliRuntime {
         fetcher: NativeFetcher::default(),
@@ -75,7 +76,7 @@ pub(super) async fn new_gateway(config: Option<engine_v2::VersionedConfig>) -> O
 
             for (subgraph_name, subgraph) in config.subgraph_configs.iter() {
                 if let Some(limit) = subgraph.rate_limit {
-                    let name = &config.graph[config.graph[*subgraph_name].name];
+                    let name = &graph[graph[*subgraph_name].name];
                     key_based_config.insert(
                         RateLimitKey::Subgraph(name.clone().into()),
                         GraphRateLimit {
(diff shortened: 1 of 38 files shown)


engine/crates/engine-v2/config/src/v6.rs
```diff
@@ -0,0 +1,123 @@
...
+    AuthConfig, AuthProviderConfig, CacheConfig, CacheConfigTarget, CacheConfigs, EntityCaching, GraphRateLimit,
+    Header, HeaderForward, HeaderInsert, HeaderRemove, HeaderRenameDuplicate, HeaderRule, HeaderRuleId, HeaderValue,
+    JwksConfig, JwtConfig, NameOrPattern, OperationLimits, PathId, RateLimitConfig, RateLimitKey, RateLimitRedisConfig,
+    RateLimitRedisTlsConfig, RateLimitStorage, RetryConfig, StringId, SubgraphConfig,
+};
+
+/// Configuration for a federated graph
+#[derive(serde::Serialize, serde::Deserialize, Debug)]
+pub struct Config {
+    pub graph: String,
```
Finistere, line 21: maybe `federated_sdl`? `graph` sounds less helpful

engine/crates/engine-v2/schema/src/builder/graph.rs
```diff
@@ -648,10 +658,15 @@ impl<'a> GraphBuilder<'a> {
         resolver_id
     }
 
-    fn push_directives(&mut self, config: &Config, directives: Directives) -> IdRange<TypeSystemDirectiveId> {
+    fn push_directives(
+        &mut self,
+        config: &Config,
+        directives: Directives,
+        graph: &mut FederatedGraphV4,
```
Finistere, line 665: doesn't need to be `mut`, no? (not a big deal)

engine/crates/federated-graph/src/lib.rs
```diff
@@ -19,27 +19,27 @@ pub enum FederatedGraph {
...
-        from_sdl(sdl)
+    pub fn into_federated_sdl(self) -> String {
+        match self {
+            FederatedGraph::Sdl(sdl) => sdl,
+            other => render_federated_sdl(&other.into_latest()).unwrap(),
+        }
     }
 
-    pub fn into_latest(self) -> FederatedGraphV3 {
+    pub fn into_latest(self) -> FederatedGraphV4 {
```
Finistere, line 37: if `FederatedGraphV4` isn't meant to be stored, I would rename `FederatedGraph` to `VersionedFederatedGraph` and `FederatedGraphV4` to `FederatedGraph`

engine/crates/engine-v2/schema/src/builder/mod.rs
```diff
@@ -27,10 +28,19 @@ use requires::*;
...
+    fn try_from(config: Config) -> Result<Self, Self::Error> {
+        let graph = federated_graph::from_sdl(&config.graph)?;
+        (config, graph).try_into()
+    }
+}
+
+impl TryFrom<(Config, FederatedGraphV4)> for Schema {
+    type Error = BuildError;
+
+    fn try_from((mut config, mut graph): (Config, FederatedGraphV4)) -> Result<Self, Self::Error> {
```
Finistere, line 40: I would not add this method. This optimization is only useful for the CLI which is kind of irrelevant IMHO. Even there the optimization likely doesn't even matter. The type signature doesn't guarantee in any way that `graph` must be from coming from the `config`

cli/crates/federated-dev/src/dev/gateway_nanny.rs
```diff
@@ -91,7 +92,7 @@ pub(super) async fn new_gateway(config: Option<engine_v2::VersionedConfig>) -> O
         entity_cache: InMemoryEntityCache::default(),
     };
 
-    let schema = config.try_into().ok()?;
+    let schema = (config, graph).try_into().ok()?;
```
Finistere, line 95: As mentioned later, think it's more harmful to have this API than the small gains in a barely used cli feature (only meant for development) this adds.

Finistere (approved):

engine/crates/engine-v2/config/src/v6.rs
tomhoule: I agree. Changing that.

engine/crates/engine-v2/schema/src/builder/graph.rs
tomhoule: Will double check

engine/crates/engine-v2/schema/src/builder/mod.rs
tomhoule: Removing that.

cli/crates/federated-dev/src/dev/gateway_nanny.rs
tomhoule: done

engine/crates/federated-graph/src/lib.rs
tomhoule: Agree we might as well do it now. Pushing that in a minute.

engine/crates/engine-v2/schema/src/builder/graph.rs
tomhoule: Indeed, remove the mut

PR #2056, opened 2024-08-24
chore(engine-v2): remove useless Names

Finistere:
Added initially to handle renaming of fields, etc. It's not used and a poor design.


tomhoule (approved):
:fire:

PR #2059, opened 2024-08-27
feat: scope entity cache entries by auth directives

obmarg:
When we're doing entity caching we need to consider whether the entity that we are caching was contained inside a type or field with `@authenticated` or `@requiresScopes`.  If it was we should include that fact in the cache key, so that we don't share a cache entry from an authenticated user with that of an unauthed user etc.

Fixes GB-7179

### DIFF
--- a/engine/crates/engine-v2/src/operation/cache_scopes.rs
+++ b/engine/crates/engine-v2/src/operation/cache_scopes.rs
@@ -1,0 +1,192 @@
+use std::collections::HashSet;
+
+use schema::{RequiredScopeSetIndex, RequiredScopesId, RequiredScopesWalker, TypeSystemDirectivesWalker};
+
+use super::{LogicalPlanId, OperationPlan, OperationWalker, PlanWalker, SelectionSetWalker};
+
+#[derive(Clone, Copy, Debug, serde::Serialize, serde::Deserialize)]
+pub struct CacheScopeId(u16);
+
+impl<'a, Item, SchemaItem> PlanWalker<'a, Item, SchemaItem> {
+    /// The cache scopes that should be applied to the current plans cache
+    pub fn cache_scopes(self) -> CacheScopes<'a> {
+        CacheScopes {
+            walker: self.walk_with((), ()),
+            current_index: self
+                .operation
+                .logical_plan_cache_scopes
+                .binary_search_by(|(plan_id, _)| plan_id.cmp(&self.logical_plan_id))
+                .unwrap_or(self.operation.logical_plan_cache_scopes.len()),
+        }
+    }
+}
+
+pub struct CacheScopes<'a> {
+    walker: PlanWalker<'a>,
+    current_index: usize,
+}
+
+impl<'a> Iterator for CacheScopes<'a> {
+    type Item = CacheScope<'a>;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        let current = self
+            .walker
+            .operation
+            .logical_plan_cache_scopes
+            .get(self.current_index)?;
+
+        if current.0 != self.walker.logical_plan_id {
+            return None;
+        }
+
+        self.current_index += 1;
+
+        Some(match self.walker.operation.cache_scopes[current.1 .0 as usize] {
+            CacheScopeRecord::Authenticated => CacheScope::Authenticated,
+            CacheScopeRecord::RequiresScopes(required_scopes_id) => CacheScope::RequiresScopes(RequiredScopeSet {
+                walker: self.walker.schema_walker.walk(required_scopes_id),
+                scope_set: self
+                    .walker
+                    .query_modifications
+                    .selected_scope_set(required_scopes_id)
+                    .expect("a scope set to be selected"),
+            }),
+        })
+    }
+}
+
+pub enum CacheScope<'a> {
+    Authenticated,
+    RequiresScopes(RequiredScopeSet<'a>),
+}
+
+pub struct RequiredScopeSet<'a> {
+    walker: RequiredScopesWalker<'a>,
+    scope_set: RequiredScopeSetIndex,
+}
+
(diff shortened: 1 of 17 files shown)


engine/crates/integration-tests/tests/federation/entity_caching/directive_scopes.rs
```diff
@@ -0,0 +1,187 @@
...
+        let engine = engine().await;
+
+        const AUTHENTICATED_QUERY: &str = "{ authenticatedProducts { upc reviews { id body } } }";
+        const UNAUTHENTICATED_QUERY: &str = "{ topProducts { upc reviews { id body } } }";
+
+        engine
+            .post(AUTHENTICATED_QUERY)
+            .header("Authorization", format!("Bearer {token}"))
+            .await
+            .into_data();
```
Finistere, line 26: why `into_data()` though?

engine/crates/engine-v2/src/operation/modifier/query.rs
```diff
@@ -20,6 +20,7 @@ pub(crate) struct QueryModifications {
     pub concrete_shape_has_error: BitSet<ConcreteObjectShapeId>,
     pub field_shape_id_to_error_ids: IdToMany<FieldShapeId, ErrorId>,
     pub root_error_ids: Vec<ErrorId>,
+    selected_scope_sets: Vec<(RequiredScopesId, RequiredScopeSetIndex)>,
```
Finistere, line 23: Would have used `matched_scopes` rather than `selected_scopes` I think

Finistere (approved):

engine/crates/engine-v2/src/operation/cache_scopes.rs
```diff
@@ -0,0 +1,192 @@
...
+            );
+        }
+
+        builder.pop_scopes(scopes_added);
+    }
+
+    builder.pop_scopes(parent_entity_scopes_added);
+}
+
+fn add_directive_scopes(directives: TypeSystemDirectivesWalker<'_>, builder: &mut CacheScopeBuilder) -> usize {
```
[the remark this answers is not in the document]
obmarg: 🤷 I don't think it matters tbqh

engine/crates/integration-tests/tests/federation/entity_caching/directive_scopes.rs
obmarg: Why not into_data?

Finistere: I'm just confused as to why you're calling this method but not doing anything with the result

obmarg: Because it has an `assert!(errors == None)` in it

engine/crates/engine-v2/src/operation/cache_scopes.rs
```diff
@@ -0,0 +1,192 @@
...
+        if Some(parent_entity_id) != last_parent_entity_id {
+            last_parent_entity_id = Some(parent_entity_id);
+            builder.pop_scopes(parent_entity_scopes_added);
+            parent_entity_scopes_added = add_directive_scopes(definition.parent_entity().directives(), builder)
+        }
+
+        let scopes_added = add_directive_scopes(definition.directives(), builder);
+
+        let current_plan_id = operation_plan.plan_id_for_field(field.item);
+        if parent_plan_id != Some(current_plan_id) {
```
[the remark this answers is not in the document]
obmarg: Yeah, the `plans_handled` stuff was written before I realised that `fields` was ordered by parent entity.  I'm still kinda into keeping it anyway though tbh, unless it proves problematic

engine/crates/integration-tests/tests/federation/entity_caching/directive_scopes.rs
Finistere: might be worth adding a method that only does that explicitly then imho. Were I to touch that code I might remove `into_data()` not understanding why it's there to be honest, no idea.
I tend to always `insta` snapshot the result to ensure I did receive what I expected.

obmarg: Maybe don't remove things you don't understand?

obmarg: I'm really not _that_ keen on adding another method that does basically the same thing as into_data

Finistere: > Maybe don't remove things you don't understand?

Well yeah, it's a bit extreme what I said. It's more the opposite, someone can change `into_data()` to not have the side-effect anymore. So I'm not a fan of relying on a side-effect of a method with a name that has no relation to the side-effect you expect.

obmarg: Why would someone remove an assert from commonly used test code?

Finistere: 🤷. My reasoning is just that the fact that I had to ask why you did it or that I would have to look at the definition to understand why you're doing this is reason enough for me to rename it to something more explicit. If you think that's useless then just leave it as is 🤷

Finistere: fair enough, I just use insta snapshots everywhere. Sometimes a bit tedious, but I find problems to be easy to spot and understand without having to touch the code.

obmarg: Yeah, I normally would use insta.  Just for these tests I don't really give a shit about the contents of the response, I just want to make sure the subgraph was called the correct number of times.

PR #2062, opened 2024-08-28
feat: add wasi hooks for access logs

pimeys:
This does the wasi side of the story. More PRs to follow. The WIT is already different in the coming PRs, so any changes proposed here will be implemented in the coming PRs.

Yeah it is huge. Mostly due to bindings.rs changes which you can skip.


engine/crates/wasi-component-loader/src/context.rs
```diff
@@ -1,23 +1,85 @@
...
+pub type ChannelLogSender = crossbeam::channel::Sender<AccessLogMessage>;
+
+/// A receiver for the logger to receive messages and write them somewhere.
+pub type ChannelLogReceiver = crossbeam::channel::Receiver<AccessLogMessage>;
+
+/// https://github.com/tokio-rs/tracing/blob/master/tracing-appender/src/non_blocking.rs#L61-L70
+const DEFAULT_BUFFERED_LINES_LIMIT: usize = 128_000;
+
+/// Creates a new channel for access logs.
+pub fn create_log_channel() -> (ChannelLogSender, ChannelLogReceiver) {
```
Finistere, line 28: the `lossy_log` boolean flag should be included in the Sender type IMO, so have both:
```
create_lossy_log_channel()-> (ChannelLogSender, ChannelLogReceiver)
create_log_channel()-> (ChannelLogSender, ChannelLogReceiver)
```
with
```
struct ChannelLogSender {
  sender: crossbeam::channel::Sender<AccessLogMessage>
  lossy: bool
}
```

I don't find it great that this boolean flag is propagated at various places

engine/crates/runtime-local/src/hooks.rs
```diff
@@ -15,22 +15,32 @@ use runtime::{
...
     gateway: Pool<GatewayComponentInstance>,
     authorization: Pool<AuthorizationComponentInstance>,
     subgraph: Pool<SubgraphComponentInstance>,
     hook_latencies: Histogram<u64>,
+    sender: ChannelLogSender,
+    lossy_log: bool,
 }
 
 impl HooksWasiInner {
+    pub fn shared_context(&self, kv: &Arc<HashMap<String, String>>) -> SharedContext {
```
Finistere, line 40: I would have just made:
```
#[derive(Clone)]
struct Context {
  kv: Arc<HashMap<String, String>>>,
  sender: ChannelLogSender
}
```
You can add the sender after calling the wasm `on_gateway_request` function, no need for `SharedContext` vs `Context` imo

engine/crates/integration-tests/src/federation/builder/test_runtime.rs
```diff
@@ -18,11 +18,14 @@ pub struct TestRuntime {
...
     pub rate_limiter: runtime::rate_limiting::RateLimiter,
     pub entity_cache: InMemoryEntityCache,
+    pub access_log_sender: runtime_local::hooks::ChannelLogSender,
+    pub access_log_receiver: runtime_local::hooks::ChannelLogReceiver,
 }
 
 impl Default for TestRuntime {
     fn default() -> Self {
         let (_, rx) = watch::channel(Default::default());
+        let (access_log_sender, access_log_receiver) = runtime_local::hooks::create_log_channel();
```
Finistere, line 28: I don't think I would put it here, `TestRuntime` will be owned by Engine. So you lose the receiver, the receiver should be added to `TestEngineV2` instead so it's accessible inside tests.

engine/crates/runtime-local/src/hooks.rs
Finistere: well saying that but I will with authorized at some point... so well, but it's in the future

pimeys: Yeah but this will annoyingly change the interface to the outside a bit too much to my liking. I like that the implementation adds the channel if needed, but the outside doesn't really get it back. I tried that first and I seriously didn't like how it looked.

pimeys: You just have to believe me for now this approach is better. We can reconsider when the feature is working and the big picture is more clear!

PR #2064, opened 2024-08-29
graphql-federated-graph: fix rendering of @join__type on interfaces without keys

tomhoule:
When you start the gateway with `--schema` and federated SDL that contains an interface with an `@join__type` directive _without keys argument_, the gateway fails to start with the following error:

```
Error: internal error: Failed to generate engine Schema:  --> 1:4
  |
1 | {  }
  |    ^---
  |
  = expected selection
```

This error comes from `graphql_federated_graph::from_sdl()`, because we try to parse an empty string as a selection set.

The SDL that triggers the error when generating the engine Schema contains something like:

```
@join__type(graph: MY_GRAPH, key: "")
```

which is not valid. But the original SDL passed to the engine contains something like:

```
@join__type(graph: MY_GRAPH)
```

which is valid.

**How does that happen?**

The root cause is the following: `graphql_federated_graph::from_sdl()` will populate a `Key` in the `FederatedGraph` for each `@join__type` directive. The key has a selection, but in the absence of a `key` argument on the directive, that selection is empty. When we re-render that graph to SDL with
`graphql_federated_graph::render_federated_sdl()`, the rendered SDL a key argument with an empty string as a selection. When that is passed again to `from_sdl()`, we run into the error that ends up being displayed.

This error on startup is a combination of a recent change and a bug fix that missed a scenario.

The recent change is #2055. The relevant part is that we now store SDL in the `engine_v2::Config` struct instead of a structured representation of the FederatedGraph. This change should not be relevant to the gateway, but it introduced the conditions for the bug to manifest: we parse the SDL to a `FederatedGraph` to construct `Config`, then render that same `FederatedGraph` to federated SDL to store in that same `Config`. When we later instantiate the `engine_v2::Schema` from that config, we call `from_sdl()` on that SDL, completing the `from_sdl() -> render federated_sdl() -> from_sdl()` sequence that triggers the error. Ideally in the future, for the gateway, we would keep the same FederatedGraph and avoid re-instantiating.

The bug fix is f34d0bd. That fixed the empty `key` on `@join__type` on objects, but not interfaces.

**The fix**

The changes in this pull request are three fold:

1. The error message for this kind of errors will be slightly more helpful. It now looks like this:
```
      Error: Error validating federated SDL: Error parsing a selection from a federated directive:  --> 1:4
     |
   1 | {  }
     |    ^---
     |
     = expected selection
```

2. In `render_federated_sdl()`, we no longer render empty strings in `key` on `@join__type` on interfaces. That completes the bug fix in f34d0bd.
3. In the gateway, we stop rendering and re-parsing the rendered SDL. Instead, we pass the original SDL into the `engine_v2::Config`.

Any of the last two changes on their own fixes the observed bug, but combining both introduces some defense in depth. If there is another bug in `render_federated_sdl()`, 3. will prevent it from affecting gateway start up.

closes GB-7322

### DIFF
--- a/cli/crates/federated-dev/src/dev.rs
+++ b/cli/crates/federated-dev/src/dev.rs
@@ -49,9 +49,10 @@ pub(super) async fn run(
     log::trace!("starting the federated dev server");
 
     let (gateway_sender, gateway) = watch::channel(
-        gateway_nanny::new_gateway(
-            graph.map(|graph| engine_config_builder::build_with_sdl_config(&config.borrow(), graph)),
-        )
+        gateway_nanny::new_gateway(graph.map(|graph| {
+            let federated_sdl = graphql_federated_graph::render_federated_sdl(&graph).unwrap();
+            engine_config_builder::build_with_sdl_config(&config.borrow(), graph, federated_sdl)
+        }))
         .await,
     );
     let (websocket_sender, websocket_receiver) = mpsc::channel(16);
--- a/cli/crates/federated-dev/src/dev/gateway_nanny.rs
+++ b/cli/crates/federated-dev/src/dev/gateway_nanny.rs
@@ -36,11 +36,10 @@ impl EngineNanny {
 
         while let Some(message) = stream.next().await {
             log::trace!("nanny received a {message:?}");
-            let config = self
-                .graph
-                .borrow()
-                .clone()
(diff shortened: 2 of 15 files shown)


pimeys (approved):
supah

tomhoule:
re:

>  Ideally in the future, for the gateway, we would keep the same FederatedGraph and avoid re-instantiating.

I discussed that with @Finistere , and I'll make it happen today after the patch release.

PR #2066, opened 2024-08-29
federated-graph: start decluttering render_federated_sdl()

tomhoule:
While working on this morning's fix, I noticed this module could really do with some love. Here's what this PR starts to address:

- Align render_federated_sdl() and render_api sdl() on `fmt::Formatter` based APIs instead of writing directly to a `String`.
- Reduce the number of different ways we render directives. This PR introduces a `DirectiveWriter` abstraction that should work in the whole crate.
- Avoid the proliferation of types that only hold FederatedGraph, an item and implement Display. Functions are more concise and easy to follow.

These changes should not affect the output in any way. We have pretty extensive tests in the graphql-federated-graph and graphql-composition crates so I am fairly confident that this is the case.

### DIFF
--- a/engine/crates/federated-graph/src/render_sdl/display_utils.rs
+++ b/engine/crates/federated-graph/src/render_sdl/display_utils.rs
@@ -1,11 +1,27 @@
 use crate::*;
-use std::{
-    fmt::{self, Display, Write},
-    iter,
-};
+use std::fmt::{self, Display, Write};
 
 pub(super) const BUILTIN_SCALARS: &[&str] = &["ID", "String", "Int", "Float", "Boolean"];
 pub(super) const INDENT: &str = "    ";
+
+/// Lets you take a routine that expects a formatter, and use it on a string.
+pub(in crate::render_sdl) fn with_formatter<F>(out: &mut String, action: F) -> fmt::Result
+where
+    F: Fn(&mut fmt::Formatter<'_>) -> fmt::Result,
+{
+    struct Helper<T>(T);
+
+    impl<T> Display for Helper<T>
+    where
+        T: Fn(&mut fmt::Formatter<'_>) -> fmt::Result,
+    {
+        fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+            (self.0)(f)
+        }
+    }
+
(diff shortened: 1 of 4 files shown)


engine/crates/federated-graph/src/render_sdl/display_utils.rs
```diff
@@ -1,12 +1,28 @@
...
-    fmt::{self, Display, Write},
-    iter,
-};
+use std::fmt::{self, Display, Write};
 
 pub(super) const BUILTIN_SCALARS: &[&str] = &["ID", "String", "Int", "Float", "Boolean"];
 pub(super) const INDENT: &str = "    ";
 
+/// Lets you take a routine that expects a formatter, and use it on a string.
+pub(in crate::render_sdl) fn with_formatter<F>(out: &mut String, action: F) -> fmt::Result
```
tomhoule, line 8: This is fairly basic, there should be something standard to do that, am I forgetting something?

Finistere: aren't you re-implementing https://doc.rust-lang.org/std/macro.write.html ?

tomhoule: It's basically a version of `write!()` that instead of `Display` arguments, takes a function that takes a `Formatter`.

Finistere: >that instead of Display arguments, takes a function that takes a Formatter.

Isn't that the definition of `impl std::fmt::Display`?

Finistere: ok got it

tomhoule: Basically I could use `write()` with small structs that implement `Display`, but I'd rather write a function than a struct + a display impl if it's going to happen 20 times in the same module.

tomhoule: That function will become unnecessary once render_federated_sdl() works like render_api_sdl(), with just one Display impl at the top.

Finistere (approved):

PR #2069, opened 2024-08-29
feat: plug new hook to the runtime with the channel logger

pimeys:
This is going to plug in the subgraph hook to the access logs. Now it's time to say a word about the wit, but only for subgraphs. This also implements a "kind of I think this is how it should look like version" of the access log writer.

### DIFF
--- a/engine/crates/engine-v2/src/execution/hooks/responses.rs
+++ b/engine/crates/engine-v2/src/execution/hooks/responses.rs
@@ -1,0 +1,39 @@
+use runtime::hooks::{
+    ExecutedGatewayRequest, ExecutedHttpRequest, ExecutedSubgraphRequest, Hooks, Operation, ResponseHooks,
+};
+use tracing::{instrument, Level};
+
+use crate::response::GraphqlError;
+
+impl<'ctx, H: Hooks> super::RequestHooks<'ctx, H> {
+    #[instrument(skip_all, ret(level = Level::DEBUG))]
+    pub async fn on_subgraph_response(&self, request: ExecutedSubgraphRequest<'_>) -> Result<Vec<u8>, GraphqlError> {
+        self.hooks
+            .responses()
+            .on_subgraph_response(self.context, request)
+            .await
+            .map_err(Into::into)
+    }
+
+    #[instrument(skip_all, ret(level = Level::DEBUG))]
+    pub async fn on_gateway_response(
+        &self,
+        operation: Operation<'_>,
+        request: ExecutedGatewayRequest,
+    ) -> Result<Vec<u8>, GraphqlError> {
+        self.hooks
+            .responses()
+            .on_gateway_response(self.context, operation, request)
+            .await
+            .map_err(Into::into)
+    }
+
+    #[instrument(skip_all, ret(level = Level::DEBUG))]
+    pub async fn on_http_response(&self, request: ExecutedHttpRequest<'_>) -> Result<(), GraphqlError> {
+        self.hooks
+            .responses()
+            .on_http_response(self.context, request)
+            .await
+            .map_err(Into::into)
+    }
+}
--- a/engine/crates/engine-v2/src/sources/graphql/subscription.rs
+++ b/engine/crates/engine-v2/src/sources/graphql/subscription.rs
@@ -1,17 +1,23 @@
 use std::borrow::Cow;
 
 use bytes::Bytes;
-use futures::TryStreamExt;
+use futures::{Future, TryStreamExt};
 use futures_util::{stream::BoxStream, StreamExt};
 use headers::HeaderMapExt;
-use runtime::fetch::{FetchRequest, Fetcher};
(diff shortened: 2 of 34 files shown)


gateway/crates/federated-server/src/server/access_logs.rs
```diff
@@ -0,0 +1,43 @@
...
+                    if let Err(e) = log.write_all(&data) {
+                        tracing::error!(target: GRAFBASE_TARGET, "error writing to access log: {e}");
+                    }
+                }
+                AccessLogMessage::Shutdown(guard) => {
+                    if let Err(e) = log.flush() {
+                        tracing::error!(target: GRAFBASE_TARGET, "error flushing access log: {e}");
+                    }
+
+                    drop(guard);
```
pimeys, line 36: This will message to graceful shutdown that we've written all lines to the file.

engine/crates/engine-v2/src/execution/hooks/responses.rs
```diff
@@ -0,0 +1,39 @@
...
+    pub async fn on_subgraph_response(&self, request: ExecutedSubgraphRequest<'_>) -> Result<Vec<u8>, GraphqlError> {
+        self.hooks
+            .responses()
+            .on_subgraph_response(self.context, request)
+            .await
+            .map_err(Into::into)
+    }
+
+    #[instrument(skip_all, ret(level = Level::DEBUG))]
+    pub async fn on_gateway_response(
```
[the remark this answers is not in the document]
Finistere: Should be something like `on_operation_response` if it's called for each operation

engine/crates/runtime/src/hooks.rs
```diff
@@ -115,6 +123,191 @@ pub trait SubgraphHooks<Context>: Send + Sync + 'static {
...
+        ResponseInfoBuilder {
+            start: SystemTime::now(),
+            connection_time: None,
+            response_time: None,
+        }
+    }
+}
+
+pub struct ResponseInfoBuilder {
+    start: SystemTime,
```
Finistere, line 153: `Instant` would be more appropriate for a duration, it has the convenient function `elapsed()` and it's guaranteed to not be negative

engine/crates/runtime-local/src/fetch.rs
```diff
@@ -36,6 +47,8 @@ impl Fetcher for NativeFetcher {
         *response.extensions_mut() = extensions;
         *response.headers_mut() = headers;
 
+        response.extensions_mut().insert(info.finalize(status.as_u16()));
```
Finistere, line 50: I wouldn't pass it through `extensions` imo, why go through the AnyMap when you can return a tuple?

engine/crates/runtime/src/hooks.rs
```diff
@@ -39,6 +40,11 @@ impl<'a, T> Anything<'a> for T where T: serde::Serialize + serde::de::Deserializ
 pub type AuthorizationVerdict = Result<(), PartialGraphqlError>;
 pub type AuthorizationVerdicts = Result<Vec<AuthorizationVerdict>, PartialGraphqlError>;
 
+pub struct ChannelBuilder<'a, Sender: Clone> {
```
Finistere, line 43: Why Is `ChannelBuilder` here? It's a hooks wasi implementation detail, so should be in `runtime-local`

gateway/crates/federated-server/src/server.rs
```diff
@@ -229,6 +249,11 @@ async fn graceful_shutdown(handle: axum_server::Handle) {
 
     tracing::info!(target: GRAFBASE_TARGET, "Shutting down gracefully...");
     handle.graceful_shutdown(Some(std::time::Duration::from_secs(3)));
+
+    let wg = WaitGroup::new();
```
Finistere, line 253: Should be done at the same place `server_runtime.graceful_shutdown()` is called. The `handle.graceful_shutdown` does not wait for requests to finish, in only sends a signal. So you have to add it _after_ the `bind(...).await`

engine/crates/engine-v2/src/sources/graphql/subscription.rs
```diff
@@ -186,3 +192,73 @@ impl GraphqlResolver {
         Ok(Box::pin(stream))
     }
 }
+
+async fn retrying_fetch<'ctx, R: Runtime, F, T>(
```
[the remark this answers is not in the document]
pimeys: Yeah, but the info is stored into http extensions. The subscription side doesn't return http::Response... The durations are calculated in the fetcher, not in this function.

engine/crates/engine-v2/src/execution/hooks/responses.rs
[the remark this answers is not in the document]
pimeys: Yeah, will rename this in the next PR.

Finistere (approved):

PR #2071, opened 2024-08-29
chore: Improve docker build time

Finistere:
To read a full explanation: https://www.tweag.io/blog/2023-08-10-rust-static-link-with-mimalloc/

It's an old problem that profoundly bothered me. As Graeme started taking a look at build times in general, got me curious again as to why cargo builds were so much slower within a docker build than without on my machine. Almost an order of magnitude.

The short version is that the musl allocator behaves poorly in multi-threaded scenarios. So the author patched musl to use mimalloc. Now on my machine build times are comparable to those without docker instead of spending an absurd amount of time in kernel threads.

|               | Gateway | CLI   |
|---------------|---------|-------|
| base          | 15m30   | 19m50 |
| mimalloc musl | 5m30    | 6m35  |
| 8 -> 16 cores | 4m      | 4m40  |

### DIFF
--- a/.github/workflows/binaries.yml
+++ b/.github/workflows/binaries.yml
@@ -158,7 +158,7 @@ jobs:
   docker-grafbase:
     needs: [test]
     if: ${{ needs.detect-change-type.outputs.build == 'true' || startsWith(github.ref, 'refs/tags/') }}
-    runs-on: ubicloud-standard-8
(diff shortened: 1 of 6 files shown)


obmarg (approved):
Excellent find

tomhoule (approved):
Awesome!

PR #2072, opened 2024-08-30
Initial implementation of patching in graphql-schema-diff

tomhoule:
Until now, you could diff two GraphQL schemas with graphql-schema-diff, and get the locations that changed, and what kind of change occurred. In https://github.com/grafbase/grafbase/pull/2014, spans were added to the changes to point out the exact content of the change, as well as a way to resolve these spans to strings.

This PR implements patching with a `patch()` function that takes a diff, the resolved spans for that diff, and a source schema, and applies the diff to the schema. The format is intentionally designed to be easy to store in a database table.

The implementation is not complete, there are missing pieces:

- directive definitions
- directives
- field arguments
- type extensions
- descriptions

I opted against finishing all of it now so we can make progress on the features we are planning and fill in the gaps later. Also to get feedback on patching earlier.


engine/crates/graphql-schema-diff/src/patch/type_definitions.rs
```diff
@@ -0,0 +1,318 @@
...
+
+        schema.push_str(": ");
+
+        if let Ok(idx) = changed_field_types.binary_search_by(|(field_name, _)| field_name.cmp(&field.name())) {
+            schema.push_str(changed_field_types[idx].1);
+        } else {
+            schema.push_str(&field.ty().to_string());
+        }
+
+        schema.push('\n');
```
obmarg, line 125: Are directives & defaults intentionally missing here?  Assume so, just checking

obmarg (approved):

engine/crates/graphql-schema-diff/src/patch/type_definitions.rs
```diff
@@ -0,0 +1,318 @@
...
+                schema.push_str(argument.name());
+                schema.push_str(": ");
+                schema.push_str(&argument.ty().to_string());
+
+                if argument.default_value().is_some() {
+                    schema.push(' ');
+                    let span = argument.default_value_span();
+                    schema.push_str(&paths.source()[span.start..span.end]);
+                }
+
```
obmarg, line 204: Argument directives also intentionally missing here?

engine/crates/graphql-schema-diff/src/patch/type_definitions.rs
tomhoule: Yes I have issues for these.

engine/crates/graphql-schema-diff/src/patch/type_definitions.rs
tomhoule: Same as fields, it won't be complete (diffing for directives isn't there yet), but I'll add the same level of support as in other places, good catch.

PR #2074, opened 2024-08-30
chore: disable incremental builds on CI

obmarg:
Incremental builds don't serve much purpose on CI: they're mostly to speed up builds for devs when you change a single file and don't want to recompile the whole crate.  On CI we are more or less always going to be building the whole crate so disabling it saves a little bit of time, and should reduce the size of our `/target` directory, helping a bit with caching.

Hard to be scientific since our CI runs are quite variable, but this does appear to shave 2-3 minutes off a full CI run (most obviously on the windows build)


pimeys (approved):

obmarg:
Actually looks like the action we use to install rust already sets this - so not entirely sure how I saw a 3-4 minute improvement in times...

Going to keep this anyway though - can't hurt

PR #2075, opened 2024-08-30
chore: remove built

obmarg:
We use the crate `built` to embed git hash & build time info into the engine-v2-schema crate.  However, this pulls in libgit2 which appears to be the #2 consumer of compile time on every platform on CI. 

Specifically:

- Windows spends a full 2 minutes compiling libgit2
- aarch-darwin spends 40 seconds compiling libgit2
- x86-64 linux spends 20 seconds

I can't be bothered getting the rest of them, but you get the idea.  It's pretty trivial to get this info without building libgit2, so this PR does that.

Interestingly this had little impact on the overall time of a CI run - although I wonder if that was just CI being slow for that individual run.  Going to see what happens on another run before I mark this as ready for review.


obmarg:
Note to self: is this actually an improvement?  Most of the actions runs I've seen appear slightly slower (that may just be random CI variance, needs investigation)

Finistere (approved):

PR #2077, opened 2024-08-30
feat: access logs hook for operation layer

pimeys:


engine/crates/engine-v2/src/engine/execute/single.rs
```diff
@@ -59,29 +62,56 @@ impl<R: Runtime> Engine<R> {
...
                 "Subscriptions are only suported on streaming transports. Try making a request with SSE or WebSockets",
-            )
+            );
+
+            (Some(metrics_attributes), None, response)
         } else {
-            self.execute_query_or_mutation(operation_plan).await
-        };
+            if let Some(ref name) = metrics_attributes.name {
+                operation_info.set_name(name.clone());
```
pimeys, line 95: This damn call here requires by-value transfer of the name. I tried to fight it a while, but decided just to pass this as a string to get it compiling.

engine/crates/runtime/src/hooks.rs
```diff
@@ -248,27 +248,92 @@ impl<'a> ExecutedSubgraphRequest<'a> {
     }
 }
 
-#[derive(Debug, Clone, Copy)]
+#[derive(Debug, Clone)]
 pub struct Operation<'a> {
-    pub name: Option<&'a str>,
+    pub name: Option<String>,
```
pimeys, line 253: See the other comment on lifetime issues. Would be great to have this as `Option<&'a str>`, but it just doesn't compile easily with that. Maybe `&'a Option<String>` could work, but that also feels a bit sus.

Finistere: won't work, you have no choice but top copy that one

engine/crates/engine-v2/src/engine/execute/single.rs
```diff
@@ -59,29 +62,56 @@ impl<R: Runtime> Engine<R> {
...
     }
 }
 
 impl<'ctx, R: Runtime> PreExecutionContext<'ctx, R> {
-    async fn execute_single(mut self, request: Request) -> (Option<OperationMetricsAttributes>, Response) {
-        let operation_plan = match self.prepare_operation(request).await {
+    async fn execute_single(
+        mut self,
+        request: Request,
+    ) -> (Option<OperationMetricsAttributes>, Option<Vec<u8>>, Response) {
```
Finistere, line 77: I would strongly consider here and before returning a struct. I would be fine with a tuple if we have a newtype, but a `Vec<u8>` without a field name is very ambiguous

engine/crates/engine-v2/src/engine/execute/prepare.rs
```diff
@@ -17,9 +18,10 @@ impl<'ctx, R: Runtime> PreExecutionContext<'ctx, R> {
     pub(super) async fn prepare_operation(
         &mut self,
         request: Request,
+        operation_info: &mut OperationBuilder,
```
Finistere, line 21: Would name those `OperationHookInfoBuilder` or smth like this imho. `OperationBuilder` is quite ambiguous here

Finistere (approved):

PR #2079, opened 2024-08-30
feat(gateway): better logs part 2

Finistere:
By default "debug" and "trace" produce way too much information with h2,
hyper etc. So forcing them to stay at info. I'm also adding the
possibility to fully customize the log filter with a custom input. The
goal is to remove `__GRAFBASE_RUST_LOG` and be able to adjust the log
levels as needed. This will also help figuring out the best default
values.

### DIFF
--- a/gateway/crates/gateway-binary/src/args/std.rs
+++ b/gateway/crates/gateway-binary/src/args/std.rs
@@ -46,11 +46,18 @@ pub struct Args {
     /// to the Grafbase API.
     #[arg(long, short, env = "GRAFBASE_SCHEMA_PATH")]
     pub schema: Option<PathBuf>,
-    /// Set the logging level
-    #[arg(long = "log", env = "GRAFBASE_LOG", default_value_t = LogLevel::default())]
-    pub log_level: LogLevel,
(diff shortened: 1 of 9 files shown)


gateway/crates/gateway-binary/src/args/std.rs
```diff
@@ -100,7 +107,7 @@ impl super::Args for Args {
 
         if let Some((token, graph_ref)) = self.grafbase_access_token.as_ref().zip(self.graph_ref.as_ref()) {
             config.telemetry.grafbase = Some(OtlpExporterConfig {
-                endpoint: "https://otel.grafbase.com".parse().unwrap(),
+                endpoint: "https://grafbase-telemetry-sink-prod.fly.dev ".parse().unwrap(),
```
tomhoule, line 110: I don't get the reason for this change

tomhoule (approved):
:100:

gateway/crates/gateway-binary/src/args/std.rs
Finistere: thanks

PR #2082, opened 2024-09-02
feat: http layer hook for access logs

pimeys:
The last big PR for access logs. This one has tests, which add to the total line count sadly.

- refactor tower layers to use async/await
- plug the http response hook to a new tower layer
- test the whole pipeline
- add a metric counter for access log queue size
- clean up wit definitions
- add a method to the shared context for trace id


engine/crates/wasi-component-loader/gateway-hooks.wit
```diff
@@ -118,14 +122,14 @@ interface types {
     }
 
     // Info about an executed gateway operation.
-    record executed-gateway-request {
+    record executed-operation-request {
```
Finistere, line 125: just `executed-operation`

engine/crates/wasi-component-loader/src/context.rs
```diff
@@ -19,22 +24,24 @@ use crate::{
...
+            match self.sender.try_send(data) {
+                Ok(_) => (),
+                Err(TrySendError::Full(AccessLogMessage::Data(data))) => return Err(LogError::ChannelFull(data)),
+                Err(_) => return Err(LogError::ChannelClosed),
             }
         } else if self.sender.send(data).is_err() {
             return Err(LogError::ChannelClosed);
         }
 
+        self.log_queue_length_counter.add(1, &[]);
```
Finistere, line 43: You don't seem to be decreasing it anywhere.

gateway/crates/federated-server/src/server.rs
```diff
@@ -163,7 +179,10 @@ pub async fn serve(
 
     // Once all pending requests have been dealt with, we shutdown everything else left (telemetry, logs)
     server_runtime.graceful_shutdown().await;
-    access_log_sender.graceful_shutdown().await;
+
+    if config.gateway.access_logs.enabled {
```
Finistere, line 183: could be a `join!`

engine/crates/wasi-component-loader/src/context.rs
Finistere: I'm not even sure if it works as you access it by name in the federated server. IMO It would be better to include the counter inside a `ChannelLogReceiver` rather than relying on globals

gateway/crates/federated-server/src/server/access_logs.rs
```diff
@@ -1,11 +1,15 @@
...
 
 use gateway_config::{AccessLogsConfig, RotateMode};
-use grafbase_telemetry::span::GRAFBASE_TARGET;
+use grafbase_telemetry::metrics::meter_from_global_provider;
 use runtime_local::hooks::{AccessLogMessage, ChannelLogReceiver};
 use tracing_appender::rolling::{RollingFileAppender, Rotation};
 
 pub(crate) fn start(config: &AccessLogsConfig, access_log_receiver: ChannelLogReceiver) -> crate::Result<()> {
+    let log_queue_length = meter_from_global_provider()
+        .i64_up_down_counter("grafbase.gateway.access_log.pending")
```
Finistere, line 10: IMHO, too much magic for my taste, supposing that works, to rely on globals to get the same counter, should be part of the `ChannelLogReceiver`

Finistere (approved):

engine/crates/wasi-component-loader/src/context.rs
pimeys: There is a test for this, it works. I'll let you hunt where the decreasing happens :)

engine/crates/engine-v2/src/execution/hooks/responses.rs
```diff
@@ -29,7 +29,7 @@ impl<'ctx, H: Hooks> super::RequestHooks<'ctx, H> {
     }
 
     #[instrument(skip_all, ret(level = Level::DEBUG))]
-    pub async fn on_http_response(&self, request: ExecutedHttpRequest<'_>) -> Result<(), GraphqlError> {
```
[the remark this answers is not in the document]
pimeys: Wait, whut?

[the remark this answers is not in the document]
Finistere: it's not called in the engine

PR #2083, opened 2024-09-03
Use web_time::Instant instead of Instant from std in runtime::hooks

tomhoule:
Otherwise the engine stops working on wasm32-unknown-unknown


obmarg (approved):

jakubadamw (approved):
Lovely

PR #2090, opened 2024-09-05
chore: a new workflow for rust prs

obmarg:
Our current GHA workflow for building all of our rust is quite slow: pretty much always over 30 minutes, often closer to 40.  This is far too long, so I've been working to fix it.  There's two main angles I've been tackling it from:

1. Trying to only build what's actually changed.
    1. I've written a small rust app, [what-rust-changed](https://github.com/grafbase/what-rust-changed) that can take the set of changes from git and map that into a list of packages that have changed.  This also correctly takes `Cargo.lock` changes into account
    2. The new workflow I've added uses this to control what it builds & tests.
2. Trying to do less things _generally_.  There's a lot of things our current setup does that I'd argue don't need to be done on every PR.

The result is a workflow that runs in just shy of 20 minutes when building everything - and should be less if you only change a subset of the repo.  I have further changes that shave another 5ish minutes off this, but I'll do those in a follow up PR.

Some notes:

- This doesn't yet replace the existing workflows for `main` builds - I'd like to get confidence in this setup before doing that.
- It also doesn't replace release builds yet - they do actually require half the things I've removed.  We could look into optimising once we have confidence in the new appproach.
- I've avoided using actions to deduplicate like we do in the existing workflows.  Actions doesn't give you timing info on sub-steps of actions so doing this makes it kinda difficult to monitor & improve things.
- I've tried to make the jobs as homogenous as possible (rather than our current setup of specific runs for each platform), because this kind of selective CI run stuff is just easier to do that way.

It could be worth examining our dependency tree after this work, since we now only build a subset of the repo we can speed up builds further by uncoupling dependencies from each other.  An obvious improvement would be avoiding pulling `engine` into  `engine-v2` but I'm sure there's others.

### DIFF
--- a/.github/workflows/rust-prs.yml
+++ b/.github/workflows/rust-prs.yml
@@ -153,10 +153,10 @@ jobs:
           sudo ln -s /bin/g++ /bin/aarch64-linux-musl-g++
 
       - name: Clippy
-        if: fromJson(needs.what-changed.outputs.rust).cargo-package-specs
+        if: fromJson(needs.what-changed.outputs.rust).cargo-build-specs
         shell: bash
         run: |
-          cargo clippy --target ${{ matrix.platform.target }} ${{ fromJson(needs.what-changed.outputs.rust).cargo-package-specs }}
+          cargo clippy --target ${{ matrix.platform.target }} ${{ fromJson(needs.what-changed.outputs.rust).cargo-build-specs }}
 
       - name: Install cargo binstall
         if: fromJson(needs.what-changed.outputs.rust).cargo-docker-test-specs
@@ -266,32 +266,6 @@ jobs:
           api_key: ${{ secrets.DATADOG_API_KEY }}
           junit_path: target/nextest/ci/junit.xml
           service: cli
-
-      # - name: Install cargo-nextest & libs linux
-      #   # if: ${{ startsWith(inputs.platform, 'linux') }}
-      #   shell: bash
-      #   run: |
-      #     curl -LsSf https://get.nexte.st/latest/${{ inputs.platform }} | tar zxf - -C ${CARGO_HOME:-~/.cargo}/bin
-      #     sudo apt-get install musl-tools
-
-      # Ok, so this is tricky.
-      # Need to:
-      # - Install all the udf/bun BS
-      # - Login to docker hub
-      # - Build WASI components if they are required by the package set
-      #   (TODO: Figure out some way to avoid doing this all the time maybe?)
-      # - Run tests _except gateway & docker tests
-      #   - Starting & stopping docker-compos if integration-tests are in the selected set.
-      #   - Upload junit
-      # - If gateway is in the build set
-      #   - Start docker comppose for gateway
-      #   - Run gateway tests
-      #   - Upload junit
-      # - name: Tests
-      #   shell: bash
-      #   run: |
-      #     # TODO: this is obviously a terrible name if I'm using clippy for the test targets...
-      #     cargo clippy ${{ fromJson(needs.what-changed.outputs.rust).cargo-package-specs }}
 
       - name: Build debug binaries
         if: fromJson(needs.what-changed.outputs.rust).cargo-bin-specs
--- a/.github/workflows/rust-prs.yml
+++ b/.github/workflows/rust-prs.yml
@@ -55,6 +55,7 @@ jobs:
         uses: dtolnay/rust-toolchain@master
         with:
           toolchain: ${{ env.RUST_VERSION }}
+          components: clippy, rustfmt
 
       - name: Install what-rust-changed
         shell: bash
@@ -95,6 +96,29 @@ jobs:
               - cli/Dockerfile
             gateway-docker:
               - gateway/Dockerfile
+
+  format:
+    if: fromJson(needs.what-changed.outputs.rust).changed-packages
+    needs: [what-changed]
+    runs-on: ubicloud-standard-2
+    steps:
+      - name: Get sources
+        uses: actions/checkout@v4
+
+      - name: Extract the Rust version to use from the `rust-toolchain.toml` file
+        shell: bash
+        run: |
+          rust_version=$(grep "channel" rust-toolchain.toml | cut -d "=" -f 2 | cut -d " " -f 2)
+          echo "RUST_VERSION=$rust_version" >> $GITHUB_ENV
+
+      - name: Install Rust
+        uses: dtolnay/rust-toolchain@master
+        with:
+          toolchain: ${{ env.RUST_VERSION }}
+
+      - name: cargo fmt
+        shell: bash
+        run: cargo fmt --check
 
   example-component:
     if: contains(fromJson(needs.what-changed.outputs.rust).changed-packages, 'integration-tests')
@@ -161,7 +185,7 @@ jobs:
   builds:
     needs: [what-changed, example-component, cli-assets]
     if: |
-      always()
+      (!cancelled())
       && fromJson(needs.what-changed.outputs.rust).changed-packages
       && (needs.example-component.result == 'skipped' || needs.example-component.result == 'success')
       && (needs.cli-assets.result == 'skipped' || needs.cli-assets.result == 'success')
@@ -438,7 +462,7 @@ jobs:
     # This job is responsible for reacting to build success or failure. It must
     # happen after the builds, hence the `needs`. But it must not be skipped
     # when the builds are cancelled or fail (hence the `if: ${{ always() }}`).
-    needs: [builds, docker-grafbase, docker-gateway]
+    needs: [format, builds, docker-grafbase, docker-gateway]
     runs-on: ubicloud-standard-2
     if: ${{ always() }}
     steps:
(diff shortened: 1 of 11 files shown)


.github/workflows/rust-prs.yml
```diff
@@ -0,0 +1,471 @@
+name: rust-pr workflow
+
+on:
+  workflow_dispatch:
+  push:
+    branches:
+      - 'obmarg/ysrtktmxktrp'
```
obmarg, line 7:
```suggestion
  push:
    branches:
      - 'obmarg/ysrtktmxktrp'
```

.github/workflows/rust-prs.yml
```diff
@@ -0,0 +1,471 @@
...
+  pull_request:
+  # TODO: Maybe add branches in here?  Not sure...
```
obmarg, line 9:
```suggestion
```

.github/workflows/rust-prs.yml
```diff
@@ -0,0 +1,471 @@
...
+  CARGO_TERM_COLOR: 'always'
+  DO_NOT_TRACK: 1
+  GRAFBASE_RUDDERSTACK_WRITE_KEY: ${{ secrets.GRAFBASE_RUDDERSTACK_WRITE_KEY }}
+  GRAFBASE_RUDDERSTACK_DATAPLANE_URL: ${{ secrets.GRAFBASE_RUDDERSTACK_DATAPLANE_URL }}
+  RUST_BACKTRACE: 1
+
+  # TODO: Suspect these might not be needed - look into it.
+  CLI_RELEASE_CLOUDFLARE_R2_ENDPOINT_URL: https://d267a8ab95268b272f5147e8939c9d38.r2.cloudflarestorage.com
+  CLI_RELEASE_CLOUDFLARE_R2_PUBLIC_URL: https://pub-b0013effd6614b32a4fb9bf338b90658.r2.dev
+  CLI_RELEASE_CLOUDFLARE_R2_BUCKET_NAME: cli-releases
```
obmarg, line 25:
```suggestion
```

.github/workflows/rust-prs.yml
```diff
@@ -0,0 +1,471 @@
...
+        run: |
+          echo ${{ needs.what-changed.outputs.rust }}
+
+      - name: Extract the Rust version to use from the `rust-toolchain.toml` file
+        shell: bash
+        run: |
+          rust_version=$(grep "channel" rust-toolchain.toml | cut -d "=" -f 2 | cut -d " " -f 2)
+          echo "RUST_VERSION=$rust_version" >> $GITHUB_ENV
+
+      # TODO: Maybe refactor this?  Not sure
```
obmarg, line 234:
```suggestion
```

.github/workflows/rust-prs.yml
```diff
@@ -0,0 +1,471 @@
...
+      - name: Login to Docker Hub
+        if: runner.os == 'Linux'
+        uses: docker/login-action@v3
+        with:
+          username: ${{ secrets.DOCKERHUB_USERNAME }}
+          password: ${{ secrets.DOCKERHUB_TOKEN }}
+
+      # TODO: maybe merge these composes into one?  Think it would download in parallel if nothing else
+      # Probably worthwhile...
+      - name: Start integration-test docker compose
```
obmarg, line 327:
```suggestion
      - name: Start integration-test docker compose
```

engine/crates/graphql-schema-diff/src/lib.rs
```diff
@@ -16,6 +16,8 @@ use self::state::*;
 use cynic_parser::type_system as ast;
 use std::collections::HashMap;
 
+// TODO: Remove me (this is just to pretend there is a change)
+
```
obmarg, line 20:
```suggestion
```

rust-toolchain.toml
```diff
@@ -2,6 +2,7 @@
 profile = "default"
 channel = "1.80.0"
 targets = [
+    # TODO: Figure out how to not install these everywhere, it's a bit crap.
```
obmarg, line 5:
```suggestion
```

.github/workflows/rust-prs.yml
```diff
@@ -0,0 +1,459 @@
...
+
+      - name: Install Rust
+        uses: dtolnay/rust-toolchain@master
+        with:
+          toolchain: ${{ env.RUST_VERSION }}
+          components: clippy, rustfmt
+
+      - name: Install what-rust-changed
+        shell: bash
+        run: cargo install --locked --git https://github.com/grafbase/what-rust-changed.git --branch obmarg/qmtxorwqlrtl
```
pimeys, line 51: Could we use bininstall here? That would avoid compiling the tool.

.github/workflows/rust-prs.yml
```diff
@@ -0,0 +1,459 @@
...
+        with:
+          registry: ghcr.io
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Build docker images
+        run: |
+          # Re-use the latest layers if possible
+          docker pull ghcr.io/grafbase/gateway:latest || true
+          docker build -f gateway/Dockerfile -t ghcr.io/grafbase/gateway:$COMMIT_SHA .
```
tomhoule, line 394: nice!

.github/workflows/rust-prs.yml
```diff
@@ -0,0 +1,459 @@
...
+      - uses: actions/upload-artifact@v4
+        with:
+          name: wrappers-dist.js
+          path: cli/wrappers/dist.js
+          retention-days: 5
+
+  builds:
+    needs: [what-changed, example-component, cli-assets]
+    if: |
+      (!cancelled())
```
tomhoule, line 179: Is there a clear reason why this is only needed for this job?

tomhoule (approved):
Awesome, both the results and the implementation

.github/workflows/rust-prs.yml
```diff
@@ -0,0 +1,459 @@
...
+          toolchain: ${{ env.RUST_VERSION }}
+          components: rustfmt
+
+      - name: cargo fmt
+        shell: bash
+        run: |
+          cargo fmt --check
+
+  example-component:
+    if: contains(fromJson(needs.what-changed.outputs.rust).changed-packages, 'integration-tests')
```
pimeys, line 115: Also `gateway-binary` and `wasi-component-loader`... We should have a simpler place to change this so we remember when more crates start depending on these examples.

.github/workflows/rust-prs.yml
obmarg: `example-component` & `cli-assets` are both optional - without doing some `if` shenanigans this wouldn't run if they were skipped.  

I did have `always()` initially, but it turns out if you put `always()` in a job it becomes uncancellable - this is the recommended alterantive that lets you cancel.

I'll add a comment though

.github/workflows/rust-prs.yml
obmarg: This was lifted straight out of the existing workflow fwiw

.github/workflows/rust-prs.yml
```diff
@@ -0,0 +1,459 @@
...
+          components: clippy, rustfmt
+
+      - name: Install musl-tools
+        if: ${{ contains(matrix.platform.target, 'linux') }}
+        shell: bash
+        run: |
+          sudo apt-get install musl musl-tools
+          # This seems like a horrible hack that might come back to bite, but lets see!
+          sudo ln -s /bin/g++ /bin/musl-g++
+          sudo ln -s /bin/g++ /bin/aarch64-linux-musl-g++
```
pimeys, line 238: 😱 

We should consider just getting rid of musl eventually... Maybe consider having a certain glibc version minimum we require, document it and avoid having other dynamic dependencies, such as openssl et.al.

pimeys (approved):
awesome

pimeys:
Btw, are you removing the CLI workflow eventually? It's here just as a comparison?

.github/workflows/rust-prs.yml
obmarg: Once https://github.com/grafbase/what-rust-changed/pull/2 is merged I'm going to tag a release of what-rust-changed.  There's a workflow in that PR that'll publish release binaries and then I'll change this to fetch those.

Don't neccesarily want to use `cargo binstall` - both because I don't want to publish to crates.io _and_ because I saw some weird issues where `cargo binstall` occasionally failed to install itself and just retried endlessly until the build timed out 👎

obmarg:
> Btw, are you removing the CLI workflow eventually? It's here just as a comparison?

That is the goal, but I don't know when.  

- My next PR will stop running it on PR builds (it's still a required step so can't remove it till this is merged), so it only runs on main & release.
- Want to continue running CLI workflow on main for a bit incase I've missed anything in this.  Once we have confidence we could probably use this action (or similar) for main builds
- This action isn't really suitable for release builds, so going to keep the CLI build for them _for now_.  At some point I'd like to rework the release builds as well but don't have it planned immediately.

.github/workflows/rust-prs.yml
obmarg: Yeah, would definitely be nice.

.github/workflows/rust-prs.yml
obmarg: (Weirdly I only saw that cargo binstall issue in one place I was using it, which I've now removed - the binstall of `cargo component` always seemed to work)

PR #2093, opened 2024-09-05
chore: add a workspace hack package & caching

obmarg:
We make quite heavy use of feature flags in this repo, both with external dependencies and with our own crates.  These are great, but it's very very easy to end up with slightly different feature sets if you're compiling subsets of the repo.  This can cause rebuilds of entire dependency trees.  I've certainly noticed this on my local machine when running test suites for individual packages, and from my testing it also happens on CI.

This PR adds a fix for that: a "workspace hack" package managed by `cargo-hakari`.  `cargo-hakari` looks at all the dependencies of all our crates and adds them to the workspace hack package, with the full set of features used everywhere. This ensures we always have a consistent set of features enabled.  

This has a few ramifications:

- It cuts down on unnecessary rebuilds locally.  A clear win
- It cuts down on unnecessary rebuilds in a CI job.  Also a clear win.
- It makes caching in CI easier, as the set of dependencies are always unified.
- It does mean every CI run has to build all the dependencies - even if they aren't directly used in the subset of packages we're building.  Not so good, but this can be mitigated by caching.

### Caching

With the workspace hack in place we need about 1gb of space to cache each target.  We have 4 targets so we can now fit a single builds cache into GitHubs 10GB of cache space.  However, if we change a few deps concurrently we're easily going over 10GB.  In my testing the GH cache is also not consistently fast.

Buildjet & ubicloud both offer GHA compatible caches.  When running on ubicloud workers the GHA cache is a little slow, so 
this PR enables the Buildjet cache for this. It was easier than ubiclouds cache and they offer 20gb by default (and also appear open to giving you more), whereas ubicloud only give 10 (and it's less clear that they will give you more - though we could email them and ask).

I have tested builds on CI both with and without caching:

- For a full rebuild of the repo
  - With this change but no caching I have seen 14 minute rebuilds (vs 16 or so without this change)
  - With caching enabled but nothing in the cache - 15m 20s
  - With caching enabled and a populated cache - 9m 31s
- For a build of _just_ gqlint (the smallest package in our repo in terms of other packages it affects)
  - With this change but no caching: 8 minutes (vs 4:40 without - quite bad)
  - I don't have full rebuild numbers for buildjet cache, but was 6m 27 to build using GH caching and an empty cache
  - With a populated cache (on buildjet) - 4m 23

So provided we combine it with caching, this seems like a very clear win.

I've also removed the quite ineffective old caching as part of this - it was filling up our available cache storage basically every day, and makes the caching that actually works harder.  See #2099 for more thinking on that.


obmarg:
Note to self: need to do some testing of partial builds with this.  Definitely shaves a lot off full builds, but if every build has to compile every 3rd party dependency that's not great...

obmarg:
One slightly annoying thing about this change: `cargo hakari generate` on macos seems to miss one of the dependencies that running it on linux requires.  I've added it in by hand here but will need to do that going forward whenever the deps change (unless we can figure out how to make them consistent)

.github/workflows/rust-prs.yml
```diff
@@ -209,12 +209,24 @@ jobs:
...
           components: clippy, rustfmt
 
+      - name: Rustup overridpe
+        shell: bash
+        # Adding this override prevents rust from reading rust-toolchain
+        # We want to do this because rust-toolchain has a bunch of targets in it
+        # we don't need here - I'm hoping this setting will prevent those from
+        # downloading
+        run: rustup override set ${{ env.RUST_VERSION }}
+
```
obmarg, line 229:
```suggestion
```

.github/workflows/rust-prs.yml
```diff
@@ -289,6 +301,12 @@ jobs:
           name: wrappers-dist.js
           path: cli/wrappers/
 
+      - name: Build
+        if: fromJson(needs.what-changed.outputs.rust).cargo-build-specs
+        shell: bash
+        run: |
+          cargo build --target ${{ matrix.platform.target }} ${{ fromJson(needs.what-changed.outputs.rust).cargo-build-specs }}
+
```
obmarg, line 309:
```suggestion
```

.github/workflows/cargo-hakari.yml
```diff
@@ -0,0 +1,50 @@
...
+    steps:
+      - name: Checkout Repository
+        uses: actions/checkout@v4
+
+      - name: Install cargo-hakari
+        uses: taiki-e/install-action@v2
+        with:
+          tool: cargo-hakari
+
+      - name: Ensure grafbase-worksace-hack is up to date
```
tomhoule, line 44:
```suggestion
      - name: Ensure grafbase-workspace-hack is up to date
```

nitpicking

.github/workflows/cargo-hakari.yml
```diff
@@ -0,0 +1,50 @@
...
+        shell: bash
+        run: cargo hakari generate --diff
+
+      - name: Ensure all packages are using grafbase-workspace-hack
+        shell: bash
+        run: cargo hakari generate --diff
```
tomhoule, line 50: Are the last two steps supposed to run the same command?

workspace-hack/Cargo.toml
```diff
@@ -0,0 +1,445 @@
...
+web-sys = { version = "0.3", default-features = false, features = ["BlobPropertyBag", "CanvasRenderingContext2d", "Document", "DomRect", "HtmlCanvasElement", "Location", "Navigator", "Performance", "RequestCredentials", "RequestMode", "ServiceWorkerGlobalScope", "Window", "Worker"] }
+
+[target.wasm32-unknown-unknown.build-dependencies]
+getrandom = { version = "0.2", default-features = false, features = ["js"] }
+gimli = { version = "0.29", default-features = false, features = ["read", "std", "write"] }
+libc = { version = "0.2", features = ["extra_traits"] }
+miniz_oxide = { version = "0.7", default-features = false, features = ["with-alloc"] }
+object = { version = "0.36", default-features = false, features = ["archive", "read_core", "unaligned", "write"] }
+rustix = { version = "0.38", features = ["net", "process", "procfs", "termios", "time"] }
+web-sys = { version = "0.3", default-features = false, features = ["BlobPropertyBag", "CanvasRenderingContext2d", "Document", "DomRect", "HtmlCanvasElement", "Location", "Navigator", "Performance", "RequestCredentials", "RequestMode", "ServiceWorkerGlobalScope", "Window", "Worker"] }
```
tomhoule, line 443: I've already seen Cargo.toml's that large that weren't autogenerated, sadly.

tomhoule (approved):
Brilliant

.github/workflows/cargo-hakari.yml
obmarg: No 🤦 - good spot

enescakir:
Hi @obmarg, this is Enes from Ubicloud. I noticed your note
> ubicloud only give 10 (and it's less clear that they will give you more - though we could email them and ask)

We'd be happy to offer you more cache space. Could you drop us a note at [support@ubicloud.com](mailto:support@ubicloud.com)?

PR #2105, opened 2024-09-06
Replace async-graphql-parser with cynic-parser in graphql-federated-graph

tomhoule:
For `from_sdl()`. We are standardizing on cynic-parser, and we want to make that function more robust, progressively, so this has to happen.

### DIFF
--- a/engine/crates/federated-graph/src/from_sdl.rs
+++ b/engine/crates/federated-graph/src/from_sdl.rs
@@ -61,7 +61,7 @@ struct State<'a> {
 
 impl<'a> State<'a> {
     fn field_type(&mut self, field_type: &'a ast::Type<'a>) -> Result<Type, DomainError> {
-        use cynic_parser::{common::WrappingType, type_system as ast};
+        use cynic_parser::common::WrappingType;
 
         self.type_wrappers.clear();
         self.type_wrappers.extend(field_type.wrappers());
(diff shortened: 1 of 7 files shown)


engine/crates/federated-graph/src/from_sdl.rs
```diff
@@ -96,38 +107,31 @@ impl<'a> State<'a> {
...
-                } else if let Some(number) = number.as_f64() {
-                    Value::Float(number)
-                } else {
-                    unreachable!()
-                }
-            }
-            async_graphql_value::ConstValue::String(s) => Value::String(self.insert_string(s)),
-            async_graphql_value::ConstValue::Boolean(b) => Value::Boolean(*b),
-            async_graphql_value::ConstValue::Enum(enm) => expected_enum_type
+            ast::Value::Null => Value::String(self.insert_string("null")),
```
obmarg, line 112: Are we representing nulls as a string?  That seems... odd.

engine/crates/federated-graph/src/from_sdl.rs
```diff
@@ -479,46 +470,49 @@ fn ingest_entity_keys(parsed: &ast::ServiceDocument, state: &mut State<'_>) -> R
...
+{
+    let is_federated_entity = parent_directives()
+        .filter(|dir| dir.name() == JOIN_TYPE_DIRECTIVE_NAME)
+        .any(|dir| dir.get_argument("key").is_some());
 
     let type_subgraph_ids = if !is_federated_entity {
-        parent_directives
-            .iter()
-            .filter(|dir| dir.node.name.node == JOIN_TYPE_DIRECTIVE_NAME)
+        parent_directives()
```
obmarg, line 511: I may be missing some reason why this wouldn't work, but iirc I made `Iter: Clone + Copy` so I expect parent_directives could just be an `ast::iter::Iter<'a, ast::Directive<'a>>`

engine/crates/federated-graph/src/from_sdl/value.rs
```diff
@@ -0,0 +1,51 @@
...
+        executable_ast::Value::List(v) => {
+            ast::Value::List(v.into_iter().map(executable_value_to_type_system_value).collect())
+        }
+        executable_ast::Value::Object(v) => ast::Value::Object(
+            v.into_iter()
+                .map(|(name, value)| (name, executable_value_to_type_system_value(value)))
+                .collect(),
+        ),
+        executable_ast::Value::Null => ast::Value::Null,
+    }
```
obmarg, line 20: I really need to unify and/or otherwise tidy the Value stuff up in the parser sometime...

obmarg (approved):
Nice 🎉

engine/crates/federated-graph/src/from_sdl.rs
tomhoule: Indeed, not from this PR but I can't remember a good reason to do that. I'll look into fixing that.

engine/crates/federated-graph/src/from_sdl/value.rs
tomhoule: I wonder about Value::Variable. I can't think of a scenario where it would be in a SDL document. That would be a reason to have two enums.

obmarg: Yeah I think two enums makes sense, but I probably want something closer to the `Value`/`ConstValue` approach that async-graphql uses.  Haven't sat down and given it proper thought though.

One thing I do like about the current separation is that they use different `ValueId` types, since that should help prevent mistakes if you end up needing to use those ids.  Not sure how likely that is really though.

I also kind of wanted to refactor the `Value`s themselves to be a bit more lazy - currently calling `Argument::value` (or whatever) could be kind of expensive since it constructs the whole heirarchy of objects & lists etc...

Lots to do around Values basically - probably the bit I'm the least happy with currently.

engine/crates/federated-graph/src/from_sdl.rs
tomhoule: It looks like this particular instance doesn't implement Clone

obmarg: ah well, damn

obmarg: Fixed that here: https://github.com/obmarg/cynic/pull/1030

PR #2112, opened 2024-09-07
chore(engine-v2-schema): Use codgen to generate walkers

Finistere:
Walkers and the id+Index combo are great, they only suffer from a single drawback, they're extremely tedious to write.
This led to the `SchemaWalker` use of generics to be oriented towards reducing the amount of code we have to write and it's only exacerbated in `engine-v2` for operations with heavy use of `Deref`, forwarding to an embedded SchemaWalker, and few walkers because I've refactored the underlying structs so many times.

All of this came at the cost of reduced readability, using "go to definition" anywhere in `schema.walk(id)` would not lead to the walker. Requesting the type from rust-analyzer would produce something like `SchemaWalker<'a, ID>`. This produced hard to read code for anyone but me such as:
```rust
self
    .schema_walker
    .walk(&self.schema_walker.as_ref()[*id])
    .serialize(serializer)
```

@obmarg  found an excellent solution for this, generating most of the walker code from a GraphQL schema in `registry-v2` several months ago. Since then I've been working here and there on applying on `engine-v2-schema`, especially the last few weeks. Within that effort, I've tried to improve the readability of this design as much as I could. I ended up with the following design:

### Code generation

Code is generated with `engine-v2-codegen`, this is the output for [EnumValue](https://github.com/grafbase/grafbase/blob/gb-7452-use-code-generation-in-engine-v2-schema-like-registry-v2/engine/crates/engine-v2/schema/src/generated/enum_value.rs):
```rust
//! ===================
//! !!! DO NOT EDIT !!!
//! ===================
//! Generated with: `cargo run -p engine-v2-codegen`
//! Source file: <engine-v2-codegen dir>/domain/schema.graphql
use crate::{
    generated::{TypeSystemDirective, TypeSystemDirectiveId},
    prelude::*,
    StringId,
};
use walker::{Iter, Walk};

/// Generated from:
///
/// ```custom,{.language-graphql}
/// type EnumValue @meta(module: "enum_value") @indexed(id_size: "u32", max_id: "MAX_ID") {
///   name: String!
///   description: String
///   directives: [TypeSystemDirective!]!
/// }
/// ```
#[derive(Debug, serde::Serialize, serde::Deserialize)]
pub struct EnumValueRecord {
    pub name_id: StringId,
    pub description_id: Option<StringId>,
    pub directive_ids: Vec<TypeSystemDirectiveId>,
}

#[derive(PartialEq, Eq, PartialOrd, Ord, Clone, Copy, Hash, serde::Serialize, serde::Deserialize, id_derives::Id)]
#[max(MAX_ID)]
pub struct EnumValueId(std::num::NonZero<u32>);

#[derive(Clone, Copy)]
pub struct EnumValue<'a> {
    pub(crate) schema: &'a Schema,
    pub(crate) id: EnumValueId,
}

impl std::ops::Deref for EnumValue<'_> {
    type Target = EnumValueRecord;
    fn deref(&self) -> &Self::Target {
        self.as_ref()
    }
}

impl<'a> EnumValue<'a> {
    /// Prefer using Deref unless you need the 'a lifetime.
    #[allow(clippy::should_implement_trait)]
    pub fn as_ref(&self) -> &'a EnumValueRecord {
        &self.schema[self.id]
    }
    pub fn id(&self) -> EnumValueId {
        self.id
    }
    pub fn name(&self) -> &'a str {
        &self.schema[self.as_ref().name_id]
    }
    pub fn description(&self) -> Option<&'a str> {
        self.as_ref().description_id.map(|id| self.schema[id].as_ref())
    }
    pub fn directives(&self) -> impl Iter<Item = TypeSystemDirective<'a>> + 'a {
        self.as_ref().directive_ids.walk(self.schema)
    }
}

impl Walk<Schema> for EnumValueId {
    type Walker<'a> = EnumValue<'a>;
    fn walk<'a>(self, schema: &'a Schema) -> Self::Walker<'a>
    where
        Self: 'a,
    {
        EnumValue { schema, id: self }
    }
}

impl std::fmt::Debug for EnumValue<'_> {
    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
        f.debug_struct("EnumValue")
            .field("name", &self.name())
            .field("description", &self.description())
            .field("directives", &self.directives())
            .finish()
    }
}
```

 The whole schema is defined by [engine-v2-codegen/domain/schema.graphql](https://github.com/grafbase/grafbase/blob/gb-7452-use-code-generation-in-engine-v2-schema-like-registry-v2/engine/crates/engine-v2/codegen/domain/schema.graphql) which generates the whole `generated` module from `engine-v2-schema`. Generated code usually includes the `XRecord` which is stored in the `Schema`, a `XId` and a walker simply named `X`. It's ensured that `XRecord` field names do not clash with the walker methods, `name_id` stores the `StringId` and `name()` is the walker method. This allows the walker to `Deref` to the struct to avoid most use of `.as_ref()` while keeping it clear what is retrieved and favoring the use of walkers through shorter names. Previously the struct field would have just been named `name`. To help with discoverability, `XId`, `XRecord`, the walker `X`, and the implementation of the `Walk` trait are always in the same file. The walker is also a dedicated struct rather than a type alias.

### Walk trait

Instead of the `schema.walk(id)` pattern, I've inversed the relationship to be `id.walk(schema)` with this trait:
```rust
pub trait Walk<G> {
    type Walker<'a>
    where
        Self: 'a,
        G: 'a;

    fn walk<'a>(self, graph: &'a G) -> Self::Walker<'a>
    where
        Self: 'a;
}
```
Now using "go to definition" with `walk` will lead in the right direction instead of a generic function. The `Walk` trait has another big advantage with multiple implementors being able to point to the same walker struct. This is particularly useful for input values as the walker was `SchemaWalker<'a, &'a InputValue>` as only some values could be accessed directly by an id.

To alleviate some of the pain experienced when manipulating `Vec<id>`, `Option<id> and `IdRange<Id>` I've added blanket implementations for those to allow writing the following:

```rust
ids.walk(schema)
   .map(|(input_value_definition, value)| (input_value_definition.name(), value)),
```

rather than

```rust
self.schema[*ids].iter().map(|(input_value_definition_id, value)| {
    let value = self.walk(value);
    (self.walk(*input_value_definition_id).name(), value)
})
```

I've hesitated a bit, but it makes writing walker-related code a lot easier when it can't be generated and it's more readable IMHO.

### Misc

- fixed missing use of coercion for default input values. While it shouldn't have led to any bugs today, everything expected it to be coerced (fixes GB-7453).
- I left a `schema.walk` method because it was just too much work to clean up `engine-v2` for this PR
- Removed the embedded `SchemaWalker` inside the various operation walkers and replaced with `&'a Schema`. Since walkers aren't defined through generics anymore, had no choice but to rework it (in the right direction).

Eventually, I'll do the same for the operation walkers but the underlying structs are more complex, so well for another time...


Cargo.toml
```diff
@@ -227,7 +229,6 @@ strip = "none"
 debug = true
 
 [profile.release]
-strip = "symbols"
```
tomhoule, line 230: Is this change intentional?

Finistere: ah yeah, forgot to mention it. We're stripping symbols today which makes stack trace pretty much useless today. Out of curiosity I ran benchmarks with & without, and with `abort`/`unwind` for panic and saw no difference. So this parameter is just us hurting ourselves for the standard release

engine/crates/engine-v2/codegen/src/formatter.rs
```diff
@@ -0,0 +1,73 @@
...
+        use xshell::cmd;
+
+        let code = cmd!(self.shell, "rustfmt")
+            .stdin(code.clone())
+            .read()
+            .inspect_err(|_| {
+                tracing::error!("Failed to format file:\n{code}");
+            })?;
+        // Transform 'serde :: Serialize' to 'serde::Serialize' and other similar cases.
+        let code = code.replace(" :: ", "::");
```
obmarg, line 27: Does rustfmt not do this for you?

engine/crates/engine-v2/id-derives/src/id.rs
```diff
@@ -27,11 +27,27 @@ pub fn derive_id(input: proc_macro::TokenStream) -> proc_macro::TokenStream {
...
+        }
+
+        impl #impl_generics From<#inner_ty> for #ident #ty_generics #where_clause {
+            fn from(value: #inner_ty) -> Self {
                 #max_check
                 Self(
-                    #inner_ty::try_from(value + 1).ok().and_then(|value| value.try_into().ok()).expect(#too_many_error)
+                    (value + 1).try_into().expect(#too_many_error)
                 )
             }
```
obmarg, line 52: I may be missing something, but why are we only doing the `+ 1` in the `From<#inner_ty>` impl?  That kind of inconsistency seems like it'd cause off by one errors...

Also why do we need a conversion from i32 _at all_ - surely we never need negative numbers?

engine/crates/engine-v2/schema/src/builder/input_values.rs
```diff
@@ -1,48 +1,48 @@
...
-    ) -> Result<SchemaInputValue, InaccessibleEnumValue> {
+    ) -> Result<SchemaInputValueRecord, InaccessibleEnumValue> {
         Ok(match value {
-            Value::Null => SchemaInputValue::Null,
-            Value::String(id) | Value::UnboundEnumValue(id) => SchemaInputValue::String(id.into()),
-            Value::Int(n) => SchemaInputValue::BigInt(n),
-            Value::Float(f) => SchemaInputValue::Float(f),
-            Value::Boolean(b) => SchemaInputValue::Boolean(b),
+            Value::Null => SchemaInputValueRecord::Null,
+            Value::String(id) | Value::UnboundEnumValue(id) => SchemaInputValueRecord::String(id.into()),
```
obmarg, line 18: Not really a comment on this PR really, but converting enum values into strings is often wrong?

I assume it's fine here because we put these into variables and not the query literal?  But might be worth a comment saying that if so.

Finistere:
Yeah, directives are a nested Vec now. The reason is that `TypeSystemDirectiveId` is an enum now, like `SubgraphId` and `DefinitionId`. So instead of `TypeSystemDirectiveId` pointing to a `TypeSystemDirective` enum inside a vec, the TypeSystemDirectiveId` is the enum directly. It felt easier to manipulate IMHO. 

The `deduplicated` is probably the least obvious thing in code generation. It's used to tell that the type being indexed is deduplicated and thus we cannot rely on `IdRange` but have to use a `Vec<Id>`, in a similar way to how `Vec<StringId>`  is used rather than `IdRange<StringId>`. It avoids the needs to specify on the fields whether to use a `Vec` or `IdRange`.

engine/crates/engine-v2/schema/src/entity.rs
```diff
@@ -0,0 +1,43 @@
...
+        }
+    }
+
+    pub fn directives(&self) -> impl Iter<Item = TypeSystemDirective<'a>> + 'a {
+        let (schema, directive_ids) = match self {
+            EntityDefinition::Object(item) => (item.schema, &item.as_ref().directive_ids),
+            EntityDefinition::Interface(item) => (item.schema, &item.as_ref().directive_ids),
+        };
+        directive_ids.walk(schema)
+    }
```
tomhoule, line 42: This module looks a lot like the generated one, is it still needed?

Finistere: it's adding a few functions to the generated walker, like `name()`. I found it too risky to complexify the graphql schema even more to auto-generate those, so just writing them by hand

tomhoule:
> The deduplicated is probably the least obvious thing in code generation. It's used to tell that the type being indexed is deduplicated and thus we cannot rely on IdRange but have to use a Vec<Id>, in a similar way to how VeThe deduplicated is probably the least obvious thing in code generation. It's used to tell that the type being indexed is deduplicated and thus we cannot rely on IdRange but have to use a Vec<Id>, in a similar way to how Vec<StringId> is used rather than IdRange<StringId>. It avoids the needs to specify on the fields whether to use a Vec or IdRange.c<StringId> is used rather than IdRange<StringId>. It avoids the needs to specify on the fields whether to use a Vec or IdRange.

Ah ok I get it. Deduplicated is a synonym for interned then, right?

> Yeah, directives are a nested Vec now. The reason is that TypeSystemDirectiveId is an enum now, like SubgraphId and DefinitionId. So instead of TypeSystemDirectiveId pointing to a TypeSystemDirective enum inside a vec, the TypeSystemDirectiveId is the enum directly. It felt easier to manipulate IMHO.

That makes sense — directives is always a pain point when I think about them in the context of FederatedGraph. That's a good approach to try :+1:

engine/crates/engine-v2/schema/src/entity.rs
Finistere: and yes all of those are used. Those extra functions are more or less good, but would require reworking the `engine-v2` operation part to change them

tomhoule: Ah got it. I should have read more closely.

tomhoule: Thanks!

Finistere:
> Ah ok I get it. Deduplicated is a synonym for interned then, right?

Hum yes, hesitated to use `interned` for objects & interfaces (for possible_types & interfaces fields), but seeing your answer would it have been better?

engine/crates/engine-v2/id-newtypes/src/range.rs
```diff
@@ -86,10 +86,10 @@ where
         let Some(first) = ids.next() else {
             return Some(Self::empty());
         };
-        let start = usize::from(*first);
+        let start: usize = (*first).into();
```
obmarg, line 89: I'm being a massive nitpicker but I really thought this was one place where `usize::from` is unambiguously better... 

(I prefer the clarity of `usize::from` in most of the places you've changed, but struggling to see any benefit to using into here since you have to specify the destination type anyway)

tomhoule (approved):
That's going to make things a lot more consistent :100:

tomhoule:
> Hum yes, hesitated to use interned for objects & interfaces (for possible_types & interfaces fields), but seeing your answer would it have been better?

I'm fine with deduplicated now that I understand it, no need to change imo

engine/crates/engine-v2/codegen/src/formatter.rs
Finistere: it doesn't :/

obmarg: 👎

engine/crates/engine-v2/id-derives/src/id.rs
Finistere: I might have done stupid stuff somewhere, but somehow it didn't accept `ObjectDefinition::from(0)`, got annoyed and added it. But after the coercion fix I'm not creating as many of those anyway. So removing it, you're right.

The `+1` was always done in the `From` until now, are you suggesting creating a `new` method? Or do I miss something?

engine/crates/engine-v2/id-newtypes/src/range.rs
Finistere: I agree, did for `IdOperations` stuff when I copy pasted from your `regsitry-v2` code initially, but don't need it anymore. I'm going to revert those to like before. I agree `usize::from` is a lot better than `.into()`

engine/crates/engine-v2/id-derives/src/id.rs
obmarg: > The +1 was always done in the From until now, are you suggesting creating a new method? Or do I miss something?

Of the three impls I have highlighted only the last one contains a `+ 1`.  The first two do not - that seems wrong?

engine/crates/engine-v2/schema/src/builder/input_values.rs
obmarg: But then if we try and write this SchemaInputValueRecord back into GraphQL it'll be `{a: "ACTIVE"}` which is not the same?  Is there some way in which that is not a bug?

engine/crates/engine-v2/id-derives/src/id.rs
Finistere: the first two use the last one. so `From<usize>` does: `Self::from(u32::try_from(value))`

engine/crates/engine-v2/schema/src/builder/input_values.rs
Finistere: hum, that's a good remark. Today it's not a problem since we always pass everything through variables, so serialized as JSON. Will do it in another PR then, you're right that I don't see how we could no keep that case...

engine/crates/engine-v2/id-derives/src/id.rs
obmarg: Hmm...  Could be clearer tbh.

Finistere: yeah agreed, it's not super obvious in the macro

engine/crates/engine-v2/schema/src/builder/input_values.rs
obmarg: > Today it's not a problem since we always pass everything through variables

Yeah, this is what I was saying in my initial comment - I assume it's not actually a problem, but a comment here explaining why its not a problem would clarify (although if you want to properly fix it then that is also cool)

Finistere: yeah good point, taking a look on it, it's probably best to just keep track of them and only treat them as strings when serializing

engine/crates/engine-v2/walker/src/iter.rs
```diff
@@ -0,0 +1,144 @@
...
+            for (left, right) in self.clone().zip(other.clone()) {
+                match left.cmp(&right) {
+                    Ordering::Equal => continue,
+                    other => return other,
+                }
+            }
+            Ordering::Equal
+        })
+    }
+}
```
obmarg, line 99: Not saying this is wrong, but an iterator implementing `Ord` seems a bit odd to me... Worth a comment explaining why it's needed if nothing else?

Finistere: Adding a comment later, I'm relying on it to compare field arguments in `@requires` since they can. This allows me to deduplicate/intern those required fields and make set manipulations at lot easier when determining the everything that is required during planning

engine/crates/engine-v2/codegen/src/formatter.rs
```diff
@@ -0,0 +1,73 @@
...
+        Ok(Self {
+            shell: xshell::Shell::new()?,
+            doc_re: Regex::new(r#"(?m)^(?<spaces>\s*)#\[doc\s*=\s*"(?<doc>.*)"\]$"#)?,
+        })
+    }
+
+    pub(super) fn format(&self, code: String) -> anyhow::Result<String> {
+        use xshell::cmd;
+
+        let code = cmd!(self.shell, "rustfmt")
```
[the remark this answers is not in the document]
Finistere: maybe, this works 🤷 Took a quick look and their doc doesn't mention anything regarding using it as a lib.

PR #2116, opened 2024-09-11
engine-v2: return `null` for inaccessible enum values

tomhoule:
closes GB-7473


engine/crates/engine-v2/schema/src/lib.rs
```diff
@@ -249,23 +249,28 @@ impl std::fmt::Debug for Schema {
...
-)]
+#[derive(Debug, Clone, Copy, PartialEq, Eq, strum::Display, serde::Serialize, serde::Deserialize)]
 pub enum ScalarType {
     String,
     Float,
     Int,
     BigInt,
     JSON,
     Boolean,
+    Enum(EnumDefinitionId),
```
tomhoule, line 260: This, or move that out of `ScalarType`?

engine/crates/engine-v2/src/response/write/deserialize/scalar.rs
```diff
@@ -3,22 +3,40 @@ use serde::{de::DeserializeSeed, Deserialize};
...
+        let ScalarTypeSeed(ctx, value) = self;
+
+        match value {
             ScalarType::String => String::deserialize(deserializer).map(Into::into),
             ScalarType::Float => f64::deserialize(deserializer).map(Into::into),
             ScalarType::Int => i32::deserialize(deserializer).map(Into::into),
             ScalarType::BigInt => i64::deserialize(deserializer).map(Into::into),
             ScalarType::JSON => Box::<serde_json::Value>::deserialize(deserializer).map(Into::into),
             ScalarType::Boolean => bool::deserialize(deserializer).map(Into::into),
+            ScalarType::Enum(enum_definition_id) => {
```
tomhoule, line 26: Is that stage in the request lifecycle the right place to do that?

engine/crates/integration-tests/tests/federation/basic/enums.rs
```diff
@@ -0,0 +1,31 @@
...
+        "recommendedTeas": [
+          {
+            "id": 0,
+            "name": "Earl Grey",
+            "style": "BLACK"
+          },
+          {
+            "id": 7,
+            "name": "Tuóchá",
+            "style": null
```
obmarg, line 25: Style isn't nullable in your schema so is setting this to null correct?  Would have kind of expected it to behave like errors and bubble up to the nearest nullable or something like that.

tomhoule: Good point. It won't emit an error but we should bubble up the nullability. That means I probably haven't picked the right place for that change in response processing.

engine/crates/engine-v2/schema/src/definition.rs
```diff
@@ -29,7 +29,6 @@ impl<'a> Definition<'a> {
     pub fn scalar_type(&self) -> Option<ScalarType> {
         match self {
             Definition::Scalar(scalar) => Some(scalar.ty),
-            Definition::Enum(_) => Some(ScalarType::String),
```
tomhoule, line 32: I wonder if this line has ramifications outside of the code paths I touched.

tomhoule: Looks like that's the only usage. Should be fine.

engine/crates/engine-v2/src/response/write/deserialize/enum.rs
```diff
@@ -0,0 +1,30 @@
+use schema::EnumDefinitionId;
+use serde::{de::DeserializeSeed, Deserialize};
+use walker::Walk as _;
+
+use crate::response::ResponseValue;
+
+use super::SeedContext;
+
+pub(crate) struct EnumTypeSeed<'parent, 'ctx>(pub &'parent SeedContext<'ctx>, pub EnumDefinitionId);
```
Finistere, line 9: Hum realizing `ScalarTypeSeed` isn't a great name , would have used `EnumValueSeed` here, but it's really just a nitpick

Finistere (approved):
Nice!

engine/crates/engine-v2/src/response/write/deserialize/enum.rs
tomhoule: pushed

PR #2117, opened 2024-09-12
engine-v2: enforce inaccessibility of inaccessible arguments

tomhoule:
An argument with `@inaccessible` should be treated as if it wasn't defined at all. What I found out testing this is that so far, we have not been validating against the usage of unknown arguments (they were just ignored). So there are two tests in this commit: one for the general case, and one for inaccessible. The implementation is the same, becausue we were already (correctly) removing inaccessible arguments in the conversion from FederatedGraph to engine_v2::Schema.


yoav-lavi (approved):
Nice!

PR #2118, opened 2024-09-12
cli: replace handlebars with format!() in dev and federated dev

tomhoule:
The assumption is that we don't need HTML escaping there.


obmarg (approved):
💯

yoav-lavi (approved):

PR #2119, opened 2024-09-12
feat: implement a rolling logger with time and size limits

pimeys:
Rolling logger with more sanity compared to `tracing-appender` and all the other similar crates. No locks, no loops, just a simple strategy:

- Write to `access.log` from the start
- When reaching rotation, moves the `access.log` to `access.log.start_timestamp`
- Continue logging to `access.log`

Implements duration and size based rotation strategies.

Config syntax, rotate daily:

```toml
[gateway.access_logs]
enabled = true
path = "/foo/bar"
rotate = "daily"
```

rotate by max size

```toml
[gateway.access_logs]
enabled = true
path = "/foo/bar"
rotate.size = 10_000_000 # bytes
```


gateway/crates/rolling-logger/src/strategy.rs
```diff
@@ -0,0 +1,78 @@
+use std::time::{Duration, SystemTime};
+
+#[derive(Debug, Clone, Copy)]
+pub(crate) enum StrategyKind {
+    Never,
+    Minutely,
+    Hourly,
+    Daily,
+    MaxSize(u64),
```
tomhoule, line 9: Nitpick, but with the max size strategy, the log files are going to grow slightly larger than max size in most cases, since it rotates only after max size is exceeded. Maybe rename to size based?

tomhoule (approved):
Neat!

PR #2126, opened 2024-09-13
feat(gateway-config): Support specifying access log size as 2kiB

Finistere:
Support:

```toml
[gateway.access_logs]
enabled = true
path = "./logs"
rotate.size = "200MiB"
```


pimeys (approved):
this is really cool.

PR #2136, opened 2024-09-16
fix(engine-v2): fix endless recursion in object debug

pimeys:
Debug impl of an object adds the object fields to the output. If one of these fields references the same object, we have endless recursion.

This fix just prints the field name, not the whole field.


Finistere:
nice catch

PR #2142, opened 2024-09-18
Support inline fragments in selection sets of `@requires` and `@provides`

tomhoule:
... in composition and FederatedGraph.

Example: `myField String @requires("otherField { ... on Frog { jumpHeight } ... on Toad { wartsCount } }")`

For the federation audit. Technically, to fix the parent in engine-v2 we only need the graphql-federated-graph part (no composition), but we use our composition in integration-tests, and we want the feature anyway, so it's implemented in composition too in this PR.

closes GB-7563


engine/crates/federated-graph/src/federated_graph/v4.rs
```diff
@@ -122,6 +134,73 @@ pub enum Value {
     List(Box<[Value]>),
 }
 
+#[derive(serde::Serialize, serde::Deserialize, Clone)]
+pub struct Object {
```
tomhoule, line 138: I had to redefine these in v4.rs because they contain `Key`s, which contain the new type. That cascaded.

pimeys (approved):

PR #2150, opened 2024-09-19
feat: a federation audit test runner

obmarg:
We're working on fixing all the failing tests from [graphql-federation-gateway-audit](https://github.com/the-guild-org/graphql-federation-gateway-audit).  But working on fixing them is a bit annoying at the moment - you need to `cargo build` then copy a binary over then run a make command.  It's _fine_, but not the fastest feedback loop.

I really just want to use `cargo nextest` like we do with everything else.

Thankfully it's quite easy to integrate with the setup in `graphql-federation-gateway-audit` - behind all the makefiles & scripts there's a single server that exposes all the subgraphs & tests via an API.  

There's a few different components in this PR:

1. A custom test runner that reads test details from the federation-gateway-audit API and can run the tests against an inline engine, using the same setup we use for our integration-tests.
2. A test that keeps a cached list of the `federation-gateway-audit` tests up to date.  I found running `cargo nextest` annoyingly slow without this because every single test has to make like 15 http requests on startup to get the full list of tests (even though most of the invocations will only need a single test).
3. A test that keeps a locally cached copy of the `federation-gateway-audit` repository up to date, so nobody accidentally runs tests against an out of date copy.

Currently this expects users to run `federation-gateway-audit` in the background themselves (via `npm start serve`).  I'd like to automate this in a future PR, but finding a good way to do that across tests is proving annoying so wanted to get this in first.


tomhoule (approved):
Fantastic! Maybe I just missed that part, but we still have to copy the binary over for now, right?

pimeys:
Demo and discussion tomorrow 👏🏾

pimeys:
P.s. awesome!

engine/crates/federation-audit-tests/tests/checkout.rs
```diff
@@ -0,0 +1,78 @@
...
+use std::process::Command;
+
+#[test]
+fn ensure_upstream_is_up_to_date() {
+    // The tests in this crate are powered by the upstream graphql-federation-gateway-audit repo
+    //
+    // I don't want people to have to manually fetch that and keep it up to date, because that sucks.
+    //
+    // But I also don't want to do the update inline in the audit_tests - because they'll probably
+    // be run in parallel and who can be bothered with cross process locking.
```
Finistere, line 12: > who can be bothered with cross process locking.

was going to suggest that xD
Doing that kind of thing with `OnceLock` / `tokio::sync::OnceCell` in clickhouse-integration-tests

PR #2151, opened 2024-09-19
fix: plan from types if interface does not provide keys

pimeys:


engine/crates/engine-v2/src/operation/logical_planner/selection_set.rs
```diff
@@ -480,32 +511,48 @@ impl<'schema, 'a> SelectionSetLogicalPlanner<'schema, 'a> {
...
         // Inserting into its parent selection set in order.
         let mut field_ids = std::mem::take(
             &mut self.operation[parent_selection_set_id].field_ids_ordered_by_parent_entity_id_then_position,
         );
+
         let extra_parent_entity_id = Some(self.schema[definition_id].parent_entity_id);
         let extra_query_position = self.operation[id].query_position();
-        let i = field_ids
+
+        let extra_field_id = field_ids
```
Finistere, line 555: It's not the `extra_field_id` but the `extra_field_position` / `position`

engine/crates/engine-v2/src/operation/logical_planner/selection_set.rs
```diff
@@ -515,24 +562,30 @@ impl<'schema, 'a> SelectionSetLogicalPlanner<'schema, 'a> {
...
-        unplanned_fields: &HashMap<FieldId, FieldDefinition<'schema>>,
+        unplanned_fields: &mut HashMap<FieldId, FieldDefinition<'schema>>,
         planned_selection_set: &mut PlannedSelectionSet,
         candidates: &mut HashMap<ResolverDefinitionId, ChildPlanCandidate<'schema>>,
     ) -> LogicalPlanningResult<()>
     where
         'schema: 'field,
     {
-        for (&id, definition) in unplanned_fields {
+        let mut plan_changes = Vec::new();
```
Finistere, line 579: I would rename it to `interface_fields_to_replan` or something like this, `plan_changes` is very general while it's use is quite specific

engine/crates/engine-v2/src/operation/logical_planner/selection_set.rs
```diff
@@ -553,10 +606,153 @@ impl<'schema, 'a> SelectionSetLogicalPlanner<'schema, 'a> {
...
+    /// Replans the fields within a selection set by resetting their current plans and adding
+    /// new fields to be planned. This method may be invoked when the structure of the selection
+    /// set changes or when new fields become available for planning.
+    fn replan_fields(
+        &mut self,
+        selection_set_id: SelectionSetId,
+        plan_changes: Vec<(FieldId, Vec<FieldDefinition<'schema>>)>,
+        unplanned_fields: &mut HashMap<FieldId, FieldDefinition<'schema>>,
+    ) {
+        for (field_id, add_to_plan) in plan_changes {
```
Finistere, line 649: would have used `interface_field_id` instead of `field_id` and renamed the function `replan_interface_fields()`

engine/crates/engine-v2/src/operation/logical_planner/selection_set.rs
```diff
@@ -553,10 +606,153 @@ impl<'schema, 'a> SelectionSetLogicalPlanner<'schema, 'a> {
...
+            }
+        }
+    }
+
+    /// Resets the planning state for a given field identified by `field_id`.
+    ///
+    /// This function updates the logical plan ID and resolved requirements for the specified
+    /// field, allowing it to be re-planned in subsequent operations. If the field has a selection
+    /// set, this process is recursive, resetting all fields within that selection set.
+    fn reset_field_plans(&mut self, field_id: FieldId) {
```
Finistere, line 698: would have used `mark_field_as_never_planned()` or something like this. Not a "reset" imho

Finistere (approved):

PR #2160, opened 2024-09-20
fix: union interface queries need to be explicit with types

pimeys:
What we do here:

- subgraph can query only the types it implements, so we need to make sure we render fragments correctly
- we have now information is interface implemented for all available types, this is based on `@join__implements` directive
- composition now renders `@join__implements` correctly for objects
- and also for interfaces, which we are not currently using but might be required in the future


engine/crates/composition/src/ingest_subgraph.rs
```diff
@@ -1,4 +1,3 @@
-//! This is a separate module because we want to use only the public API of [Subgraphs] and avoid
```
tomhoule, line 1: I think this line was deleted by mistake

engine/crates/composition/src/emit_federated_graph.rs
```diff
@@ -169,17 +169,27 @@ fn emit_input_value_definitions(input_value_definitions: &[InputValueDefinitionI
...
-                ctx.out.objects[object_id.0].implements_interfaces.push(implementee);
+                let object = &mut ctx.out.objects[object_id.0];
+                object.implements_interfaces.push(implementee);
+
+                for subgraph_id in ctx
+                    .subgraphs
+                    .subgraphs_implementing_interface(implementee_name, implementer_name)
+                {
+                    object
+                        .join_implements
```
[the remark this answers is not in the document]
tomhoule: Yes I'll simplify later.

tomhoule (commented):
Are we sure other implementations _always_ emit join__implements, and not only when the interface isn't implemented in all subgraphs? I haven't checked

pimeys:
> Are we sure other implementations _always_ emit join__implements, and not only when the interface isn't implemented in all subgraphs? I haven't checked

Yes. We can find one example in the guild tests.

Subgraph A: https://github.com/the-guild-org/graphql-federation-gateway-audit/blob/main/src/test-suites/requires-interface/a.subgraph.ts
Subgraph B: https://github.com/the-guild-org/graphql-federation-gateway-audit/blob/main/src/test-suites/requires-interface/b.subgraph.ts

Now, let's change this, let's add the `SecondAddress` in subgraph A implementing the `Address` interface. We have in A:
```graphql
interface Address {
  id: ID!
}

type HomeAddress implements Address @key(fields: "id") {
  id: ID!
  city: String @shareable
}

type WorkAddress implements Address @key(fields: "id") {
  id: ID!
  city: String @shareable
}

type SecondAddress implements Address @key(fields: "id") {
  id: ID!
  city: String @shareable
}
```
and in B:
```graphql
interface Address {
  id: ID!
}

type HomeAddress implements Address @key(fields: "id") {
  id: ID!
  city: String @shareable
}

type WorkAddress implements Address @key(fields: "id") {
  id: ID!
  city: String @shareable
}

type SecondAddress implements Address @key(fields: "id") {
  id: ID!
  city: String @shareable
}
```
This composes into the following types in the supergraph:
```graphql
type HomeAddress implements Address
  @join__implements(graph: A, interface: "Address")
  @join__implements(graph: B, interface: "Address")
  @join__type(graph: A, key: "id")
  @join__type(graph: B, key: "id")
{
  id: ID!
  city: String
}

type SecondAddress implements Address
  @join__implements(graph: A, interface: "Address")
  @join__implements(graph: B, interface: "Address")
  @join__type(graph: A, key: "id")
  @join__type(graph: B, key: "id")
{
  id: ID!
  city: String
}

type WorkAddress implements Address
  @join__implements(graph: A, interface: "Address")
  @join__implements(graph: B, interface: "Address")
  @join__type(graph: A, key: "id")
  @join__type(graph: B, key: "id")
{
  id: ID!
  city: String
}
```
The composition is done if you execute `make run-grafbase-gateway TEST_SUITE=requires-interface`. You find the composed schema in `gateways/grafbase-gateway/supergraph.graphql`.

tomhoule:
Sounds good :+1: Weird that `@join__implements` is always there but `@join__field` only when it isn't defined in all subgraphs.

engine/crates/federated-graph/src/from_sdl.rs
```diff
@@ -349,6 +352,80 @@ fn ingest_object_interfaces(
...
+        Some(Definition::Interface(interface_id)) => *interface_id,
+        _ => {
+            let error = DomainError("Broken invariant: join__implements points to a non-interface type".to_owned());
+
+            return Err(error);
+        }
+    };
+
+    Ok(Some((subgraph_id, interface_id)))
+}
```
tomhoule, line 427: Nice :+1:

engine/crates/engine-v2/src/sources/graphql/request/prepare.rs
```diff
@@ -196,23 +258,93 @@ impl QueryBuilderContext {
...
+                entity,
+                self.subgraph_id,
+                selection_set,
+            );
+
+            match rendering {
+                SelectionSetRendering::InterfaceWithPartialFragment(interface, subgraph_id) => {
+                    for r#type in interface.possible_types_ordered_by_typename() {
+                        if !r#type.subgraph_implements_interface(&subgraph_id, &interface.id()) {
+                            continue;
```
tomhoule, line 279: I don't understand this part — do we skip rendering the inline fragment in the subgraph request if we can't be sure the subgraph has all the implementers of the interface?

tomhoule (approved):
I'm not sure I understand the engine-v2 part, approving for the composition part

engine/crates/engine-v2/schema/src/builder/graph.rs
```diff
@@ -364,6 +384,46 @@ impl<'a> GraphBuilder<'a> {
             }
         }
 
+        // Adding all fully implemented interfaces per subgraph.
+        for interface_id in (0..self.graph.interface_definitions.len()).map(InterfaceDefinitionId::from) {
+            // Collecting per subgraph all objects defining this interface.
+            let mut found_ids: HashMap<SubgraphId, Vec<ObjectDefinitionId>> = HashMap::new();
```
Finistere, line 390: I don't think this algorithm isn't accurate. The following wouldn't be correctly taken into account:
```graphql
type Horse @join_type(graph: A) @join_implements(graph: A, interface: "Animal") implements Animal

type Dog @join_type(graph: B) @join_implements(graph: B, interface: "Animal") implements Animal

interface Animal
```

`Dog` is only available in subgraph `B`, so within `B` all the objects implement the interface `Animal` and there is no need to keep track of it.

IMHO the algorithm should do:
```
for interface in ... {
  let not_fully_implement_in_subgraphs: HashSet<SubgraphId> = HashSet::new();
  for object in interface.possible_types {
    for subgraph in object.subgraph {
        if does_not_implement_interface_in_subgraph(object, interface, subgraph) {
          not_fully_implement_in_subgraphs.insert(subgraph)
        }
    }
  }
}
```

Your algorithm is stricter, it requires that `Horse` implements `Animal` even in `B` where it's not present.

engine/crates/engine-v2/codegen/domain/schema.graphql
```diff
@@ -52,6 +59,7 @@ type InterfaceDefinition
   possible_types: [ObjectDefinition!]!
   possible_types_ordered_by_typename: [ObjectDefinition!]!
   directives: [TypeSystemDirective!]!
+  fully_implemented_in: [Subgraph!]!
```
Finistere, line 62: nitpick: As you rely on it, can you add a comment that it's sorted?

engine/crates/engine-v2/src/sources/graphql/request/prepare.rs
```diff
@@ -29,7 +33,8 @@ impl PreparedGraphqlOperation {
         operation_type: OperationType,
         plan: PlanWalker<'_>,
     ) -> Result<PreparedGraphqlOperation, Error> {
-        let mut ctx = QueryBuilderContext::default();
+        let mut ctx = QueryBuilderContext::new(plan.subgraph_id());
```
Finistere, line 36: IMO it would be better to provide `subgraph_id` to the `build()` function instead of adding subgraph_id & endpoint to the PlanWalker. The caller is `GraphqlResolver` which has `definition: GraphqlRootFieldResolverDefinition<'_>` with everything you need already. Same comment for the `PreparedFederationEntityOperation` underneath. That would also guarantee you to you have `SubgraphId` instead of an option.

engine/crates/engine-v2/src/sources/graphql/request/prepare.rs
```diff
@@ -164,6 +224,7 @@ impl QueryBuilderContext {
 
     fn write_selection_set(
         &mut self,
+        maybe_selection_set_type: Option<SelectionSetType>,
```
Finistere, line 227: You can remove the `maybe_entity_id` it's the same thing as `maybe_selection_set_type`, only makes the signature heavier than necessary and it's confusing as I'm not even sure you're providing the same thing for both. IMHO

engine/crates/engine-v2/src/sources/graphql/request/prepare.rs
```diff
@@ -196,23 +258,93 @@ impl QueryBuilderContext {
...
-                )?;
-                buffer.indent += 1;
-            }
-            for field in fields {
-                self.write_field(buffer, field)?;
-            }
-            if maybe_entity_id != Some(entity_id) {
-                buffer.indent -= 1;
-                indent_write!(buffer, "}}\n")?;
+            let fields = fields.collect_vec();
```
Finistere, line 263: Why allocate a vec? `write_entity_fields` can just accept an `impl Iterator<Item = PlanWalker<'_, FieldId>>`

engine/crates/engine-v2/src/sources/graphql/request/prepare.rs
pimeys: So what happens here is we have something like a union in the return type, or interface. Let's say this is `Product`, which is a union.

We do a query:

```graphql
query {
  products {
    ... on Node {
      id
    }
  }
}
```

This now went to subgraph B as-is. Subgraph B has ovens and toasters, but only toaster implements `Node` in B. Both, the ovens and toasters are part of the `Product` union.

This is our federated query:

```graphql
type Query
  @join__type(graph: A)
  @join__type(graph: B)
{
  products: [Product] @join__field(graph: A)
}
```

And in subgraph A this looks like:

```graphql
type Query {
  products: [Product]
}
```

So we try to query with an interface fragment. But the interface is implemented only in federated level, not in all subgraphs. The subgraph A returns products, which can be of oven or of toaster. If we send down to the subgraph the query as-is (as we do in main right now):

```graphql
query {
  products {
    ... on Node {
      id
    }
  }
}
```

This would return an error because the query returns ovens and toasters, but oven does not implement node in A:

```json
{
  "data": {
    "products": [
      null,
      null,
      {
        "id": "toaster1"
      },
      {
        "id": "toaster2"
      }
    ]
  },
  "errors": [
    {
      "message": "Error decoding response from upstream: Missing required field named 'id' at line 1 column 23",
      "locations": [
        {
          "line": 2,
          "column": 3
        }
      ],
      "path": [
        "products",
        0
      ],
      "extensions": {
        "code": "SUBGRAPH_INVALID_RESPONSE_ERROR"
      }
    },
    {
      "message": "Error decoding response from upstream: Missing required field named 'id' at line 1 column 26",
      "locations": [
        {
          "line": 2,
          "column": 3
        }
      ],
      "path": [
        "products",
        1
      ],
      "extensions": {
        "code": "SUBGRAPH_INVALID_RESPONSE_ERROR"
      }
    }
  ]
}
```

Now, what we do in these first three matches is we take a look into the query fragment, and then we check this subgraph does it implement all possible instances of the interface. If not, we change the query to explicitly expand into all the fragments, so the correct query to the subgraph becomes:

```graphql
query {
  products {
    ... on Oven {
      id
    }
    ... on Toaster {
      id
    }
  }
}
```

```json
{
  "data": {
    "products": [
      {
        "id": "oven1"
      },
      {
        "id": "oven2"
      },
      {
        "id": "toaster1"
      },
      {
        "id": "toaster2"
      }
    ]
  }
}
```

Which returns the correct data. This is now repeated in the first three matches so that we never try to send a fragment to this subgraph that it cannot represent.

engine/crates/engine-v2/src/sources/graphql/request/prepare.rs
```diff
@@ -196,23 +258,93 @@ impl QueryBuilderContext {
...
+                        self.write_type_fields(buffer, r#type.name(), &fields)?;
+                    }
+                }
+                SelectionSetRendering::UnionWithPartialFragment(union) => {
+                    for r#type in union.possible_types_ordered_by_typename() {
+                        self.write_type_fields(buffer, r#type.name(), &fields)?;
+                    }
+                }
+                SelectionSetRendering::InterfaceWithObjects(interface_id, subgraph_id) => {
+                    if let EntityDefinition::Object(ref object) = entity {
```
Finistere, line 291: IMHO would be clearer if all the logic for "SelectionSetRendering" was in a single place. It's not obvious to understand here in which case we are with `InterfaceWithObjects`. Maybe just a nitpick, so feel free to ignore me, but I feel like something like this would be clearer to understand and follow:
```rust
match selection_set_type {
   None => ...
   Some(SelectionSetType::Interface(parent)) => match entity {
      EntityDefinitionId::Interface(child) => ...
   }
   ...
}

```
instead of separating the logic somewhere with an `entity_id` and `subgraph_id` match which can be removed.

Finistere (approved):

engine/crates/engine-v2/src/sources/graphql/request/prepare.rs
Finistere: I think that's false at this point we know already that not every object in this subgraph implements the interface. But we don't want to skip them. you must filter on object that are in said subgraph. So:

```rust
                   for object in interface.possible_types_ordered_by_typename() {
                        if !object.only_resolvable_in.contains(subgraphs) && !object.only_resolvable_in.is_empty() {
                            continue;
                        }
```

engine/crates/engine-v2/src/sources/graphql/request/prepare.rs
```diff
@@ -196,23 +258,93 @@ impl QueryBuilderContext {
...
+                        if !r#type.subgraph_implements_interface(&subgraph_id, &interface.id()) {
+                            continue;
+                        }
+
+                        self.write_type_fields(buffer, r#type.name(), &fields)?;
+                    }
+                }
+                SelectionSetRendering::UnionWithPartialFragment(union) => {
+                    for r#type in union.possible_types_ordered_by_typename() {
+                        self.write_type_fields(buffer, r#type.name(), &fields)?;
```
Finistere, line 287: Here it's also broken and hidden by the logic being separated. This case is parent is an union and child an interface. So you should end up  with the same code as before if the interface isn't fully implemented in the subgraph

engine/crates/engine-v2/src/sources/graphql/request/prepare.rs
Finistere: However, you can use `subgraph_implements_interface` to create fewer fragments. You only need to add object type conditions for those that don't implement the interface so:
```rust
add_interface_fragment = false
for object in interface.possible_types_ordered_by_typename() { 
   if !object.only_resolvable_in.contains(subgraphs) && !object.only_resolvable_in.is_empty() { continue; }
   if !r#type.subgraph_implements_interface(&subgraph_id, &interface.id()) {
     write_entity_fields(...) // for object
   } else {
add_interface_fragment = true
}
   
}

if add_interface_fragment {
     write_entity_fields(...) // for object
}
```

engine/crates/engine-v2/src/sources/graphql/request/prepare.rs
Finistere: Ah got it

PR #2171, opened 2024-09-23
fix: remove gateway access token cli arg

obmarg:
A customer was trying to create their own docker image for running `grafbase-gateway` last week, but they were running into some confusing 401s.  

This turned out to be because they were using our base image, which sets the docker entrypoint to `grafbase-gateway`, but also providing `grafbase-gateway` in the `CMD`.  Essentially running `grafbase-gateway grafbase-gateway ...`.  Normally I'd expect this to complain about a bad CLI argument, but the gateway is set up to accept `GRAFBASE_ACCESS_TOKEN` as a positional argument, so the second `grafbase-gateway` was being taken as the token, and they were getting a 401.

This was quite confusing, and it's not really best practice to accept a token as a CLI argument either for security purposes.  So, this PR switches it to _just_ rely on the `GRAFBASE_ACCESS_TOKEN` environment variable instead.

### DIFF
--- a/gateway/crates/gateway-binary/src/args/std.rs
+++ b/gateway/crates/gateway-binary/src/args/std.rs
@@ -24,7 +24,6 @@ use super::{log::LogStyle, LogLevel};
     group(
         ArgGroup::new("graph-ref-with-access-token")
             .args(["graph_ref"])
-            .requires("grafbase_access_token")
     )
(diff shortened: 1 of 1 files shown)


tomhoule (approved):
wow. I wasn't even aware we had that.

PR #2175, opened 2024-09-25
`@skip` and `@include` support for Federation

yoav-lavi:
# Description

## Features 
- Implements `@skip` and `@include` for fields, inline fragments and fragment spreads (for Federation)

# Type of change

- [ ] 💔 Breaking
- [x] 🚀 Feature
- [ ] 🐛 Fix
- [ ] 🛠️ Tooling
- [ ] 🔨 Refactoring
- [ ] 🧪 Test
- [ ] 📦 Dependency


engine/crates/engine-v2/src/operation/bind/selection_set.rs
```diff
@@ -159,7 +185,7 @@ impl<'schema, 'p, 'binder> SelectionSetBinder<'schema, 'p, 'binder> {
         for Positioned { node: selection, .. } in &selection_set.items {
             match selection {
                 engine_parser::types::Selection::Field(field) => {
-                    self.register_field(ty, field)?;
+                    self.register_field(ty, field, parent_query_modifier_indicators.clone())?;
```
yoav-lavi, line 188: I don't love this clone but not sure we have a good alternative here

engine/crates/engine-v2/src/response/write/deserialize/object/concrete.rs
```diff
@@ -218,7 +228,7 @@ impl<'de, 'ctx, 'seed> ConcreteObjectFieldsSeed<'ctx, 'seed> {
             }
         }
 
-        if response_fields.len() < self.field_shape_ids.len() {
+        if response_fields.len() < self.field_shape_ids.len() - self.skipped_ids.len() {
```
yoav-lavi, line 231: Since we know all skipped fields will be omitted this should theoretically be fine (and not mask any field that should be there being missing)

gateway/crates/gateway-binary/tests/integration_tests.rs
```diff
@@ -195,6 +196,25 @@ impl Client {
         &self.endpoint
     }
 
+    pub fn gql_with_variables<Response>(
```
yoav-lavi, line 199: I didn't end up using this but I guess it could be useful later

engine/crates/engine-v2/src/response/write/deserialize/object/concrete.rs
yoav-lavi: @Finistere please take a second look at this to make sure I'm not missing a case, thanks!

engine/crates/engine-v2/schema/src/lib.rs
```diff
@@ -269,3 +269,12 @@ impl ScalarType {
...
 }
+
+impl Graph {
+    pub fn scalar_definition_by_name(&self, name: &str) -> Option<ScalarDefinitionId> {
+        self.scalar_definitions
+            .iter()
+            .position(|definition| definition.ty.to_string() == name)
+            .map(ScalarDefinitionId::from)
+    }
+}
```
pimeys, line 280: We could do either a lookup map for these in the graph, or we could just mark the ID of boolean when generating the schema?

pimeys: And could this return `DefinitionId` already?

pimeys (commented):
Oh yeah and approving so I'm not blocking with this review. @Finistere might want to take a look still?

pimeys (approved):

engine/crates/engine-v2/schema/src/lib.rs
yoav-lavi: For the latter sure, for the former this was a suggestion from @Finistere so there might be some reason not to do so, but I agree in general. @Finistere WDYT?

We could probably do this separately though to avoid delaying

engine/crates/engine-v2/src/operation/bind/selection_set.rs
```diff
@@ -27,7 +28,11 @@ pub(super) struct SelectionSetBinder<'schema, 'parsed, 'binder> {
     #[allow(clippy::type_complexity)]
     fields: HashMap<
         (SafeResponseKey, FieldDefinitionId),
-        (QueryPosition, Vec<&'parsed Positioned<engine_parser::types::Field>>),
+        (
+            QueryPosition,
+            Vec<&'parsed Positioned<engine_parser::types::Field>>,
+            Vec<QueryModifierIndicator>,
+        ),
```
obmarg, line 35: I may be misunderstanding this, but what happens if the same field appears under completely separate `QueryModifierIndicator`s?  Can this handle that?

obmarg: I wrote a failing test that suggests it can't:

```
#[test]
fn skip_include_again() {
    let response = runtime().block_on(async move {
        let engine = Engine::builder().with_subgraph(FakeGithubSchema).build().await;

        engine
            .post(
                r#"
                    query PullRequests($a: Boolean!, $b: Boolean!) {
                        pullRequestsAndIssues(filter: { search: "1" }) {
                            ... on PullRequest @include(if: $a) {
                                title
                            }
                            ... on PullRequest @include(if: $b) {
                                title
                            }
                        }
                    }
                    "#,
            )
            .variables(json!({"a": true, "b": false}))
            .await
    });

    insta::assert_json_snapshot!(response, @r###"
    {
      "data": {
        "pullRequestsAndIssues": [
          {
            "title": "Creating the thing"
          },
          {
            "title": "Some bot PR"
          },
          {}
        ]
      }
    }
    "###);
}
```

yoav-lavi: @obmarg is the snapshot what you expect or the result?

obmarg: The snapshot is what should be returned.  But it just returns a list of empty objects.

obmarg: You can copy and paste that into your test file and it should be runnable if you want to see.

yoav-lavi: Theoretically this is the same as 

```
{
  title @include(if: true)
  title @include(if: false)
 }
 ```
 which are merged to one field (as in your example) - When discussing with @Finistere we said that if any directive sets the field to skip that should affect all directives (which is the behavior when you set multiple directives or both include and skip to the same field) - So as far as I can tell it's working as expected

Comment (yoav-lavi) [reply engine/crates/engine-v2/src/operation/bind/selection_set.rs:35]: The spec doesn't really define the former, it does define the latter (multiple on one field rather than multiple instances of one field)

Comment (yoav-lavi) [reply engine/crates/engine-v2/src/operation/bind/selection_set.rs:35]: In general the behavior is that a single true (or false for include) for a field overrides anything else

Comment (yoav-lavi) [reply engine/crates/engine-v2/src/operation/bind/selection_set.rs:35]: (but it does consider all rules for the same field if that's what you're asking, or at least should unless I missed something)

Comment (obmarg) [reply engine/crates/engine-v2/src/operation/bind/selection_set.rs:35]: That is not how skip and include are specified as working though.

Comment (yoav-lavi) [reply engine/crates/engine-v2/src/operation/bind/selection_set.rs:35]: e.g. 

```
{ 
  title @skip(if: false) @include(if:true)
 }
 ```
 and so forth will look at all directives and omit the field only if one of them is the "off" value for the relevant directive

yoav-lavi: @obmarg how do you mean?

>Neither @skip nor @include has precedence over the other. In the case that both the @skip and @include directives are provided on the same field or fragment, it must be queried only if the @skip condition is false and the @include condition is true. Stated conversely, the field or fragment must not be queried if either the @skip condition is true or the @include condition is false.

Regarding merging multiple instances of the same field it doesn't really specify anything, and that's how our engine works (as in there isn't multiple instances of the same field when the query is planned, they're merged)

obmarg: The spec is pretty clear that you look at each individual selection individually and decide whether or not to include that selection based on the directive that is applied to it.  See step 3.a & 3.b here: http://spec.graphql.org/October2021/#CollectFields()

yoav-lavi: Actually now that I reread our conversation @Finistere I think you might've meant this as well. I'd misread since I'd been going by the behavior of multiple instances on the same field. I'll look into this tomorrow

yoav-lavi: Good catch @obmarg

engine/crates/engine-v2/src/operation/bind/modifier.rs
```diff
@@ -14,6 +13,30 @@ impl<'schema, 'p> super::Binder<'schema, 'p> {
...
+        input_value_ids: Vec<(QueryInputValueId, FieldSkippingDirective)>,
+    ) {
+        self.generate_modifiers_for_type_system_directives(field_id, argument_ids, field_definition);
+        self.generate_modifiers_for_executable_directives(field_id, input_value_ids);
+    }
+
+    fn generate_modifiers_for_executable_directives(
+        &mut self,
+        field_id: FieldId,
+        input_value_ids: Vec<(QueryInputValueId, FieldSkippingDirective)>,
```
Finistere, line 25: could just be `QueryModifierRule`

engine/crates/engine-v2/src/response/write/deserialize/object/concrete.rs
Finistere: looks sane to me, but you don't need to allocate array, just the count is enough

engine/crates/engine-v2/src/operation/bind/field.rs
```diff
@@ -35,6 +36,7 @@ impl<'schema, 'p> Binder<'schema, 'p> {
         definition_id: FieldDefinitionId,
         Positioned { pos, node: field }: &'p Positioned<engine_parser::types::Field>,
         selection_set_id: Option<SelectionSetId>,
+        query_modifier_indicators: Vec<QueryModifierIndicator>,
```
Finistere, line 39: Could just be `QueryModifierRule` by coercing them earlier

Finistere (approved):

engine/crates/engine-v2/src/operation/bind/selection_set.rs
```diff
@@ -258,7 +298,38 @@ impl<'schema, 'p, 'binder> SelectionSetBinder<'schema, 'p, 'binder> {
...
+
+    fn directives_to_query_modifier_indicators(
+        &mut self,
+        directives: &[Positioned<Directive>],
+    ) -> BindResult<Vec<QueryModifierIndicator>> {
+        let mut indicators = Vec::new();
+        for directive in directives {
+            let directive_name = directive.name.node.as_str();
+            if matches!(directive_name, "skip" | "include") {
+                let argument = directive.arguments.first().expect("must exist");
```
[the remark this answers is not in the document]
yoav-lavi: Done

engine/crates/engine-v2/src/operation/modifier/mod.rs
```diff
@@ -26,6 +32,10 @@ pub(crate) enum QueryModifierRule {
         directive_id: AuthorizedDirectiveId,
         definition_id: DefinitionId,
     },
+    Skip {
```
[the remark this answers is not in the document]
yoav-lavi: I'd rather avoid having `"skip"` and `"include"` as raw strings everywhere though, so I think we should  have at least one enum that defines these

engine/crates/engine-v2/schema/src/lib.rs
```diff
@@ -269,3 +269,12 @@ impl ScalarType {
         })
     }
 }
+
+impl Graph {
+    pub fn scalar_definition_by_name(&self, name: &str) -> Option<ScalarDefinitionId> {
+        self.scalar_definitions
+            .iter()
+            .position(|definition| definition.ty.to_string() == name)
```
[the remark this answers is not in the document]
yoav-lavi: The graph can't be indexed by StringId though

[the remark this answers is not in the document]
yoav-lavi: Done

engine/crates/engine-v2/src/operation/modifier/mod.rs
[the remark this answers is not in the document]
Finistere: I don't understand your point with strings. You just need `QueryModifierRule::Skip` and `QueryModifierRule::Include` variants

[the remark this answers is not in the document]
Finistere: I think you can remove a lot of enums/structs with this and coercing as soon as possible

engine/crates/engine-v2/src/response/write/deserialize/object/concrete.rs
yoav-lavi: Done

engine/crates/engine-v2/src/operation/bind/selection_set.rs
yoav-lavi: Done

engine/crates/engine-v2/src/operation/bind/selection_set.rs
```diff
@@ -242,7 +279,10 @@ impl<'schema, 'p, 'binder> SelectionSetBinder<'schema, 'p, 'binder> {
             })?;
 
         let ty = self.bind_selection_set_type(parent, &fragment.node.type_condition)?;
-        self.register_selection_set_fields(ty, &fragment.node.selection_set)?;
+
+        let query_modifier_indicators = self.directives_to_query_modifier_indicators(&spread.directives)?;
```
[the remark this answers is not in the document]
yoav-lavi: Done

engine/crates/engine-v2/src/operation/bind/selection_set.rs
yoav-lavi: Fixed

engine/crates/engine-v2/schema/src/lib.rs
yoav-lavi: Fixed the definition ID issue, we'll tackle creating a lookup map separately

engine/crates/engine-v2/src/operation/modifier/query.rs
```diff
@@ -150,6 +153,51 @@ where
...
+            // additional query modifiers guaranteed to be for different instances of the field.
+            // since we merge the field instances anyhow, any one instance of a field that should not be skipped is enough to include the field. (we use `all` below to avoid a case of no rules)
+            // within the same field instance, more than one value ID means the others are derived from parent fragments, so we should check if any of the values mark the field as skipped.
+            // (which is the inverse logic of multiple instances of the same field)
+            let skip_field = self
+                .operation
+                .query_modifiers
+                .iter()
+                .filter(|modifier| matches!(modifier.rule, QueryModifierRule::SkipInclude { .. }))
+                .all(|modifier| match &modifier.rule {
```
Finistere, line 170: would have used `filter_map(...).reduce(|(a, b)| a && b)`

engine/crates/engine-v2/src/operation/bind/selection_set.rs
```diff
@@ -151,21 +157,24 @@ impl<'schema, 'p, 'binder> SelectionSetBinder<'schema, 'p, 'binder> {
         &mut self,
         ty: SelectionSetType,
         selection_set: &'p Positioned<engine_parser::types::SelectionSet>,
+        parent_directive_value_ids: (&[QueryInputValueId], &[QueryInputValueId]),
```
Finistere, line 160: would be simpler to read with something like:

```rust
struct ExecutableDirectives {
  skip_input_values_ids: Vec<QueryInputValueId>,
  include_input_value_ids: Vec<QueryInputValueId>
}

impl ExecutableDirectives {
  fn extend(&mut self, parent_directives: &ExecutableDirectives) { ... }
}
```

engine/crates/engine-v2/src/operation/bind/selection_set.rs
```diff
@@ -251,14 +289,62 @@ impl<'schema, 'p, 'binder> SelectionSetBinder<'schema, 'p, 'binder> {
...
+        include_input_value_ids.extend_from_slice(parent_directive_value_ids.1);
+
+        self.register_selection_set_fields(
+            ty,
+            &fragment.selection_set,
+            (&skip_input_value_ids, &include_input_value_ids),
+        )
+    }
+
+    fn directives_to_value_ids(
```
Finistere, line 315: would name `bind_executable_directives` or smth like this imho with the proposed struct a bit before

yoav-lavi: Done

engine/crates/engine-v2/src/operation/bind/selection_set.rs
yoav-lavi: Done

engine/crates/engine-v2/src/operation/modifier/query.rs
yoav-lavi: `any` will short circuit in this case, so it'd possibly be more efficient

PR #2176, opened 2024-09-25
test: use audit tests to compare composition outputs

obmarg:
Runnable with

```sh
CLICOLOR_FORCE=1 cargo nextest run -p federation-audit-tests  --no-fail-fast --test composition_comparisons --run-ignored all
```

### DIFF
--- a/engine/crates/engine-v2/codegen/src/loader.rs
+++ b/engine/crates/engine-v2/codegen/src/loader.rs
@@ -81,7 +81,7 @@ pub(super) fn load(path: PathBuf) -> anyhow::Result<domain::Domain> {
                         // Add any explicitly defined field name or leave empty to be generated
                         // afterwards.
                         record_field_name: parse_record_field_name(field.directives()).unwrap_or_default(),
-                        description: field.description().map(|s| s.description().to_cow().into_owned()),
+                        description: field.description().map(|s| s.to_string()),
                         type_name: field.ty().name().to_string(),
                         wrapping: field.ty().wrappers().collect(),
                     })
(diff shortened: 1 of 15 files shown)


engine/crates/engine-v2/codegen/src/loader.rs
```diff
@@ -226,7 +226,7 @@ fn parse_union_kind(name: &str, directives: Iter<'_, Directive<'_>>) -> domain::
         }
     } else {
         domain::UnionKind::Record(domain::RecordUnion {
-            indexed: parse_indexed(name, directives),
+            indexed: parse_indexed(name, directives.clone()),
```
yoav-lavi, line 229: Do we need this clone?

tomhoule (approved):
:fire:

yoav-lavi (approved):
Looks good, see comment

engine/crates/engine-v2/codegen/src/loader.rs
yoav-lavi: As in do we want to add a clone here specifically for testing? It's probably a non issue but thought I'd ask

obmarg: We're not adding a clone specifically for testing.  We are adding a clone because I updated `cynic-parser` and `Iter` (which directives is an instance of) is no longer `Copy`.  This code will not compile if I remove the `clone`.  So yes we need this clone.

It is incredibly inexpensive, and this is also just codegen code that is not performance critical.  

tldr: yes

yoav-lavi: I see, was wondering why it changed. Ok sounds good

PR #2179, opened 2024-09-27
feat(engine-v2): parse and compose `@join__unionMember` directives

pimeys:
This will be needed when the planning work is done.


tomhoule (approved):

engine/crates/engine-v2/codegen/domain/schema.graphql
```diff
@@ -113,12 +113,19 @@ type EnumValue @meta(module: "enum_value") @indexed(id_size: "u32", max_id: "MAX
   directives: [TypeSystemDirective!]!
 }
 
+type JoinMemberDefinition @meta(module: "join_member", derive: ["Clone"]) {
```
Finistere, line 116: for stuff that is obviously `Copy` you can use `@copy` which will add `Copy` to the struct and generate a walker that copies the struct rather than keep a ref.

PR #2180, opened 2024-09-27
graphql-federated-graph: introduce TypeDefinition

tomhoule:
This is a refactoring I have had on my mind for quite some time. It should enable simpler code in from_sdl(), rendering, as well as composition eventually.

The main idea is that we need a Definition or TypeDefinition entity, like in other parts of the codebase (graphql-schema-diff for example). That definition can hold a name, a directive and a description. For scalars, enums and union, that would be the whole of the Scalar, Enum or Union struct, so these can eventually disappear. That will be more concise and enable a lot more code reuse between different kinds of definitions.

This PR is only a very partial first step, because I want to get feedback before going further.

The other important aspect of this PR is the federated_graph::v4::view module. This is an abstraction I have been thinking about for a while as an alternative to the walker pattern. In the long term, we should either keep this pattern if it is nice to work with, or switch to the same approach to code generation for schema walkers as engine-v2. In the short term, this is already an improvement on the status quo.


engine/crates/federated-graph/src/render_sdl/render_federated_sdl.rs
```diff
@@ -203,7 +207,12 @@ pub fn render_federated_sdl(graph: &FederatedGraph) -> Result<String, fmt::Error
         let mut members = union.members.iter().peekable();
 
         while let Some(member) = members.next() {
-            sdl.push_str(&graph[graph[*member].name]);
+            sdl.push_str(
+                graph
+                    .through(*member)
+                    .through(|member| member.type_definition_id)
+                    .str(|def| def.name),
```
tomhoule, line 214: We could make this a bit nicer to read with one of the derive crates for getters:

```rust
graph
	.through(*member)
        .through(Object::type_definition_id)
        .str(TypeDefinition::name)
```

engine/crates/federated-graph/src/federated_graph/v4/ids.rs
```diff
@@ -0,0 +1,38 @@
...
+            impl std::ops::IndexMut<$name> for FederatedGraph {
+                fn index_mut(&mut self, index: $name) -> &mut $out {
+                    &mut self.$storage[index.0]
+                }
+            }
+        )*
+    }
+}
+
+id_newtypes!(type_definitions[TypeDefinitionId] -> TypeDefinitionRecord,);
```
tomhoule, line 38: The other ids are still defined in older versions of FederatedGraph, but these are going away soon, so let's make a place to define them here.

engine/crates/composition/src/emit_federated_graph.rs
```diff
@@ -244,7 +245,7 @@ fn emit_fields<'a>(
         let mut end_field_id = None;
 
         if let federated::Definition::Object(id) = definition {
-            let object_name = ctx.out.objects[id.0].name;
+            let object_name = ctx.out.through(id).view(|obj| obj.type_definition_id).name;
```
obmarg, line 248: Been staring at for 5 minutes or so, but not sure I get this through/view stuff...

tomhoule: Hehe that's not a good sign. The `.view()` methods are intended to return a `View`, that is basically a record struct + its id. The `through` methods are just a shortcut (with the `NestedView` helper) to avoid having to write `graph.view(graph.view(interface_id).type_definition_id)`.

And the `.str()` methods are a shortcut to avoid having to write: `let interface_name = &graph[graph.view(graph.view(interface_id).type_definition_id).name]`. Which is a line I wrote in this PR before this abstraction, but that I found disgusting.

tomhoule: We could unify `View` and `NestedView` if we made `View` more like the other abstractions we have and added a `graph: &FederatedGraph` field to it.

tomhoule: Maybe "follow" is a better name than "through". Basically take an id, dereference the FederatedGraph with it and the next view in the chain is whatever got dereferenced.

tomhoule: I think we could maybe decouple the closure part and make it its own method, have something more orthogonal. I'll give that a try and see if it ends up more readable.

tomhoule: I rearranged the methods on View & co a bit so instead of `view()`, `through()` and `str()` with different signatures between FederatedGraph and NestedView, we only have `FederatedGraph::at()` and `NestedView::then()`, plus some more `Deref` magic.

The change is all in this commit: https://github.com/grafbase/grafbase/pull/2180/commits/b6f8cbf094a68fa8b78f8110b96676af297636a5

engine/crates/composition/src/emit_federated_graph.rs
```diff
@@ -248,7 +249,7 @@ fn emit_fields<'a>(
         let mut end_field_id = None;
 
         if let federated::Definition::Object(id) = definition {
-            let object_name = ctx.out.objects[id.0].name;
+            let object_name = ctx.out.at(id).then(|obj| obj.type_definition_id).name;
```
obmarg, line 252: Yeah this seems more obvious for sure.

obmarg (approved):
TypeDefinition changes seem like a good move to me.

The View/ViewNested stuff is... interesting.  Happy to see how it works out at least

engine/crates/federated-graph/src/federated_graph/v4/view.rs
```diff
@@ -0,0 +1,97 @@
...
+    }
+
+    /// View the record with the given ID.
+    pub fn view<Id, Record>(&self, id: Id) -> View<Id, &Record>
+    where
+        Id: Copy,
+        FederatedGraph: Index<Id, Output = Record>,
+    {
+        View { id, record: &self[id] }
+    }
```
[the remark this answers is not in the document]
tomhoule: Yes I was starting to have the same thoughts. You'd only want `View` if you don't want to store references to the graph everywhere when you're passing `NestedView`s around. But we don't pass these around much, and that's probably unnecessary to worry about the performance cost of these structs being one pointer larger. One type fewer here would make things easier to understand... So yes let's simplify, I think it's a better tradeoff.

That module is starting to look more like the old Walker structs, with the problems you and Benjamin noticed with readability from the generic structs instead of concrete types.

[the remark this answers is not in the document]
tomhoule: I'm starting the next PR now, dropping View in that one.

PR #2181, opened 2024-09-27
fix(tracing): don't track hooks or trusted docs if they are not enabled

pimeys:
Also we should add the prepare_operation span. This will grow bigger when we enable the new planner and the user has a large schema.

### DIFF
--- a/engine/crates/engine-v2/src/engine/trusted_documents.rs
+++ b/engine/crates/engine-v2/src/engine/trusted_documents.rs
@@ -140,6 +140,7 @@ impl<'ctx, R: Runtime> PreExecutionContext<'ctx, R> {
     }
 }
 
+#[tracing::instrument(skip_all)]
 async fn handle_trusted_document_query<'ctx, 'r, R: Runtime>(
     engine: &'ctx Engine<R>,
     client_name: &'ctx str,
@@ -166,6 +167,7 @@ async fn handle_trusted_document_query<'ctx, 'r, R: Runtime>(
 /// Handle a request using Automatic Persisted Queries.
 /// We don't cache anything here, we only rely on the operation cache. We might want to use an
 /// external cache for this one day, but not another in-memory cache.
+#[tracing::instrument(skip_all)]
 async fn handle_apq<'r, 'f>(
     query: &'r str,
     ext: &'r PersistedQueryRequestExtension,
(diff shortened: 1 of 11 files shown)


engine/crates/engine-v2/src/engine/trusted_documents.rs
```diff
@@ -28,7 +27,6 @@ pub(crate) struct OperationDocument<'a> {
 impl<'ctx, R: Runtime> PreExecutionContext<'ctx, R> {
     /// Determines what document should be used for the request and provides an appropriate cache
     /// key for the operation cache and as a fallback a future to load said document.
-    #[instrument(skip_all)]
```
Finistere, line 31: I think we should rather intrusment`handle_trusted_document_query` & `handle_apq` rather than the implementations here IMHO, contrary to hooks when there's no ambiguity of whether it's implemented or not.

engine/crates/wasi-component-loader/src/hooks/authorization.rs
```diff
@@ -210,7 +222,10 @@ impl AuthorizationComponentInstance {
         edges: Vec<(String, Vec<String>)>,
         metadata: String,
     ) -> crate::Result<Vec<Result<(), crate::GuestError>>> {
+        let span = tracing::info_span!(AUTHORIZE_EDGE_POST_EXECUTION_HOOK_FUNCTION);
+
         self.call3_one_output(
```
Finistere, line 227: maybe we should add the instrument to the `call3_one` etc. variants? we would just name to set `otel.name` to the function name so that hey appear correctly on otel data

Finistere (approved):

engine/crates/wasi-component-loader/src/hooks/authorization.rs
pimeys: is otel.name then the value that gets shown in our tracing UI?

Finistere: yeah it's the span name that will be exported by gateway

Finistere: it overrides the tracing span name if present

engine/crates/engine-v2/src/engine/trusted_documents.rs
Finistere: but up to you 🤷

engine/crates/wasi-component-loader/src/hooks/authorization.rs
pimeys: Awesome, then there's no const restriction and I can send a dynamic string...

PR #2183, opened 2024-09-27
federated-graph: migrate more fields to TypeDefinition

tomhoule:
That already allows us to delete federated_graph::Scalar.


pimeys (approved):
a-w-e-s-o-m-e

PR #2187, opened 2024-09-30
fix(tracing): remove spammy spans and nest subgraph things under subgraph request

pimeys:
So this goes from this trace:

Screenshot 2024-09-30 at 14-21-26 julius-federated - Trace spans - Grafbase

Into this:

Screenshot 2024-09-30 at 14-22-33 julius-federated - Trace spans - Grafbase

And as a comparison, if disabling the access log hooks, we get this:

Screenshot 2024-09-30 at 14-23-23 julius-federated - Trace spans - Grafbase

And as a final thing, I added a span for the actual http request and renamed the subgraph span with the subgraph name, so now we look like this:

image

### DIFF
--- a/engine/crates/engine-v2/src/sources/graphql/request/execute.rs
+++ b/engine/crates/engine-v2/src/sources/graphql/request/execute.rs
@@ -11,6 +11,7 @@ use runtime::{
     rate_limiting::RateLimitKey,
 };
 use tower::retry::budget::Budget;
+use tracing::Instrument;
 use web_time::Duration;
 
 use crate::{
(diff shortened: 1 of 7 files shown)


obmarg (approved):

engine/crates/engine-v2/src/sources/graphql/request/execute.rs
```diff
@@ -68,11 +69,12 @@ pub(crate) async fn execute_subgraph_request<'ctx, 'a, R: Runtime>(
             timeout: endpoint.config.timeout,
         }
     };
-    ctx.record_request(&request);
 
     let fetcher = ctx.engine.runtime.fetcher();
+    let http_span = ctx.create_subgraph_request_span(&request);
```
Finistere, line 74: I would have created the HTTP span immediately with the context and only added `instrument` here. That way we would not need to have both
`http_span.record_http_status_code(response.status());` and `ctx.record_http_response(&response);`. Having a single "context" to record what happens and update the relevant spans & metrics feels simpler to me imho

Finistere: I know you don't have the http request when creating the span and need to put everything as `Empty`, but that feels still cleaner than having multiple things to call here to me IMHO. Don't have a strong opinion though.

PR #2197, opened 2024-10-06
chore(engine-v2): rework hook interface

Finistere:
Use generics inside of engine-v2 instead of `Vec<u8>` for on subgraph/operation response hook. Flattened the interface for `Hooks`, only kept `authorized` as is at it'll be refactored later anyway. Also moved the gateway request body timeout to a better place inside engine-v2


engine/crates/engine-v2/src/response/mod.rs
```diff
@@ -75,11 +75,11 @@ pub(crate) struct RequestErrorResponse {
...
-impl From<RequestErrorResponse> for Response {
-    fn from(resp: RequestErrorResponse) -> Self {
-        Self::RequestError(resp)
-    }
-}
+// impl<E> From<RequestErrorResponse> for Response<E> {
+//     fn from(resp: RequestErrorResponse) -> Self {
+//         Self::RequestError(resp)
+//     }
+// }
```
pimeys, line 82: Remove.

pimeys (approved):
Maybe try to get that rwlock out too in this PR for the hook cache, if possible.

PR #2206, opened 2024-10-08
GB-7696: Add support for the `message` argument to the `publish` command

jakubadamw:
# Description

The presence of the argument will cause a newly produced schema version to be annotated with the provided message.

# Type of change

- [ ] 💔 Breaking
- [X] 🚀 Feature
- [ ] 🐛 Fix
- [ ] 🛠️ Tooling
- [ ] 🔨 Refactoring
- [ ] 🧪 Test
- [ ] 📦 Dependency

### DIFF
--- a/cli/crates/cli/src/cli_input/publish.rs
+++ b/cli/crates/cli/src/cli_input/publish.rs
@@ -36,6 +36,10 @@ pub struct PublishCommand {
     #[arg(long, default_value_t = 4000)]
     pub(crate) dev_api_port: u16,
 
+    /// The message to annotate the publication with
+    #[arg(long)]
+    pub(crate) message: Option<String>,
+
     /// Add a header to the introspection request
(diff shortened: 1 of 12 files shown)


cli/crates/cli/src/cli_input/publish.rs
```diff
@@ -36,6 +36,10 @@ pub struct PublishCommand {
     #[arg(long, default_value_t = 4000)]
     pub(crate) dev_api_port: u16,
 
+    /// The message to annotate the publication with
+    #[arg(long)]
```
fbjork, line 40: can we add the shortcut -m for --message?

tomhoule (approved):
:100:

cli/crates/cli/src/cli_input/publish.rs
jakubadamw: Sure!

PR #2207, opened 2024-10-08
chore(engine-v2/qp): Initial cost estimation

Finistere:
This adds `estimate_resolver_costs` method to estimate the cost of each resolver taking into account all of its dependencies. I've changed the dot_graph generation to have one simple format for tests and a "pretty" one that heavily relies on colors and shapes to convey the information. The former is ideal for insta snapshot while the later makes it easier to read big graphs. I've adapted the tests to use this and print the pretty format on snapshot errors.

This is the pretty graph for the new test:
graphviz(8)

- blue: resolvers
- turquoise: between a field and resolver, the latter resolving the former
- orange: requirement
- black: operation field structure

Only edges with a non-zero cost are shown


engine/crates/engine-v2/query-planning/src/cost_estimation.rs
```diff
@@ -0,0 +1,156 @@
...
+                            *edge_cost = requirements_cost
+                        }
+                    }
+                    _ => (),
+                }
+            }
+        }
+    }
+
+    fn find_common_ancestor(&self, mut left: NodeIndex, mut right: NodeIndex) -> NodeIndex {
```
pimeys, line 114: What if these two functions would return an `Option<NodeIndex>` so we don't panic in weird cases. Maybe log the error? I'm just a bit worried one of these functions has a bug which leads to the panic...

engine/crates/engine-v2/query-planning/src/cost_estimation.rs
```diff
@@ -0,0 +1,156 @@
...
+    /// composition should ensure, we will end up in a finite number of iterations until
+    /// convergence. But how many is unclear and a composition error shouldn't lead to an infinite
+    /// loop, so we're capping the max number of iterations to an arbitrary seemingly large enough
+    /// number.
+    ///
+    /// Today each iterations goes naively over all nodes again, we can certainly be smarter than
+    /// that.
+    #[instrument(skip_all)]
+    #[allow(unused)]
+    pub(crate) fn estimate_resolver_costs(&mut self) {
```
pimeys, line 31: Should we call this automatically when we create a new graph, or is it something the user must do when planning?

engine/crates/engine-v2/query-planning/src/graph.rs
```diff
@@ -47,42 +50,44 @@ impl<'ctx, Op: Operation> std::ops::Index<Op::FieldId> for OperationGraph<'ctx,
...
 
 impl<'ctx, Op: Operation> OperationGraph<'ctx, Op> {
+    #[instrument(skip_all)]
     pub fn new(schema: &'ctx Schema, operation: &'ctx mut Op) -> OperationGraph<'ctx, Op> {
         Self::builder(schema, operation).build()
     }
 
+    /// Use https://dreampuf.github.io/GraphvizOnline
+    /// or `echo '..." | dot -Tsvg` from graphviz
+    pub fn to_prettty_dot_graph(&self) -> String {
```
pimeys, line 60:
```suggestion
    pub fn to_pretty_dot_graph(&self) -> String {
```

engine/crates/engine-v2/query-planning/src/tests/schema1.rs
```diff
@@ -0,0 +1,85 @@
...
+        }
+    }
+    "#,
+    );
+
+    let mut graph = OperationGraph::new(&schema, &mut operation);
+    insta::assert_snapshot!("graph", graph.to_dot_graph(), &graph.to_prettty_dot_graph());
+
+    graph.estimate_resolver_costs();
+    insta::assert_snapshot!("cost", graph.to_dot_graph(), &graph.to_prettty_dot_graph());
```
pimeys, line 84: What are we testing here? It's the same graph on both sides, or?

pimeys (approved):

engine/crates/engine-v2/query-planning/src/cost_estimation.rs
Finistere: good point, should return an internal planning error or something, rather than outright panicking

engine/crates/engine-v2/query-planning/src/cost_estimation.rs
Finistere: We won't call this from the engine, it's separate so I can have a unit tests for it. Engine will just call a single function

engine/crates/engine-v2/query-planning/src/tests/schema1.rs
Finistere: it's `insta::assert_snapshot!(<snapshot name>, <snapshot data>, <description>)`

Finistere: so it's mostly a sanity check when changing code

PR #2209, opened 2024-10-08
chore(engine-v2/qp): Add initial Steiner tree algorithm

Finistere:
Add the simplest possible Steiner tree algorithm with test graphs with known optimal cost.


pimeys (approved):
Discussed in Slack. Understood what it does. Let's go with this and tweak later if needed.

PR #2210, opened 2024-10-09
GB-7719: Add support for disabling APQs and configuring the operation cache limit

jakubadamw:
# Description

Add support for disabling APQs and configuring the operation cache limit.

# Type of change

- [ ] 💔 Breaking
- [X] 🚀 Feature
- [ ] 🐛 Fix
- [ ] 🛠️ Tooling
- [ ] 🔨 Refactoring
- [ ] 🧪 Test
- [ ] 📦 Dependency

### DIFF
--- a/engine/crates/gateway-core/src/lib.rs
+++ b/engine/crates/gateway-core/src/lib.rs
@@ -70,6 +70,13 @@ pub struct Gateway<Executor: self::Executor> {
     authorizer: Box<dyn Authorizer<Context = Executor::Context>>,
     operation_metrics: EngineMetrics,
     rate_limiter: Box<dyn RateLimiterInner>,
+    apq_enabled: bool,
+}
+
+#[derive(Clone, Copy, PartialEq)]
+pub enum AutomaticallyPersistedQueriesConfig {
+    Enabled,
+    Disabled,
 }
 
 impl<Executor> Gateway<Executor>
@@ -89,6 +96,7 @@ where
         trusted_documents: runtime::trusted_documents_client::Client,
         meter: grafbase_telemetry::otel::opentelemetry::metrics::Meter,
         rate_limiter: Box<dyn RateLimiterInner>,
+        apq_config: AutomaticallyPersistedQueriesConfig,
     ) -> Self {
         Self {
             executor,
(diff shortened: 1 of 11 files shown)


engine/crates/gateway-core/src/lib.rs
```diff
@@ -70,6 +70,13 @@ pub struct Gateway<Executor: self::Executor> {
     authorizer: Box<dyn Authorizer<Context = Executor::Context>>,
     operation_metrics: EngineMetrics,
     rate_limiter: Box<dyn RateLimiterInner>,
+    apq_enabled: bool,
+}
+
+#[derive(Clone, Copy, PartialEq)]
+pub enum AutomaticallyPersistedQueriesConfig {
```
fbjork, line 77:
```suggestion
pub enum AutomaticPersistedQueriesConfig {
```

tomhoule:
It looks like this is SDL-config only, but that is going away. Shouldn't the toml config from `gateway/crates/config` be supported?

jakubadamw:
> It looks like this is SDL-config only, but that is going away. Shouldn't the toml config from `gateway/crates/config` be supported?

Oops, I don't know how I missed that. Fixing.

gateway/crates/config/src/apq.rs
```diff
@@ -0,0 +1,6 @@
+#[derive(Debug, Clone, PartialEq, Default, serde::Deserialize)]
+#[serde(deny_unknown_fields)]
+pub struct AutomaticPersistedQueries {
+    /// If automatic persisted queries should be enabled.
+    pub enabled: bool,
```
tomhoule, line 5: Sorry I didn't re-read the rest of the PR, but is APQ being off by default intended? Intuitively I would have expected the opposite

tomhoule (approved):

crates/gateway-config/src/apq.rs
```diff
@@ -0,0 +1,19 @@
+#[derive(Debug, Clone, PartialEq, serde::Deserialize)]
+#[serde(deny_unknown_fields)]
```
Finistere, line 2: You can use `#[serde(default)]` to fill missing fields from a `AutomaticPersistedQueries::default()` instead of having to specify in twice

crates/gateway-config/src/operation_caching.rs
```diff
@@ -0,0 +1,28 @@
+#[derive(Debug, Clone, PartialEq, serde::Deserialize)]
+#[serde(deny_unknown_fields)]
```
Finistere, line 2: Same comment on `#[serde(default)]`

Finistere (approved):

PR #2214, opened 2024-10-09
chore(engine-v2/qp): operation graph fixes & snapshot diffs for graph changes

Finistere:
Snapshot the diff between graphs after cost estimation & pruning rather than the graph itself, it's a lot more readable. Keep track of "extra" fields (not present in the original query) and "indispensable" ones (must be retrieved, so present in the query string or needed for `@authorized` typically).

Adding a simple pruning strategy to remove any resolvers that don't lead to any scalar fields, useful for shared root fields as we'll only know later if there is something that could be provided.

```graphql
query {
  products {
    price
  }
}
```

Here we have resolvers from `name` and `category` because they can provide `products`, but they can't provide anything else further down. So they're useless:
graphviz(9)

Now we generate:
graphviz(10)

I now also avoid exponentially building new nested resolvers which could generate huge graphs. Unfortunately, this means I need to rework the cost estimation as it relied on resolvers having a single parent, which they won't anymore.

```graphql
query {
  products {
    reviews {
      stars
    }
  }
}
```

`reviews` comes from an entity join and products can be provided by multiple subgraphs. We ended up generating an individual `reviews` revolver for each of them, which would get even worse at lower depths:

graphviz(11)

Now we correctly detect the common resolver:
graphviz(12)


engine/crates/engine-v2/query-planning/src/graph/builder.rs
```diff
@@ -104,162 +97,218 @@ impl<'ctx, Op: Operation> OperationGraphBuilder<'ctx, Op> {
...
+                    dependent_ix: query_field_ix,
                     required_field_set,
                 })
             }
+            self.graph
+                .node_weight_mut(query_field_ix)
+                .unwrap()
+                .as_query_field_mut()
+                .unwrap()
+                .query_depth = query_depth;
```
pimeys, line 167: Lots of unwraps here...

engine/crates/engine-v2/query-planning/src/steiner_tree/shortest_path.rs
```diff
@@ -116,15 +116,16 @@ impl<'a, N, E> SteinerTreeAlg<'a, N, E> for ShortestPathAlg<'a, N, E> {
...
-struct ShortestPath {
-    parent: NodeIndex,
+struct ShortestPath<Id> {
+    parent: Id,
     cumulative_cost_from_root: Cost,
 }
 
 /// This assumes that the graph is a DAG.
-fn compuste_shortest_paths<N, E, F>(graph: &ResolverGraph<N, E>, root: NodeIndex, edge_cost: F) -> Vec<ShortestPath>
+fn compuste_shortest_paths<G, F>(graph: &G, root: G::NodeId, edge_cost: F) -> Vec<ShortestPath<G::NodeId>>
```
pimeys, line 125:
```suggestion
fn compute_shortest_paths<G, F>(graph: &G, root: G::NodeId, edge_cost: F) -> Vec<ShortestPath<G::NodeId>>
```

engine/crates/engine-v2/query-planning/src/tests/entities.rs
```diff
@@ -0,0 +1,97 @@
...
+            reviews {
+              stars
+            }
+          }
+        }
+    "#,
+    );
+
+    let graph = OperationGraph::new(&schema, &mut operation);
+    insta::assert_snapshot!("nested_join-graph", graph.to_dot_graph(), &graph.to_pretty_dot_graph());
```
pimeys, line 96: I'm not sure do I like these tests. You can't really see what should happen, just that these two things must be the same. Maybe the output is just too long to paste here, but for the reader it immediately tells what is expected, and also how to construct the next test.

pimeys (approved):

engine/crates/engine-v2/query-planning/src/tests/entities.rs
Finistere: It's a snapshot with an output as follows. It's unreadable as is, but it's easy to understand the difference. The third argument is the description of the snapshot. It gets printed whenever the test fails. This description is the readable format when using graphviz to show the graph.

```text
digraph {
    0 [ root]
    1 [ stars]
    2 [ reviews]
    3 [ products]
    4 [ Root#category]
    5 [ products@Root#category]
    6 [ Root#name]
    7 [ products@Root#name]
    8 [ Root#price]
    9 [ products@Root#price]
    10 [ FedEntity#review]
    11 [ reviews@FedEntity#review]
    12 [ stars@FedEntity#review]
    13 [ *id]
    14 [ id@Root#category]
    15 [ id@Root#name]
    16 [ id@Root#price]
    0 -> 3 [ label = "Field" ]
    0 -> 4 [ label = "CreateChildResolver(1)" ]
    0 -> 4 [ label = "HasChildResolver" ]
    4 -> 5 [ label = "CanProvide(0)" ]
    5 -> 3 [ label = "Provides" ]
    0 -> 6 [ label = "CreateChildResolver(1)" ]
    0 -> 6 [ label = "HasChildResolver" ]
    6 -> 7 [ label = "CanProvide(0)" ]
    7 -> 3 [ label = "Provides" ]
    0 -> 8 [ label = "CreateChildResolver(1)" ]
    0 -> 8 [ label = "HasChildResolver" ]
    8 -> 9 [ label = "CanProvide(0)" ]
    9 -> 3 [ label = "Provides" ]
    3 -> 2 [ label = "Field" ]
    9 -> 10 [ label = "CreateChildResolver(1)" ]
    3 -> 10 [ label = "HasChildResolver" ]
    10 -> 11 [ label = "CanProvide(0)" ]
    11 -> 2 [ label = "Provides" ]
    2 -> 1 [ label = "Field" ]
    11 -> 12 [ label = "CanProvide(0)" ]
    12 -> 1 [ label = "Provides" ]
    7 -> 10 [ label = "CreateChildResolver(1)" ]
    5 -> 10 [ label = "CreateChildResolver(1)" ]
    10 -> 13 [ label = "Requires" ]
    3 -> 13 [ label = "Field" ]
    5 -> 14 [ label = "CanProvide(0)" ]
    14 -> 13 [ label = "Provides" ]
    7 -> 15 [ label = "CanProvide(0)" ]
    15 -> 13 [ label = "Provides" ]
    9 -> 16 [ label = "CanProvide(0)" ]
    16 -> 13 [ label = "Provides" ]
}```

engine/crates/engine-v2/query-planning/src/graph/builder.rs
Finistere: yeah fixed in next PR

PR #2217, opened 2024-10-10
Make subgraph name in `check` non-optional

yoav-lavi:
# Description

## Breaking

- Make the subgraph name option in `check` non-optional as a part of sunsetting standalone graphs

# Type of change

- [x] 💔 Breaking
- [ ] 🚀 Feature
- [ ] 🐛 Fix
- [ ] 🛠️ Tooling
- [ ] 🔨 Refactoring
- [ ] 🧪 Test
- [ ] 📦 Dependency


cli/crates/cli/src/cli_input/check.rs
```diff
@@ -4,10 +4,9 @@ use super::ProjectRef;
 pub struct CheckCommand {
     #[arg(help = ProjectRef::ARG_DESCRIPTION)]
     pub project_ref: ProjectRef,
-    /// The name of the subgraph to check. This argument is always required in a federated graph
-    /// context, and it should not be used in a single graph context.
+    /// The name of the subgraph to check
     #[arg(long("name"))]
-    pub(crate) subgraph_name: Option<String>,
+    pub(crate) subgraph_name: String,
```
obmarg, line 9: Would it be worth allowing this as a positional argument if it's now required?

yoav-lavi: We could, although a flag has the benefit of not having a specific order and one positional argument is what we usually do, so not sure

obmarg: was just a thought, feel free to ignore if it's not obviously a good idea

yoav-lavi: I think one positional is more obvious in this case since otherwise people will confuse the ordering, but we can think about it and possibly change it later

PR #2230, opened 2024-10-21
fix(engine): mark cpu-heavy parts as blocking in tokio

Finistere:
Executing the cpu-heavy parts in `block_inplace` helps tokio achieve better latencies overall at the cost of higher CPU usage.

Added it behind a `tokio` feature flag so it can be disabled for CF

### DIFF
--- a/engine/crates/engine-v2/src/utils/mod.rs
+++ b/engine/crates/engine-v2/src/utils/mod.rs
@@ -1,3 +1,20 @@
 mod pool;
 
 pub(crate) use pool::*;
+
+/// Until engine-v2 is entirely remove
+#[cfg(feature = "tokio")]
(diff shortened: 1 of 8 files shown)


engine/crates/engine-v2/src/utils/mod.rs
```diff
@@ -1,3 +1,20 @@
 mod pool;
 
 pub(crate) use pool::*;
+
+/// Until engine-v2 is entirely remove
```
tomhoule, line 5: entirely native?

tomhoule (approved):
lgtm!

engine/crates/engine-v2/src/utils/mod.rs
Finistere: yeah that was the idea. removed it though, to just use tokio directly

PR #2234, opened 2024-10-22
use Rust 1.82 build image

fbjork:
This also updates the docker image to use Debian bookworm.

Closes: GB-7807


gateway/Dockerfile
```diff
@@ -35,13 +28,12 @@ COPY ./gqlint ./gqlint
...
 
-RUN cargo build --release --target "$(rustc -vV | sed -n "s|host: ||p")" --bin grafbase-gateway &&\
-    mv "/grafbase/target/$(rustc -vV | sed -n "s|host: ||p")/release/grafbase-gateway" /grafbase/target/release/grafbase-gateway
+RUN cargo build --release --bin grafbase-gateway
 
 #
 # === Final image ===
 #
-FROM alpine:3.20
+FROM debian:bookworm
```
obmarg, line 36: Not that fussed if no, but could we use `bookworm-slim` as the base of the final image?

gateway/Dockerfile
```diff
@@ -53,9 +45,9 @@ LABEL org.opencontainers.image.url='https://grafbase.com' \
 WORKDIR /grafbase
 
 # used curl to run a health check query against the server in a docker-compose file
-RUN apk add --no-cache curl
+RUN apt update && apt install -y curl
```
obmarg, line 48: Not sure how much difference it makes, but I remember in previous job we used to delete apt stuff in the same line as `apt update` to keep layer size down.

```suggestion
RUN apt update && apt install -y curl && rm -rf /var/lib/apt/lists/*
```

gateway/Dockerfile
fbjork: do we want to use a specific version?

gateway/Dockerfile
```diff
@@ -1,17 +1,10 @@
 #
 # === Build image ===
 #
-FROM rust:1.80-alpine3.20 AS chef
-
-# Patching musl to use mimalloc, see docker/README.md.
-# Source: https://github.com/tweag/rust-alpine-mimalloc
-COPY docker/build.sh docker/mimalloc.diff /tmp/
```
fbjork, line 8: can we remove the docker folder now?

gateway/Dockerfile
```diff
@@ -1,17 +1,10 @@
...
-RUN /tmp/build.sh
-ENV LD_PRELOAD=/usr/lib/libmimalloc.so
-ENV RUSTFLAGS="-C target-feature=+crt-static"
+FROM rust:1.82-bookworm AS chef
 
 COPY rust-toolchain.toml rust-toolchain.toml
-RUN apk add --no-cache musl-dev && cargo install --target "$(rustc -vV | sed -n "s|host: ||p")" cargo-chef
+RUN cargo install --target "$(rustc -vV | sed -n "s|host: ||p")" cargo-chef
```
obmarg, line 7: Is this `--target` even needed now?  Would have thought not.

```suggestion
RUN cargo install cargo-chef
```

obmarg (approved):

gateway/Dockerfile
pimeys: Well the build image doesn't have latest debian, and points to bookworm, so we do too. This one will be supported for a _loooong_ time.

PR #2237, opened 2024-10-22
Coordinator futures

Finistere:
This PR improves our latencies with two improvements:
- use of `spawn_blocking`/ `block_inplace` in the most CPU-consuming parts to improve Tokio's scheduling. This improved my tail latencies in a few cases, but hard to measure accurately. This comes at the cost of additional CPU use.
- The core inner loop of the operation execution is roughly:
```rust
let subgraph_requests = FuturesUnordered::new();
while let Some(...) = subgraph_requests.next().await {
   // ingest response
   // prepare next subgraph requests
}
```
The problem is that we're not awaiting the `subgraph_requests` while doing the ingestion/preparation so it's not moved forward by Tokio. I've changed that so they're executed simultaneously

### DIFF
--- a/engine/crates/engine-v2/src/execution/coordinator.rs
+++ b/engine/crates/engine-v2/src/execution/coordinator.rs
@@ -146,13 +146,12 @@ impl<'ctx, R: Runtime> ExecutionContext<'ctx, R> {
         );
 
         OperationExecution {
-            initial_futures: Some(ExecutionPlanFutureSet::new()),
             state: self.new_execution_state(),
             executed_operation_builder,
             response: ResponseBuilder::new(self.operation.root_object_id),
             ctx: self,
         }
-        .run()
+        .run(VecDeque::new())
         .await
     }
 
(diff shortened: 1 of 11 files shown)


engine/crates/engine-v2/src/execution/coordinator.rs
```diff
@@ -350,90 +353,154 @@ impl SubscriptionResponse {
...
+            }
+        }
+
+        let mut state = State::Execution(self);
+
+        futures_util::pin_mut!(futures);
+        let self_ = loop {
+            state = match state {
+                State::Ingestion(mut ingestion_fut) => {
+                    let mut fut = unsafe { Pin::new_unchecked(&mut ingestion_fut) };
```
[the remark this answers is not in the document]
Finistere: Good point, took it from `future_util::pin_mut` hacking my way through, I should find a different way. IIRC `Box::pin` would work, just wanted to try without first and left it as is

engine/crates/engine-v2/src/execution/coordinator.rs
```diff
@@ -350,90 +353,153 @@ impl SubscriptionResponse {
...
+        let mut state = State::Execution(self);
+
+        futures_util::pin_mut!(futures);
+        let self_ = loop {
+            state = match state {
+                State::Ingestion(mut ingestion_fut) => {
+                    let next_task = futures_util::select_biased! {
+                        ingestion_result = ingestion_fut => Ok(ingestion_result),
+                        execution_result = futures.next() => Err(execution_result),
+                    };
```
pimeys, line 397: `Either` would be better here rather than `Result`. It's a bit weird because the second one is not really an error. We have this elsewhere too, but we should maybe make these more clear in the future.

pimeys (approved):

engine/crates/engine-v2/src/execution/coordinator.rs
```diff
@@ -350,90 +353,153 @@ impl SubscriptionResponse {
...
+                            for fut in next_futures {
+                                futures.push(fut)
+                            }
+                            State::Execution(self_)
+                        }
+                    }
+                }
+                State::Execution(self_) => {
+                    if let Some(result) = results.pop_front() {
+                        State::Ingestion(Box::pin(self_.ingest_execution_result(result).boxed().fuse()))
```
obmarg, line 420: Looks like we're boxing this twice?  Do we need to do that?

Finistere: oh yes my bad, definitely not

engine/crates/engine-v2/src/execution/coordinator.rs
Finistere: fair enough, I just wish it was part of the std

Finistere: fixed

PR #2239, opened 2024-10-22
Update deps & fix benchmarks

Finistere:
We're fixing dependency versions quite strictly for a reason that escapes me. We miss updates like this one: https://github.com/serde-rs/json/pull/1206. So relaxed a few of those. And fixed the benchmarks now that we're using `spawn_blocking` inside the engine

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -14,20 +14,11 @@ dependencies = [
 
 [[package]]
 name = "addr2line"
-version = "0.22.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "6e4503c46a5c0c7844e948c9a4d6acd9f50cccb4de1c48eb9e291ea17470c678"
-dependencies = [
- "gimli 0.29.0",
-]
-
-[[package]]
-name = "addr2line"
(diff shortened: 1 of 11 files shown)


fbjork (approved):

pimeys (approved):

Cargo.toml
```diff
@@ -159,12 +159,12 @@ fnv = "1.0.7"
 sha2 = "0.10.8"
 
 # Serde
-serde = { version = "1.0.199", features = ["derive"] }
+serde = { version = "1", features = ["derive"] }
 serde_dynamo = "4.2.14"
-serde_json = { version = "1.0.116", features = ["preserve_order"] }
+serde_json = { version = "1", features = ["preserve_order"] }
```
jakubadamw, line 164: This wasn't a fixed version. `=1.0.116` would have been one. `1.0.116` means 1.0.116 or anything newer within the same major version. So, `cargo update` would have sufficed here.

Finistere: ah my bad then, misunderstood it. should I revert the Cargo.toml then?

PR #2243, opened 2024-10-23
fix: propagate the correct span to subgraphs

obmarg:
I've been using traces to investigate an issue today, and had enabled propagation to see whether it was us or subgraphs that were slow.  But in my tracing view the subgraph HTTP span was marked as a sibling of  our corresponding HTTP span rather than as a child.

This fixes that by propagating the http span to children rather than just using the current span.

Before:

CleanShot 2024-10-23 at 14 04 40@2x

After:

CleanShot 2024-10-23 at 14 04 59@2x

### DIFF
--- a/engine/crates/engine-v2/src/sources/graphql/request/execute.rs
+++ b/engine/crates/engine-v2/src/sources/graphql/request/execute.rs
@@ -2,7 +2,10 @@ use std::borrow::Cow;
 
 use bytes::Bytes;
 use futures::Future;
-use grafbase_telemetry::{graphql::GraphqlResponseStatus, otel::tracing_opentelemetry::OpenTelemetrySpanExt as _};
+use grafbase_telemetry::{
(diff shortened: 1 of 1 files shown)


engine/crates/engine-v2/src/sources/graphql/request/execute.rs
```diff
@@ -36,6 +39,8 @@ pub(crate) async fn execute_subgraph_request<'ctx, 'a, R: Runtime>(
 ) -> ExecutionResult<SubgraphResponse> {
     let endpoint = ctx.endpoint();
 
+    let http_span = SubgraphHttpRequestSpan::new(endpoint.url(), &http::Method::POST);
```
Finistere, line 42: we're already creating that span a bit later: `    let http_span = ctx.create_subgraph_request_span(&request);
`

Finistere: ah we should move the propagator later after the request creation

pimeys (approved):

PR #2246, opened 2024-10-24
fix(engine): Generate compact subgraph queries

Finistere:
Helps when subgraph log the queries to stdout at least.
Also adding a feature flag `pprof` for the gateway to have quickly a profile, it's quite convenient.


engine/crates/integration-tests/tests/federation/auth/authenticated.rs
```diff
@@ -101,22 +101,22 @@ fn not_authenticated() {
 
         // We shouldn't have requested the field.
         let requests = engine.drain_graphql_requests_sent_to::<SecureSchema>();
-        insta::assert_json_snapshot!(requests, @r###"
+        insta::assert_json_snapshot!(requests, @r#"
         [
           {
-            "query": "query {\n  check {\n    anonymous\n  }\n}\n",
+            "query": "query{ check{ anonymous}}",
```
pimeys, line 107: Do we need the whitespace after `{`'s?

pimeys (approved):

engine/crates/integration-tests/tests/federation/auth/authenticated.rs
Finistere: It's just to handle correctly multiple fields. one space won't matter, the actual problematic part were newlines

gateway/crates/gateway-binary/src/main.rs
```diff
@@ -19,6 +19,13 @@ mod telemetry;
 const THREAD_NAME: &str = "grafbase-gateway";
 
 fn main() -> anyhow::Result<()> {
+    #[cfg(feature = "pprof")]
+    let guard = pprof::ProfilerGuardBuilder::default()
+        .frequency(5000)
+        .blocklist(&["libc", "libgcc", "pthread", "vdso"])
+        .build()
+        .unwrap();
```
obmarg, line 27: Can these pprof bits go in functions (possibly with the `cfg inside the functions?)

Minor complaint, just seems irrelevant most of the time so would be good to not clutter up main so much

engine/crates/engine-v2/src/sources/graphql/request/prepare.rs
```diff
@@ -97,7 +89,7 @@ impl PreparedFederationEntityOperation {
         );
         query.push_str("query");
         query.push('(');
-        write!(query, "${entities_variable_name}: [_Any!]!")?;
```
obmarg, line 100: Nitpicking a bit, but surely a few spaces for readability would be ok...

obmarg (approved):

engine/crates/engine-v2/src/sources/graphql/request/prepare.rs
Finistere: fair enough

gateway/crates/gateway-binary/src/main.rs
Finistere: separated it from the rest, not great, but better than before at least

PR #2262, opened 2024-10-29
feat(engine-v2/qp): Add initial solver

Finistere:
This PR fully reworks the cost estimation, adjusts the Steiner tree algorithm API, and finally adds a solver that combines all of this to generate a solution for the operation. The solution isn't used anywhere yet though.

The Steiner tree algorithm itself is mostly the same, relying on the shortest path heuristic but it's adjusted for what I think should provide decent performance and to support the new cost estimation method.

Cost estimation previously on a bunch of logic, I worked on it multiple times thinking I could have some simple heuristic, but it's just as difficult as solving the operation itself! As a reminder, we need to estimate the cost of requirements for edges as those are only necessary if we take said edge. There is no need to retrieve `id` field for a federation entity request if we can just retrieve the field from the parent resolver. We cannot express "conditionally required" nodes in the Steiner tree problem, so we have to take those into account by adjusting the cost of incoming edges. The only way to estimate correctly how costly requirements are is... to generate a Steiner tree! Given the current operation Steiner tree, how costly is adding those new requirements?

As this extra cost estimation means computing a Steiner tree it can quickly become costly. So at the various steps I ensure to only compute what I need (premature optimization? never heard of it). Each query field is marked as indispensable if it's always needed for example.

Now let's see some examples, with this schema:
```graphql
enum join__Graph {
  CATEGORY @join__graph(name: "category", url: "http://localhost:4200/shared-root/category")
  NAME @join__graph(name: "name", url: "http://localhost:4200/shared-root/name")
  PRICE @join__graph(name: "price", url: "http://localhost:4200/shared-root/price")
  REVIEW @join__graph(name: "review", url: "http://localhost:4200/shared-root/review")
}

type Product
  @join__type(graph: CATEGORY, key: "id")
  @join__type(graph: NAME, key: "id")
  @join__type(graph: PRICE, key: "id")
  @join__type(graph: REVIEW, key: "id")
{
  id: ID!
  reviews: [Review] @join__field(graph: REVIEW)
  category: String @join__field(graph: CATEGORY)
  name: String @join__field(graph: NAME)
  price: Float @join__field(graph: PRICE)
}

type Review @join__type(graph: REVIEW)
{
  stars: Int @join__field(graph: REVIEW)
}

type Query
  @join__type(graph: CATEGORY)
  @join__type(graph: NAME)
  @join__type(graph: PRICE)
{
  product: Product
  products: [Product]
}
```

first with a relatively simple shared root fields query:
```graphql
query {
  products {
    price
    category
    id
    name
  }
}
```
We generate the following operation graph from it:
graphviz(13)

Which generates the following Steiner graph (custom name for the graph we're solving with the Steiner tree algorithm). We add an extra dummy root node at the top to always have an incoming edge.:
graphviz(14)

After running the solver we get the following:
graphviz(15)

The Steiner tree nodes are all marked in green, notice how only one resolver provides the `id` field.

Now let's make it a bit more complex:
```graphql
query {
  products {
    name
    reviews {
      stars
    }
  }
}
```
We have a nested join here with `reviews` being only providable through an entity join. It's requirement `id` can be provided by 3 subgraphs. We also request `name` only providable directly by the `NAME` subgraph, all others need an entity join for it. So we have the following operation graph:
graphviz(16)

The solver generates this initial Steiner graph:
graphviz(17)

And we get the following solution:
graphviz(18)


engine/crates/engine-v2/query-planning/src/graph/prune.rs
```diff
@@ -0,0 +1,44 @@
+use crate::{Edge, Node, Operation};
```
Finistere, line 1: only moved code

PR #2271, opened 2024-10-31
chore: check licenses for every PR

pimeys:
This is annoying when it happens only on release.

### DIFF
--- a/gateway/about.toml
+++ b/gateway/about.toml
@@ -23,3 +23,5 @@ targets = [
     "x86_64-pc-windows-msvc",
     "x86_64-unknown-linux-musl",
 ]
+
+workarounds = ["ring"]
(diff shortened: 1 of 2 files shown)


gateway/about.toml
```diff
@@ -23,3 +23,5 @@ targets = [
     "x86_64-pc-windows-msvc",
     "x86_64-unknown-linux-musl",
 ]
+
+workarounds = ["ring"]
```
pimeys, line 27: I have no clue why this never triggered in our release pipelines. But Ring has a _weird_ license combo (nothing special in the licenses they use, but it's a long combination of OpenSSL et.al. licenses that we can't really parse without workarounds). 

https://embarkstudios.github.io/cargo-about/cli/generate/workarounds.html#ring

tomhoule (approved):

PR #2273, opened 2024-10-31
graphql-schema-validation: fix validation of directive definition argument types

tomhoule:
Validation of the types of arguments inside directive definitions worked only for types defined before the directive definition. That would lead to incorrect validation errors about scalar or input types not existing. Validation of argument types in directive definitions now properly takes all types into account.

The solution is, like for fields of types, to move validation of the arguments to the second pass of schema validation, where all types are known.

closes GB-7859


engine/crates/validation/src/validate/type_definition.rs
```diff
@@ -115,3 +118,21 @@ pub(crate) fn validate_definitions_second_pass<'a>(ast: &'a ast::ServiceDocument
         }
     }
 }
+
+fn validate_directiv_definition_arguments(
```
tomhoule, line 122: typo

Finistere (approved):

PR #2274, opened 2024-10-31
chore(engine-v2/qp): run initial part of QP on feature flag

Finistere:
Adapter between the current `engine_v2::Operation` and query planning crate. Only running it with a feature flag that I use manually for now.

### DIFF
--- a/engine/crates/engine-v2/query-planning/src/graph.rs
+++ b/engine/crates/engine-v2/query-planning/src/graph.rs
@@ -21,13 +21,17 @@ use crate::dot_graph::Attrs;
 
 pub type Cost = u16;
 
-pub trait Operation: std::fmt::Debug {
+pub trait Operation {
     type FieldId: From<usize> + Into<usize> + Copy + std::fmt::Debug + Ord;
 
     fn field_ids(&self) -> impl ExactSizeIterator<Item = Self::FieldId> + 'static;
     fn field_defintion(&self, field_id: Self::FieldId) -> Option<FieldDefinitionId>;
     fn field_satisfies(&self, field_id: Self::FieldId, requirement: RequiredField<'_>) -> bool;
-    fn create_extra_field(&mut self, requirement: RequiredField<'_>) -> Self::FieldId;
+    fn create_extra_field(
+        &mut self,
+        petitioner_field_id: Self::FieldId,
+        requirement: RequiredField<'_>,
+    ) -> Self::FieldId;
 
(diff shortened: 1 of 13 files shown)


engine/crates/engine-v2/query-planning/src/graph.rs
```diff
@@ -37,14 +41,14 @@ pub trait Operation: std::fmt::Debug {
...
 pub struct OperationGraph<'ctx, Op: Operation> {
     pub(crate) schema: &'ctx Schema,
-    pub(crate) operation: &'ctx mut Op,
+    pub(crate) operation: Op,
     root_ix: NodeIndex,
     pub(crate) graph: StableGraph<Node<Op::FieldId>, Edge>,
 }
 
 impl<'ctx, Op: Operation> OperationGraph<'ctx, Op> {
     #[instrument(skip_all)]
```
pimeys, line 50: I know this is not part of this pull request, but if we use instrumentation like we do here, should we enforce them to be in debug level only? I haven't seen this span in traces, but wouldn't be surprised if we had a random `new` appearing in a customer demo...

engine/crates/engine-v2/src/operation/mod.rs
```diff
@@ -69,7 +71,7 @@ where
     }
 }
 
-#[derive(serde::Serialize, serde::Deserialize, IndexedFields)]
+#[derive(Clone, serde::Serialize, serde::Deserialize, IndexedFields)]
```
pimeys, line 74: Is cloning this really necessary? Smells like it's an expensive operation. Are we cloning it somewhere?

engine/crates/engine-v2/src/operation/prepare.rs
```diff
@@ -82,6 +82,9 @@ impl Operation {
             }
         };
 
+        #[cfg(feature = "qp")]
+        crate::operation::plan::plan(schema, operation.clone()).unwrap();
```
pimeys, line 86: Oh, here's the clone. Is this temporary or do we have to clone it?

pimeys (approved):

engine/crates/engine-v2/src/operation/prepare.rs
Finistere: Temporary, I just want to see that QP doesn't fail on tests for now

engine/crates/engine-v2/query-planning/src/graph.rs
Finistere: hum ok, fair enough

PR #2278, opened 2024-10-31
feat: derived components in message signing

obmarg:
This adds support for signing all the relevant derived components when message singing is enabled.  The implementation of this was trivial, but testing was a bit painful - axum really didn't want to give us the full URL, which made any of the components that require it fail.

Signature parameters will come in the next PR.

Part of GB-7858

### DIFF
--- a/engine/crates/runtime-local/src/fetch/signing.rs
+++ b/engine/crates/runtime-local/src/fetch/signing.rs
@@ -60,7 +60,6 @@ pub struct SigningParameters {
     expiry: Option<Duration>,
     include_headers: Option<HashSet<String>>,
     exclude_headers: HashSet<String>,
-    #[expect(unused)]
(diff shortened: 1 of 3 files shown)


engine/crates/runtime-local/src/fetch/signing.rs
```diff
@@ -121,19 +119,25 @@ impl SigningParameters {
...
             .collect::<Result<Vec<_>, _>>()?;
 
-        // TODO: Handle derived_components & signature parameters
+        covered_components.extend(
+            self.derived_components
+                .iter()
+                .map(derived_component_to_message_component),
+        );
+
+        // TODO: Handle signature parameters
```
obmarg, line 140: Coming in next PR

Finistere (approved):

pimeys (approved):

PR #2279, opened 2024-10-31
cli: rename PROJECT_REF arguments to GRAPH_REF everywhere

tomhoule:
In the implementation, the two names referred to two different things:

- ProjectRef: `<account>/<graph>@<branch>`
- GraphRef: `<graph>@<branch>`

ProjectRef is renamed to FullGraphRef in this commit. In all commands that took a project_ref argument, it is renamed to graph_ref, so the public API surface only mentions graph refs.

closes GB-7860


cli/crates/cli/src/cli_input/graph_ref.rs
```diff
@@ -1,16 +1,15 @@
 use std::{borrow::Cow, fmt, str};
 
 use graph_ref::GraphRef;
-
 /// Parsed project reference. A project reference is a string of the form `account/project@branch`.
 #[derive(Clone, Debug)]
-pub struct ProjectRef {
+pub struct FullGraphRef {
```
fbjork, line 6: Update comment above

tomhoule: good catch

cli/crates/cli/src/cli_input/graph_ref.rs
```diff
@@ -41,24 +40,24 @@ impl str::FromStr for ProjectRef {
             return Err("The account name is missing before '/'.");
         }
 
-        let (project, branch) = match rest.split_once('@') {
+        let (graph, branch) = match rest.split_once('@') {
             Some((project, branch)) => (project, Some(branch)),
```
obmarg, line 44:
```suggestion
            Some((graph, branch)) => (graph, Some(branch)),
```

cli/crates/cli/src/subgraphs.rs
```diff
@@ -2,7 +2,7 @@ use crate::{cli_input::SubgraphsCommand, errors::CliError, output::report};
 
 #[tokio::main]
 pub(super) async fn subgraphs(cmd: SubgraphsCommand) -> Result<(), CliError> {
-    let project_ref = cmd.project_ref;
+    let project_ref = cmd.graph_ref;
```
obmarg, line 5: should this be graph_ref now as well?

obmarg (approved):

cli/crates/cli/src/subgraphs.rs
tomhoule: definitely, changing that too

PR #2287, opened 2024-11-02
chore(engine-v2/qp): Generate new OperationPlan

Finistere:
Generate a new `OperationPlan` from the query planning. Doing a "few" little changes:

- codegen: Split the different cases of scalars and add support for `scalar FieldRef @ref(target: Field)` which allows one to do `required_fields: IdRange<FieldRefId>` pointing to a shared `Vec<FieldId>` instead of `required_fields: Vec<FieldId>` which `IdRange<FieldRefId>` being 2 * u32 rather than 3 words. Yeah well, maybe was a stupid idea.
- query-planning: Adding missing composite type, before for a query like `{ author { name } }` I wouldn't keep track of the "QueryNode" `author` anymore, only which resolvers should be used and scalars like `name`. But I do need `author`. And also slightly adapted the API to have cleaner tests with less logs by solving the graph only once.

The OperationGraph code is generated with the codegen:
```graphql
scalar Operation
  @domain(
    destination: "src/plan"
    root_module: "plan"
    context_name: "ctx"
    context_type: "PlanContext"
    visibility: "(crate)"
    imports: [{ module: "schema", domain: "schema" }]
  )

scalar PositionedResponseKey @copy @prelude
scalar Location @copy @prelude
scalar String @prelude
scalar QueryInputValue @indexed @prelude @record
scalar FieldRef @ref(target: "Field")

type Plan @indexed(id_size: "u16") @meta(module: "plan") {
  resolver_definition: ResolverDefinition!
  fields: [Field!]!
  required_fields: [FieldRef!]!
}

# ----------------
# - SelectionSet -
# ----------------

union Field @indexed(id_size: "u32") @meta(module: "field") @variants(remove_suffix: true) = DataField | TypenameField

"In opposition to a __typename field this field does retrieve data from a subgraph"
type DataField @meta(module: "field/data") {
  key: PositionedResponseKey!
  location: Location!
  definition: FieldDefinition!
  arguments: [FieldArgument!]!
  selection_set_fields: [Field!]!
  required_fields: [FieldRef!]!
}

type FieldArgument @meta(module: "field/argument") @indexed(id_size: "u16") {
  definition: InputValueDefinition!
  value: QueryInputValue!
}

"__typename field"
type TypenameField @meta(module: "field/typename") {
  key: PositionedResponseKey!
  location: Location!
  type_condition: CompositeType!
}

type VariableDefinition @meta(module: "variable") @indexed(id_size: "u16") {
  name: String!
  name_location: Location!
  default_value: QueryInputValue
  ty: Type!
}

# ------------
# - Modifier -
# ------------

scalar QueryModifierRule @prelude

type QueryModifier @meta(module: "modifier") {
  rule: QueryModifierRule!
  impacted_fields: [FieldRef!]!
}

scalar ResponseModifierRule @prelude

type ResponseModifier @meta(module: "modifier") {
  rule: ResponseModifierRule!
  impacted_fields: [FieldRef!]!
}
```

We now make a clear distinction between structs generated during the binding phase used as query planning input and the ones used for the `OperationPlan`. We don't generate dependencies between plans anymore, everything is kept at the field level. Plans depend on fields as defined by the resolver definition and fields may depend on other fields. Once QueryModifiers are applied we will infer from it the plan dependencies.

Fixes: GB-7899


engine/crates/engine-v2/codegen/src/domain/scalar.rs
```diff
@@ -1,28 +1,91 @@
...
+        name: String,
+        id_struct_name: String,
+        span: cynic_parser::Span,
+        in_prelude: bool,
+        external_domain_name: Option<String>,
+        target: Box<Definition>,
+    },
 }
 
+// #[derive(Debug)]
```
Finistere, line 32: Removing it another PR

obmarg (approved):

PR #2291, opened 2024-11-04
feat: allow JWK format message signing keys

obmarg:
The first pass of message signing expected keys in PEM format.  These are fairly standard, but very opaque.  Even in my testing I often found that I'd have a PEM file but its contents wouldn't be in PKCS8 format so things would fail and it would not be obvious why.

JWKs are a lot more inspectable since they are just JSON - it's much easier to tell at a glance what kind of key you have etc.  So it seems like a good idea to support these as well - which this PR does.

It's a lot more manual code than I'd initially hoped: 

- The obvious JWK crate on crates.io doesn't support all the things we need, and seems somewhat dead.  (There are PRs adding support for the things we need, but they've been sitting open for years)
- The elliptic-curve library does contain some JWK support code, but it is broken.  If you provide it a fairly standard JWK that has some fields it wasn't expecting it'll fail with a very unhelpful "crypto error"
- There doesn't seem to be any crate that supports ed25519 JWKs, so I hand rolled that.


tomhoule (approved):
:100:

PR #2303, opened 2024-11-06
chore: add a user in dockerfile

obmarg:
Apparently semgrep is complaining that this Dockerfile doesn't have a user in it.  It's only for tests so it really doesn't matter, but it's also easy to fix.

Fixes GB-7912

### DIFF
--- a/engine/crates/integration-tests/data/sse-subgraph/Dockerfile
+++ b/engine/crates/integration-tests/data/sse-subgraph/Dockerfile
@@ -1,4 +1,8 @@
 FROM oven/bun:1.1.27-alpine
+
+RUN adduser -g wheel -D grafbase -h "/grafbase" && mkdir -p /grafbase && chown grafbase: /grafbase
+USER grafbase
+
 WORKDIR /grafbase
 COPY . .
 


engine/crates/integration-tests/data/sse-subgraph/Dockerfile
```diff
@@ -1,4 +1,8 @@
 FROM oven/bun:1.1.27-alpine
+
+RUN adduser -g wheel -D grafbase -h "/grafbase" && mkdir -p /grafbase && chown grafbase: /grafbase
```
tomhoule, line 3: I'm not familiar with the colon in `chown grafbase: /grafbase` part, what does it do? (it's a case of RTFM but I'm lazy)

tomhoule (approved):

engine/crates/integration-tests/data/sse-subgraph/Dockerfile
obmarg: I wasn't familiar either, this was a copy & paste from the gateway dockefile.

But looked it up: it sets the group to whatever the users login group is.

fbjork (approved):

engine/crates/integration-tests/data/sse-subgraph/Dockerfile
tomhoule: TIL. Thanks!

PR #2305, opened 2024-11-07
feat: handle cost directive in composition

obmarg:
This adds support for the `@cost` demand control directive to composition.  This doesn't actually do anything yet, it just makes sure we validate it during ingestion & output it into the composed SDL.

Part of GB-7642

### DIFF
--- a/engine/crates/composition/tests/composition/cost_basic/api.graphql
+++ b/engine/crates/composition/tests/composition/cost_basic/api.graphql
@@ -1,0 +1,19 @@
+enum Blah @cost(weight: 5) {
+    BLAH
+}
+
+type Query @cost(weight: 1) {
+    foo(name: String @cost(weight: 100)): String
+    id: ID! @cost(weight: 2)
+}
+
+interface Account @cost(weight: 3) {
+    foo(name: String @cost(weight: 200)): String
+    id: ID! @cost(weight: 4)
+}
+
+input Bar {
+    x: String @cost(weight: 7)
+}
+
+scalar Foo @cost(weight: 6)
--- a/engine/crates/composition/tests/composition/cost_basic/api.graphql
+++ b/engine/crates/composition/tests/composition/cost_basic/api.graphql
@@ -1,19 +1,19 @@
-enum Blah @cost(weight: 5) {
+enum Blah {
     BLAH
 }
 
-type Query @cost(weight: 1) {
-    foo(name: String @cost(weight: 100)): String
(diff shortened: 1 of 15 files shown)


engine/crates/composition/tests/composition/cost_basic/api.graphql
```diff
@@ -0,0 +1,19 @@
+enum Blah @cost(weight: 5) {
+    BLAH
+}
+
+type Query @cost(weight: 1) {
+    foo(name: String @cost(weight: 100)): String
+    id: ID! @cost(weight: 2)
```
tomhoule, line 7: I'm not familiar with the directive yet so I don't know: is the directive supposed to be in the public API exposed by the router or is it an internal federation directive? (like `@key` & co)

tomhoule: The part I don't understand is that in the implementation below it's marked as non public, but it's in this snapshot.

tomhoule (approved):
Just a question about whether `@cost` should appear in the public exposed API schema. Everything looks good :+1:

engine/crates/composition/tests/composition/cost_basic/api.graphql
tomhoule: Ah I guess what's happening is that it comes back as `Directive::Other` because `@cost` isn't handled in `from_sdl()` in the federated-graph crate.

obmarg: Yeah, good spot.  It was not meant to be in here, it's just very easy to not notice how many spapshots you're updating when you use UPDATE_EXPECT.

engine/crates/composition/tests/composition/cost_basic/federated.graphql
```diff
@@ -0,0 +1,45 @@
...
+
+directive @join__implements(graph: join__Graph!, interface: String!) repeatable on OBJECT | INTERFACE
+
+directive @join__unionMember(graph: join__Graph!, member: String!) repeatable on UNION
+
+enum join__Graph {
+    A @join__graph(name: "a", url: "http://example.com/a")
+}
+
+scalar Foo @cost(weight: 6)
```
pimeys, line 27: Should we add the `@cost` directive declaration to the headers?

obmarg: Good point, added it

PR #2308, opened 2024-11-08
Remove versioning in graphql-federated-graph

tomhoule:
Only v4 left. Not renaming it for now to keep a clean diff. First PR in a batch of simplifications made possible by the removal of managed graphs.

closes GB-7947


engine/crates/federated-graph/src/federated_graph/v4/ids.rs
```diff
@@ -1,10 +1,10 @@
-use super::FederatedGraph;
+use super::*;
 
 macro_rules! id_newtypes {
     ($($storage:ident [ $name:ident ] -> $out:ident,)*) => {
         $(
             #[derive(Debug, Clone, Copy, PartialOrd, Ord, PartialEq, Eq, Hash, serde::Serialize, serde::Deserialize)]
-            pub struct $name(usize);
+            pub struct $name(pub usize);
```
obmarg, line 7: Don't really mind either way, but why have this pub if we have the `From` impls (or alternatively why the from impls if this is pub?

obmarg (approved):

engine/crates/federated-graph/src/federated_graph/v4/ids.rs
tomhoule: Good catch, the answer is laziness. I wanted to get to a compiling state faster and that helped a lot, but I should finish converting the existing ID constructors and patterns to not use the field. I'll do that and make the inner usize private again before merging.

tomhoule: Just pushed that change, let's see what the tests say.

PR #2309, opened 2024-11-08
Dev command internal demo

yoav-lavi:
# Dev Command Internal Demo Release

> [!IMPORTANT]
> The implementation of the dev command is still a WIP, this PR and release is to be able to demo the usage internally and receive feedback. The current implementation contains demo specific aspects, specifically the output handler.

https://github.com/user-attachments/assets/584a9618-7709-40a5-ba5e-12daf5e81daf

This release includes the following features:

* Both configurations
* Configuration merging
* Subgraph retrieval from API
* Subgraph retrieval from overrides
* Composition
* Integration with the federated server
* Dev headers for introspection
* Environment variable interpolation (on the normal config fields where this is supported and for introspection header values)
* A flag to set the port

Still missing:

* Hot reloading
* Subgraph hot reloading
* Pathfinder
* Various optimizations and cleanup, possibly a separate crate for dev

## Usage

```
Usage: grafbase dev [OPTIONS]
Options:
      -r, --graph-ref <GRAPH_REF> The graph reference following the format "org/graph@branch"
      -c, --gateway-config <GATEWAY_CONFIG>  The path of the gateway configuration file
      -o, --graph-overrides <GRAPH_OVERRIDES>  The path of the graph overrides configuration file
      -p, --port <PORT> The port on which the server will listen
```

## Configuration

The following values are added for the graph overrides (using these in the normal configuration will be a hard error, currently will panic in some cases):

```toml
[subgraphs.inventory]
introspection_url = "https://grafbase-federation-demo-inventory.fly.dev/graphql"
schema_path = "../../test.sdl"

[subgraphs.inventory.introspection_headers]
Accept = "application/json"
```

- `introspection_url` - A url from which the subgraph SDL can be introspected.
- `schema_path ` - A file containing the subgraph SDL (this is used if both are specified as this is faster)
- `introspection_headers` - Headers to use specifically for introspection. Introspection currently does not use the header rules in addition to this, may change in a future iteration. Will be unused if `schema_path ` is present.

The URL for a subgraph is determined by (each defaulting to the next): the remote URL, the configuration `url` field, the introspection URL. One of these must be specified.

### Listening Address

The listening address will default to `127.0.0.1:5000`, but can be overridden using a flag or the configuration. As this is meant for personal dev use, we will always use localhost but will determine the port from the flag, then default to the port specified in the configuration listening address and finally default to `5000`.

### Known Issues

- As currently we don't have a hard error for dev fields being in the normal configuration, adding `file` in the normal config will result in either an incorrect relative path or a panic, depending on whether the overrides config exists. Will be fixed in the next iteration.


cli/crates/cli/src/cli_input/dev.rs
```diff
@@ -0,0 +1,19 @@
+use super::ProjectRef;
+use clap::Parser;
+use std::path::PathBuf;
+
+/// Start a local development server
+#[derive(Debug, Parser)]
+pub struct DevCommand {
+    #[arg(short('r'), long("graph-ref"), help = ProjectRef::ARG_DESCRIPTION)]
+    pub(crate) graph_ref: Option<ProjectRef>,
```
Finistere, line 9: We have a `GraphRef` in the `graph-ref` crate, you might want to re-use it, wrapping it in something like `GraphRefWithAccountRef` for example

gateway/crates/config/src/lib.rs
```diff
@@ -186,6 +186,15 @@ pub struct SubgraphConfig {
     pub entity_caching: Option<EntityCachingConfig>,
     /// Subgraph specific message signatures config
     pub message_signatures: Option<MessageSignaturesConfig>,
+    #[cfg(feature = "dev")]
+    /// The path of an SDL schema file for the subgraph.
+    pub file: Option<PathBuf>,
```
Finistere, line 191: I find `file` to be too ambiguous, it should be `schema_path` IMO as we already use `_path` suffix and `schema` to explicit about what it is

yoav-lavi: Fair enough will change it

cli/crates/cli/src/cli_input/dev.rs
Finistere: up to you, just wanted to mention it in case you weren't aware of it^^

gateway/crates/config/src/lib.rs
yoav-lavi: Done

PR #2329, opened 2024-11-12
chore: replace codegen directives with derives

obmarg:
I was working on composition last week and noticed that we had to parse directives in two different places.  I started working to consolidate those, but then had another thought: why are we writing directive deserialisation by hand.  So, I've introduced some derives that can do it for us.

This PR brings in that code in `engine-v2-codegen`: a relatively low impact place to add it, where if anything goes wrong it _shouldn't_ have much impact.


Finistere (approved):
Thank you

PR #2330, opened 2024-11-12
chore: consolidate parsing of `@cost` & `@deprecated`

obmarg:
This PR has two changes:

1. It pulls the crate I introduced in #2329 into composition so we can derive directive parsing based on structs.
2. It adds structs to cover `@cost` & `@deprecated` into `federated-graph` and updates the directive handling in `composition` & `federated-graph` to use them.

If we like this approach we can expand to the other directives later, this seemed like a small enough change to gather feedback & make sure things actually work.

### DIFF
--- a/engine/crates/federated-graph/src/directives/deprecated.rs
+++ b/engine/crates/federated-graph/src/directives/deprecated.rs
@@ -1,0 +1,28 @@
+use cynic_parser_deser::ValueDeserialize;
+
+#[derive(ValueDeserialize)]
+pub struct DeprecatedDirective<'a> {
+    pub reason: Option<&'a str>,
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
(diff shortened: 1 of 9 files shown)


Finistere (approved):

engine/crates/federated-graph/src/directives/deprecated.rs
```diff
@@ -0,0 +1,28 @@
...
+
+        assert_eq!(value.reason, None);
+    }
+
+    #[test]
+    fn test_parsing_with_reason() {
+        let doc = directive_test_document("@deprecated(reason: \"because I wanted to\")");
+        let value = parse_from_test_document::<DeprecatedDirective<'_>>(&doc).unwrap();
+
+        assert_eq!(value.reason, None);
```
obmarg, line 26: This is clearly not correct 🤔

tomhoule (approved):
Much nicer than what we've been doing so far. I'm all for porting the other directives.

PR #2333, opened 2024-11-12
Streamline repo structure

tomhoule:
- Move all crates to a new `crates/` directory, except for the top level `grafbase` (CLI) and `grafbase-gateway` binary crates.
- Make sure the name of the containing directory matches the crate name for every crate.


Cargo.toml
```diff
@@ -1,28 +1,24 @@
...
-    "engine/crates/engine-v2/id-newtypes",
-    "engine/crates/engine-v2/parser",
-    "engine/crates/engine-v2/value",
-    "graph-ref",
-    "graphql-lint",
-    "gqlint",
-    "workspace-hack",
+    "cli",
+    "gateway",
+    "crates/*",
```
fbjork, line 6: can't we do "crates/**/*" or "crates/engine-v2/*" here?

Cargo.toml
```diff
@@ -1,28 +1,24 @@
...
+    "crates/engine-v2/query-solver",
+    "crates/engine-v2/id-derives",
+    "crates/engine-v2/id-newtypes",
+    "crates/engine-v2/parser",
+    "crates/engine-v2/value",
 ]
 exclude = [
-    "engine/crates/gateway-v2",
-    "engine/crates/wasi-component-loader/examples",
+    "crates/gateway-v2",
```
fbjork, line 20: can we call this just gateway?

Cargo.toml
```diff
@@ -1,28 +1,24 @@
...
-    "engine/crates/engine-v2/parser",
-    "engine/crates/engine-v2/value",
-    "graph-ref",
-    "graphql-lint",
-    "gqlint",
-    "workspace-hack",
+    "cli",
+    "gateway",
+    "crates/*",
+    "crates/engine-v2/auth",
```
fbjork, line 7: can we rename engine-v2 to just engine?

Cargo.toml
tomhoule: No, because engine-v2 isn't a repository that contains only crates (just `src/` is enough to break that), that wouldn't work.

Cargo.toml
tomhoule: That's going to make a huge diff, so I'd prefer to do that as a separate PR.

Cargo.toml
tomhoule: that line is dead code, I just removed it.

Finistere (approved):

PR #2338, opened 2024-11-12
feat: handle the `@listSize` directive in composition

obmarg:
As per the title of the PR.  Also made a couple of tweaks while I was in here:

1. Removed trailing commas from lists when rendering SDL.  They didn't hurt, but I thought they looked kind of bad?  Feel free to disagree though.
2. Changed composition tests & composition in the CLI to print a report on GraphQL parse failures, instead of just a message.  Definitely useful in tests, but debatable whether we always want to pay the price of building this.  Comments welcome.  Looks like this:

CleanShot 2024-11-12 at 15 04 02@2x

Fixes GB-7642


crates/graphql-composition/src/subgraphs.rs
```diff
@@ -74,17 +74,23 @@ const BUILTIN_SCALARS: [&str; 5] = ["ID", "String", "Boolean", "Int", "Float"];
...
     fn source(&self) -> Option<&(dyn std::error::Error + 'static)> {
-        self.0.source()
+        self.error.source()
     }
 }
 
 impl std::fmt::Display for IngestError {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        std::fmt::Display::fmt(&self.0, f)
+        if f.alternate() {
```
tomhoule, line 90: I've never used that API, good idea.

engine/crates/composition/tests/composition/list_size/subgraphs/a.graphql
```diff
@@ -0,0 +1,11 @@
+type Query {
+  foo: [String!] @listSize(assumedSize: 10)
+  bar(slice: Int!): BarCollection
+    @listSize(slicingArguments: ["slice"], sizedFields: ["length"], requireOneSlicingArgument: false)
```
tomhoule: not saying we should validate that right away, but since I'm getting familiar with this directive: shouldn't `sizeFields` references a field that exists on `BarCollection`?

tomhoule (approved):
Really nice! Much better to add directives this way. And nice work on the error parsing formatting and the trailing commas too.

engine/crates/composition/tests/composition/list_size/subgraphs/a.graphql
obmarg: Oh yes - I did mean to have it reference a correct field, but fucked up.

PR #2344, opened 2024-11-13
chore(engine): Fix child-type-mismatch

Finistere:
Had to add the `JoinFields` struct in federated graph to keep track of the `type` in `@join__field`. I'll change that in a further PR to keep track of all the `@join_field` information in a `directives` attributes.

Explained the problem there https://github.com/grafbase/grafbase/pull/2344/files#diff-ac90207e5a538e9a94040951df4e2fcc504f9ecad005a706274a4a1709764479 near the end. In short, the subgraph has a field of type `ID!` and in the supergraph it ends being `ID`. That's fine most of time except when you send an union to the subgraph where it can break with something like:
```graphql
{ ... on User { id } ... on Account { id } }
```
In the supergraph both have the same type `ID`, but in the subgraph one has `ID!` and the other has `ID`, so the query fails. So we need to send the following instead:
```graphql
{ ... on User { alias: id } ... on Account { id } }
```

### DIFF
--- a/crates/engine/src/operation/solve/solver/adapter.rs
+++ b/crates/engine/src/operation/solve/solver/adapter.rs
@@ -1,7 +1,10 @@
 use std::borrow::Cow;
 
 use id_newtypes::IdRange;
-use schema::{ObjectDefinitionId, Schema, SchemaField};
+use im::HashMap;
+use schema::{
+    CompositeTypeId, FieldDefinition, FieldDefinitionId, ObjectDefinitionId, Schema, SchemaField, SubgraphId,
+};
 use walker::Walk;
 
 use crate::{
@@ -9,13 +12,14 @@ use crate::{
         BoundExtraField, BoundField, BoundFieldArgument, BoundFieldArgumentId, BoundFieldId, BoundOperation,
         QueryInputValueRecord, QueryPosition,
     },
-    response::SafeResponseKey,
+    response::{ResponseKeys, SafeResponseKey},
 };
 
 pub(super) struct OperationAdapter<'a> {
     pub schema: &'a Schema,
     pub operation: &'a mut BoundOperation,
     tmp_response_keys: Vec<SafeResponseKey>,
+    field_with_distinct_type_to_unique_key: HashMap<FieldDefinitionId, SafeResponseKey>,
 }
 
 impl<'a> OperationAdapter<'a> {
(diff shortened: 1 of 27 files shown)


crates/engine/src/operation/solve/solver/adapter.rs
```diff
@@ -171,3 +264,50 @@ impl<'a> OperationAdapter<'a> {
...
     }
 }
+
+/// Generating a response key for an extra field. The only constraint is that it needs to be unique
+/// within the selection set we send to a subgraph. Afterwards we identify extra fields through
+/// their associated `SchemaFieldId` so the response key doesn't matter past deserialization.
+fn generate_new_extra_field_key(
+    response_keys: &mut ResponseKeys,
+    selection_set: &[SafeResponseKey],
+    definition: FieldDefinition<'_>,
```
obmarg, line 274: What's the difference between this function and the very similarly named/written one above?  Answer in a comment please.

crates/engine/src/operation/solve/solver/adapter.rs
```diff
@@ -142,14 +147,102 @@ impl<'a> query_solver::Operation for OperationAdapter<'a> {
...
+    /// ```ignore,graphql
+    /// {
+    ///   ... on User {
+    ///     id
+    ///     name
+    ///   }
+    ///   ... on Admin {
+    ///     id
+    ///     name
+    ///   }
```
obmarg, line 189: I said it on slack, but this doesn't look like a legit query given the types that you've outlined.  If they're actually meant to be in a union/interface together can you make that clear.  And if not, can you eexplain how it works please.

crates/graphql-federated-graph/src/from_sdl.rs
```diff
@@ -1925,3 +1980,179 @@ fn test_missing_type() {
...
+
+        union Account
+            @join__unionMember(graph: B, member: "Admin")
+            @join__unionMember(graph: B, member: "User")
+         = User | Admin
+    "#]];
+
+    let actual = crate::render_sdl::render_federated_sdl(&super::from_sdl(sdl).unwrap()).unwrap();
+
+    expected.assert_eq(&actual);
```
obmarg, line 2157: This looks like (effectively) a roundtrip test - would it be better just written as one?  Like just `input == render(from_sdl(input))`

obmarg (approved):

crates/graphql-federated-graph/src/from_sdl.rs
Finistere: yeah should you're right, copy-pasted the input from rover so it's not exactly the same, should use the output

Finistere: well fixing this in a later PR

PR #2346, opened 2024-11-14
fix: redo `@listSize` in composition

obmarg:
I added support for `@listSize` in #2338 - but once I started trying to use it in engine I realised that I'd done a bad job.  I'd kept the two `[String]` arguments as strings - but these refer to fields & arguments so would be better converted into the corresponding Ids so that the engine doesn't have to do a bunch of string lookups when it's trying to calculate the complexity.

This was harder than it sounds, because the `sizedFields` argument refers to the fields of another object, which isn't necessarily available at the point where we're handling most of the directives.  Got there in the end though.

Fixes GB-7642 (again)

### DIFF
--- a/crates/engine/schema/src/builder/graph.rs
+++ b/crates/engine/schema/src/builder/graph.rs
@@ -791,8 +791,8 @@ impl<'a> GraphBuilder<'a> {
                         reason_id: reason.map(Into::into),
                     })
                 }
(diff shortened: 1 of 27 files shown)


tomhoule (approved):
Looks good, with the context of what we discussed yesterday. After the refactoring, we'll probably still need the list sizes separate in CompositionIr but we should be able to bring them back in Directive in the federated graph.

PR #2347, opened 2024-11-14
chore: pull complexity control into engine schema

obmarg:
As the title says.  Had to rework some codegen stuff so it could handle these.

Part of GB-7643

### DIFF
--- a/crates/engine/codegen/src/generation/object/walker.rs
+++ b/crates/engine/codegen/src/generation/object/walker.rs
@@ -272,51 +272,18 @@ impl quote::ToTokens for WalkerFieldMethod<'_> {
                         self.as_ref().#field.iter().copied()
                     }
                 },
-                AccessKind::Ref => quote! {
+                AccessKind::Ref | AccessKind::IdRef => quote! {
                     impl Iter<Item = &'a #ty> + 'a {
                         self.as_ref().#field.iter()
                     }
                 },
-                AccessKind::IdRef => {
-                    if list_as_id_range {
-                        quote! {
-                            impl Iter<Item = &'a #ty> + 'a {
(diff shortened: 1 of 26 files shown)


crates/engine/codegen/src/generation/object/walker.rs
```diff
@@ -272,51 +272,18 @@ impl quote::ToTokens for WalkerFieldMethod<'_> {
...
-                        }
-                    } else {
-                        quote! {
-                            impl Iter<Item = #ty<'a>> + 'a {
-                                self.as_ref().#field.walk(self.#ctx)
-                            }
+                AccessKind::IdWalker | AccessKind::ItemWalker | AccessKind::RefWalker => {
+                    quote! {
+                        impl Iter<Item =  #ty<'a>> + 'a {
+                            self.as_ref().#field.walk(self.#ctx)
```
obmarg, line 283: These changes will result in a few more `as_ref()` calls in the generated code - but I'm not convinced it was worth having all of this code to save a handful of characters.

crates/engine/schema/src/builder/graph.rs
```diff
@@ -15,11 +15,15 @@ pub(crate) struct GraphBuilder<'a> {
     ctx: &'a mut BuildContext,
     field_sets: FieldSetsBuilder,
     required_scopes: Interner<RequiresScopesDirectiveRecord, RequiresScopesDirectiveId>,
+    // deferred_list_directive_fields: Vec<(ListSizeDirectiveId, Vec<FieldId>)>
     graph: Graph,
```
obmarg, line 19:
```suggestion
    graph: Graph,
```

Finistere (approved):

PR #2348, opened 2024-11-15
Dev command internal release 2 + Fix `create`

yoav-lavi:
- Adds a hot reloader to the dev command, currently only handles configuration changes and reloads everything (remote subgraphs are cached).
- Some cleanup
- Removes a field to fix `create`

Note the the implementation and structure of the dev command is still a WIP, this release is to be able to test functionality internally.


crates/grafbase-local-backend/src/dev.rs
```diff
@@ -82,231 +90,50 @@ pub async fn start(
         }
     };
 
+    let (reload_sender, reload_receiver) = mpsc::channel::<(String, Config)>(1);
+
     let server_config = ServerConfig {
         listen_addr: Some(listen_address),
         config_path: None,
         config_hot_reload: false,
```
yoav-lavi, line 98: This will be removed and `config_path` will be used to determine whether to turn on the gateway's internal hot reloading for a static schema (which only currently changes rate limiter options), I had done this already but apparently have reverted it at some point

cli/src/output/log_formatters.rs
```diff
@@ -0,0 +1,43 @@
...
+            tracing::Level::WARN => "warn".yellow().bold(),
+            tracing::Level::INFO => "info".green().bold(),
+            tracing::Level::DEBUG => "debug".purple().bold(),
+            tracing::Level::TRACE => "trace".blue(),
+        };
+
+        match *metadata.level() {
+            tracing::Level::ERROR => write!(writer, "{} - ", level)?,
+            tracing::Level::WARN => write!(writer, "{}  - ", level)?,
+            tracing::Level::INFO => write!(writer, "{}  - ", level)?,
```
tomhoule, line 34: Is the spacing difference before the hyphen intentional?

tomhoule (approved):
nice!

cli/src/output/log_formatters.rs
yoav-lavi: Yeah, it's to align the dash since `info` and `warn` have one less letter

PR #2352, opened 2024-11-17
composition: simplify entity interfaces in composition IR

tomhoule:
Remove object_fields_from_entity_interfaces. That required changing the parent definition from DefinitionId to StringId in FieldIr, because we may not have the definition for the implementers of the interface yet when we compose the interface.


Finistere:
Asked on Slack, I don't understand why we're _adding_ fields 🤷 If a field from the interface is missing, shouldn't that be a fatal error?

Finistere (approved):
Ok got it, if you don't do it will add a comment later for this. I didn't realize you were adding fields provided from the objects

tomhoule:
Adding that comment now.

PR #2356, opened 2024-11-19
feat: complexity cost calculation

obmarg:
This PR implements _most_ of the cost calculation for complexity control.  It's relatively simple conceptually: we walk the query, assigning a cost to each of the types & fields that are present.  We either use a default cost or a cost configured in the schema.  Fields inside lists are multiplied by the expected length of the list.

Where it gets slightly more complicated is integrating it with the engine as it stands:

- I'd hoped to do this _before_ planning, but the engine is currently set up such that you can't use variables until after planning has happened.  I looked into working around this but it'd either make complexity control much more complicated or require a bit of rework around caching.  Just going to see how we get on with this approach for now and can rework later if we need to.
- Input values aren't set up to be easy to use with anything other than `Deserialize` impls at the moment.  For calculating the cost of input values we kind of need to walk the input value structure, but that is very painful right now.  I'm going to handle that in the next PR instead.
- I also found a bunch of dead code, which I've removed in this PR.

Fixes GB-7643
Fixed GB-7645


crates/engine/src/prepare/error.rs
```diff
@@ -38,6 +40,8 @@ impl From<PrepareError> for GraphqlError {
             PrepareError::Plan { err, .. } => err.into(),
             PrepareError::Solve { err, .. } => err.into(),
             PrepareError::NormalizationError => GraphqlError::new(err.to_string(), ErrorCode::InternalServerError),
+            // TODO: Is this a good error code? Not sure
```
Finistere, line 43: Think should be `OperationValidationError`

crates/engine/src/prepare/operation/complexity_control.rs
```diff
@@ -0,0 +1,192 @@
...
+
+    Ok(())
+}
+
+struct ComplexityContext<'a> {
+    default_list_size: usize,
+    variables: &'a Variables,
+}
+
+fn selection_set_complexity(
```
Finistere, line 48: I'm a bit confused as to how you apply `sizedFields` to be honest. Maybe it makes sense, but feels like the traversing of the selection sets and the propagation of the estimated list size from the `@listSize` directive are too intertwined to make it clear what's happening to me. I don't find the enum `ChildCount` to be straightforward either.

Finistere (approved):

crates/engine/src/prepare/error.rs
Finistere: for errors in the binding, operation limits we use `OperationValidationError`. `BadRequest` is used for stuff like improper content-type, ill-formed graphql request, etc.

crates/engine/src/prepare/operation/complexity_control.rs
obmarg: What are you confused about?

Finistere: Well, reading it it's not obvious to me if the logic is correctly applied. Maybe it's just a question of time, but I don't find `ChildCount` to be a very clear enum typically. For the sizeFields argumentsI would have expected something like this without an enum to propagate

```
fn compute_field_complexity(field, parent_count) {
  for field in selection_set {
     if field in sizeFields {
        compute_selection_set_complexity(field.subselection(), count)
     } else {
        compute_field_complexity(field, count)
     }
  }
  ...
}

fn compute_selection_set_complexity(selection_set, parent_count) {
  for field in selection_set {
        compute_field_complexity(field, count)
  }
  ...
}
```

But maybe I'm missing something 🤷

PR #2357, opened 2024-11-19
fix(engine): complex-entity-call test

Finistere:
The original goal was to fix `complex-entity-call` audit test which had this:

```graphql
type Category
    @join__type(graph: PRICE, key: "id tag")
    @join__type(graph: PRODUCTS, key: "id")
{
    id: String!
    tag: String
    mainProduct: Product! @join__field(graph: PRODUCTS)
}
```

We treated `tag` as only available in `PRICE` subgraph and not `PRODUCTS`. The problem is that our interpretation logic of the federated graph was distributed across the federated-graph and the engine. So it quickly became tricky to understand exactly what to change and how. So I changed the federated graph to only be a structured representation of the federated graph, keeping directives as is. Still parsing the selection sets, but not interpreting the meaning of the directives.

I initially wanted to avoid modifying the composition, but it was impossible to do so. So had to change it also, while iterating on this I stumbled upon many small issues like `type` missing in `@join__field`, interface fields not being composed together correctly, missing `@join__type` / `@join__field`, etc. So while iterating on tests I compared with rover to see what they do.

As I had to rework so much stuff for the federated directives, I went all the way and improved the directive handling overall. Now directives are handled at the end in composition, making it a lot easier to add ones like `@listSize` that require the federated graph definitions & fields to be available. All directives are now in a single enum, making it simpler to manipulate.

I've added a simple bash script to run and update the composition snapshots, it was quite painful to do without.
https://github.com/grafbase/grafbase/pull/2357/files#diff-cde56a53ec028ff4607dd6ba0ba638af333e6c1e12e2e253740b787e14f03f0c

I've changed the federation prelude to use `join__FieldSet` instead of `String`:
```graphql
directive @join__field(
    graph: join__Graph
    requires: join__FieldSet
    provides: join__FieldSet
) on FIELD_DEFINITION
```

All in all, this PR could be summarized as changing the `FederatedGraph` to use the following to store *all* directives:
```rust
pub enum Directive {
    Authenticated,
    Deprecated {
        reason: Option<StringId>,
    },
    Inaccessible,
    Policy(Vec<Vec<StringId>>),
    RequiresScopes(Vec<Vec<StringId>>),
    Cost {
        weight: i32,
    },
    JoinField(JoinFieldDirective),
    JoinType(JoinTypeDirective),
    JoinUnionMember(JoinUnionMemberDirective),
    JoinImplements(JoinImplementsDirective),
    Authorized(AuthorizedDirective),
    Other {
        name: StringId,
        arguments: Vec<(StringId, Value)>,
    },
    ListSize(ListSize),
}
```
And adapt engine & composition for it.


crates/engine/query-solver/src/operation/builder.rs
```diff
@@ -75,6 +77,38 @@ impl<'ctx, Op: Operation> OperationGraphBuilder<'ctx, Op> {
             }
         }
 
+        for (id, node_ix) in self.field_nodes.iter().copied().enumerate() {
```
Finistere, line 80: Moved code for clarity

crates/engine/codegen/domain/schema.graphql
```diff
@@ -77,7 +77,7 @@ type InterfaceDefinition
 type FieldDefinition @meta(module: "field") @indexed(id_size: "u32", max_id: "MAX_ID") {
   name: String!
   description: String
-  parent_entity: EntityDefinition! @meta(debug: false)
```
Finistere, line 80: unused and it's actually helpful

crates/graphql-composition/src/compose/scalar.rs
```diff
@@ -5,6 +5,9 @@ pub(crate) fn merge_scalar_definitions<'a>(
     definitions: &[DefinitionWalker<'a>],
     ctx: &mut Context<'a>,
 ) {
+    if first.name().as_str() == "join__FieldSet" {
```
tomhoule, line 8: I wonder if we should have special treatment in a general way for scalars defined by federation directives (there are others IIRC)

tomhoule (approved):
Very nice bundle of improvements :100:

crates/engine/schema/src/builder/graph.rs
```diff
@@ -811,67 +747,67 @@ impl<'a> GraphBuilder<'a> {
...
+                    let list_size_id = self.graph.list_size_directives.len().into();
+                    self.graph.list_size_directives.push(ListSizeDirectiveRecord {
+                        assumed_size: *assumed_size,
+                        slicing_argument_ids: slicing_arguments
+                            .iter()
+                            .filter_map(|id| self.ctx.idmaps.input_value.get(*id))
+                            .collect(),
+                        sized_field_ids: sized_fields
+                            .iter()
+                            .filter_map(|id| self.ctx.idmaps.field.get(*id))
```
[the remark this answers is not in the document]
Finistere: Next PR will remove those, yes^^

crates/graphql-composition/src/compose/scalar.rs
Finistere: yeah think so, this was really adhoc

PR #2368, opened 2024-11-20
chore: complexity control tests

obmarg:
1. Adds integration tests of complexity control
2. Fixes a few issues I found in complexity control
3. Fixes an issue I found in the dynamic graphql mock


crates/integration-tests/tests/federation/complexity_control.rs
```diff
@@ -0,0 +1,343 @@
...
+                )
+
+                connectionField(first: Int): Connection @listSize(
+                    slicingArguments: "first"
+                    sizedFields: "items"
+                )
+            }
+
+            # TODO: add arguments that cost
+            # TODO: add inpput object fields that cost
```
obmarg, line 319: Will do these once I've implemented that bit

crates/engine/src/prepare/operation/complexity_control.rs
```diff
@@ -135,48 +147,53 @@ fn calculate_child_count<'a>(
...
 
     let mut slicing_arguments = directive.slicing_arguments().peekable();
     if slicing_arguments.peek().is_some() {
-        let argument_size = slicing_arguments
+        let slicing_arguments = slicing_arguments
             .filter_map(|argument| field.argument(argument.name())?.value(context.variables).as_usize())
-            .max();
+            .collect::<Vec<_>>();
+
+        if directive.require_one_slicing_argument && slicing_arguments.len() != 1 {
```
Finistere, line 174: shouldn't that be `!slicing_arguments.is_empty()`? or is two arguments also an error?

Finistere (approved):

crates/engine/src/prepare/operation/complexity_control.rs
Finistere: Ok. well might be worth a comment imho and/or rename the error to `ExactlyOne` rather than `One`. Up to you 🤷

PR #2370, opened 2024-11-20
fix(engine): Fix requires-with-argument

Finistere:
There was an error with the query partition cycles handling. It wouldn't
detect correctly that fields within the same partition had were
dependent on each other. I had a simple check that would only process
the fields if there were at least two sibling partitions, but failed to
take into nested ones.


tomhoule (approved):
After thinking about it for a bit, I think I get it. Good catch!

PR #2375, opened 2024-11-21
feat: add complexity into the graphql operation span

obmarg:


crates/engine/src/response/mod.rs
```diff
@@ -44,6 +44,7 @@ pub(crate) enum Response<OnOperationResponseHookOutput> {
 
 pub(crate) struct ExecutedResponse<OnOperationResponseHookOutput> {
     operation: Arc<CachedOperation>,
+    attributes: GraphqlOperationAttributes,
```
Finistere, line 47: `operation_attributes`?

Finistere (approved):

crates/engine/src/prepare/operation/mod.rs
```diff
@@ -94,3 +102,17 @@ fn mutation_not_allowed_with_safe_method<OnOperationResponseHookOutput>() -> Res
         )],
     )
 }
+
+fn validate_prepared_operation(operation: &PreparedOperation, settings: &Settings) -> PrepareResult<()> {
+    let ComplexityControl::Enforce { limit, .. } = settings.complexity_control else {
+        return Ok(());
+    };
+    let Some(complexity) = operation.complexity else {
+        return Ok(());
```
Finistere, line 111: Shouldn't that be an `InternalError` or smth?

obmarg: Although I should at the least add an error message in probably.

Finistere: yeah but silently having a problem sounds worse to me. At least there should be a `tracing::error!` in that case

PR #2377, opened 2024-11-21
cli: remove --header argument in publish

tomhoule:
What is up with that argument? It is not used, and there is no introspection involved in publish, so it is misleading.

image


yoav-lavi (approved):

obmarg:
> What is up with that argument?

It used to be used for the refreshing introspection calls in `publish --dev`.  But yeah, removable.

tomhoule:
> It used to be used for the refreshing introspection calls in publish --dev. But yeah, removable.

Aaah ok, that makes sense, I thought it just landed there by accident. And yes, since it doesn't do anything anymore, let's remove.

PR #2396, opened 2024-11-26
Dev3

yoav-lavi:


yoav-lavi:
Things left to do after this PR:
- Various optimizations can be made (e.g. using the subgraph cache for local subgraphs as well if reloading from a file, extending it to support file based subgraphs, using an internal function from our introspection crate so it doesn't prettify)
- Hints should be added for errors
- Either sync the internal html file with the pathfinder assets or stop bundling the .hbs
- Error or disallow using dev fields in the normal configuration outside of the dev command (can be done in multiple ways)

crates/grafbase-local-backend/src/dev.rs
```diff
@@ -109,40 +126,46 @@ pub async fn start(
...
 
     Ok(())
 }
 
-// temporary output handler for internal testing until we move output to the CLI and use a proper terminal crate.
-// none of us uses Windows, right?
-fn output_handler(mut receiver: Receiver<String>) {
-    // gray
-    println!("\x1b[90mWarning: This command is in beta, expect missing features, bugs or breaking changes\x1b[0m\n");
+fn output_handler(mut receiver: Receiver<String>) -> Result<(), Box<dyn std::error::Error>> {
```
yoav-lavi, line 153: Currently leaving this as an internal handler since this will be merged into the normal CLI

crates/grafbase-local-backend/src/dev.rs
```diff
@@ -53,32 +67,35 @@ pub async fn start(
...
         graph_ref.as_ref(),
-        &dev_configuration,
+        &dev_configuration.overridden_subgraphs,
+        &dev_configuration.merged_configuration,
         &mut subgraphs,
         graph_overrides_path,
     )
     .await?;
 
+    let subgraph_cache = Box::leak(Box::new(subgraph_cache));
```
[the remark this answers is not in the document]
yoav-lavi: Apparently I didn't copy the comment when I copied the leak, the subgraph cache lives for the duration of the run so leaking it makes it easier to pass around with multiple consumers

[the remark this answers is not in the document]
yoav-lavi: Although now that it's an Arc this might just be a leftover, let me see if I can remove that

crates/grafbase-local-backend/src/dev/hot_reload.rs
```diff
@@ -1,18 +1,253 @@
...
+            sender,
+            subgraph_cache,
+            overridden_subgraphs,
+            merged_configuration,
+            graph_overrides_path,
+        )
+        .await
+    }
+
+    async fn introspection_poller(
```
obmarg, line 75: Does this function need to be async?

Also might be clearer if it was named `spawn_introspection_poller` - hard to tell at the callsite whether it spawns something or actually does the polling.

crates/grafbase-local-backend/src/dev/hot_reload.rs
```diff
@@ -1,18 +1,253 @@
...
+            }
+        });
+
+        Ok(())
+    }
+
+    #[allow(clippy::too_many_arguments)]
+    async fn schema_file_watcher(
+        &mut self,
+        sender: mpsc::Sender<(String, Arc<Config>)>,
```
obmarg, line 185: Same comments here

crates/grafbase-local-backend/src/dev.rs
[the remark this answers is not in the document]
yoav-lavi: Done, removed. Good catch

crates/grafbase-local-backend/src/dev/hot_reload.rs
yoav-lavi: RE the first point - WDYM? It calls multiple async functions and does the polling in parallel

yoav-lavi: renamed

crates/grafbase-local-backend/src/dev/hot_reload.rs
yoav-lavi: renamed

crates/grafbase-local-backend/src/dev/hot_reload.rs
yoav-lavi: Done

PR #2397, opened 2024-11-26
graphql-schema-diff: overhaul path strings

tomhoule:
Until now, the paths in a change have been of the format `a.b.c`, just nesting a path. This PR introduces a more structured representation, along with a new string representation, that should address three problems and allow us to represent any path within a GraphQL schema.

1. We need to be able to address items that don't have their own names. These are schema and type extensions, as well as repeatable directives. So I introduced indexing, like `:schema[2]`, `Query[0]`, `@tag[3]`, etc.

2. Introduce tokens to distinguish between different namespaces that could be addressed by a path. On type definitions, we want to distinguish paths leading to fields (or enum values, etc.) and paths to directives on the types. To give an example, disambiguate cases like this:
  ```graphql
  # What does the path "Test.key" refer to?
  type Test @key(fields: "id") {
    id: ID!
    key: String!
  }
  ```
  Similar problem at the top level. This is valid GraphQL:

  ```graphql
  type test { id: ID! }
  directive @test
  ```
  So for directives, we have the `@` prefix. This gives us `test` vs `@test` in the example above.

3. This results in a small grammar. So there should be a single way to parse and print it, and that should be tested. So that is done in the new `path` module.

This PR started as a port of https://github.com/grafbase/api/blob/37969f88372f774f973e247174b1d5ef6c2c0f20/services/graphql/src/application/schema_proposal/comment/sdl_map/path.rs, with a suggestion from @obmarg on a more readable syntax (`@directive` instead of `d:directive`). I changed how indexes are represented (`type[0]` instead of `t:type:0` for example) and had to introduce special syntax (`&`) for interface implementations, since those were missing.

The path module in grafbase/api will be immediately replaced with the implementation in this PR.

closes GB-7919

### DIFF
--- a/crates/graphql-schema-diff/src/patch/schema_definitions.rs
+++ b/crates/graphql-schema-diff/src/patch/schema_definitions.rs
@@ -14,7 +14,12 @@ pub(super) fn patch_schema_definition<T: AsRef<str>>(
     let mut new_mutation_type = None;
     let mut new_subscription_type = None;
 
-    for change in paths.iter_exact([""; 3]) {
+    let prefix = match definition_or_extension {
+        DefinitionOrExtension::Extension => ":schema[0]",
+        DefinitionOrExtension::Definition => ":schema",
+    };
+
+    for change in paths.iter_exact([prefix, "", ""]) {
         match change.kind() {
             ChangeKind::ChangeQueryType => {
                 new_query_type = Some(change.resolved_str());
--- a/crates/graphql-schema-diff/src/path/parse.rs
+++ b/crates/graphql-schema-diff/src/path/parse.rs
@@ -1,0 +1,173 @@
+use super::*;
+
+impl<'a> Path<'a> {
+    pub fn parse(s: &'a str) -> ParseResult<Path<'a>> {
+        let mut segments = s.split('.');
+
+        let first_segment = segments.next().filter(|s| !s.is_empty()).ok_or(ParseError)?;
+
+        // Is it a schema definition or extension?
+        if let Some(suffix) = first_segment.strip_prefix(":schema") {
+            if segments.next().is_some() {
+                return Err(ParseError);
+            }
+
+            if suffix.is_empty() {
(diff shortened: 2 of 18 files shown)


crates/graphql-schema-diff/src/patch/schema_definitions.rs
```diff
@@ -14,7 +14,12 @@ pub(super) fn patch_schema_definition<T: AsRef<str>>(
     let mut new_mutation_type = None;
     let mut new_subscription_type = None;
 
-    for change in paths.iter_exact([""; 3]) {
+    let prefix = match definition_or_extension {
+        DefinitionOrExtension::Extension => ":schema[0]",
```
tomhoule, line 18: This is of course not correct in general, but it's no less correct than what we had before. Extensions support in graphql-schema-diff has to be implemented, and the first step is to represent extensions in paths like as done in this PR.

crates/graphql-schema-diff/src/path/parse.rs
```diff
@@ -0,0 +1,173 @@
...
+    pub fn parse(s: &'a str) -> ParseResult<Path<'a>> {
+        let mut segments = s.split('.');
+
+        let first_segment = segments.next().filter(|s| !s.is_empty()).ok_or(ParseError)?;
+
+        // Is it a schema definition or extension?
+        if let Some(suffix) = first_segment.strip_prefix(":schema") {
+            if segments.next().is_some() {
+                return Err(ParseError);
+            }
```
obmarg, line 13: Take it we don't care about directives on schema definitions?

obmarg (approved):

crates/graphql-schema-diff/src/path/parse.rs
tomhoule: It's not implemented yet ([issue](https://linear.app/grafbase/issue/GB-7389/graphql-schema-diff-implement-directive-addition-and-removal-as-well))

PR #2398, opened 2024-11-26
chore(engine): Simplify id_derives

Finistere:
Remove the max id stuff, useless. And added support to just have `u32`
instead of `NonZero<u32>`. I put a lot of `NonZero`, but not sure it's
that useful.

### DIFF
--- a/crates/engine/codegen/domain/schema.graphql
+++ b/crates/engine/codegen/domain/schema.graphql
@@ -19,7 +19,7 @@ type RootOperationTypes @meta(module: "root") {
 }
 
 union Definition @id @meta(module: "definition") @variants(remove_suffix: true) =
-    ObjectDefinition
+  | ObjectDefinition
   | InterfaceDefinition
   | UnionDefinition
   | EnumDefinition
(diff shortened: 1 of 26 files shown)


pimeys (approved):

crates/engine/codegen/domain/schema.graphql
```diff
@@ -19,20 +19,20 @@ type RootOperationTypes @meta(module: "root") {
...
   | ScalarDefinition
 
 union EntityDefinition @id @meta(module: "entity") @variants(remove_suffix: "Definition") =
-    ObjectDefinition
+  | ObjectDefinition
   | InterfaceDefinition
 
 "Composite type is the term previously used by the GraphQL spec to describe this union."
 union CompositeType @id @meta(module: "composite_type") @variants(remove_suffix: "Definition") =
-    ObjectDefinition
```
obmarg, line 35: How are you formatting this file?

Finistere: prettier I think?

PR #2399, opened 2024-11-26
feat: http-client for hooks

pimeys:
This adds a HTTP client functionality to the gateway hooks. The idea is you can run one or multiple requests asynchronously using the host runtime from the guest.

WIT additions:

```wit
interface types {
    resource http-client {
        execute: static func(request: http-request) -> result<http-response, http-error>;

        execute-many: static func(requests: list<http-request>) -> list<result<http-response, http-error>>;
    }

    record http-request {
        method: http-method,
        url: string,
        headers: list<tuple<string, string>>,
        body: list<u8>,
        timeout-ms: option<u64>,
    }
    
    enum http-method {
        get,
        post,
        put,
        delete,
        patch,
        head,
        options,
        connect,
        trace,
    }

    record http-response {
        status: u16,
        version: http-version,
        headers: list<tuple<string, string>>,
        body: list<u8>,
    }
    
    enum http-version {
        http-0-9,
        http-1-0,
        http-1-1,
        http-2-0,
        http-3-0,
    }

    variant http-error {
        timeout,
        request(string),
        connect(string),
    }
}
```

Usage in Rust guest, single request:

```rust
impl gateway_request::Guest for Component {
    fn on_gateway_request(context: Context, _: Headers) -> Result<(), ErrorResponse> {
        let address = std::env::var("MOCK_SERVER_ADDRESS").unwrap();

        let request = HttpRequest {
            method: HttpMethod::Get,
            url: address,
            headers: Vec::new(),
            body: Vec::new(),
            timeout_ms: None,
        };

        let response = HttpClient::execute(&request).unwrap();
        let body = String::from_utf8(response.body).unwrap();

        context.set("HTTP_RESPONSE", &body);

        Ok(())
    }
}
```

multiple requests:

```rust
pub(super) fn authorize_user(current_user_id: usize, user_ids: Vec<usize>) -> Vec<Result<(), Error>> {
    let requests = user_ids
        .into_iter()
        .map(|user_id| {
            tracing::info!(
                "Authorizing access to user of user {} for user {}",
                user_id,
                current_user_id
            );

            HttpRequest {
                method: HttpMethod::Post,
                url: String::from("http://localhost:4001/authorize-user"),
                headers: vec![("Content-Type".to_string(), "application/json".to_string())],
                body: serde_json::to_vec(&AuthorizeUserRequest {
                    current_user_id,
                    user_id,
                })
                .unwrap(),
                timeout_ms: Some(1000),
            }
        })
        .collect_vec();

    HttpClient::execute_many(&requests)
        .into_iter()
        .map(|result| match result {
            Ok(response) => {
                let body: AuthorizationResponse =
                    serde_json::from_slice(&response.body).expect("Failed to deserialize authorization response");

                if body.authorized {
                    Ok(())
                } else {
                    Err(error("Unauthorized"))
                }
            }
            Err(err) => {
                tracing::error!("Auth service request failure: {err:?}");
                Err(error("Unauthorized"))
            }
        })
        .collect()
}
```

Spans are recorded per request under a hook call parent, we see they are running in parallel:

image (7)

Every HTTP request adds to the exponential histogram `grafbase.hook.http_request.duration` the time it took to make the request. The histogram also acts as a counter and holds the following attributes:

- `http.response.status_code`
- `http.request.method`
- `http.route`
- `network.protocol.version`
- `server.address`
- `server.port`
- `url.scheme`


crates/wasi-component-loader/src/lib.rs
```diff
@@ -127,6 +128,7 @@ impl ComponentLoader {
                 headers::map(&mut types)?;
                 context::map(&mut types)?;
                 context::map_shared(&mut types)?;
+                http_client::map(&mut types)?;
```
Finistere, line 131: IMO `inject_mapping` or something like this would be clearer than `map`. When I read `map` here I understand something like `Iterator::map` / `Operation::map`, not "mapping"

crates/wasi-component-loader/src/http_client.rs
```diff
@@ -0,0 +1,362 @@
...
+};
+use tracing::{field::Empty, info_span, Instrument};
+use wasmtime::{
+    component::{ComponentType, Lift, LinkerInstance, Lower, ResourceType},
+    StoreContextMut,
+};
+
+static HTTP_CLIENT: LazyLock<reqwest::Client> = LazyLock::new(reqwest::Client::new);
+
+static REQUEST_METRICS: LazyLock<Histogram<u64>> = LazyLock::new(|| {
```
Finistere, line 25: Not a fan of that static implicitly relying on a global setup so deep down. All of this should be provided by the ComponentLoader with an `Arc` instead of a deep static lock hiding dependencies. Same for the http client, no need for a static lock. Just setup one in the ComponentLoader and clone it.

Finistere (approved):

crates/wasi-component-loader/src/http_client.rs
```diff
@@ -0,0 +1,362 @@
...
+    Head,
+    #[component(name = "options")]
+    Options,
+    #[component(name = "connect")]
+    Connect,
+    #[component(name = "trace")]
+    Trace,
+}
+
+impl AsRef<str> for HttpMethod {
```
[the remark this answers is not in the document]
Finistere: Ah right, good point

PR #2401, opened 2024-11-27
cli: make publish idempotent

tomhoule:
Now the CLI will report when you publish a subgraph and neither the subgraph schema nor the url changes (they are identical to what is already published). In that case, no deployment will be created in the schema registry. This is not an error, so the exit code remains 0.

The noise in the diff is from updating the platform's GraphQL schema (in particular the api keys removal).

Companion platform PR: https://github.com/grafbase/api/pull/4269

### DIFF
--- a/cli/src/output/report.rs
+++ b/cli/src/output/report.rs
@@ -187,6 +187,10 @@ pub(crate) fn publish_command_success(subgraph_name: &str) {
     println!("🧩 {subgraph_name} published successfully");
 }
 
+pub(crate) fn publish_no_change() {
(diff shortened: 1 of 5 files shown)


Finistere (approved):

obmarg:
> The noise in the diff is from updating the platform's GraphQL schema (in particular the api keys removal).

Wonder if it'd be worth automating this update.  I made an action a while back that makes it quite easy ([example of using it over here](https://github.com/obmarg/cynic/blob/main/.github/workflows/update-schemas.yaml))

PR #2402, opened 2024-11-27
chore: make engine use cynic-parser

obmarg:
We decided a while ago to standardise on cynic-parser as it's faster & a bit more ergonomic than `async-graphql-parser` (and therefore `engine-parser`, which is just a fork of `async-graphql-parser`.  So this PR swaps the parser in `engine` to `cynic-parser`.

It's a fairly mechanical swap, the only complicating factor is cynic-parser tracking Spans as byte offsets, whereas we need line & column info for the `position` field in GraphQL errors.  So I've added some translation code for that.

Running our benchmarks this sees a 13-17% decrease in time in the benchmarks without an operation cache.  Although I wouldn't expect to see similar results in production since the operation cache bypasses parsing and waiting on subgraphs is likely to be much slower than parsing ever was.

### DIFF
--- a/crates/engine/src/operation/bind/error.rs
+++ b/crates/engine/src/operation/bind/error.rs
@@ -1,125 +1,96 @@
-use crate::{
-    operation::{Location, LocationError},
-    response::GraphqlError,
-    ErrorCode,
-};
+use cynic_parser::Span;
+
+use crate::{response::GraphqlError, ErrorCode};
+
+use super::{Location, ParsedOperation};
 
 #[derive(thiserror::Error, Debug)]
 pub enum BindError {
     #[error("Unknown type named '{name}'")]
-    UnknownType { name: String, location: Location },
+    UnknownType { name: String, span: Span },
     #[error("The field `{field_name}` does not have an argument named `{argument_name}")]
     UnknownArgument {
         field_name: String,
         argument_name: String,
-        location: Location,
+        span: Span,
     },
     #[error("{container} does not have a field named '{name}'")]
     UnknownField {
         container: String,
         name: String,
-        location: Location,
+        span: Span,
     },
     #[error("Unknown fragment named '{name}'")]
-    UnknownFragment { name: String, location: Location },
+    UnknownFragment { name: String, span: Span },
     #[error("Field '{name}' does not exists on {ty}, it's a union. Only interfaces and objects have fields, consider using a fragment with a type condition.")]
-    UnionHaveNoFields {
-        name: String,
-        ty: String,
-        location: Location,
-    },
+    UnionHaveNoFields { name: String, ty: String, span: Span },
     #[error("Field '{name}' cannot have a selection set, it's a {ty}. Only interfaces, unions and objects can.")]
-    CannotHaveSelectionSet {
-        name: String,
-        ty: String,
-        location: Location,
-    },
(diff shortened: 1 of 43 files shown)


crates/engine/src/operation/bind/error.rs
```diff
@@ -70,31 +55,29 @@ pub enum BindError {
...
         name: String,
-        location: Location,
+        location: Span,
     },
     #[error("Missing argument named '{name}' for directive '{directive}'")]
     MissingDirectiveArgument {
         name: String,
         directive: String,
-        location: Location,
+        location: Span,
```
Finistere, line 70: maybe `span` instead?

crates/engine/src/operation/parse/mod.rs
```diff
@@ -1,58 +1,83 @@
...
+use self::error::{ParseError, ParseResult};
+use super::Location;
+use crate::response::GraphqlError;
 
 pub(crate) struct ParsedOperation {
     pub name: Option<String>,
-    pub definition: OperationDefinition,
-    pub fragments: HashMap<engine_value::Name, Positioned<FragmentDefinition>>,
+    document: cynic_parser::ExecutableDocument,
+    line_offsets: Rc<offsets::LineOffsets>,
```
Finistere, line 18: You need a `Rc`? aren't you passing it everywhere with a ref?

crates/engine/src/prepare/operation/before_variables.rs
```diff
@@ -13,15 +13,22 @@ impl<'ctx, R: Runtime> PrepareContext<'ctx, R> {
         request: &Request,
         document: &str,
     ) -> PrepareResult<CachedOperation> {
-        let parsed_operation = crate::operation::parse(self.schema(), request.operation_name.as_deref(), document)?;
+        // TODO: Make the number here configurable probably?
```
Finistere, line 16: ah nice was thinking of this. we have `request_body_limit` in the config which serves a similar purpose, just it's for the whole body.

Finistere (approved):

crates/engine/src/operation/bind/error.rs
obmarg: https://github.com/grafbase/grafbase/blob/4b0bcb18deeb6983adc7a8e7df25b58f82ed0547/crates/engine/src/operation/bind/error.rs#L7

crates/engine/src/operation/parse/mod.rs
obmarg: If you scroll down to the parse function you'll see the problem - I need `ParsedOperation` to own `LineOffsets` but I also need a `LineOffsets` to generate the errors if I can't build a `ParsedOperation`.

An RC seemed preferable to cloning the whole vec (even though the vec will probably be small in most cases)

Finistere: well up to you, from a quick look I think I would return `(name, document)` from the `parse_iml` and create the `ParsedOperation` + validation inside `parse`.  Seems like you would avoid the clone that way

obmarg: Way ahead of you https://github.com/grafbase/grafbase/pull/2402/commits/ad372831afa00436b533a67486f9b562f310e34f

Finistere (approved):

PR #2407, opened 2024-11-27
feat: Hooks Rust SDK

pimeys:
# Grafbase Hooks Rust SDK

Implements two new crates

- `grafbase-hooks` is an SDK to help building gateway hooks.
- `grafbase-hooks-derive` is the proc macro crate, providing the `grafbase_hooks` macro for the SDK

There's now only one wit file, which is included in the `grafbase-hooks` repo and which generates the mappings when you compile the hooks using the SDK.

The `HttpClient` type is gone, the sdk provides `http_client` module with simple functions `execute` and `ececute_many`.

The `SharedContext` does not have `log_access` anymore, but the user can call `access_log::send` from any point of their code.

All hooks give `&mut self`, and the hooks module is instantiated when a new wasm instance is created. The user can now store data in their hooks if they want.

See the example projects how this thing works, and read the RFC:

https://linear.app/grafbase/document/rust-api-for-hooks-9b0611f74d4a

- crate: https://crates.io/crates/grafbase-hooks
- docs: https://docs.rs/grafbase-hooks/latest/grafbase_hooks/

Questions not answered this week:

- The HookImplementation enum is in two spots now. I'd like to have it in one place, but don't yet know where.
- The grafbase cli commands are not yet done, still thinking about those. Meanwhile the user can use cargo-component to build the hooks.

As a bonus, if the user forgets to add the `grafbase_hooks` proc macro, we have a pretty nice error message from rustc:

```
error[E0277]: Missing grafbase_hooks macro on the hooks implementation
   --> crates/simple/src/lib.rs:30:17
    |
30  | register_hooks!(MyHooks);
    |                 ^^^^^^^ On this trait implementation
    |
    = help: the trait `HookImpls` is not implemented for `MyHooks`
    = note: Add #[grafbase_hooks] to the trait implementation
```

Closes: GB-8090


.config/hakari.toml
```diff
@@ -25,5 +25,8 @@ platforms = [
 
 workspace-hack-line-style = "workspace-dotted"
 
+[traversal-excludes]
+workspace-members = ["grafbase-hooks", "grafbase-hooks-derive"]
```
pimeys, line 29: So... The test hooks want to use the grafbase-hooks crate from the workspace for easier tests. Now, if they require the workspace hack, it means there's suddenly tokio in the dependency chain. Tokio doesn't compile nicely to wasm32-wasip1, so we should exclude these two crates from hakari.

Luckily the SDK and derive have very small amount of dependencies.

obmarg (approved):

.config/hakari.toml
tomhoule: Does it actually fail to compile? My understanding is that cargo hakari is for Cargo.lock, the deps won't end up being compiled, since they aren't used.

tomhoule (approved):
nice!

.config/hakari.toml
pimeys: Yeah. So the dependency for workspace hack starts compiling axum and tokio in the examples without this.

pimeys: And wasip1 target fails hard there.

PR #2411, opened 2024-11-29
graphql-schema-diff: optional nested diffs for additions

tomhoule:


tomhoule:
Not going to need this after all, so I'm going to close this PR. It would still be a legit feature if someone else needs it.

PR #2420, opened 2024-12-03
chore(engine): Better handling of entity errors

Finistere:
When preparing a request for a federated entity, we may or may not have all the requirements. A subgraph request may have partially failed and we only have fields for some of the entities. the `null-keys` tests a subgraph simply returns `null` for one of the entities for example.

This means that for each entity we need to keep track of whether we could retrieve all the requirements or not. I wasn't handling this. While reading the response, I was treating a non-present field as null. Which led to sending invalid representations to subgraphs. The whole response handling wasn't really build on the idea of treating each entity differently. I was treating it as a list of entities that must be in the same order as the inputs. This made it more painful than necessary to treat entities separately.

So in this PR I've added a `InputObjectId` which indexes the `InputResponseObjectSet`, the input objects of each resolver. Now it's up to the federated entity resolver to keep trackof which object to update for which entity. Which makes it a lot easier to treat them differently. For this I had to rewrite a good chunk of the federated entity resolver, but should be clearer.

I've removed a bunch of intermediate layers for the  subgraph response handling giving pretty much direct access to the `Rc<RefCell<SubgraphResponse>>` instead of `ResponseWriter` stuff. So less indirection overall.

I tried to make more sense out of the serde error management also, keeping them only for invalid subgraph response errors. Still not great, but should be a bit simpler now.

Unfortunately this breaks `requires-with-fragments` tests now that we're handling errors in a better way, so have to fix this in another PR.


crates/engine/src/resolver/graphql/cache.rs
```diff
@@ -0,0 +1,178 @@
+use futures::future::join_all;
```
Finistere, line 1: mostly moved code with new output types and sometimes merging functions

crates/engine/src/resolver/graphql/deserialize/errors.rs
```diff
@@ -86,7 +64,16 @@ where
...
     }
 }
+
+#[derive(serde::Deserialize)]
+struct SubgraphGraphqlError {
+    message: String,
+    #[serde(default)]
+    path: serde_json::Value,
+    #[serde(default)]
+    extensions: serde_json::Value,
```
tomhoule, line 78: Nitpick: can't we assume this is a `Map<String, serde_json::Value>`?

crates/engine/src/resolver/graphql/deserialize/errors.rs
```diff
@@ -86,7 +64,16 @@ where
...
+        self.response.borrow_mut().set_subgraph_errors(errors);
         Ok(errors_count)
     }
 }
+
+#[derive(serde::Deserialize)]
+struct SubgraphGraphqlError {
+    message: String,
+    #[serde(default)]
+    path: serde_json::Value,
```
tomhoule, line 76: Also nitpick, but we should be able to assume `Vec<String>` for the path (https://spec.graphql.org/draft/#sec-Errors.Error-Result-Format)

crates/engine/src/response/object_set.rs
```diff
@@ -69,7 +69,7 @@ pub(crate) struct InputResponseObjectSet {
 impl InputResponseObjectSet {
     pub(crate) fn with_response_objects(mut self, refs: Arc<ResponseObjectSet>) -> Self {
         let n = self.indices.len();
-        self.indices.reserve(refs.len());
+        self.indices.reserve_exact(refs.len());
```
tomhoule, line 72: Just curious, why reserve_exact here?

tomhoule (approved):
nice!

crates/engine/src/resolver/graphql/deserialize/errors.rs
Finistere: We could, but we propagate it as is, so this avoids dealing with invalid formats. It'll be up to the clients to deal with it if this happens.

crates/engine/src/resolver/graphql/deserialize/errors.rs
Finistere: Kind of the same reason actually, we don't want deserialization to fail because of an invalid data format. This would stop the whole deserialization and not consume the rest of the JSON. So we need to accept anything and deal with problems as they come. So this actually needs to be also reworked (doing that currently for the data part). Thanks for pointing it out^^

crates/engine/src/response/object_set.rs
Finistere: well, premature optimization to be honest. will remove it. We'll never write more in this function than `refs.len()` and it's unlikely to call this function multiple times. But who knows 🤷

crates/engine/src/resolver/graphql/deserialize/errors.rs
tomhoule: makes sense :+1:

crates/engine/src/response/object_set.rs
tomhoule: I was wondering because maybe it's pointing a flaw in my understanding. I thought there wasn't a performance advantage to reserve_exact(), and it was just a way to fine-tune the capacity without really an upside. Looking at the docs, I didn't know that part

> unlike [reserve](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.reserve), this will not deliberately over-allocate to speculatively avoid frequent allocations.

https://doc.rust-lang.org/std/vec/struct.Vec.html#method.reserve_exact

So my mental model was wrong. I'd use reserve_exact() here too given that.

PR #2425, opened 2024-12-04
fix(engine): add a ton of serde failure tests

Finistere:
In the engine I'm directly deserializing into the response without passing through an intermediate type like `serde_json::Value`. This means validating the data while the deserializer parses it. I had only partially understood how to properly do it though.

During deserialization we must accept possible type, that's the only way drive forward the deserialization. Meaning that if for:
```graphql
type User {
   id: ID!
   name: String
   age: Int
}
```

the subgraph returns
```json
{"id": "1", "name": [], "age": 25}
```

we have to send back the following
```json
{"id": "1", "name": null, "age": 25}
```

But before this PR I would handle this correctly, expecting a String for `name` and the deserializer would get stuck at the `[` token having no sense anymore for the parent map visitor. So instead of expecting a String, now I implement a `Visitor` that accepts every possible type. So we have something like this now:

```rust
    fn visit_string<E>(self, v: String) -> Result<Self::Value, E>
    where
        E: Error,
    {
        Ok(ResponseValue::String {
            value: v.into_boxed_str(),
        })
    }

    fn visit_seq<A>(self, mut seq: A) -> Result<Self::Value, A::Error>
    where
        A: serde::de::SeqAccess<'de>,
    {
        // Try discarding the rest of the list, we might be able to use other parts of
        // the response.
        while seq.next_element::<IgnoredAny>()?.is_some() {}
        Ok(ResponseValue::Unexpected)
    }
```

If we visit a `String` we're happy, if we visit a list we discard it, moving the deserializer forward and return a `Unexpected` value. This new value represents something we didn't expect in the response. Not storing anything would be too much work, so using this. If it's a field that will be returned to the client, so not only for a `@requires`, we will add a GraphQL error and propagate null if necessary. So for the client, if we reach this value despite null propagation, we will return `null`. For subgraph requirements, if we reach this value, we consider this to be an error and requirements could not be provided.

From there I added a ton of similar tests to check tons of different possible errors. A few tests are missing, but will do them later. 

This helped me detect a few problems and move all invalid subgraph response error message to traces with the response path. Now we only return a generic error message back the client. They can't do anything about it anyway.


pimeys (approved):
That is a lot of characters in a PR. And a lot of boilerplate, but I get it why we need that. Good amount of tests.

crates/integration-tests/tests/federation/deser/any.rs
```diff
@@ -0,0 +1,310 @@
...
+    let response = run(
+        REQUIRED_ANY_SCHEMA,
+        QUERY,
+        json!({"data": {"user": {"something": null, "valid": "yes"}, "dummy": "yes"}}),
+    );
+    insta::assert_json_snapshot!(response, @r#"
+    {
+      "data": {
+        "user": {
+          "something": null,
```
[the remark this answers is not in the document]
Finistere: oops, that's broken... missed it... thanks

crates/engine/schema/src/lib.rs
```diff
@@ -258,23 +258,25 @@ impl std::fmt::Debug for Schema {
...
-)]
+#[derive(Debug, Clone, Copy, PartialEq, Eq, serde::Serialize, serde::Deserialize)]
 pub enum ScalarType {
     String,
     Float,
     Int,
     BigInt,
-    JSON,
     Boolean,
+    Any,
```
obmarg, line 268: This seems like a potentially confusing name given there's a literal `Any` scalar which has a specific format.

`Unknown` or something might be less ambiguous?

Finistere: what? which `Any` are you talking about?

PR #2428, opened 2024-12-05
fix: main test build used cargo-component and failed

pimeys:


obmarg (approved):

obmarg:
Have been working on unifying main & branch CI builds in between other things btw - so hopefully soon this should be less of a problem

PR #2429, opened 2024-12-05
feat: simpler entity resolvers in dynamic mocks

obmarg:
I kinda had something like this in mind when I was writing the dynamic mocks, but I didn't need them so didn't want to spend the time.

Since #2427 has started using entity resolvers now seems as good a time as any to revisit.

### DIFF
--- a/crates/graphql-mocks/src/dynamic/entity_resolvers.rs
+++ b/crates/graphql-mocks/src/dynamic/entity_resolvers.rs
@@ -1,0 +1,62 @@
+use async_graphql::{dynamic::ResolverContext, ServerError};
+
+pub struct EntityResolverContext<'a> {
+    pub inner_context: &'a ResolverContext<'a>,
+    pub typename: String,
+    pub representation: serde_json::Map<String, serde_json::Value>,
+}
+
+impl<'a> EntityResolverContext<'a> {
+    pub(super) fn new(inner_context: &'a ResolverContext<'a>, representation: serde_json::Value) -> Self {
+        let serde_json::Value::Object(representation) = representation else {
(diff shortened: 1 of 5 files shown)


crates/graphql-mocks/src/dynamic/entity_resolvers.rs
```diff
@@ -0,0 +1,62 @@
...
+            repr,
+        }
+    }
+
+    pub fn add_error(&self, error: ServerError) {
+        self.inner_context.query_env.errors.lock().unwrap().push(error);
+    }
+
+    /// Returns the key of the entity we need to resolve
+    pub fn key(&self) -> serde_json::Value {
```
Finistere, line 25: Would name it `representation`, it also includes fields from `@requires` not just from `@key`

Finistere (approved):

crates/graphql-mocks/src/dynamic/entity_resolvers.rs
obmarg: hmm, fair point.  Although the representation usually also includes the __typename....

obmarg: Tempted to just not offer this as a convenience

Finistere: ah didn't realize it, could just remove it yeah

obmarg: yeah, just going to bin this.  Could implement something that looks at the actual keys defined for the entity, but not worth it unless we need it.

PR #2431, opened 2024-12-05
chore(engine): Rename SafeResponseKey to ResponseKey

Finistere:


crates/engine/src/response/key.rs
```diff
@@ -0,0 +1,126 @@
...
+impl ResponseKey {
+    pub(crate) fn with_position(self, query_position: QueryPosition) -> PositionedResponseKey {
+        PositionedResponseKey {
+            query_position: Some(query_position),
+            response_key: self,
+        }
+    }
+}
+
+/// A ResponseKey is guaranteed to exist inside ResponseKeys
```
Finistere, line 40: Moved code, same logic

crates/engine/src/response/key.rs
```diff
@@ -0,0 +1,126 @@
+use std::cmp::Ordering;
```
Finistere, line 1: Merged two files together.

tomhoule (approved):

PR #2432, opened 2024-12-06
Configure unused-crate-dependencies lint back to default (allow)

tomhoule:
As discussed. The cost to benefit ratio isn't great, and there are false negatives.


pimeys (approved):
yass

PR #2434, opened 2024-12-06
chore: adjust the dev start output, make its log style follow the style of gateway

pimeys:
image

The logs were not in the same style as the gateway logs, and looked out of place when the dev logs were mixed with gateway logs.

New logs:

image

Old logs:

image


crates/grafbase-local-backend/src/dev/configurations.rs
```diff
@@ -82,6 +82,8 @@ pub async fn get_and_merge_configurations(
         .map(|config| config.subgraphs.into_keys().collect::<HashSet<_>>())
         .unwrap_or_default();
 
+    merged_configuration.graph.introspection = true;
```
tomhoule, line 85: Since that might be confusing, could we emit a warning saying we force enable introspection in cases where introspection was explicitly disabled in the configuration?

tomhoule (approved):

crates/grafbase-local-backend/src/dev/configurations.rs
pimeys: Yep.

cli/Makefile.toml
```diff
@@ -102,8 +102,9 @@ mac_alias = "mac-profile-dev"
 [tasks.fetch-assets]
 description = "Fetch the Pathfinder tar.gz"
 script = '''
+CACHE_BREAKER=$(cat /dev/urandom | tr -dc 'a-zA-Z0-9' | fold -w 10 | head -n 1)
```
tomhoule, line 105: Will that work on MacOS? I don't think they have /dev/urandom

pimeys: Yes https://en.wikipedia.org/wiki//dev/random#macOS,_iOS_and_other_Apple_OSes

PR #2436, opened 2024-12-06
chore(engine): Use ResponseKey to identify requirements

Finistere:
Today I rely on an id generated during schema generation as a unique identifier for requirements, so it's tied to the `FieldDefinitionId` which doesn't work well if we want to replace object fields with interface ones or the opposite. It's unnecessary as I can rely on the `ResponseKey`, but made things simpler. Now I don't support multiple `@authorized` post-execution directives on the same field and batching is less good. But well, we'll see how to improve that later.

I've removed the initial alias support, I can't add it for "free" anymore and we have no use for it today.

### DIFF
--- a/crates/engine/src/response/read/mod.rs
+++ b/crates/engine/src/response/read/mod.rs
@@ -1,10 +1,12 @@
 use std::sync::Arc;
 
+use crate::operation::RequiredFieldSet;
+
 use super::{InputResponseObjectSet, ResponseBuilder};
 mod ser;
 mod view;
 
-use schema::{FieldSetRecord, Schema};
+use schema::Schema;
 pub(crate) use view::*;
 
 impl ResponseBuilder {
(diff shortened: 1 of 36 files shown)


crates/engine/src/response/read/mod.rs
```diff
@@ -13,10 +15,11 @@ impl ResponseBuilder {
         &'a self,
         schema: &'a Schema,
         response_object_set: Arc<InputResponseObjectSet>,
-        selection_set: &'a FieldSetRecord,
+        selection_set: RequiredFieldSet<'a>,
     ) -> ResponseObjectsView<'a> {
+        tracing::warn!("{selection_set:#?}");
```
Finistere, line 20: removed in next PR

tomhoule (approved):
As usual, can't say I get every detail, but the big picture looks good :+1:

PR #2438, opened 2024-12-07
FederatedGraph: walk back TypeDefinition{Record,Id}

tomhoule:
There is a halfway done refactoring from Enum, Scalar, Object, Interface, etc. to a unified TypeDefinition in the graphql-federated-graph crate. The idea was to share the common fields between all type definitions (name, description, directives) in a single uniform struct, and store them all together. Many functions could then work on type definitions directly without duplicating logic bbtween different kinds of definitions, while still allowing specialized handling for specific kinds of definitions. The expectation was that it would simplify logic across graphql-composition and graphql-federated-graph, and make graphql-federated-graph more readable by reducing the amount of types in the public API and the amount of fields in FederatedGraph. It would also make the `Definition` enum unnecessary, since you could use a `TypeDefinitionId` instead.

After discussions with @Finistere, some experimentation and a lot of thinking, I don't think this is going to pan out. A few reasons:

- When looking for a specific object, interface, etc., it's better to search in a shorter Vec containing only the kinds of definitions we are interested in.
- There are many places where we do want type definition specific logic.
- More precise ID types as a default are a better idea
- Separate type definition collections better match the model of `engine_schema`.
- The code size savings are not that great.


crates/engine/config/src/lib.rs
```diff
@@ -190,7 +190,7 @@ where
         type Value = FederatedGraph;
 
         fn visit_str<E: serde::de::Error>(self, v: &str) -> Result<Self::Value, E> {
-            federated_graph::from_sdl(v).map_err(E::custom)
+            FederatedGraph::from_sdl(v).map_err(E::custom)
```
tomhoule, line 193: I just want to not have `FederatedGraph::from_sdl()` and `from_sdl()` as a free function in the public API. Not sure picking this one is the right call from an ergonomics perspective.

Finistere: I prefer `FederatedGraph::from_sdl(v)` because I'll have that type imported anyway already.

Finistere (approved):

PR #2441, opened 2024-12-09
fix(engine): Replace object field by parent interface field.

Finistere:
With a schema like:
```graphql
type Query {
  node: Node! @join__field(graph: B)
}

interface Node {
  id: ID
}

type User implements Node @join__type(graph: A) {
  id: ID
  name: String
}
```

and a query:
```graphql
{
  node {
    ... on User { name }
  }
}
```

We need to retrieve the `User.id` but the `B` graph cannot provide it, instead it can provide `Node.id` which represents the same underlying data. So adding the necessary code to add it. It's not perfect and will improve it a bit later, but for now the new problem is that we don't track in the composition that fields we add to an object from entity interfaces are only resolvable through the interface

### DIFF
--- a/crates/engine/query-solver/src/lib.rs
+++ b/crates/engine/query-solver/src/lib.rs
@@ -11,7 +11,9 @@ pub(crate) mod solve;
 pub use error::*;
 pub(crate) use operation::*;
 pub use petgraph;
-use schema::{CompositeTypeId, FieldDefinitionId, ObjectDefinitionId, Schema, SchemaField, SubgraphId};
+use schema::{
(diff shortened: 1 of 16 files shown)


tomhoule:
> but for now the new problem is that we don't track in the composition that fields we add to an object from entity interfaces are only resolvable through the interface

how is that represented in SDL?

Finistere:
```graphql
type Regular implements Account
  @join__implements(graph: A, interface: "Account")
  @join__type(graph: A, key: "id")
{
  id: ID!
  isMain: Boolean!
  name: String! @join__field
  isActive: Boolean! @join__field
}
```

`@join__field` without any args

tomhoule:
interesting, never noticed that

tomhoule (approved):
:100:

PR #2443, opened 2024-12-09
fix(engine): better handling of entity interface fields

Finistere:
We weren't handling entity interface fields well. They must have `@join__field` without any arguments to indicate that they're not resolvable through the object.  So had to change composition & the schema to take this into account. The latter assumed that a field was always resolvable by some subgraph typically. Not greatest code, but works.


pimeys (approved):

crates/graphql-composition/tests/composition/entity_interfaces_with_requires/federated.graphql
```diff
@@ -32,7 +32,7 @@ type Book implements Media
     @join__implements(graph: A, interface: "Media")
 {
     id: ID!
-    reviews: [Review!]! @join__field(graph: B, requires: "title")
+    reviews: [Review!]! @join__field
```
tomhoule, line 35: Is it expected here that the requires is gone?

tomhoule (approved):

crates/graphql-composition/tests/composition/entity_interfaces_with_requires/federated.graphql
Finistere: Yes, because it's not `Book.reviews` that has a requirement, but `Media.reviews`.

PR #2453, opened 2024-12-10
fix(engine): Replace interface fields by object fields.

Finistere:
Replace an interface field by all of its implementor's fields.
Hacky at the plan/bound operation layer, but well will improve that later.

The goal roughly is that we have something like this:
```graphql
type Query {
  node: Node @join__type(graph: B)
}

interface Node @join__type(graph: A) @join__type(graph: B) {
  id: ID!
  name: String! @join__field(graph: A)
}

type A implements Node @join__type(graph: A, key: "id") @join__type(graph: B, key: "id") {
  id: ID!
  name: String!
}
```

and a request like:

```
{
  node {
    name
  }
}
```

we actually need to execute:

```
{
  node {
   ... on A { name }
  }
}
```

for all implementors as `name` isn't accessible from the subgraph B, but is for each individual implementor.

To solve this I re-use a similar logic as before, trying to plan alternative fields. This time however I need to clone the whole sub-selection of the field also as it might be planned differently for each field and need to delete old ones. I also need now to plan alternative fields in the right order as nested fields obviously can't be planned until their parent one is. There are then a few hacks to make it work with the current operation/query-solver split.


pimeys (approved):
🚀

PR #2455, opened 2024-12-11
chore: remove OperationCacheFactory trait

obmarg:
To do query warming I need to reach into the operation cache to fetch queries to prewarm.  I wanted to do this in the runtime, but currently the runtime doesn't have access to the operation cache, only an OperationCacheFactory.  

I can't see much reason for this factory.  Things seem to work fine without it, so I've removed it.  If I've missed something please let me know and I'll revert.

### DIFF
--- a/crates/engine/src/engine.rs
+++ b/crates/engine/src/engine.rs
@@ -1,4 +1,4 @@
-use ::runtime::{hooks::Hooks as _, operation_cache::OperationCacheFactory};
+use ::runtime::hooks::Hooks as _;
 use bytes::Bytes;
 use engine_auth::AuthService;
 use futures::{StreamExt, TryFutureExt};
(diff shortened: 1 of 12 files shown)


crates/engine/src/engine.rs
```diff
@@ -47,7 +47,7 @@ impl<R: Runtime> Engine<R> {
         Self {
             auth,
             retry_budgets: RetryBudgets::build(&schema),
-            operation_cache: runtime.operation_cache_factory().create().await,
+            // operation_cache: runtime.operation_cache_factory().create().await,
             schema,
```
obmarg, line 51:
```suggestion
            schema,
```

crates/engine/src/prepare/context.rs
```diff
@@ -67,6 +68,10 @@ impl<'ctx, R: Runtime> PrepareContext<'ctx, R> {
         self.engine.runtime.metrics()
     }
 
+    pub fn operation_cache(&self) -> &'ctx impl OperationCache<Arc<CachedOperation>> {
```
Finistere, line 71: Just a nitpick, could have been `&'ctx R::OperationCache` instead

Finistere (approved):

PR #2457, opened 2024-12-11
chore: rework engine & config updates in gateway

obmarg:
I need to hook into the engine creation process in the gateway to add query warming.  Prior to this change that happened in three places: inside the GraphUpdater when using a graph ref to fetch from the GDN & a couple of places inside GraphFetchMethod::start when using a schema file.  This made it a bit awkward to add in query warming.

This refactors things a bit so that GraphUpdater & GraphFetchMethod::start both just send a `GraphDefinition` on a channel, where a new `EngineReloader` (name suggestions welcome) is listening and will handle the update.  There's also some logic to cancel a running reload if another one comes in afterwards.

I also separated out config updates from graph updates in the `reload_signal` used by federated dev - config updates now go over their own `tokio::sync::watch` channel, rather than sometimes being passed as an argument and sometimes being passed through a channel along with a GraphDef.

Finally, I changed the engine watcher to not contain an `Option`.  As far as I could tell the case where this is `None` could only be ever be observed by the websocket handler - we only ever set it to `Some(engine)` and we have a `watcher.changed().await` before we start up the main server.  So it seems like a reasonable simplification.

### DIFF
--- a/crates/federated-server/src/server/engine_reloader.rs
+++ b/crates/federated-server/src/server/engine_reloader.rs
@@ -7,18 +7,17 @@ use tokio::{
     sync::{mpsc, watch},
     task::JoinHandle,
 };
-use tokio_stream::{
-    wrappers::{ReceiverStream, WatchStream},
-    Stream,
-};
+use tokio_stream::{wrappers::WatchStream, Stream};
 
 use crate::server::gateway;
 
-use super::gateway::{EngineSender, EngineWatcher, GatewayRuntime, GraphDefinition};
+use super::{
+    gateway::{EngineSender, EngineWatcher, GatewayRuntime, GraphDefinition},
+    graph_fetch_method::GraphStream,
+};
 
 /// Handles graph and config updates by constructing a new engine
 pub(super) struct EngineReloader {
-    graph_sender: GraphSender,
     engine_watcher: EngineWatcher,
 }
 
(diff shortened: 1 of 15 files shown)


obmarg:
Note to self: need to test that I haven't broken the dev command

crates/federated-server/src/server/engine_reloader.rs
```diff
@@ -0,0 +1,140 @@
...
+    context: Context,
+) {
+    let mut update_handle: Option<JoinHandle<()>> = None;
+    let mut graph_definition = None;
+
+    pin!(updates);
+
+    while let Some(update) = updates.next().await {
+        if let Some(update_handle) = update_handle.take() {
+            update_handle.abort();
```
obmarg, line 97: Update handle could be named better

crates/federated-server/src/server/engine_reloader.rs
```diff
@@ -0,0 +1,140 @@
...
+    engine_sender: EngineSender,
+}
+
+async fn update_loop(
+    updates: impl Stream<Item = Update>,
+    mut current_config: gateway_config::Config,
+    context: Context,
+) {
+    let mut update_handle: Option<JoinHandle<()>> = None;
+    let mut graph_definition = None;
```
obmarg, line 91: Should pass graph_definition in and avoid this Option

gateway/src/main.rs
```diff
@@ -52,3 +52,12 @@ fn main() -> anyhow::Result<()> {
         result
     })
 }
+
+fn config_channel(config: gateway_config::Config) -> watch::Receiver<gateway_config::Config> {
+    let (sender, receiver) = watch::channel(config);
+
+    // Drop the sender so the channel never closes
+    std::mem::drop(sender);
```
obmarg, line 60: This should be leak, not drop 🤦‍♂️

crates/federated-server/src/server/engine_reloader.rs
```diff
@@ -0,0 +1,143 @@
...
+enum Update {
+    Graph(GraphDefinition),
+    Config(gateway_config::Config),
+}
+
+impl EngineReloader {
+    pub async fn spawn(
+        gateway_config: watch::Receiver<gateway_config::Config>,
+        mut graph_stream: GraphStream,
+        hot_reload_config_path: Option<PathBuf>,
```
Finistere, line 35: A bit confused why a `hot_reload_config_path` needs to be propagated down when `EngineReloader` seems like the struct that should handle all reloads to be honest. But seems like it was like this before so well. 🤷

crates/federated-server/src/server/graph_fetch_method.rs
```diff
@@ -21,81 +15,60 @@ pub enum GraphFetchMethod {
...
-                            let gateway = gateway::generate(
-                                GraphDefinition::Sdl(sdl),
-                                &config,
-                                hot_reload_config_path.clone(),
-                                hooks.clone(),
-                            )
-                            .await?;
-                            sender.send(Some(Arc::new(gateway)))?;
+                // Leak the sender so the channel never closes
+                Box::leak(Box::new(sender));
```
Finistere, line 59: Same nitpick, if you leak would say that this function is only called once

Finistere (approved):

crates/federated-server/src/server/engine_reloader.rs
obmarg: Yeah, weird isn't it.  Essentially there are two hot reloading mechanisms at the moment: 

1. The one used in the gateway to hot reload rate limiting stuff.  That's what this `hot_reload_config_path` is used for.
2. The one used by `grafbase dev` which is powered by the `watch::Receiver` here.

Seems like it should be cleaned up sometime.  But I've already spent a while on this PR and want to get onto other things, so a problem for the future.

Finistere: might be worth adding a quick comment here IMHO with exactly what you said.
thanks for the explanation!

crates/federated-server/src/server/graph_fetch_method.rs
obmarg: Yeah, good point.  Although I wonder if I'd actually be better off doing this without leaking, just incase someone decides to call this multiple times...

PR #2463, opened 2024-12-12
feat: in memory prewarming on GDN updates

obmarg:
The initial implementation of query warming: this will warm the entire in-memory cache on schema or config updates in federated dev & the gateway, if warming is enabled.  A future PR will add more configuration to this, at the moment it's just on/off.

One thing that may be of concern with this feature: a lot of the query binding etc. code assumes it's occurring within the context of a request.  But warming is never occurring in a request.  I've just set all the associated values to null, but this seems like it could have some consequences.

Fixes GB-8098

### DIFF
--- a/crates/engine/src/engine/cache/warming.rs
+++ b/crates/engine/src/engine/cache/warming.rs
@@ -2,40 +2,41 @@ use std::sync::Arc;
 
 use runtime::operation_cache::OperationCache;
 
-use crate::{prepare::PrepareContext, CachedOperation};
+use super::{DocumentKey, Key};
 
-use super::{
-    cache::{DocumentKey, Key},
-    EarlyHttpContext, Engine, Runtime,
+use crate::{
+    engine::{EarlyHttpContext, Engine, ResponseFormat, Runtime},
+    prepare::PrepareContext,
+    CachedOperation,
 };
 
-pub struct PrewarmOperation {
+pub struct OperationForWarming {
     document: String,
     operation_name: Option<String>,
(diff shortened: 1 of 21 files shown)


crates/engine/src/engine/cache/warming.rs
```diff
@@ -0,0 +1,89 @@
...
+                }
+                Err(err) => {
+                    tracing::warn!(
+                        "Could not plan operation {}: {err}",
+                        operation.operation_name.unwrap_or_default()
+                    );
+                }
+            }
+
+            futures_lite::future::yield_now().await
```
Finistere, line 74: nice

crates/engine/src/engine/cache/warming.rs
```diff
@@ -0,0 +1,89 @@
...
+
+use super::{DocumentKey, Key};
+
+use crate::{
+    engine::{EarlyHttpContext, Engine, ResponseFormat, Runtime},
+    prepare::PrepareContext,
+    CachedOperation,
+};
+
+pub struct OperationForWarming {
```
Finistere, line 13: Would just have named this something like `RawOperation` and included the struct directly into the `CachedOperation`. You're adding exactly the same fields on both side as far as I've understood

Finistere (approved):

crates/engine/src/engine/cache/warming.rs
obmarg: Not a bad suggestion, but I don't want to wait on CI again.  Will do in a future PR if I remember.

PR #2464, opened 2024-12-12
feat: allow specific port in graphql-mocks

obmarg:
`graphql-mocks` was originally written for use in our tests.  As a result it always just picks a random port to run on.  But I.'ve been using it for building demo graphs recently where this property is quite annoying.

This is a quick refactor that adds a builder API to allow you to specify your own port.  It's a bit messy, but this crate in general is quite messy.  Should maybe tidy it up sometime.


tomhoule (approved):
nice!

PR #2465, opened 2024-12-12
fix: add a default otel `service.name`

obmarg:
I've been using traces to test things locally lately by feeding things into jaeger.  But if I forget to add a service name into my config my traces appear in jaeger under `empty-service-name`.  This is because we default to an empty string for this config value.

It seems better to have an actual default for this: I propose `grafbase-gateway` but I'm open to other suggestions.  I'm assuming this won't have an impact on our hosted otel stuff, but if someone more familiar with it could confirm that'd be good.

### DIFF
--- a/crates/gateway-config/src/telemetry/mod.rs
+++ b/crates/gateway-config/src/telemetry/mod.rs
@@ -19,7 +19,7 @@ pub use exporters::{BatchExportConfig, OpenTelemetryExportersConfig, StdoutExpor
 #[serde(default, deny_unknown_fields)]
 pub struct TelemetryConfig {
     /// The name of the service
-    pub service_name: String,
+    pub service_name: Option<String>,
     /// Additional resource attributes
     pub resource_attributes: HashMap<String, String>,
     /// Global exporters config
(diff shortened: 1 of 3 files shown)


crates/gateway-config/src/telemetry/mod.rs
```diff
@@ -19,7 +19,7 @@ pub use exporters::{BatchExportConfig, OpenTelemetryExportersConfig, StdoutExpor
 #[serde(default, deny_unknown_fields)]
 pub struct TelemetryConfig {
     /// The name of the service
-    pub service_name: String,
+    pub service_name: Option<String>,
```
Finistere, line 22: Would add the default here. I find it way too deep otherwise, risking inconsistencies between traces/metrics/etc. Just need to implement `Default` for TelemetryConfig as we have `#[serde(default)]`

Finistere (approved):

crates/gateway-config/src/telemetry/mod.rs
obmarg: I don't want to put that much effort in.

Finistere: I mean the times, it takes to have this whole discussion and writing the unwrap, you would have written the `Default` :D  but anyway, approved

Finistere: fair enough :D

Finistere: Thank you!

PR #2474, opened 2024-12-17
chore: more traces for caching 

obmarg:
As I mentioned in my demo last week, it's not super obvious from traces whether or not you hit the operation cache.  This is because it all happens under the "prepare operation" span.  We could put a `is_cached` property on this span, but I think there's an opportunity to show users how they could/couldn't benefit from better caching.  This PR is an attempt to do that:

1. Added a span on `build_cached_operation` - which builds the operation that will be cached.  I named it build operation because `build_cached_operation` seems a little ambiguous to me  - very much sounds like it could be building an operation that _was_ cached.
2. Added some spans on fetching a trusted document, since that also seemed like a faintly sensible thing.
3. Moved the spans I'd added on bind/parse/plan etc. to debug spans, so we can turn them on if we want (I think they're interesting) but users don't see them by default.

Fixes GB-8193

### DIFF
--- a/crates/engine/src/prepare/trusted_documents.rs
+++ b/crates/engine/src/prepare/trusted_documents.rs
@@ -10,6 +10,7 @@ use futures::{future::BoxFuture, FutureExt};
 use grafbase_telemetry::grafbase_client::X_GRAFBASE_CLIENT_NAME;
 use runtime::trusted_documents_client::TrustedDocumentsError;
 use std::borrow::Cow;
+use tracing::Instrument;
 
 use super::PrepareContext;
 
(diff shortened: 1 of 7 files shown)


crates/engine/src/prepare/trusted_documents.rs
```diff
@@ -10,14 +10,36 @@ use futures::{future::BoxFuture, FutureExt};
...
+        }
+    }
+}
+
+pub(super) enum DocumentOrFuture<'a> {
+    Document(Cow<'a, str>),
+    Future(DocumentFuture<'a>),
+}
+
+fn wrap_document(doc: &str) -> DocumentOrFuture<'_> {
```
Finistere, line 41: Could have been a `From`

Finistere (approved):

crates/engine/src/prepare/trusted_documents.rs
obmarg: I don't think that would have been better

PR #2475, opened 2024-12-17
feat: partial cache warming config

obmarg:
Users might not want to warm the whole cache - this adds config for that.

Fixes GB-8191

### DIFF
--- a/crates/federated-server/src/server/engine_reloader.rs
+++ b/crates/federated-server/src/server/engine_reloader.rs
@@ -2,6 +2,7 @@ use std::{path::PathBuf, sync::Arc};
 
 use engine::{Engine, OperationForWarming};
 use futures_lite::{pin, StreamExt};
+use rand::{seq::SliceRandom, thread_rng};
 use runtime_local::HooksWasi;
 use tokio::{
     sync::{mpsc, watch},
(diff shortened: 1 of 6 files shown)


crates/federated-server/src/server/engine_reloader.rs
```diff
@@ -139,24 +134,46 @@ async fn update_loop(
...
+            cache.entry_count(),
+        )
+    };
+
+    if config.operation_caching.warming_percent >= 100 {
+        return operations;
+    }
+
+    // If we're not taking 100% of the list we shuffle the list and take a percentage of it
+    operations.shuffle(&mut thread_rng());
```
Finistere, line 173: No idea whether the current LRU cache makes it possible, but with `warming_percent` I would have expected the _top_ X% to be warmed, not a random subset of the cache. So would change the the documentation of the `gateway-config` to make it more obvious IMHO.

Finistere (approved):

crates/federated-server/src/server/engine_reloader.rs
Finistere: Not even sure to see why you're shuffling at all to be honest.

obmarg: > Not even sure to see why you're shuffling at all to be honest if having a random subset is the goal.

What would the alternative be?

Finistere: Well ok, you do get a higher probably that a query is cached by at least one worker supposing that they receive in the same/similar order. But as load is balanced across your worker they will each still need to replan the missing 90% if you're not using shared cache.

I'm just kinda mitigated of the usefulness of replanning a random % of plans. You can't predict in any way how much of your traffic will stay cached. You might end up with 90% traffic cached or 0.1% if your traffic.

PR #2477, opened 2024-12-18
chore: extract what-rust-changed into its own action

obmarg:
I changed the grafbase PR action to do partial builds a while back, using what-rust-changed to decide what we need to build.  This moves some of the logic for that into a resuable action for what-rust-changed.  I initially wrote this because I was planning on using it again somewhere, and intended to avoid repeating the pattern once again.  But in the end I didn't need to do that.  This still seems like a reasonable cleanup to make though.

Depends on https://github.com/grafbase/what-rust-changed/pull/7.


Finistere (approved):

.github/workflows/federation-audit.yml
```diff
@@ -32,36 +32,12 @@ jobs:
...
-          git checkout $HEAD_REF
-          cargo metadata --locked > /tmp/target.metadata.json
-          CHANGED_FILES=$(git diff --no-commit-id --name-only -r $MERGE_BASE HEAD)
-          CHANGES=$(echo $CHANGED_FILES | xargs what-rust-changed /tmp/base.metadata.json /tmp/target.metadata.json)
-          echo "rust=$CHANGES" >> "$GITHUB_OUTPUT"
-          echo "Done.  Output Contents:"
-          echo ""
-          cat $GITHUB_OUTPUT
+        id: rust
+        uses: grafbase/what-rust-changed@obmarg/yosxprrlrxpm
```
obmarg, line 37:
```suggestion
        uses: grafbase/what-rust-changed@v0.2.0
```

PR #2478, opened 2024-12-19
Rework operation processing

Finistere:
Gigantic PR and hopefully the last one I need to do of this size for the engine. Initially, the `query-solver` crate relied on a trait to retrieve operation data (fields, selection sets, etc.):
```rust
pub trait Operation {
    type FieldId: From<usize> + Into<usize> + Copy + std::fmt::Debug + Ord;

    fn root_object_id(&self) -> ObjectDefinitionId;
    fn field_ids(&self) -> impl ExactSizeIterator<Item = Self::FieldId> + 'static;
    fn root_selection_set(&self) -> impl ExactSizeIterator<Item = Self::FieldId> + '_;
    fn subselection(&self, field_id: Self::FieldId) -> impl ExactSizeIterator<Item = Self::FieldId> + '_;
    // (...)
}
```

That was a stupid idea as there are too many interactions and the trait itself was also entirely useless with a half-baked implementation in `query-solver` test and the real deal inside `engine` crate.

So moved all the parsing & binding steps into `engine/operation` and removed any planning-related logic. So now we keep fragments and they're only flattened during the planning. This allowed me to solve a few over-simplifications I did and the engine now treats correctly type conditions.

This change implied reworking a significant portion of the `query-solver` and `engine` code and took me significantly longer than I expected. But now the separation is a lot cleaner and would have made things like complexity control easier.

Fixes: GB-8164, GB-8172, GB-8162


tomhoule (approved):

crates/engine/codegen/domain/operation.graphql
```diff
@@ -0,0 +1,74 @@
...
+    name: "operation"
+    destination: "operation/src/model"
+    root_module: "model"
+    context_name: "ctx"
+    context_type: "OperationContext"
+    imports: [{ module: "schema", domain: "schema" }]
+  )
+
+scalar ResponseKey @copy
+scalar ResponseKey @copy
```
tomhoule, line 12: Duplicated line?

crates/engine/codegen/domain/query_plan.graphql
```diff
@@ -0,0 +1,89 @@
...
+}
+
+scalar PartitionDataField @record @indexed
+scalar PartitionTypenameField @record @indexed
+
+union PartitionField @id @meta(module: "field") @variants(names: ["Data", "Typename"]) =
+  | PartitionDataField
+  | PartitionTypenameField
+
+# "In opposition to a __typename field this field does retrieve data from a subgraph"
```
tomhoule, line 52: Intentionally left in?

crates/engine/codegen/domain/operation.graphql
Finistere: ah yes thx

crates/integration-tests/tests/federation/basic/skip_include.rs
```diff
@@ -48,14 +48,8 @@ fn skip_include() {
...
-          {
-            "author": {}
-          },
-          {
-            "author": {
-              "id": "123"
-            }
-          },
+          {},
+          {},
```
tomhoule, line 52: Interesting, is it an arbitrary choice or is the behaviour of having both skip and include specified?

crates/engine/codegen/domain/query_plan.graphql
Finistere: no leftover, wasn't sure if I would keep it or not here. it just has too many specificities for code generation. thx

pimeys (approved):

crates/integration-tests/tests/federation/basic/skip_include.rs
Finistere: It's the type condition that wasn't tracked correctly, the only thing left after skip/include is:
```
                    fragment IssueFragment on Issue {
                        author {
                            ...AuthorFragment
                        }
                    }
```
and the first two items weren't `Issue`s but `PullRequest`

Finistere: I was merging fields too aggressively which led me to "forget" that type conditions applied

PR #2489, opened 2025-01-06
feat: redis operation caching

obmarg:
Fixes GB-8194

### DIFF
--- a/crates/runtime-local/src/operation_cache/redis.rs
+++ b/crates/runtime-local/src/operation_cache/redis.rs
@@ -1,0 +1,66 @@
+use redis::{AsyncCommands, RedisResult};
+use runtime::operation_cache::OperationCache;
+
+use crate::redis::Pool;
+
+#[derive(Clone)]
+pub struct RedisOperationCache {
+    pool: Pool,
+    key_prefix: String,
+}
+
+impl RedisOperationCache {
+    pub fn new(pool: Pool, key_prefix: &str) -> Self {
+        RedisOperationCache {
+            pool,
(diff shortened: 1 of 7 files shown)


crates/runtime-local/src/operation_cache/tiered.rs
```diff
@@ -0,0 +1,50 @@
...
+            });
+        }
+    }
+
+    async fn get(&self, key: &String) -> Option<V> {
+        if let Some(value) = self.in_memory.get(key).await {
+            return Some(value);
+        };
+
+        self.distributed.as_ref()?.get(key).await
```
pimeys, line 48: Should we store the value to the in-memory cache here, so we don't need to query redis every time we can't find the item from memory?

obmarg: Oh yeah, probably

crates/runtime-local/src/operation_cache/redis.rs
```diff
@@ -0,0 +1,66 @@
...
+        }
+    }
+}
+
+impl<V> OperationCache<V> for RedisOperationCache
+where
+    V: Clone + Send + Sync + 'static + serde::Serialize + serde::de::DeserializeOwned,
+{
+    async fn insert(&self, key: String, value: V) {
+        let Ok(value) = serde_json::to_vec(&value) else { return };
```
Finistere, line 26: You could just use `postcard` here, no need to pay for JSON no?

Finistere (approved):

PR #2500, opened 2025-01-14
engine: implement trusted documents incremental adoption

tomhoule:
The aim of this change is to offer an incremental adoption path for trusted documents.

Until now, trusted documents would be either disabled, or fully enforced. However in practice, organizations follow a more incremental migration path. The following options are now available in `grafbase.toml`:

```toml
[trusted_documents]
enabled = true # default: false
enforced = true # default: false
bypass_header_name = "my-header-name" # default null
bypass_header_value = "my-secret-is-{{ env.SECRET_HEADER_VALUE }}" # default null
document_id_unknown_log_level = "error" # default: info
document_id_and_query_mismatch_log_level = "off" # default: info
inline_document_unknown_log_level = "warn" # default: info
```

The `bypass_header_name` and `bypass_header_value` settings are unchanged.

The `enabled` setting is now much weaker. When trusted documents are only `enabled`, the gateway will not enforce trusted documents, but will still fetch and cache them. This is useful for organizations that want to gradually adopt trusted documents.

The `enforced` setting matches the previous behavior: when `true`, the gateway will enforce trusted documents. Additionally, we now allow trusted document requests with only an inline document, without document id, as long as the query is a trusted document. This is useful for organizations that want to enforce trusted documents, but have not yet migrated all their clients to send the document id.

The three new log options allow monitoring proper usage of trusted documents before you enforce them:

- `document_id_unknown_log_level`: logged when a document id is present in the request, but the corresponding trusted document cannot be found.
- `document_id_and_query_mismatch_log_level`: logged when both a query document and a document id are sent, when the retrieved trusted document does not match the inline query document.
- `inline_document_unknown_log_level`: logged when an inline document (a string in the `"query"` key of a the request) is sent, but no trusted document is found with the id matching the sha256 hash of the inline document.

They can all take `off`, `error`, `warn`, `info`, or `debug` as values. The default is `info`.

closes GB-7799
closes GB-6180

### DIFF
--- a/crates/engine/config/src/trusted_documents.rs
+++ b/crates/engine/config/src/trusted_documents.rs
@@ -1,0 +1,33 @@
+pub use gateway_config::LogLevel;
+
+use serde::{Deserialize, Serialize};
+
+#[derive(Debug, Deserialize, Serialize)]
+pub struct TrustedDocumentsConfig {
+    pub document_id_unknown_log_level: LogLevel,
+    pub document_id_and_query_mismatch_log_level: LogLevel,
+    pub inline_document_unknown_log_level: LogLevel,
+}
+
+impl Default for TrustedDocumentsConfig {
+    fn default() -> Self {
+        gateway_config::TrustedDocumentsConfig::default().into()
+    }
+}
+
+impl From<gateway_config::TrustedDocumentsConfig> for TrustedDocumentsConfig {
+    fn from(
+        gateway_config::TrustedDocumentsConfig {
+            document_id_unknown_log_level,
+            document_id_and_query_mismatch_log_level,
+            inline_document_unknown_log_level,
+            ..
+        }: gateway_config::TrustedDocumentsConfig,
+    ) -> Self {
+        TrustedDocumentsConfig {
+            document_id_unknown_log_level,
+            document_id_and_query_mismatch_log_level,
+            inline_document_unknown_log_level,
+        }
+    }
+}
--- a/crates/engine/src/prepare/trusted_documents.rs
+++ b/crates/engine/src/prepare/trusted_documents.rs
@@ -9,12 +9,15 @@ use config::LogLevel;
 use futures::{future::BoxFuture, FutureExt, TryFutureExt as _};
 use grafbase_telemetry::grafbase_client::X_GRAFBASE_CLIENT_NAME;
 use operation::{extensions::PersistedQueryRequestExtension, Request};
-use runtime::trusted_documents_client::{TrustedDocumentsEnforcementMode, TrustedDocumentsError};
+use runtime::{
+    operation_cache::OperationCache as _,
+    trusted_documents_client::{TrustedDocumentsEnforcementMode, TrustedDocumentsError},
+};
 use sha2::Digest;
 use std::borrow::Cow;
 use tracing::Instrument;
 
-use super::{OperationDocument, PrepareContext};
+use super::{CacheKey, OperationDocument, PrepareContext};
 
 pub(super) struct ExtractedOperationDocument<'a> {
     pub key: DocumentKey<'a>,
@@ -50,7 +53,7 @@ fn wrap_document(doc: &str) -> DocumentOrFuture<'_> {
 impl<'ctx, R: Runtime> PrepareContext<'ctx, R> {
     /// Determines what document should be used for the request and provides an appropriate cache
     /// key for the operation cache and as a fallback a future to load said document.
-    pub(super) fn extract_operation_document<'r, 'f>(
+    pub(super) async fn extract_operation_document<'r, 'f>(
         &mut self,
         request: &'r Request,
     ) -> Result<ExtractedOperationDocument<'f>, GraphqlError>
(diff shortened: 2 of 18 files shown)


gateway/changelog/unreleased.md
```diff
@@ -0,0 +1,30 @@
+# Unreleased
```
tomhoule, line 1: I'm allowing myself to introduce this file, so we can write bits of release notes when we still have the context in mind, as opposed to rushing it at release time.

crates/engine/config/src/trusted_documents.rs
```diff
@@ -0,0 +1,33 @@
...
+impl Default for TrustedDocumentsConfig {
+    fn default() -> Self {
+        gateway_config::TrustedDocumentsConfig::default().into()
+    }
+}
+
+impl From<gateway_config::TrustedDocumentsConfig> for TrustedDocumentsConfig {
+    fn from(
+        gateway_config::TrustedDocumentsConfig {
+            unknown_trusted_document_id_log_level,
```
fbjork, line 21: perhaps `unknown_id_log_level` to be shorter?

crates/engine/config/src/trusted_documents.rs
```diff
@@ -0,0 +1,33 @@
...
+            document_id_and_query_mismatch_log_level,
+            unknown_inline_document_log_level,
```
fbjork, line 23: `inline_document_unknown_log_level` perhaps?

crates/gateway-config/src/lib.rs
```diff
@@ -784,15 +770,19 @@ mod tests {
 
         let config = toml::from_str::<Config>(input).unwrap();
 
-        insta::assert_debug_snapshot!(config.trusted_documents, @r###"
+        insta::assert_debug_snapshot!(config.trusted_documents, @r#"
         TrustedDocumentsConfig {
             enabled: true,
+            enforce_trusted_documents: false,
```
fbjork, line 776: perhaps just `enforce`?

crates/engine/config/src/trusted_documents.rs
tomhoule: done

crates/engine/config/src/trusted_documents.rs
tomhoule: done

crates/gateway-config/src/lib.rs
tomhoule: done

crates/engine/src/prepare/trusted_documents.rs
```diff
@@ -79,26 +82,57 @@ impl<'ctx, R: Runtime> PrepareContext<'ctx, R> {
...
                         },
-                        document_or_future: wrap_document(document),
+                        content: wrap_document(document),
+                    })
+                } else if let Some(query) = request.query.as_deref().filter(|query| !query.is_empty()) {
+                    // Just an inline document.
+                    let Some(client_name) = client_name else {
+                        return Err(missing_client_name_error());
+                    };
+                    let hash = sha2::Sha256::digest(query.as_bytes());
```
Finistere, line 92: why sha256 rather than blake3? Do we want sha256 specifically or a cryptographic hash?
I guess the `dbg!` underneath can be removed?

Finistere: Ah we assume the doc_id is a hex(sha256(query))

tomhoule: It's because that's what apollo persisted queries use. We have to take a guess on the document id, basically, in this branch. It's best effort.

tomhoule: The dbg should go, obviously, good catch.

crates/engine/src/prepare/trusted_documents.rs
```diff
@@ -215,3 +357,47 @@ async fn handle_apq<'r, 'f>(
         ErrorCode::PersistedQueryNotFound,
     ))
 }
+
+/// When a request contains a trusted document id, but the trusted document is not found in GDN. Default: INFO.
+fn log_unknown_trusted_document_id<R: Runtime>(engine: &Engine<R>, document_id: &str) {
+    const MESSAGE: &str = "Unknown trusted document";
+
+    match engine.schema.settings.trusted_documents.document_id_unknown_log_level {
```
Finistere, line 365: you could have a `Into<tracing::Level>` and use `tracing::event!(level, "Unknown trusted document", document_Id)` instead

crates/engine/src/prepare/trusted_documents.rs
```diff
@@ -115,13 +149,81 @@ impl<'ctx, R: Runtime> PrepareContext<'ctx, R> {
...
+                                document: Cow::Borrowed(query),
+                            }
+                        })
+                        .unwrap_or_else(|| DocumentKey::TrustedDocumentId {
+                            operation_name,
+                            client_name: Cow::Borrowed(client_name),
+                            doc_id: doc_id.clone(),
+                        }),
+                    content: DocumentOrFuture::Future(
+                        handle_trusted_document_query_and_check_that_query_matches(
```
Finistere, line 214: The check against the query shouldn't be hidden. The goal of the future is to prevent fetching a trusted document if we have the query in cache already. But here you're doing something different. If both the query and document id are provided, shouldn't you always check if they're consistent?

Finistere: or do we only want to check it if not in operation cache?

crates/engine/src/prepare/trusted_documents.rs
tomhoule: I tried that first. The level has to be a compile-time constant, so that doesn't work.

crates/engine/src/prepare/trusted_documents.rs
```diff
@@ -5,18 +5,20 @@ use crate::{
...
+use sha2::Digest;
 use std::borrow::Cow;
 use tracing::Instrument;
 
 use super::{OperationDocument, PrepareContext};
 
 pub(super) struct ExtractedOperationDocument<'a> {
     pub key: DocumentKey<'a>,
-    document_or_future: DocumentOrFuture<'a>,
+    content: DocumentOrFuture<'a>,
```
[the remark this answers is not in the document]
tomhoule: Hm, that's an artifact from the rebase after the big refactoring PR I think, will return the name to document_or_future / text_or_future, whatever works best.

crates/engine/src/prepare/trusted_documents.rs
Finistere: hum indeed, sad

Finistere (approved):

Finistere:
nice work

crates/engine/src/prepare/trusted_documents.rs
```diff
@@ -278,22 +281,22 @@ async fn handle_trusted_document_query_and_check_that_query_matches<'ctx, 'r, R:
     document_id: Cow<'r, str>,
     inline_document: Option<&'r str>,
 ) -> Result<Cow<'r, str>, GraphqlError> {
-    let fetch_trusted_doc_result = handle_trusted_document_query(engine, client_name, document_id.clone()).await;
+    let document_id: String = document_id.into_owned();
+
+    let trusted_doc = match engine.runtime.operation_cache().get(&document_id).await {
```
Finistere, line 286: It's not the right cache key, we should have better types, this is the right one:
```
let cache_key = CacheKey::document(self.schema(), &DocumentKey::TrustedDocumentId {
    operation_name,
    client_name,
    doc_id,
});
```

tomhoule: Good catch, just pushed that.

PR #2508, opened 2025-01-17
gateway: pass websocket connection_init payload to subgraphs

tomhoule:
The gateway implements the [graphql-transport-ws protocol](https://github.com/graphql/graphql-over-http/blob/main/rfcs/GraphQLOverWebSocket.md). In that protocol, the first message on a websocket connection will be `connection_init`. That message contains an optional `payload` field that is a JSON object with arbitrary values. Typically, information that would go in HTTP headers for queries and mutations flows instead in that `connection_init` message's payload. This is because browsers can't send custom headers with the websocket HTTP upgrade request. So headers like `Authorization` are passed there instead.

Until this PR, the gateway expected a payload of the shape:

```ts
type ConnectionInit = {
  type: 'connection_init',
  payload: { headers: Map<string, string> }
}
```

It would then handle the contents of the `headers` key like headers, passing them through subgraph header rules, extracting JWTs from Authorization if so configured, etc. Eventually, these headers did land in the connection_init payload of requests made to subgraph, at the top level.

This is limiting, because both from the perspective of the client and the subgraphs, the payload is an arbitrary JSON object. We want subscriptions that work from a given client to a given subgraph to also work from the same client to the same subgraph with the Grafbase gateway in the middle.

This PR makes the gateway pass on the connection_init payload verbatim by default, with an option to disable it under:

```toml
[websockets]
forward_connection_init_payload = false
```

The gateway still attempts to extract headers for itself from the root of the connection_init payload and from its `headers` key if present.

This PR also contains new infrastructure to test websocket subscriptions in integration-tests, since the existing websocket tests were deleted with the CLI test suite.

closes GB-8244
closes GB-8262


crates/integration-tests/src/federation/builder/router.rs
```diff
@@ -2,12 +2,19 @@ use std::sync::Arc;
...
 use engine::Engine;
+use engine_axum::websocket::{WebsocketAccepter, WebsocketService};
 
 use super::TestRuntime;
 
 pub(super) fn build(engine: Arc<Engine<TestRuntime>>) -> Router {
+    let (websocket_sender, websocket_receiver) = tokio::sync::mpsc::channel(16);
+    let (_, engine_watcher) = tokio::sync::watch::channel(engine.clone());
+    let websocket_accepter = WebsocketAccepter::new(websocket_receiver, engine_watcher);
+    tokio::spawn(websocket_accepter.handler());
```
tomhoule, line 13: should add a FIXME here too to shutdown the task at the end of the test

crates/integration-tests/src/federation/websocket_request.rs
```diff
@@ -0,0 +1,69 @@
...
+
+        Box::pin(async move {
+            let handler = self.router.into_make_service();
+            let listener = tokio::net::TcpListener::bind("127.0.0.1:0").await.unwrap();
+
+            let mut url: Url = format!("http://{}", listener.local_addr().unwrap()).parse().unwrap();
+            url.set_path("/ws");
+            url.set_scheme("ws").unwrap();
+
+            // FIXME: we should shut down the server when the stream is dropped.
```
tomhoule, line 32: Haven't spent time on this, but should do before merging. Current idea would be a wrapper for the stream that contains a channel, then select on the channel in the spawned task. That way when the stream is dropped, the task exits.

crates/engine/src/websocket.rs
```diff
@@ -39,24 +36,34 @@ pub struct SubscribeEvent {
...
 
 #[derive(Debug, Default, serde::Deserialize)]
-pub struct InitPayload {
-    #[serde(default, deserialize_with = "deserialize_as_hash_map")]
-    pub headers: http::HeaderMap,
-}
+pub struct InitPayload(pub(crate) serde_json::Map<String, serde_json::Value>);
+
+impl InitPayload {
+    pub(crate) fn to_headers(&self) -> http::HeaderMap {
```
tomhoule, line 42: This is best effort. Maybe trying a bit too hard, and too optimistic (ignores errors). I'm very open to better ideas.

Finistere: I just wonder how useful it is, sounds like something that would be worth dedicated rules/hook when asked for rather than trying to guess headers

tomhoule: I'm happy with completely removing that part of the logic, and passing empty headers to the request context constructor.

tomhoule: If that sounds best.

crates/integration-tests/src/federation/builder/router.rs
tomhoule: Or better, do it before merging. Maybe something for the weekend.

crates/engine/src/websocket.rs
Finistere: Not entirely sure to be honest, but it just feels arbitrary. Think we should probably just treat them separately and provide similar looking rules/hooks in the future to better handle it. Like `on_gateway_http_request` & `on_gateway_websocket_request` for example. So would remove it yeah

crates/integration-tests/src/federation/websocket_request.rs
```diff
@@ -0,0 +1,69 @@
...
+
+            request.headers_mut().insert(
+                "Sec-WebSocket-Protocol",
+                HeaderValue::from_str("graphql-transport-ws").unwrap(),
+            );
+
+            let (connection, _) = async_tungstenite::tokio::connect_async(request).await.unwrap();
+
+            let (client, actor) = graphql_ws_client::Client::build(connection)
+                .payload(self.init_payload.unwrap_or_else(|| serde_json::Map::default().into()))?
```
Finistere, line 45: shouldn't the default be `null`?

Finistere (approved):
nice work

crates/integration-tests/src/federation/websocket_request.rs
tomhoule: Yeah, can be null (https://github.com/enisdenjo/graphql-ws/blob/master/PROTOCOL.md#connectioninit), will change to null when I deal with the fixmes, before merging.

PR #2512, opened 2025-01-20
engine: tolerate unknown trusted doc ids in Allow mode

tomhoule:
In allow / audit mode, before this commit, we try to fetch the trusted document every time there is a trusted document id in the query, and we return an error if it doesn't exist. This is not in the spirit of allow mode: we have a setting to log, in that case. We should allow for a soft transition.

This commit changes the behaviour to try and fetch the trusted document from the trusted document id, but without hard error in case of failure, unless there is no inline document — there would be nothing to execute in that case, so we can't proceed to handle the request.

closes GB-8282


fbjork (approved):

tomhoule:
This should really be tested, adding a test.

tomhoule:
> This should really be tested, adding a test.

done

PR #2521, opened 2025-01-21
feat: schema file hot reload

pimeys:
This will be enabled for everybody loading the schema file from a file. Just start the gateway with the `-s` flag and it's running. Queries are warmed accordingly as we do with the GDN updates.

Some notes on implementation. I decided on polling, due to notify doesn't really work in certain scenarios:

- macOS with Docker doesn't have the events
- Any network filesystem does not have the events
- Possibly windows doesn't have the events
- And we are not sure about kubernetes config map filesystem...

We poll every five seconds, check the file metadata and its modified at. If this is not available, we load the file contents, hash it to memory and compare with that. This implementation should support the broadest possible amount of use cases.


crates/federated-server/src/server/graph_updater/schema_file.rs
```diff
@@ -0,0 +1,138 @@
...
+
+    match metadata.modified() {
+        // If the file system supports modification date, we can use it to check for changes with less resources.
+        Ok(modified) => Ok(Either::Left(modified)),
+        // This is a fallback for file systems that do not support modification date. It loads and hashes the file
+        // contents to check for changes.
+        Err(_) => {
+            tracing::debug!("The file system with the schema file does not support modification date. The schema hot reload will use a bit more CPU by hashing the file contents.");
+
+            match tokio::fs::read_to_string(schema_path).await {
```
tomhoule, line 108: Depending on how often that will happen (does it happen with configmaps?), there's an optimization to avoid excessive memory usage (with a tradeoff) that would be worth it here imo, to minimize the memory overhead: we don't need to read the whole file into memory, you can hash it chunk after chunk, and only if the hash changed, read the whole file into memory..

tomhoule (approved):

tomhoule:
Also I really think we'll have to make this configurable, you don't want to re-hash every 5 seconds if your schema is 150MB.

pimeys:
> Also I really think we'll have to make this configurable, you don't want to re-hash every 5 seconds if your schema is 150MB.

The hashing of schema should not happen at all. It's only if the file system does not support modification metadata, otherwise we just read the modified at timestamp.

After optimizing the read, I'd like to know how useful this configuration would be. We were discussing earlier with @fbjork how nice this feature is because it does _not_ add any configuration.

crates/federated-server/src/server/graph_updater/schema_file.rs
pimeys: Should not happen at all. I'm not sure how useful this part of the code even is, but this code is run only if the filesystem does not return metadata, especially of modification date is missing. So rare, but I want to write this optimization still.

PR #2523, opened 2025-01-22
fix(engine): Improved query-solver performance

Finistere:
Significantly improves the performance of the query planning. The slow part before was the cost computation of dispensable/extra requirements that aren't strictly necessary to the query, but might be depending on the path.

To estimate it correctly in the general case, there is no way but to run a Steiner tree algorithm and see the additional cost those requirements would incur. I had already been careful about this part, re-using allocations. But we were spending a huge amount of time in the cloning step, from the current Steiner tree to the temporary one for requirement cost estimation, and then re-computing the tree. In this PR i've optimized this part as much as I could, most importantly reducing the total memory size we need to copy and using `Vec` of `Copy` elements which does not need to go through `Clone` and can copy memory directly.

The important gains were from doing as little work as possible to no work at all. My first step was to reduce the number of times we needed to compute that extra cost, by better grouping of edges requiring the same fields and bypassing the Steiner tree algorithm if the result was obvious.

Finally, I managed to entirely remove this part by being a bit smarter and more specific for this use of resolvers requiring one or multiple entity key fields (without nesting). If the key fields are always providable at no cost for any path we may take, we don't need to compute their cost anymore. We only need to add the required fields to the Steiner tree if we end up needing them to include them in the final query plan. 

With all of this, we should be ~20x faster than the default Apollo configuration and their native query planner in the worst-case benchmark. They still generate a better plan for now though. Need to use the better Steiner tree algorithm for this. :)

```
query_plan1/3@1:2-2     time:   [26.841 µs 26.867 µs 26.893 µs]
                        thrpt:  [7.6952 MiB/s 7.7027 MiB/s 7.7101 MiB/s]
                 change:
                        time:   [-45.480% -45.269% -45.065%] (p = 0.00 < 0.05)
                        thrpt:  [+82.033% +82.713% +83.418%]
                        Performance has improved.
Found 4 outliers among 100 measurements (4.00%)
  2 (2.00%) high mild
  2 (2.00%) high severe
query_plan1/4@1:3-4     time:   [407.74 µs 408.17 µs 408.62 µs]
                        thrpt:  [15.619 MiB/s 15.635 MiB/s 15.652 MiB/s]
                 change:
                        time:   [-99.660% -99.593% -99.492%] (p = 0.00 < 0.05)
                        thrpt:  [+19590% +24457% +29301%]
                        Performance has improved.
Found 12 outliers among 100 measurements (12.00%)
  3 (3.00%) low severe
  2 (2.00%) low mild
  2 (2.00%) high mild
  5 (5.00%) high severe
query_plan1/5@1:4-4     time:   [2.1390 ms 2.1412 ms 2.1436 ms]
                        thrpt:  [10.736 MiB/s 10.748 MiB/s 10.759 MiB/s]
                 change:
                        time:   [-98.424% -95.277% +1965.0%] (p = 0.62 > 0.05)
                        thrpt:  [-95.157% +2017.3% +6246.3%]
                        No change in performance detected.
Found 4 outliers among 100 measurements (4.00%)
  1 (1.00%) high mild
  3 (3.00%) high severe
query_plan1/7@1:4-4     time:   [3.0058 ms 3.0076 ms 3.0095 ms]
                        thrpt:  [7.6467 MiB/s 7.6515 MiB/s 7.6562 MiB/s]
                 change:
                        time:   [-99.380% -98.140% +784.42%] (p = 0.61 > 0.05)
                        thrpt:  [-88.693% +5277.7% +16016%]
                        No change in performance detected.
Found 3 outliers among 100 measurements (3.00%)
  1 (1.00%) low mild
  2 (2.00%) high mild
query_plan1/9@3:3-4     time:   [5.5262 ms 5.5287 ms 5.5314 ms]
                        thrpt:  [1.1538 MiB/s 1.1543 MiB/s 1.1549 MiB/s]
                 change:
                        time:   [-99.657% -98.971% +4883.4%] (p = 0.61 > 0.05)
                        thrpt:  [-97.993% +9613.8% +29046%]
                        No change in performance detected.
Found 2 outliers among 100 measurements (2.00%)
  1 (1.00%) high mild
  1 (1.00%) high severe
```


crates/engine/query-solver/src/solve/steiner_tree/shortest_path/tree.rs
```diff
@@ -0,0 +1,125 @@
...
+    graph::GraphIndex,
+    visit::{EdgeRef, GraphBase, NodeCount, NodeIndexable},
+};
+
+use crate::{solve::steiner_tree::SteinerGraph, Cost};
+
+use super::{Reusable, SteinerContext};
+
+#[derive(Clone)]
+pub(super) struct ShortestPathSteinerTree<G: GraphBase<NodeId: Copy, EdgeId: Copy>> {
```
Finistere, line 11: Mostly copy-pasted, split it to a new file + more generics for a bit more safety in the indices

pimeys (approved):
whew, this was a bit heavy to read. Tests are green so approved, but at some point I really need to just start going through the code and grasping what's going on.

tomhoule (approved):
Very nice result! Will also have to understand this with a better approach than reading diffs at some point.

PR #2524, opened 2025-01-22
fix(engine): Properly merge default object on subgraph request error

Finistere:
When a subgraph request fails entirely we add nulls to the fields it
 was supposed to retrieve and propagate the error. This didn't behave
 that well with shared roots. Today the null is not placed at the right
 place and should be reworked later, so don't have a good test for it
 right now.


crates/integration-tests/Cargo.toml
```diff
@@ -68,6 +69,7 @@ features = ["full"]
 workspace = true
 
 [dev-dependencies]
+mimalloc = "0.1"
```
pimeys, line 72: why you need this in tests?

pimeys (approved):

crates/integration-tests/Cargo.toml
Finistere: ah wrong PR, adding it for the benchmarks to have more accurate results as we're using mimalloc in the gateway

PR #2526, opened 2025-01-22
feat: allow providing graph-ref always

pimeys:


gateway/src/args.rs
```diff
@@ -23,6 +26,72 @@ pub(crate) trait Args {
     fn hot_reload(&self) -> bool;
 
     fn log_style(&self) -> LogStyle;
+
+    fn graph_ref(&self) -> Option<&GraphRef>;
+
+    fn dashboard_analytics_enabled(&self) -> bool {
```
Finistere, line 32: would rename it to `can_export_analytics_to_platform()` or something like this. This sounds like a feature the user chooses to use or not.

gateway/src/args/std.rs
```diff
@@ -60,37 +48,46 @@ pub struct Args {
...
+
+                Ok(GraphFetchMethod::FromGraphRef {
+                    access_token: AsciiString::from_ascii(access_token)?,
+                    graph_ref,
+                })
+            }
         }
     }
 
+    fn dashboard_analytics_enabled(&self) -> bool {
```
Finistere, line 87: guess no need to override here?

Finistere (approved):

gateway/src/args.rs
fbjork: I would use the word telemetry rather than analytics

PR #2527, opened 2025-01-23
chore: Bump Gateway to 0.27.1

pimeys:
This also adds helm chart to the release process:

- instead of using `latest` tag, use the current version tag
- bump the minor version of the chart

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3135,7 +3135,7 @@ dependencies = [
 
 [[package]]
 name = "grafbase-gateway"
-version = "0.27.0"
+version = "0.27.1"
 dependencies = [
  "anyhow",
  "ascii",
--- a/gateway/CHANGELOG.md
+++ b/gateway/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## [0.27.1] - 2025-01-23
+
+[CHANGELOG](changelog/0.27.1.md)
+
 ## [0.27.0] - 2025-01-22
 
 [CHANGELOG](changelog/0.27.0.md)
--- a/gateway/Cargo.toml
+++ b/gateway/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "grafbase-gateway"
-version = "0.27.0"
+version = "0.27.1"
 edition.workspace = true
 license = "MPL-2.0"
 homepage.workspace = true
--- a/gateway/changelog/0.27.1.md
+++ b/gateway/changelog/0.27.1.md
@@ -1,0 +1,5 @@
+## Fixes
+
+- Enable `--graph-ref` together with `--schema`, so one can use telemetry without using GDN
+- Query planner performance improvements
+- Properly merge default object on subgraph request error
--- a/gateway/helm/README.md
+++ b/gateway/helm/README.md
@@ -2,6 +2,17 @@
 
 Helm chart for installing Grafbase Gateway.
 
+## Usage
+
+It is recommended to not use this chart as-is due to it pointing to the latest unstable version.
+
+Instead, find the latest Grafbase Gateway version you want to use, and override the tag in your values.yaml:
+
+```yaml
+image:
+  tag: <VERSION>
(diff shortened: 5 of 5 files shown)


fbjork (approved):

.github/workflows/gateway-partial-release.yml
```diff
@@ -162,3 +162,30 @@ jobs:
             sed -i 's/{{LATEST_VERSION}}/${{ env.VERSION_BUMP }}/g' gateway
             ./upload.sh gateway
           fi
+
+      - name: Package the helm chart
```
pimeys, line 166: https://github.com/helm/chart-releaser-action/issues/107#issuecomment-2133797963

gateway/helm/Chart.yaml
```diff
@@ -2,4 +2,4 @@ apiVersion: v2
 name: gateway
 description: A Helm chart for the Grafbase Gateway
 type: application
-version: 0.1.0
+version: 0.2.0
```
cristifalcas, line 5: I have a question about the release process.
The helm chart code did not change in this PR.
- it's a new release because we have new code between now and the previous release
- or, we bump the helm chart version because of the new application version

cristifalcas: I missed initially the version change from latest. 
Ignore my previous comment

gateway/Makefile.toml
```diff
@@ -17,7 +17,12 @@ description = "Bumps the grafbase-gateway package to the specified version"
 dependencies = ["install-cargo-release"]
 script = '''
 VERSION=${@}
+HELM_VERSION=$(semver -i minor $(sed -nE 's/version:\s*(.*)/\1/p' helm/Chart.yaml))
 cargo release version $VERSION --execute --no-confirm -p grafbase-gateway
+# Update the tag in the helm chart
+sed -i 's/^\(\s*tag:\s*\).*/\1'"$VERSION"'/' helm/values.yaml
+# Update the helm chart version
+sed -i 's/^\(\s*version:\s*\).*/\1'"$HELM_VERSION"'/' helm/Chart.yaml
```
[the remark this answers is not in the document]
pimeys: The semver bumps it to the next minor, then we write it back.

gateway/helm/Chart.yaml
cristifalcas: This is an interesting question. 
I don't think it's a requirement, and personally I don't see the benefits.
1. The meaning of the change will be opaque to the users, if they don't read the helm code.
2. In practice, this forces users automatically to follow the `latest`, which can be achieved with proper versioning of the container image
4. I would prefer for me, as a user, to have control over the version. Usually this version is the container version that I would like to use
5. if the company follows best practices with versioning, this can be achieved with containers tagged with ":0.27", "0.27.1", etc

Having said that, this definitely should be discussed and for now we should follow the existing workflows, meaning that this LGTM

pimeys: We do tag the containers. Maybe I'll write to the chart readme to change the tag to the version they want to use.

gateway/helm/values.yaml
```diff
@@ -11,7 +11,7 @@ image:
   repository: ghcr.io/grafbase/gateway
 
   # Image tag to deploy. Default: latest.
-  tag: latest
+  tag: 0.27.1
```
cristifalcas, line 14: See my comments in https://github.com/grafbase/grafbase/pull/2527/files#r1926804701
For me it's unusual to have individual helm charts with minor releases for the application

gateway/helm/Chart.yaml
fbjork: How do popular charts like this one manage their helm releases? https://artifacthub.io/packages/helm/bitnami/postgresql

pimeys:
I will not do the helm chart dance today. It's just lot of waiting for CI and I'm not sure for what benefit. Added a little snippet to the gateway helm chart readme on how to choose the container tag.

PR #2528, opened 2025-01-23
Scope the CLI version to the grafbase crate

tomhoule:
Until this commit, the CLI inherits its version from the workspace. So the workspace version is the CLI version. Other crates used to inherit it but that is no longer the case. The cli version shouldn't have precedence over versions for other crates that are decoupled from it, like the gateway or grafbase-composition.

So this commit scopes the version of the cli to the Cargo.toml of the grafbase crate.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,7 +16,11 @@ members = [
     "crates/engine/id-derives",
     "crates/engine/id-newtypes",
 ]
-exclude = ["crates/wasi-component-loader/examples", "examples/hooks-template", "examples/query_plan"]
+exclude = [
+    "crates/wasi-component-loader/examples",
+    "examples/hooks-template",
+    "examples/query_plan",
+]
 
 [patch.crates-io]
 multipart-stream = { git = "https://github.com/grafbase/multipart-stream-rs-fork", rev = "06ff198e4041c8a8c1c93e580c260d597727c193" } # http-1.0-fix-multipart-mixed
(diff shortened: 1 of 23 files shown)


Cargo.toml
```diff
@@ -16,7 +16,11 @@ members = [
     "crates/engine/id-derives",
     "crates/engine/id-newtypes",
 ]
-exclude = ["crates/wasi-component-loader/examples", "examples/hooks-template", "examples/query_plan"]
+exclude = [
+    "crates/wasi-component-loader/examples",
```
fbjork, line 20: are these excludes still needed?

fbjork (approved):

Finistere (approved):

Cargo.toml
Finistere: When adding the query_plan one needed it for sure, did you try compiling it?

Finistere:
```
❯ cargo build
error: current package believes it's in a workspace when it's not:
current:   /home/brabier/github/grafbase/grafbase/examples/query_plan/subgraph/Cargo.toml
workspace: /home/brabier/github/grafbase/grafbase/Cargo.toml

this may be fixable by adding `examples/query_plan/subgraph` to the `workspace.members` array of the manifest located at: /home/brabier/github/grafbase/grafbase/Cargo.toml
Alternatively, to keep it out of the workspace, add the package to the `workspace.exclude` array, or add an empty `[workspace]` table to the package's manifest.
```

Finistere: so yes they're still necessary

tomhoule: Ah true, I only built from the workspace side, didn't try to build the examples. Will fix.

PR #2529, opened 2025-01-23
feat: basis of the extensions sdk

pimeys:


crates/grafbase-sdk/src/lib.rs
```diff
@@ -0,0 +1,96 @@
...
+
+use semver::Version;
+use std::sync::LazyLock;
+
+/// The minimum version of Grafbase Gateway that can run this SDK.
+pub const MINIMUM_HOST_VERSION: Version = Version::new(0, 48, 0);
+
+/// The version of the SDK.
+pub static GUEST_VERSION: LazyLock<Version> =
+    LazyLock::new(|| Version::parse(std::env!("CARGO_PKG_VERSION")).expect("must be valid semver"));
```
tomhoule, line 20: I'm not sure this works for libraries when included in another crate. Have you tried / seen some docs?

crates/grafbase-sdk/src/lib.rs
```diff
@@ -0,0 +1,96 @@
...
+
+    Version::new(major, minor, patch)
+}
+
+struct Component;
+
+/// Register the extension to the Grafbase Gateway. This macro must be called in the extension
+/// crate root for the local extension implementation.
+///
+/// The first parameter is the type of the extension, and the second parameter is the extension
```
tomhoule, line 54: This line is confusing. Maybe link to the `ExtensionType` enum for the first, and change "the extension type" to "the struct that implements the `Extension` trait"?

tomhoule (approved):
Solid start!

crates/grafbase-sdk/src/lib.rs
pimeys: Augh, you're probably right. I need to to validate this and think it again.

pimeys: Tested by creating a binary crate:

```toml
[package]
name = "extension-test"
version = "0.4.0"
edition.workspace = true
license.workspace = true
homepage.workspace = true
keywords.workspace = true
repository.workspace = true

[dependencies]
grafbase-sdk = { path = "../grafbase-sdk" }

[lints]
````

SDK:

```toml
[package]
name = "grafbase-sdk"
version = "0.1.0"
edition.workspace = true
license.workspace = true
homepage.workspace = true
keywords.workspace = true
repository.workspace = true
```

test code:

```rust
fn main() {
    println!("sdk version: {}", &*grafbase_sdk::GUEST_VERSION);
    println!("this crate version: {}", std::env!("CARGO_PKG_VERSION"));
}
```

output:

```
sdk version: 0.1.0
this crate version: 0.4.0
```

So it seems to work.

crates/grafbase-sdk/src/lib.rs
```diff
@@ -0,0 +1,96 @@
...
+
+    if version < MINIMUM_HOST_VERSION {
+        -(pack_version(&MINIMUM_HOST_VERSION) as i32)
+    } else {
+        pack_version(&GUEST_VERSION) as i32
+    }
+}
+
+fn pack_version(version: &Version) -> u32 {
+    let major = version.major as u32;
```
Finistere, line 34: You should mask the relevant bits before shifting, so `let major = (major & 0xFF) << 16`
I would strongly consider using an`i64` instead of `i32`, You could use 20 bits per number for this. It's far from unheard to have versions in the hundreds. Otel-collector is at `0.118` for example.

Finistere: A back and forth unit test would be nice also :D

crates/grafbase-sdk/src/lib.rs
```diff
@@ -0,0 +1,96 @@
...
+        #[doc(hidden)]
+        #[export_name = "register-extension"]
+        pub extern "C" fn __register_extension(host_version: u32) -> i32 {
+            let version_result = grafbase_sdk::check_host_version(host_version);
+
+            if version_result < 0 {
+                return version_result;
+            }
+
+            match $extension_type {
```
Finistere, line 95: I'm not sure what to think of the extension type thing, I would need play around to see. But it doesn't seem great that both the engine and the user have to specify it.

crates/grafbase-sdk/src/types.rs
```diff
@@ -0,0 +1,113 @@
...
+    /// The name of the directive.
+    pub fn name(&self) -> &str {
+        &self.0.name
+    }
+
+    /// The directive arguments. The output is a Serde structure, that must map to
+    /// the arguments of the directive.
+    ///
+    /// Error is returned if the directive argument does not match the output structure.
+    pub fn args<'de, T>(&'de self) -> Result<T, DecodeError>
```
Finistere, line 20: We shouldn't expose `DecodeError` as part of our API. Should be `impl Error` for example. Would also just `arguments` instead of `args` IMHO.

Finistere (approved):

crates/grafbase-sdk/src/lib.rs
tomhoule: Awesome!

crates/grafbase-sdk/src/lib.rs
pimeys: I forgot WASM supports 64-bit integers, only the pointers are 32-bit... Yeah will do.

crates/grafbase-sdk/src/types.rs
pimeys: It's a bit annoying with `impl Error` because the user needs to define the error type. Maybe we should have our own Error enum with thiserror here?

crates/grafbase-sdk/src/lib.rs
Finistere: Yeah, maybe we should store the implementation in an `Box<dyn Any>` with this ExtensionType, not really sure.As you said we'll iterate on it 🤷

crates/grafbase-sdk/src/lib.rs
Finistere: nice idea for the version though

crates/grafbase-sdk/src/lib.rs
pimeys: I get a rash when thinking about using `Box<dyn Any>`...

PR #2530, opened 2025-01-24
gateway: default to the new domain for hosted trusted documents

tomhoule:
This is the URL used by clients by default. It contains trusted documents uploaded to the platform at grafbase.com. It is overridden in self-hosted enterprise platform deployments.

Note that the path of documents also changed.

closes GB-8327


tomhoule:
Need to update some tests.

fbjork (approved):

PR #2532, opened 2025-01-24
graphql-composition: ingest directive definitions

tomhoule:
As a first use case, this implements the federation spec rule that any directive composed using `@composeDirective` must have the same definition in all subgraphs.

closes GB-8317

### DIFF
--- a/crates/graphql-composition/src/subgraphs/directives/directive_definition.rs
+++ b/crates/graphql-composition/src/subgraphs/directives/directive_definition.rs
@@ -1,0 +1,176 @@
+use super::*;
+use bitflags::bitflags;
+use std::fmt::Display;
+
+pub(crate) struct DirectiveDefinition {
+    pub(crate) subgraph_id: SubgraphId,
+    pub(crate) name: StringId,
+    pub(crate) locations: DirectiveLocations,
+    pub(crate) arguments: Vec<InputValueDefinition>,
+}
+
+#[derive(Debug, PartialEq, Clone)]
+pub(crate) struct InputValueDefinition {
+    pub(crate) name: StringId,
+    pub(crate) r#type: FieldTypeId,
+    pub(crate) default_value: Option<Value>,
+    pub(crate) directives: Vec<Directive>,
+}
+
+#[derive(Debug, PartialEq, Clone)]
+pub(crate) struct Directive {
+    pub(crate) name: StringId,
+    pub(crate) arguments: Vec<(StringId, Value)>,
+}
+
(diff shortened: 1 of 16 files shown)


crates/graphql-composition/src/subgraphs/directives/directive_definition.rs
```diff
@@ -0,0 +1,105 @@
+use super::*;
+use bitflags::bitflags;
+use std::fmt::Display;
+
+pub(crate) struct DirectiveDefinition {
+    pub(crate) subgraph_id: SubgraphId,
+    pub(crate) name: StringId,
+    pub(crate) locations: DirectiveLocations,
+    pub(crate) arguments: Vec<DirectiveDefinitionArgument>,
```
Finistere, line 9: It's a list of `InputValueDefinition` like field arguments, no need for extra type I think. but don't remember how you represent the rest in composition

Finistere (approved):

crates/graphql-composition/src/subgraphs/directives/directive_definition.rs
tomhoule: Hm true, it's even in the spec. Unlike graphql-federated-graph, we don't have a unified `InputValueDefinition` in graphql_composition::Subgraphs, they're scattered in different HashMaps and such, so I'll have to think about how we can deduplicate.

tomhoule: But for now I'll rename it to InputValueDefinition at least.

tomhoule: Also have to validate that it matches. We should also match the directives applied to the input value definitions.

tomhoule: Done, and pushed.

crates/graphql-composition/src/subgraphs/directives/directive_definition.rs
```diff
@@ -0,0 +1,176 @@
...
+
+pub(crate) struct DirectiveDefinition {
+    pub(crate) subgraph_id: SubgraphId,
+    pub(crate) name: StringId,
+    pub(crate) locations: DirectiveLocations,
+    pub(crate) arguments: Vec<InputValueDefinition>,
+}
+
+#[derive(Debug, PartialEq, Clone)]
+pub(crate) struct InputValueDefinition {
```
tomhoule, line 13: Will move to its own module as it gets more use.

PR #2533, opened 2025-01-24
feat: very basic rest extension that can do a GET

pimeys:
Adds a very simple REST extension, which should show how an extension could look like. Notable things:

```rust
#[derive(grafbase_sdk::ResolverExtension)]
struct MyExtension;
```

This will generate the needed initialization code, and fail because the struct needs to implement the `Extension` and `Resolver` traits. The user implements these traits, the code compiles and should be ready to use in the gateway.

### DIFF
--- a/crates/grafbase-sdk/src/host_io/http.rs
+++ b/crates/grafbase-sdk/src/host_io/http.rs
@@ -2,11 +2,15 @@
 //! Functions are synchronous from the guest's perspective, yet they run asynchronously on the host side.
 //! While waiting for request completion, the host thread can execute other tasks concurrently.
 
-use serde::Serialize;
+use std::string::FromUtf8Error;
+
+pub use crate::wit::{HttpError, HttpMethod, HttpVersion};
+pub use http::StatusCode;
+pub use serde_json::Error as JsonDeserializeError;
 pub use url::Url;
 
 use crate::wit::HttpClient;
-pub use crate::wit::{HttpError, HttpMethod, HttpResponse, HttpVersion};
+use serde::Serialize;
 
 /// Executes a single HTTP request and returns a result containing either an `HttpResponse` or an `HttpError`.
 ///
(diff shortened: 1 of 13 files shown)


extensions/rest/src/lib.rs
```diff
@@ -0,0 +1,182 @@
...
+        })?;
+
+        let filter = self.create_filter(rest.selection)?;
+        let inputs = RcIter::new(core::iter::empty());
+
+        let filtered = filter.run((Ctx::new([], &inputs), Val::from(data)));
+
+        let mut results = FieldOutput::new();
+
+        for result in filtered {
```
Finistere, line 168: That's weird to me,  the number of field outputs must be equal to the number of field inputs. So if there are multiple results, the only thing that makes sense to me is to wrap them in a list and then push it, combined, to the output.

Finistere (approved):

crates/grafbase-sdk/src/host_io/http.rs
```diff
@@ -333,3 +344,45 @@ impl Default for BatchHttpRequest {
...
+impl HttpResponse {
+    /// Returns the status code of the HTTP response.
+    pub fn status(&self) -> http::StatusCode {
+        *OnceCell::new()
+            .get_or_init(|| http::StatusCode::from_u16(self.0.status).expect("must be valid, this comes from reqwest"))
+    }
+
+    /// Returns the headers of the HTTP response.
+    pub fn headers(&self) -> &[(String, String)] {
+        &self.0.headers
```
tomhoule, line 360: For ergonomics, we could return an `http::HeaderMap` here, but it's fine for the stage we're at now.

extensions/rest/src/lib.rs
```diff
@@ -0,0 +1,182 @@
...
+        }
+    }
+}
+
+impl Extension for RestExtension {
+    fn new(schema_directives: Vec<Directive>) -> Result<Self, Box<dyn std::error::Error>> {
+        let mut endpoints = Vec::<RestEndpoint>::new();
+
+        for directive in schema_directives {
+            if directive.name() != REST_ENDPOINT_DIRECTIVE_NAME {
```
tomhoule, line 64: Shouldn't the host pass only the relevant directives here?

tomhoule: (maybe not at the moment, but in principle)

tomhoule (approved):
Nice! Already feels usable

crates/grafbase-sdk/src/host_io/http.rs
pimeys: Yeah, and no right now. It's annoying to build and let's see if we need it.

PR #2539, opened 2025-01-27
graphql-composition: ingest directive definitions

tomhoule:
Include the relevant directive definitions from the composed subgraphs in FederatedGraph, and render them when displaying the FederatedGraph.

This lets us implement the mechanism for finely tracking which directive definitions we need to emit, including definitions from the join spec.

closes GB-8317


crates/graphql-federated-graph/src/federated_graph/directive_definitions.rs
```diff
@@ -0,0 +1,88 @@
...
+
+#[derive(Debug, Clone)]
+pub struct DirectiveDefinition {
+    pub name: StringId,
+    pub locations: DirectiveLocations,
+    pub arguments: InputValueDefinitions,
+    pub repeatable: bool,
+}
+
+bitflags::bitflags! {
```
tomhoule, line 11: This struct just moved.

crates/graphql-composition/src/emit_federated_graph/federation_builtins.rs
```diff
@@ -0,0 +1,257 @@
...
+        let name = ctx.insert_str("join__Graph");
+        ctx.out.push_scalar_definition(federated::ScalarDefinitionRecord {
+            name,
+            directives: Vec::new(),
+            description: None,
+        })
+    };
+
+    // directive @join__unionMember(graph: join__Graph!, member: String!) repeatable on UNION
+    {
```
tomhoule, line 32: Ideally we'll do what we do in this function with SDL instead of verbose constructors when `@link` support is implemented.

crates/graphql-composition/src/compose/directive_definitions.rs
```diff
@@ -0,0 +1,62 @@
...
+        .filter(|definition| ctx.subgraphs.is_composed_directive(definition.name))
+        .collect();
+
+    definitions.sort_by_key(|definition| definition.name);
+    definitions.dedup_by_key(|definition| definition.name);
+
+    // Emit
+    for definition in definitions {
+        let name = ctx.insert_string(definition.name);
+        let mut arguments_start: Option<federated::InputValueDefinitionId> = None;
```
Finistere, line 20: you could just do:
```rust
let start = input_value_definitions.len();
for { ... }
let end = input_value_definitions.len();
```
and the range will be good

crates/graphql-composition/src/emit_federated_graph.rs
```diff
@@ -46,80 +52,83 @@ pub(crate) fn emit_federated_graph(mut ir: CompositionIr, subgraphs: &Subgraphs)
...
     emit_input_value_definitions(&ir.input_value_definitions, &mut ctx);
     emit_fields(&ir.fields, &mut ctx);
 
     emit_union_members_after_objects(&ir.union_members, &mut ctx);
+    federation_builtins::emit_federation_builtins(&mut ctx);
+
+    emit_directives_and_implements_interface(&mut ctx, ir);
 
-    emit_directives_and_implements_interface(ctx, ir);
+    ctx.out.enum_values.sort_unstable_by_key(|v| v.enum_id);
```
Finistere, line 65: You're sure about keeping `ctx`? If I remember correctly I was taking ownership over `ctx` because any modifications after this method wouldn't be correct

crates/graphql-federated-graph/src/render_sdl/directive_definition.rs
```diff
@@ -0,0 +1,34 @@
...
+pub(crate) fn display_directive_definition(
+    directive_definition: &DirectiveDefinition,
+    directives_filter: fn(&Directive, graph: &FederatedGraph) -> bool,
+    graph: &FederatedGraph,
+    f: &mut fmt::Formatter<'_>,
+) -> fmt::Result {
+    f.write_str("directive @")?;
+    f.write_str(&graph[directive_definition.name])?;
+
+    if directive_definition.arguments.1 > 0 {
```
Finistere, line 14: Something like `if !graph[directive_definition.arguments].is_empty()` would be clearer IMHO

Finistere (approved):

crates/graphql-composition/src/emit_federated_graph.rs
tomhoule: IR is being `mem::take()`n from but I haven't seen anything problematic with the context. Will double check. I ended up putting the emit_cost_directive() and emit_list_size_directive() calls inside, so emit_directives_and_implements_interface() could still take ownership.

Finistere: Should have added a comment at the time I guess 😅  Maybe I'm remembering something wrong

PR #2540, opened 2025-01-27
feat: grafbase extension init

pimeys:
Creates a basic project structure with `grafbase extension init`.


cli/src/extension.rs
```diff
@@ -0,0 +1,124 @@
...
+        .expect("must_exist")
+        .to_string_lossy()
+        .to_case(Case::Kebab);
+
+    let mut params = HashMap::new();
+    params.insert("name", name.clone());
+    params.insert("sdk_version", sdk_cargo_toml.package.version);
+
+    let mut handlebars = Handlebars::new();
+    handlebars.register_template_string("Cargo.toml", source)?;
```
Finistere, line 100: I prefer https://github.com/rinja-rs/askama which is type-safe, but up to you

Finistere (approved):

PR #2541, opened 2025-01-28
chore(extension): Initial runtime trait + schema integration

Finistere:


Cargo.toml
```diff
@@ -163,6 +163,7 @@ federated-server = { path = "crates/federated-server" }
 gateway-config = { path = "crates/gateway-config" }
 rolling-logger = { path = "crates/rolling-logger" }
 grafbase-sdk = { path = "crates/grafbase-sdk" }
+extension = { path = "crates/extension" }
```
pimeys, line 166: You can and should use this from the workspace now.

pimeys (approved):

Cargo.toml
Finistere: it is the workspace, not sure what you mean

PR #2547, opened 2025-01-28
feat: grafbase extension build

pimeys:


Cargo.toml
```diff
@@ -22,6 +21,7 @@ exclude = [
     "crates/wasi-component-loader/examples",
     "examples/hooks-template",
     "examples/query_plan",
+    "extensions/rest",
```
pimeys, line 24: We can't compile the extension with wasm32-wasip2 target if it's in the workspace. We suddenly start compiling tokio and fail...

pimeys: We might wanna move extensions to their own repo at some point...

Finistere (approved):

PR #2551, opened 2025-01-29
feat: extension wasi instance loader

pimeys:
This started to get a bit bloated, so what's done now is:

- wasi loader for extensions, a test for an extension
- pooling and runtime for extensions
- config, missing some stuff like env vars and headers, which will come later
- we just pass an empty vector for extensions, the loading part comes in another PR

We'll have a call tomorrow morning with @Finistere on what's next.


crates/gateway-config/src/extensions.rs
```diff
@@ -0,0 +1,67 @@
...
+    pub version: String,
+    pub networking: bool,
+    pub stdout: bool,
+    pub stderr: bool,
+    pub environment_variables: bool,
+    pub max_pool_size: usize,
+}
+
+#[derive(Debug, Clone)]
+pub struct WasiExtensionsConfig {
```
Finistere, line 21: You could have added `HashMap<String, WasiExtensionConfig>` to the top level config and used a custom`impl Deserialize` with the enum to hide this logic.
I don't think I would have added the default location logic with `grafbase_extensions/...` here but in the wasi component loader IMHO. `{name}-{version}` sounds like a possible way to have name clashes, considering semver allows metadata information past the patch version. Would instead use `{name}/{version}` sounds a lot safer to me

Finistere (approved):

crates/gateway-config/src/extensions.rs
pimeys: We will change this completely as discussed in our call today. Not in this PR though.

crates/grafbase-sdk/wit/world.wit
```diff
@@ -41,6 +41,20 @@ world sdk {
         resolver
     }
 
+    // A sender for the system access log.
```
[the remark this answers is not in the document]
pimeys: https://linear.app/grafbase/issue/GB-8372/enabledisable-access-logs-through-extension-configuration

crates/gateway-config/src/extensions.rs
pimeys: It will be done with this issue:

https://linear.app/grafbase/issue/GB-8373/extension-loading-in-the-gateway-reload

PR #2552, opened 2025-01-29
feat(extension): basic initial integration-tests

Finistere:
Connecting the whole extension logic together with engine support


extensions/rest/example/grafbase.toml
```diff
@@ -0,0 +1,2 @@
+[extensions]
+rest = { path = "/home/brabier/github/grafbase/grafbase/extensions/rest/build" }
```
tomhoule, line 2: Should that stay hardcoded to your home path?

extensions/rest/example/schema.graphql
```diff
@@ -0,0 +1,23 @@
+enum join__Graph {
+  A @join__graph(name: "a")
+}
+
+enum extension__Link {
+  REST
+    @extension__link(
+      url: "file:///home/brabier/github/grafbase/grafbase/extensions/rest/build"
```
tomhoule, line 8: Same here

extensions/rest/example/grafbase.toml
Finistere: just a dummy example for us, it's not meant to stay

crates/engine/src/resolver/mod.rs
```diff
@@ -1,47 +1,4 @@
...
-//!     }
-//! }
-//! ```
-//!
-//! Execution plans define what to do at runtime for a given query. They only depend on the
-//! operation and thus can be cached and do not depend on any context. On the other hand,
-//! Executors are context (variables, response, headers, etc.) depend and built from the execution plans
-//!
-//! The executor for the catalog plan would have a single response object root and the price plan
-//! executor will have a root for each product in the response.
```
tomhoule, line 44: Is this doc comment not applicable anymore?

crates/graphql-federated-graph/src/federated_graph.rs
```diff
@@ -58,42 +58,22 @@ pub struct FederatedGraph {
 
 impl FederatedGraph {
     #[cfg(all(feature = "from_sdl", feature = "extension"))]
-    pub async fn from_sdl_with_extensions(sdl: &str) -> Result<Self, crate::DomainError> {
+    pub async fn from_sdl(sdl: &str) -> Result<Self, crate::DomainError> {
```
tomhoule, line 61: We'll have to land back on an extensions free API for the next release of the crate, I think. But based on what we discussed this morning, the async part will move to engine::Schema building, right?

tomhoule (approved):
Nice test setup!

crates/graphql-federated-graph/src/federated_graph.rs
Finistere: yeah removing that stuff and just imitating the `join__Graph` handling in a PR right now

crates/engine/src/resolver/mod.rs
Finistere: Names are wrong and not that accurate

PR #2559, opened 2025-01-30
feat: generate extension catalog, generate the wasi loader on gateway launch

pimeys:
I need some schema to test this with. With just a random schema the gateway crashes with:

```
[crates/engine/schema/src/generated/subgraph/virtual_.rs:53:10] &self.schema = Schema { .. }
thread 'main' panicked at crates/engine/schema/src/generated/subgraph/virtual_.rs:53:28:
index out of bounds: the len is 0 but the index is 0
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

### DIFF
--- a/crates/federated-server/src/server/gateway.rs
+++ b/crates/federated-server/src/server/gateway.rs
@@ -10,7 +10,7 @@ use runtime_local::wasi::{
     extensions::{Directive, ExtensionConfig, ExtensionType, WasiExtensions},
     hooks::{ChannelLogSender, HooksWasi},
 };
-use std::{env, fs::File, path::PathBuf, sync::Arc};
+use std::{env, fs::File, ops::Not, path::PathBuf, sync::Arc};
 use tokio::sync::watch;
 use ulid::Ulid;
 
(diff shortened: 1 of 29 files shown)


crates/federated-server/src/server/gateway.rs
```diff
@@ -76,20 +82,163 @@ pub(super) async fn generate(
...
+        let mut config = ExtensionConfig {
+            id: extension_id,
+            name: extension.manifest.name.clone(),
+            extension_type,
+            schema_directives: Vec::new(),
+            max_pool_size: extension_config.max_pool_size(),
+            wasi_config,
+        };
+
+        for subgraph in schema.subgraphs() {
```
pimeys, line 148: Could we do this without `O(n^2)` double loop?

tomhoule: I don't see how, if you need to access every directive inside every subgraph. Extension schema directives are nested inside the subgraphs, right?

tomhoule (approved):
lgtm, but @Finistere has worked with that code more

crates/federated-server/src/server/gateway.rs
pimeys: Yeah, maybe it's not an issue. Just bothers me to write this code.

Finistere: It's not `O(n^2)` it's `O(n)`, you're iterating over every extension schema directive once.

Finistere: ah didn't see the outer loop, my bad

Finistere: You could just remove it from this loop and do `gateway_config.extensions.get(catalog[extension_id].manifest.name).schema_directives.push(directive)`

Finistere: Think you should rely on `ExtensionId` defined by the catalog inside the wasi stuff, so just have it in the same order. That way you bypass the hashmap stuff

Finistere (approved):

crates/federated-server/src/server/gateway.rs
pimeys: The issue is I do need the name of the subgraph too...

PR #2560, opened 2025-01-31
fix(websocket): Propagate headers according to header rules

Finistere:


pimeys (approved):
Is the data format for headers a standard, or something we tell the clients to handle accordingly in the subgraph?

Finistere:
I'm not sure to follow, a websocket connection is initialized by a standard HTTP payload requesting an upgrade. The headers are coming from that initial HTTP payload.

pimeys:
> I'm not sure to follow, a websocket connection is initialized by a standard HTTP payload requesting an upgrade. The headers are coming from that initial HTTP payload.

I mean this stuff, so you have data sent in a certain ways, and you need to look for the headers:

https://github.com/grafbase/grafbase/pull/2560/files#diff-d538322b6a0a24314d4914d05b0bda6e8fa5a87d69d45113b34179c4e18a582fR70-R87

PR #2563, opened 2025-01-31
gateway: config option for subscription transport to subgraphs

tomhoule:
When the gateway receives a subscription request, it can use either server-sent events or websockets to communicate with the subgraph serving the subscription. This is independent from the protocol used by the client to connect to the gateway. Until now, the logic was that if a `websocket_url` was defined in the subgraph configuration, the gateway would use a websocket connection, otherwise default to server sent events. This is opaque, and not flexible enough.

This pull request introduces a new subgraph configuration option in `grafbase.toml`:

```toml
[subgraphs.my-subgraph]
subscriptions_protocol = "websockets"
```

The value must be either "websockets" or "server_sent_events". The gateway will follow the choice of protocol for the subgraph. If this option is omitted, the previous logic applies.

Implementation note: I had to introduce enums to the engine schema codegen, I hope I didn't miss the mark.

closes GB-8386

### DIFF
--- a/crates/engine/codegen/domain/schema.graphql
+++ b/crates/engine/codegen/domain/schema.graphql
@@ -245,6 +245,7 @@ type ExtensionDirective @meta(module: "directive/extension") @indexed(id_size: "
 # -------------
 
 scalar SubgraphConfig
+scalar SubscriptionsProtocol @copy
 
 union Subgraph @id @meta(module: "subgraph") @variants(empty: ["Introspection"], remove_suffix: "Subgraph") =
   | GraphqlEndpoint
@@ -258,6 +259,8 @@ type GraphqlEndpoint @meta(module: "subgraph/graphql") @indexed(id_size: "u16") 
   config: SubgraphConfig!
   "Schema directives applied by the given subgraph"
   schema_directives: [TypeSystemDirective!]!
+  "The protocol to use for subscriptions from this subgraph"
+  subscriptions_protocol: SubscriptionsProtocol!
 }
 
 "Virtual subgraphs have no dedicated support on the engine side, everything is resolved through extensions."
--- a/crates/gateway-config/src/subscriptions_protocol.rs
+++ b/crates/gateway-config/src/subscriptions_protocol.rs
@@ -1,0 +1,38 @@
+use serde::{Deserialize, Serialize};
+
+#[derive(Debug, Deserialize, PartialEq, Eq, Clone, Copy, Serialize)]
+#[serde(rename_all = "snake_case")]
+pub enum SubscriptionsProtocol {
+    ServerSentEvents,
+    Websocket,
+}
+
+#[cfg(test)]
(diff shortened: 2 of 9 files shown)


crates/engine/codegen/domain/schema.graphql
```diff
@@ -267,6 +269,11 @@ type VirtualSubgraph @meta(module: "subgraph/virtual_") @indexed(id_size: "u16")
   schema_directives: [TypeSystemDirective!]!
 }
 
+enum SubscriptionsProtocol @meta(module: "subgraph/subscriptions_protocol") {
```
Finistere, line 272: You didn't need to create an enum here, you could have created the enum outside and used a scalar here, but why not^^

crates/engine/codegen/domain/schema.graphql
```diff
@@ -267,6 +269,11 @@ type VirtualSubgraph @meta(module: "subgraph/virtual_") @indexed(id_size: "u16")
...
+  SERVER_SENT_EVENTS
+  WEBSOCKETS
```
Finistere, line 274: Why plural for websocket?

crates/gateway-config/src/subscriptions_protocol.rs
```diff
@@ -0,0 +1,38 @@
+use serde::{Deserialize, Serialize};
+
+#[derive(Debug, Deserialize, PartialEq, Eq, Clone, Serialize)]
+#[serde(rename_all = "snake_case")]
+pub enum SubscriptionsProtocol {
+    ServerSentEvents,
+    Websockets,
```
Finistere, line 7: Shouldn't be plural IMHO

Finistere (approved):

crates/engine/codegen/domain/schema.graphql
tomhoule: Just by feel. I don't care either way, happy to change to singular.

crates/gateway-config/src/subscriptions_protocol.rs
tomhoule: Will change.

Finistere: for `wesocket` I mean, `ServerSentEvents` is indeed plural from what I saw

crates/engine/codegen/domain/schema.graphql
tomhoule: Aah, I see what you mean, should have thought of that. I'll revert that part then, better not add unnecessary code.

tomhoule: It was still a nice refresher on the codegen :)

Finistere: basically a scalar is used whenever it's something external. You can add the `@copy` also which will avoid the generation of a method

PR #2565, opened 2025-01-31
graphql-composition: generic @link support

tomhoule:
Composition previously only dealt with `@link` for the federation directives. Now, we take into account every `@link` and internalize them in the Subgraphs struct. The federated graph output is not changed at the moment, because only the only linked schema with an effect is the federation schema. But this is the infrastructure we need for arbitrary directives from extensions.

closes GB-8313

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,6 +18,7 @@ target
 # misc
 .DS_Store
 *.pem
+/.cargo
 
 # debug
 npm-debug.log*
@@ -47,14 +48,10 @@ yarn-error.log*
 # Jujutsu
 .jj/
 
-# Direnv
-.envrc
-.env
-.direnv/
-
 # Nix
 /result
 
 # CLI assets
 /cli/assets
+
 .aider*
--- a/crates/graphql-composition/tests/composition/namespaced_directives_basic/subgraphs/alerts.graphql
+++ b/crates/graphql-composition/tests/composition/namespaced_directives_basic/subgraphs/alerts.graphql
@@ -1,0 +1,25 @@
+extend schema @link(url: "https://example.com/kafka/v2.3")
+
+type Alert @key(fields: "id") {
+  id: ID!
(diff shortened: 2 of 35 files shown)


.gitignore
```diff
@@ -47,14 +48,10 @@ yarn-error.log*
 # Jujutsu
 .jj/
 
-# Direnv
-.envrc
-.env
-.direnv/
```
pimeys, line 53: ?

crates/graphql-composition/tests/composition/namespaced_directives_basic/subgraphs/alerts.graphql
```diff
@@ -0,0 +1,25 @@
...
+}
+
+enum Severity {
+  FINE
+  OK
+  GOOD
+}
+
+type Mutation {
+  createAlert(message: String!, severity: Severity!): Alert @kafka__post(topic: "alerts")
```
pimeys, line 16: Am I missing something or should this directive be in the federated graph somewhere?

pimeys (approved):

.gitignore
tomhoule: There was duplication, just deduplicated.

crates/graphql-composition/tests/composition/namespaced_directives_basic/subgraphs/alerts.graphql
tomhoule: No, as long as it hasn't been registered as an extension directive (see https://linear.app/grafbase/document/extensions-sdl-ffd3e6c1fe3b). For these, there will be a transformation, they are not passed into the federated graph verbatim.

.gitignore
tomhoule: And added `.cargo/`, that's the one that got committed by mistake on my end.

PR #2566, opened 2025-01-31
fix: load extensions in parallel

pimeys:
They take a hot moment to load. If using more than a few extensions, this will add up. They can be loaded in parallel, so that's what we do here.

### DIFF
--- a/crates/runtime-local/src/wasi/extensions.rs
+++ b/crates/runtime-local/src/wasi/extensions.rs
@@ -2,6 +2,7 @@ mod pool;
 
 use engine_schema::Subgraph;
 use extension_catalog::ExtensionId;
+use futures_util::StreamExt;
 use gateway_config::WasiExtensionsConfig;
 use runtime::{
     error::{PartialErrorCode, PartialGraphqlError},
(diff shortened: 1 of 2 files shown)


tomhoule (approved):

crates/runtime-local/src/wasi/extensions.rs
```diff
@@ -21,36 +23,63 @@ use super::guest_error_as_gql;
...
+                None => Ok(None),
+            }
+        }));
+    }
 
-        let inner = WasiExtensionsInner { instance_pools };
+    let mut pools = HashMap::new();
 
-        Ok(Self(Some(Arc::new(inner))))
+    while let Some(result) = creating_pools.next().await {
```
Finistere, line 72: Would create a stream and use https://docs.rs/futures-util/latest/futures_util/stream/trait.StreamExt.html#method.buffer_unordered with limit = https://doc.rust-lang.org/stable/std/thread/fn.available_parallelism.html
This will execute everything in parallel.

Finistere (approved):

crates/runtime-local/src/wasi/extensions.rs
Finistere: so 100 extension -> 100 extensions being built all at once today

PR #2567, opened 2025-01-31
gateway: configurable websocket route

tomhoule:
The gateway has a configuration option for the graphql endpoint:

```toml
[graph]
path = "/custom" # default: "/graphql"
```

But the websocket endpoint is always exposed at `/ws`. This PR introduces a `websocket_path` option to configure this.

```toml
[graph]
path = "/custom" # default: "/graphql"
websocket_path = "/subscriptions" # default: "/ws"
```

closes GB-8392

### DIFF
--- a/crates/integration-tests/src/federation/builder/router.rs
+++ b/crates/integration-tests/src/federation/builder/router.rs
@@ -3,18 +3,22 @@ use std::sync::Arc;
 use axum::{extract::State, response::IntoResponse, routing::get, Router};
 use engine::Engine;
 use engine_axum::websocket::{WebsocketAccepter, WebsocketService};
+use gateway_config::Config;
 
 use super::TestRuntime;
 
-pub(super) fn build(engine: Arc<Engine<TestRuntime>>) -> Router {
+pub(super) fn build(engine: Arc<Engine<TestRuntime>>, config: &Config) -> Router {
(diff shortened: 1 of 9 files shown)


fbjork (approved):

pimeys (approved):
remember to do a docs pr!

crates/integration-tests/src/federation/builder/router.rs
```diff
@@ -3,18 +3,22 @@ use std::sync::Arc;
...
     let websocket_accepter = WebsocketAccepter::new(websocket_receiver, engine_watcher);
     tokio::spawn(websocket_accepter.handler());
 
+    let graphql_path = config.graph.path.as_deref().unwrap_or("/graphql");
+    let websocket_path = config.graph.websocket_path.as_deref().unwrap_or("/ws");
+
     Router::new()
-        .route("/graphql", get(execute).post(execute))
-        .route_service("/ws", WebsocketService::new(websocket_sender))
+        .route(graphql_path, get(execute).post(execute))
```
Finistere, line 20: Ideally we would use the federated-server, but well non-trivial amount of work

tomhoule: Yeah, I agree the tests aren't as convincing if we're reproducing the router but not actually using it.

Finistere (approved):

PR #2573, opened 2025-02-03
graphql-composition: use insta for tests

tomhoule:
Each test root is identified by a `test.md` file which can have a description of the purpose of the test. The contents of the test.md will end up in the snapshot for the federated graph, directly next to the diff, which I find pretty neat for documenting what the test does when you are reviewing changes to the snapshot.


Finistere (approved):
Nice! ❤️

PR #2578, opened 2025-02-04
graphql-composition: optional grafbase extensions support

tomhoule:
closes GB-8331
closes GB-8398

### DIFF
--- a/crates/graphql-composition/src/subgraphs/directives/link.rs
+++ b/crates/graphql-composition/src/subgraphs/directives/link.rs
@@ -1,0 +1,83 @@
+pub(crate) struct LinkUrl {
+    #[expect(unused)]
+    pub(crate) url: url::Url,
+    pub(crate) name: Option<String>,
+    pub(crate) version: Option<String>,
+}
+
+/// https://specs.apollo.dev/link/v1.0/#@link.url
+pub(crate) fn parse_link_url(url: &str) -> Option<LinkUrl> {
+    // Must be a url, or treated as an opaque identifier (which is valid).
+    let url = url::Url::parse(url).ok()?;
+
+    let segments = url.path_segments()?;
+
+    let mut reversed_segments = segments.rev();
+
+    let Some(maybe_version_or_name) = reversed_segments.next() else {
+        return Some(LinkUrl {
+            url,
+            name: None,
+            version: None,
+        });
+    };
+
+    if is_valid_version(maybe_version_or_name) {
+        let name = reversed_segments
+            .next()
+            .filter(|s| is_valid_graphql_name(s))
+            .map(String::from);
+
+        let version = Some(maybe_version_or_name.to_owned());
+
+        Some(LinkUrl { url, name, version })
+    } else if is_valid_graphql_name(maybe_version_or_name) {
+        let name = Some(maybe_version_or_name.to_owned());
+
+        Some(LinkUrl {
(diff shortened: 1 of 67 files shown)


crates/graphql-composition/src/subgraphs/directives/link.rs
```diff
@@ -0,0 +1,83 @@
+pub(crate) struct LinkUrl {
```
tomhoule, line 1: This whole module mostly just moved.

crates/graphql-federated-graph/src/directives/extension.rs
```diff
@@ -27,3 +27,37 @@ pub struct ExtensionDirective {
...
+/// ) repeatable on ENUM_VALUE
+///
+/// input extension__LinkSchemaDirective {
+///    graph: join__Graph!
+///    name: String!
+///    arguments: DirectiveArguments
+/// }
+/// ```
+#[derive(PartialEq, PartialOrd, Clone, Debug)]
+pub struct ExtensionLinkDirective {
```
Finistere, line 53: I'm confused by the purpose of this struct, to my understanding the exact same data is in `Extension`.

tomhoule: There's duplication, you're right. I had hesitations on whether it should live on `Extension` or on the enum. Let's deduplicate and move it back to extension only for now.

tomhoule: After getting into the weeds of this happening in from_sdl() and render_federated_sdl(), I think it would be easier to have it _only_ on the directive, and not in the Extension struct (same thing for join__graph on join__Graph). But I've got it working now, pushing to get this PR out.

Finistere: So does that mean I'll need to find the `join__Graph` / `extension__Link` enum in federated graph to get the info for the engine Schema? Or do I miss something?

tomhoule: Not now, it's only going to be in `Extension`. But yes, implementation-wise, I think it'd be cleaner if it were in the directive. To get to the info, you would have to do:

```rust
fn get_extension_url_and_schema_directives(extension: &Extension, graph: &FederatedGraph) -> (StringId, &[ExtensionLinkSchemaDirective]) {
	let enum_value = &graph[extension.enum_value_id];
   let extension_link = enum_value.directives.iter().find_map(|directive| match directive { ExtensionLinkDirective(directive) => Some(directive), _ => None).unwrap();
 Some( (exension_link.url, extension_link.schema_directives))
}
```

That's because now, in `from_sdl()`, the directives comes back as a `Directive::Other`.

tomhoule: At the same time, having it in `Extension` saves you that `.unwrap()`. I just wish we had a clean, uniform way of handling `join__Graph` and `extension__Link`.

Finistere: ah don't get me wrong, nothing against your proposal. Sounds good to me if you think that's cleaner

tomhoule: I'm trying to convince myself too, at the moment :p Always thinking about these data structures, it's hard to find the real sweet spots.

PR #2581, opened 2025-02-04
feat: hook up extension auth provider

pimeys:
- Adding auth extensions to the config and engine loading
- Creating a simple JWT auth extension
- The grafbase init now takes `--type`, which can be either `auth` or `resolver`

Integration tests coming in the next PR, this is already pretty bloated.

## Some points from LLM code review to think about

<details><summary>Thinking</summary>
Okay, let me start by understanding the code changes provided. The user is adding custom authentication using WASM modules to a GraphQL engine. The first thing I notice is the use of the ```#[derive(AuthenticationExtension)]``` macro. That's probably generating some code, but if there's a custom trait involved, like ```Authenticator```, the derived implementation might not handle all necessary methods or dependencies. I should check if the macro properly implements all required traits.

Looking at the ```impl Extension for {{name}}``` block, the ```new``` function returns ```Result<Self, Box<dyn std::error::Error>>```. Using a boxed error isn't ideal here because it can make error handling vague. They should define a concrete error type so callers can handle specific cases, which improves maintainability.

In the ```AuthExtensionService```, there's a check for ```auth_config``` using ```unwrap_or_default()```. If the configuration is missing, this defaults to an empty config, which might lead to unexpected behavior. They should instead handle the absence of auth config explicitly, maybe returning an error if it's required.

The ```authenticate``` method in ```AuthExtensionService``` uses ```instance_pools.get(&id)```, and if it's not found, returns an internal server error. However, if the pool isn't initialized properly, this could cause runtime issues. They should ensure that the pool is correctly initialized during setup to avoid such errors.

The ```AccessToken``` enum now includes an ```Extension``` variant, but the ```JwtToken``` struct no longer stores the signature. Previously, the signature was used for hashing in the cache key. Removing it might affect cache key uniqueness, leading to potential collisions. They need to ensure that the new ```ExtensionToken``` provides a reliable way to generate unique cache keys, possibly by including a unique identifier or hash of the claims.

In the WASM integration, the ```authenticate``` function's error handling converts guest errors into ```PartialGraphqlError```. However, if the guest error's status code is invalid, it defaults to 500. They should validate the status code properly and handle cases where the provided code isn't a valid HTTP status.

The ```ExtensionToken``` struct's ```claims``` are stored as a ```HashMap<String, serde_json::Value>```. If the claims are large or accessed frequently, this could lead to performance issues. They might consider using a more efficient data structure or caching mechanisms.

Looking at the ```CustomClaims``` handling, they insert the issuer into the claims. If the issuer is already present in the claims, this could overwrite it, causing data inconsistency. They should check for existing keys before inserting to avoid unintended overwrites.

The ```unauthorized()``` function constructs an error response but doesn't include a WWW-Authenticate header, which is recommended by HTTP specs for 401 errors. Adding this header would improve compliance with standards.

There's a potential security concern with the ```claims``` in ```ExtensionToken``` being arbitrary. Without validation or sanitization, malicious claims could be injected. They should implement validation logic to ensure only permitted claims are included.

Lastly, the code uses ```todo!()``` in the generated ```new``` and ```authenticate``` functions. These are placeholders and would cause panics if not replaced with actual implementations. They need to ensure that all ```todo!()``` macros are removed before deployment.
</details>

Here are potential issues and improvements for the custom authentication implementation:

1. **Security Considerations**
- Extension tokens allow arbitrary claims maps without validation/sanitization. This could enable claim injection attacks.
- **Critical**: Add claims allow-list validation in the authentication flow.

2. **Header Propagation**
- The authentication flow passes through original headers without sanitization. This could expose internal headers to downstream services.
- **Recommendation**: Implement allow-list for propagated headers.

3. **Testing Considerations**
- No apparent mechanism for testing authentication extensions locally. **Suggest** adding a dev-mode bypass or mock authenticator interface.

4. **Error Response Consistency**
- The ```unauthorized()``` helper doesn't include WWW-Authenticate header required by some HTTP clients. **Add** standard authentication challenge headers.


cli/tests/extension/mod.rs
```diff
@@ -142,3 +148,123 @@ fn build() {
...
+
+    insta::assert_snapshot!(&cargo_toml, @r#"
+    [package]
+    name = "test-project"
+    version = "0.1.0"
+    edition = "2021"
+    license = "Apache-2.0"
+
+    [dependencies]
+    grafbase-sdk = "0.1.5"
```
Finistere, line 174: this will fail at the next release, you can use redactions like the other test

Finistere: well not even sure if you can use redactions here, but will fail

crates/engine/src/engine/auth_extension.rs
```diff
@@ -0,0 +1,46 @@
...
+pub struct AuthExtensionService {
+    authorizer_id: AuthorizerId,
+    extension_id: ExtensionId,
+}
+
+impl AuthExtensionService {
+    pub fn new(config: AuthConfig) -> Option<Self> {
+        let mut this = None;
+
+        for (i, provider) in config.providers.iter().enumerate() {
```
Finistere, line 20: seems like this could be expressed as a `.find()`

PR #2582, opened 2025-02-05
chore(gateway): Add extra test with gqlgen subscriptions

Finistere:


crates/integration-tests/src/federation/mod.rs
```diff
@@ -45,18 +46,21 @@ impl std::ops::Deref for MockSubgraph {
...
         match self {
             DockerSubgraph::Sse => "sse",
+            DockerSubgraph::Gqlgen => "gqlgen",
         }
     }
 
     pub fn url(&self) -> Url {
         match self {
             DockerSubgraph::Sse => Url::parse("http://localhost:4092/graphql").unwrap(),
+            DockerSubgraph::Gqlgen => Url::parse("http://localhost:8080/query").unwrap(),
```
tomhoule, line 63: This test setup is paying off

tomhoule (approved):
Troubling that everything seems to work on our end, but it doesn't in the setup that matters :/ This is going to be a useful test subgraph though, and it rules things out. Very nice!

crates/integration-tests/src/federation/mod.rs
Finistere: yeah

PR #2583, opened 2025-02-05
cli: new test suite for grafbase dev

tomhoule:
So we can test extensions in grafbase dev properly. This is from scratch, since all cli tests were deleted last year.

closes GB-8423


.gitignore
```diff
@@ -18,7 +18,7 @@ target
 # misc
 .DS_Store
 *.pem
-/.cargo
+.cargo/
```
tomhoule, line 21: For `cargo install` in the cli directory.

fbjork (approved):

cli/src/backend/dev/subgraphs.rs
```diff
@@ -111,7 +111,7 @@ pub async fn get_subgraph_sdls(
                     .or_else(|| remote_urls.get(&name).map(|url| url.as_str()))
                     .or(subgraph.introspection_url.as_ref().map(|url| url.as_str()))
                 else {
-                    return Err(BackendError::NoDefinedRouteToSubgraphSdl(name.clone()));
+                    return Err(BackendError::NoOverriddenSubgraphUrl(name.clone()));
```
tomhoule, line 114: Better error message, this is only if url is missing.

PR #2586, opened 2025-02-06
feat: test framework for extensions

pimeys:
Adds test utils to the grafbase-sdk crate. Examples how to use them are in the jwt extension integration tests. We now test all extensions in the CI.


Finistere (approved):

crates/grafbase-sdk/src/test/mock.rs
```diff
@@ -0,0 +1,62 @@
...
+
+use std::{net::SocketAddr, sync::Arc};
+
+pub use builder::DynamicSchemaBuilder;
+pub use server::MockGraphQlServer;
+use url::Url;
+
+/// A dynamic GraphQL schema that can be built and executed at runtime.
+#[derive(Debug)]
+pub struct DynamicSchema {
```
Finistere, line 14: Would be nice to have a dedicated crate for it that we could re-use for both integration-tests & extension

PR #2590, opened 2025-02-06
 cli: take linked extensions with "file://" url into account in grafbase dev

tomhoule:
closes GB-8428


cli/src/backend/dev/extensions.rs
```diff
@@ -0,0 +1,159 @@
...
+            }
+        }
+    }
+
+    tracing::info!(?extensions, "Detected extensions");
+
+    extensions
+}
+
+fn try_load_extension_from_file(file_path: &str, url: &str) -> Option<DetectedExtension> {
```
Finistere, line 47: You could just parse the string into a `url::Url` and use `url.to_file_path()`, see https://github.com/grafbase/grafbase/blob/main/crates/extension-catalog/src/load.rs#L13

Finistere: Actually `load_manifest` there is exactly what you need :)

Finistere (approved):

cli/src/backend/dev/extensions.rs
tomhoule: Good catch! Should have thought of this.

PR #2591, opened 2025-02-06
chore(engine): Refactor Schema::build

Finistere:
The current engine::Schema build had become too complex as the state
wasn't obvious anymore. It was tricky to understand which part of the
graph was ready or not at which stage which led to a lot of indirect
code, especially the FieldSet handling.

So simplified this with clearer stages and we don't modify the FederatedGraph anymore.
Now it's a lot easier to add the relevant extension directive code.

The logic is the same.


tomhoule (approved):
Tidy :100:

PR #2592, opened 2025-02-07
chore: split graphql mocks into its own crate

pimeys:
This mock setup will be very useful in our integration tests, and we do not need to maintain them in two places...


crates/grafbase-sdk/src/test/runner.rs
```diff
@@ -44,8 +44,9 @@ impl TestRunner {
         let mut subgraphs = Subgraphs::default();
 
         for subgraph in config.mock_subgraphs.drain(..) {
-            subgraphs.ingest_str(subgraph.sdl(), subgraph.name(), subgraph.url().as_str())?;
-            mock_subgraphs.push(subgraph.start());
+            let mock_graph = subgraph.start().await;
```
Finistere, line 47: just a nitpick, could be done in parallel

crates/grafbase-sdk/src/test.rs
```diff
@@ -6,9 +6,8 @@
 //! - Building and loading extensions
 
 mod config;
-mod mock;
 mod runner;
 
 pub use config::{TestConfig, TestConfigBuilder};
-pub use mock::{DynamicSchema, DynamicSubgraph};
+pub use grafbase_sdk_mock::{DynamicSchema, DynamicSubgraph};
```
Finistere, line 12: Not necessary now, but we might want to wrap `async_graphql::ServerError` or re-export it later

Finistere (approved):
Nice!

crates/grafbase-sdk/src/test.rs
Finistere: I guess you could just do `pub use grafbase_sdk_mock::*`

crates/grafbase-sdk/src/test/runner.rs
pimeys: But you need to start the subgraph to get its address/port first, and you need that to ingest the subgraph. Kind of not that easy to parallelize, because you mutate the subgraphs...

crates/grafbase-sdk/src/test.rs
pimeys: Using a star in a pub use is not cool man...

PR #2607, opened 2025-02-11
chore(engine): Validate extension directive location

Finistere:
Validate that directives are applied at their correct location.


pimeys (approved):
awesome

tomhoule (approved):

PR #2612, opened 2025-02-11
feat(engine): parse, validate and inject InputValueSet

Finistere:
Adding support for `InputValueSet` which can be used with:

```graphql
extend schema @link(url: "https://specs.grafbase.com/grafbase", import: ["InputValueSet"])

directive @echo(input: InputValueSet!) on FIELD_DEFINITION
```

It's an arbitrary URL for now, our only constraint is that it should follow `@link` conventions, so supposing we want to use the namepsace `grafbase`, the path must end with `grafbase`.

It can be used to select arbitrary data from the arguments:
```graphql
type Query {
    echo(first: Int, limit: Int, after: String, filters: Filters): JSON @echo(input: "first limit filters { latest nested { id } }")
}
```

There are also two extra behaviors compared to usual field selection sets:
- `@echo(input: "*")` will inject all the arguments
- `@echo(input: "filters")` will inject the whole filters input object, not providing a subselection means providing everything.

I've also changed hooks & extensions code in runtime-local to use the same Context object. I've adjusted the extension runtime trait to work similarly to engine resolvers, they get access to the response data only for the initialization, the lifetime prohibits keeping it across a .await


Finistere:
I've also changed hooks & extensions code in runtime-local to use the same `Context` object. I've adjusted the extension runtime trait to work similarly to engine resolvers, they get access to the response data only for the initialization, the lifetime prohibits keeping it across a `.await`

tomhoule (approved):
LGTM

PR #2629, opened 2025-02-13
feat(engine): Add JsonTempalte & UrlTemplate support

Finistere:
I've dropped `Template` for now, not sure if that's useful. I'm doing a bunch of logic to handle objects/list and escaping to hopefully make it easy to use.

Only `args` is accessible today with access to the arguments.

Adding more tests in a further PR.


crates/integration-tests/tests/federation/extensions/injection/template/json.rs
```diff
@@ -0,0 +1,248 @@
...
+                "name": "Alice"
+              }
+            ]
+          }
+        }
+        "#);
+    });
+}
+
+#[ignore]
```
Finistere, line 217: Broken for now...

PR #2635, opened 2025-02-14
chore(engine): Add a bit more template tests

Finistere:


crates/engine/src/prepare/cached/query_plan/field/extension.rs
```diff
@@ -121,17 +127,28 @@ fn urlencode(s: &str) -> impl std::fmt::Display + '_ {
...
         match self.value.as_ref() {
-            serde_json::Value::Null => false,
-            serde_json::Value::Bool(b) => *b,
-            serde_json::Value::Number(n) => n.as_f64().map(|f| f <= f64::EPSILON).unwrap_or_default(),
-            serde_json::Value::String(s) => s.is_empty(),
-            serde_json::Value::Array(v) => v.is_empty(),
-            serde_json::Value::Object(o) => o.is_empty(),
+            serde_json::Value::Array(v) => !v.is_empty(),
+            serde_json::Value::Object(o) => !o.is_empty(),
+            _ => true,
```
pimeys, line 132: Wait, so null is truthy?

pimeys (approved):

crates/engine/src/prepare/cached/query_plan/field/extension.rs
Finistere: wait you're right, the truthy here is ambiguous. The goal of that function is to determine whether to render or not a section like `{{#post}}{{/post}}`. And just by writing it I realized it's the opposite, anything but non-empty object/arrays is false. So it's more like `is_container` or something like this.

Finistere: thx

PR #2643, opened 2025-02-16
Snowflake extension WIP

fbjork:


extensions/snowflake/Cargo.toml
```diff
@@ -0,0 +1,30 @@
...
+strip = true
+lto = true
+codegen-units = 1
+
+[dependencies]
+grafbase-sdk = { path = "../../crates/grafbase-sdk" }
+serde = { version = "1.0", features = ["derive"] }
+serde_json = "1.0"
+snowflake-api = "0.6"
+tokio = { version = "1", features = ["rt"] }
```
pimeys, line 21: This will not work in an extension without mio support.

extensions/snowflake/src/lib.rs
```diff
@@ -0,0 +1,143 @@
...
+}
+
+impl Extension for SnowflakeExtension {
+    fn new(schema_directives: Vec<Directive>, config: Configuration) -> Result<Self, Box<dyn std::error::Error>> {
+        let config: SnowflakeConfig = config.deserialize()?;
+        info!("Initializing Snowflake extension with account: {}", config.account);
+
+        let manager = SnowflakeManager { config };
+        let pool = Pool::builder(manager)
+            .max_size(10)
```
pimeys, line 84: Why ten connections for a wasm component that is single-threaded?

extensions/snowflake/src/lib.rs
```diff
@@ -0,0 +1,143 @@
...
+
+        // Get a connection from the pool
+        let client = self.pool.get().map_err(|e| Error {
+            message: format!("Failed to get connection from pool: {}", e),
+            extensions: vec![],
+        })?;
+
+        // Replace parameter placeholders with actual values
+        let sql = if let Some(params) = args.params {
+            let mut sql = args.sql;
```
pimeys, line 113: Are you sure the abstraction for just passing an SQL string is what we want here?

extensions/snowflake/src/lib.rs
```diff
@@ -0,0 +1,143 @@
...
+        _context: SharedContext,
+        directive: Directive,
+        field_definition: FieldDefinition,
+        inputs: FieldInputs,
+    ) -> Result<FieldOutput, Error> {
+        let args: QueryArgs = directive.arguments()?;
+        debug!("Executing Snowflake query: {}", args.sql);
+
+        // Get a connection from the pool
+        let client = self.pool.get().map_err(|e| Error {
```
pimeys, line 106: This is an async call, but you do not await the result. How are you implementing the async wasm component here?

extensions/snowflake/src/lib.rs
```diff
@@ -0,0 +1,143 @@
+use std::sync::Arc;
+use deadpool::managed::{Manager, Pool, PoolError};
+use grafbase_sdk::{
+    types::{Configuration, Directive, FieldDefinition, FieldInputs, FieldOutput},
+    Error, Extension, Resolver, ResolverExtension, SharedContext,
+};
+use serde::{Deserialize, Serialize};
+use snowflake_api::{Client as SnowflakeClient, ClientBuilder, QueryResult as SnowflakeQueryResult};
```
pimeys, line 8: This crate uses tokio and reqwest as their backend. Neither of them will work in wasip2 right now without fixing mio.

You should fork the crate and make it work together with the sync http client we have in our wit. Interestingly cursor did not think that as a valid solution.

extensions/snowflake/tests/integration_tests.rs
```diff
@@ -0,0 +1,44 @@
+use grafbase_sdk::test::{DynamicSchema, TestConfig, TestRunner};
+use indoc::indoc;
+
+#[tokio::test]
+async fn test_snowflake_query() {
```
pimeys, line 5: Did this test actually run :D

pimeys (changes requested):

extensions/snowflake/Cargo.toml
```diff
@@ -0,0 +1,30 @@
...
+deadpool = "0.9"
```
pimeys, line 22: Why do we need deadpool here? The snowflake client is using http, which has pooling already in every implementation.

PR #2671, opened 2025-02-19
extension: add fields used in registry to Manifest

tomhoule:


Finistere (approved):

crates/extension-catalog/src/load.rs
```diff
@@ -47,6 +47,11 @@ mod tests {
             sdk_version: "0.3.0".parse().unwrap(),
             minimum_gateway_version: "0.90.0".parse().unwrap(),
             sdl: Some("directive foo on SCHEMA".to_string()),
+            license: None,
```
pimeys, line 50: Could we also have these in the init and build commands already?

pimeys (approved):

crates/extension-catalog/src/load.rs
tomhoule: Done

PR #2692, opened 2025-02-20
feat: nats extension publish

pimeys:


crates/federated-server/src/server/gateway.rs
```diff
@@ -139,6 +139,11 @@ fn create_wasi_extension_configs(
             ExtensionType::Resolver => {
                 let id = ExtensionPoolId::Resolver(id);
 
+                let extension_config = match extension_config.config() {
+                    Some(config) => minicbor_serde::to_vec(config).unwrap(),
+                    None => Vec::new(),
```
tomhoule, line 144: Not important, but the extension config will always be an object, right? Would it make more sense to use `minicbor_serde::to_vec(&HashMap::new())` instead, for ergonomics?

tomhoule (approved):

PR #2709, opened 2025-02-21
feat: Allow URL/http method modifications of the subgraph request in hooks

Finistere:
Changed the `on_subgraph_request` hook to receive a `subgraph-request` resource which allows modifications of the url & method.

We should also definitely use bindgen macro for the host part later: https://docs.rs/wasmtime/latest/wasmtime/component/bindgen_examples/_6_exported_resources/index.html


tomhoule (approved):
Neat!

PR #2712, opened 2025-02-21
feat: nats extension with subscribe directive

pimeys:
We need to dedup the streams so we don't eat all the RAM with duplicate instances. That will come in a follow-up pull request.

https://linear.app/grafbase/issue/GB-8574/subscription-dedup


crates/engine/src/prepare/cached/extension/mod.rs
```diff
@@ -12,38 +12,41 @@ use crate::response::ResponseObjectsView;
...
     field_arguments: PartitionFieldArguments<'a>,
     variables: &'a Variables,
 }
 
-pub(crate) fn create_extension_directive_arguments_view<'ctx, 'resp>(
+pub(crate) fn create_extension_directive_arguments_view<'ctx>(
     schema: &'ctx Schema,
     directive: ExtensionDirective<'ctx>,
-    field_arguments: PartitionFieldArguments<'ctx>,
+    field_argumeents: PartitionFieldArguments<'ctx>,
```
Finistere, line 24: typo `arguments`

crates/grafbase-sdk/src/extension/resolver.rs
```diff
@@ -70,4 +70,30 @@ pub trait Resolver: Extension {
...
+    ) -> Result<(), Error>;
+
+    /// Retrieves the next item in a subscription stream.
+    ///
+    /// This method is called repeatedly to get sequential items from an active subscription.
+    ///
+    /// # Returns
+    ///
+    /// Returns a `Result` containing either the next `FieldOutput` value in the subscription stream or an `Error`
+    fn resolve_next_subscription_item(&mut self) -> Result<Option<FieldOutput>, Error>;
```
Finistere, line 98: I would have used an associated type like:
```rust
trait Extension {
  type Subscription<'a>: Subscription;
  async fn resolve_subscription(&mut self) -> Self::Subscription<'_>;
}

trait Subscription {
  fn next(&mut self) -> Option<...>;
}
```

crates/grafbase-sdk/src/jq_selection.rs
```diff
@@ -0,0 +1,137 @@
...
+    pub fn select(
+        &mut self,
+        selection: &str,
+        data: serde_json::Value,
+    ) -> anyhow::Result<impl Iterator<Item = anyhow::Result<serde_json::Value>> + '_> {
+        let hasher = DefaultHashBuilder::default();
+        let hasher = |val: &_| hasher.hash_one(val);
+        let hash = hasher(selection);
+
+        if !self.selection_cache.find(hash, |_| true).is_some() {
```
Finistere, line 87: the function cannot be `|_| true` it must compare two selections strings. Here you're basically saying that hash collisions don't exist :)

crates/grafbase-sdk/src/jq_selection.rs
```diff
@@ -0,0 +1,137 @@
...
+                anyhow::anyhow!("The selection is not valid jq syntax: `{error}`")
+            })?;
+
+        self.filters.push(filter);
+
+        let index = self.filters.len() - 1;
+        let hasher = DefaultHashBuilder::default();
+        let hasher = |val: &_| hasher.hash_one(val);
+
+        self.selection_cache.insert_unique(hash, index, hasher);
```
Finistere, line 133: Here again you're supposing hash collision don't exist, you must use something like `entry` with `eq` function if you can't guarantee the absence of hash collisions.

Finistere (approved):
nice work

crates/runtime/src/extension.rs
```diff
@@ -40,6 +41,14 @@ pub trait ExtensionRuntime: Send + Sync + 'static {
     where
         'ctx: 'f;
 
+    fn resolve_field_subscription<'ctx, 'f>(
+        &'ctx self,
+        context: &'ctx Self::SharedContext,
+        directive: ExtensionFieldDirective<'ctx, impl Anything<'ctx>>,
+    ) -> impl Future<Output = Result<mpsc::Receiver<Result<Data, PartialGraphqlError>>, PartialGraphqlError>> + Send + 'f
```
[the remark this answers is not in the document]
pimeys: It might be easy to do the deduplication with a channel instead of a stream in the end, we'll see tomorrow.

crates/grafbase-sdk/src/extension/resolver.rs
pimeys: Maybe after the MVP.

crates/runtime/src/extension.rs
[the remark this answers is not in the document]
pimeys: Let's see that in the next PR when the deduplication actually happens.

PR #2724, opened 2025-02-22
chore(extension): Simplify host part

Finistere:
Not obvious when looking at the line count, but this PR makes it _a lot_ easier to write the host part of Wasm. All the linking and types are generated and we just need to implement traits.

IMHO, the hooks should later be part of the sdk wit, just with a different world so that we can re-use the types more easily.

I've also moved everything related to extension from runtime-local to wasi-component-loader, now all the WIT types are kept as implementation details. IMHO that makes it clearer what's really expected of the wasi-component-loader.


pimeys (approved):
😎

PR #2732, opened 2025-02-24
chore: Use federated-server in integration-tests

Finistere:
Something we should have done ages ago. Would have made CORS/rate-limit
tests a lot easier to write and maintain.

Just got annoyed at it, so did it.


tomhoule (approved):
very nice!

PR #2736, opened 2025-02-25
feat: after-merge fixes to the subscribe api

pimeys:
The user-facing API for subscriptions is now just one method. We will handle polling behind the scenes.

The host side returns stream instead of a channel.

### DIFF
--- a/crates/grafbase-sdk/src/host_io/pubsub.rs
+++ b/crates/grafbase-sdk/src/host_io/pubsub.rs
@@ -8,11 +8,18 @@ use crate::{types::FieldOutput, Error};
 
 pub mod nats;
 
-/// A trait for subscribing to a stream of field outputs.
-pub trait Subscriber {
-    /// Gets the next field output from the stream.
+/// A trait for consuming field outputs from streams.
+///
+/// This trait provides an abstraction over different implementations
+/// of subscriptions to field output streams. Implementors should handle
+/// the details of their specific transport mechanism while providing a
+/// consistent interface for consumers.
+pub trait Subscription {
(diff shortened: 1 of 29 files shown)


crates/grafbase-sdk/src/host_io/pubsub.rs
```diff
@@ -0,0 +1,18 @@
...
+//! This module defines the core `Subscriber` trait that abstracts over different
+//! implementations of subscribing to field output streams, allowing extensions to
+//! handle field outputs in a transport-agnostic way.
+
+use crate::{types::FieldOutput, Error};
+
+pub mod nats;
+
+/// A trait for subscribing to a stream of field outputs.
+pub trait Subscriber {
```
Finistere, line 12: Would have named it `Subscription`, subscriber for me is the client part.

extensions/nats/src/subscriber.rs
```diff
@@ -0,0 +1,69 @@
...
+    host_io::pubsub::{
+        nats::{self, NatsSubscriber},
+        Subscriber,
+    },
+    jq_selection::JqSelection,
+    types::FieldOutput,
+    Error,
+};
+
+pub struct FilteringSubscriber {
```
Finistere, line 13: would have named it `FilterSubscription` or `FilteredSubscription` IMHO

Finistere (approved):
nice

crates/grafbase-sdk/src/host_io/pubsub.rs
tomhoule: I agree

tomhoule (approved):

PR #2737, opened 2025-02-25
feat: nats jetstream subscriptions

pimeys:
The difference between subscriptions and streams:

- streams are persistent
- they have to exist before connecting to them
- they can hold many subjects
- they have different consuming strategies
- they use a different client than nats core


Finistere (approved):

pimeys:
GB-8301

PR #2739, opened 2025-02-25
chore(extension): Easier extension trait definition

Finistere:
Removed a bunch of duplicated boilerplate code and the extension type. All extension state is managed by the same code in `grafbase_sdk::component` with the other implementation details of the component <-> extension mapping. Had to add an extra `AnyExtension` trait, but still feels simpler IMHO.

Also including the hooks v0.3.0 release, the CI isn't super happy to have only this for some reason.

I'm also making the grafbase cli more robust, it's an absolute pain today with breaking changes. So changing the generated `Cargo.toml` to use the local `grafbase-sdk`.

### DIFF
--- a/crates/grafbase-sdk/src/extension/resolver.rs
+++ b/crates/grafbase-sdk/src/extension/resolver.rs
@@ -1,66 +1,10 @@
 use crate::{
-    host_io::pubsub::Subscription,
-    types::{Configuration, Directive, FieldDefinition, FieldInputs, FieldOutput},
+    component::AnyExtension,
+    types::{Directive, FieldDefinition, FieldInputs, FieldOutput},
     wit::{Error, SharedContext},
 };
 
 use super::Extension;
-
-type InitFn = Box<dyn Fn(Vec<Directive>, Configuration) -> Result<Box<dyn Resolver>, Box<dyn std::error::Error>>>;
-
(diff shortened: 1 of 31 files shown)


tomhoule (approved):
feels cleaner

crates/grafbase-sdk/src/extension/resolver.rs
```diff
@@ -107,3 +51,48 @@ pub trait Resolver: Extension {
...
+            definition: FieldDefinition,
+        ) -> Result<Box<dyn Subscription>, Error> {
+            Resolver::resolve_subscription(&mut self.0, context, directive, definition)
+        }
+    }
+    crate::component::register_extension(Box::new(|schema_directives, config| {
+        <T as Extension>::new(schema_directives, config)
+            .map(|extension| Box::new(Proxy(extension)) as Box<dyn AnyExtension>)
+    }))
+}
```
pimeys, line 98: what's the point of these?

pimeys (approved):

crates/grafbase-sdk/src/extension/resolver.rs
Finistere: I can't do `impl<T: Resolver> AnyExtension for T` because one can implement both `Resolver` and `Authenticator`, so have no choice but to create a proxy type that implements `AnyExtension`

PR #2741, opened 2025-02-26
fix: utilize unfold with extension subscriptions

pimeys:
We avoid taking the instance out from the pool, and can reuse it for the next request.


tomhoule (approved):
neat!

PR #2749, opened 2025-02-27
fix: build cli and gateway always in extensions run

pimeys:
We should make sure the binaries are built always if we run extension tests.

### DIFF
--- a/.github/workflows/rust-prs.yml
+++ b/.github/workflows/rust-prs.yml
@@ -379,7 +379,7 @@ jobs:
           tool: nextest
 
       - name: Build CLI and Gateway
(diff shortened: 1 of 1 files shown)


.github/workflows/rust-prs.yml
```diff
@@ -379,7 +379,6 @@ jobs:
           tool: nextest
 
       - name: Build CLI and Gateway
-        if: needs.what-changed.outputs.cargo-bin-specs
```
tomhoule, line 382: Please let's comment this out instead, we have to remember restore it once extensions move out of this repo.

fbjork (approved):

PR #2753, opened 2025-02-27
chore: add docs, fix cbor ser

pimeys:
This starts the docs effort for NATS, but does not document anything NATS related yet. Preliminary things I found when testing:

- Minicbor is not serializing nulls correctly. We have an open PR but we need to vendor this code until it is merged
- Simplify NATS return types for static return values like booleans or integers
- document REST directives

### DIFF
--- a/cli/tests/extension/mod.rs
+++ b/cli/tests/extension/mod.rs
@@ -37,12 +37,12 @@ fn init_resolver() {
     codegen-units = 1
 
     [dependencies]
-    grafbase-sdk = "0.5.2"
+    grafbase-sdk = "0.5.4"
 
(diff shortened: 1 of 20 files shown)


cli/tests/extension/mod.rs
```diff
@@ -37,12 +37,12 @@ fn init_resolver() {
     codegen-units = 1
 
     [dependencies]
-    grafbase-sdk = "0.5.2"
+    grafbase-sdk = "0.5.4"
```
pimeys, line 40: Did two versions while on this PR :D

Finistere (approved):

PR #2757, opened 2025-02-27
feat(sdk): Add authorization

Finistere:
Adding the relevant structs for authorization extensions and reworking a bit SDK to hide more the wit.

I treat the WIT as raw data with indices like we do in engine and most other places. Extensions receive walkers instead of wit structs. This makes a nicer API as it's not behind a macro and allows us to keep extra metadata around. Typically for authorization results, instead of exposing a push method to add the decision in the right order,  I'm relying on the index of the walker to identify the authorized element. Making it a lot easier to work with IMHO.

I've wrapped the internal errors as `SdkError` and extension code uses an `Error` type wrapping the wit one, similar to `ErrorResponse` works already.

I've changed the error extensions to be CBOR serialized instead of JSON as they're now not directly accessible anyway.


crates/grafbase-sdk/wit/world.wit
```diff
@@ -255,33 +247,89 @@ interface types {
...
+
+    record union-directive-site {
+        union-name: string,
+        arguments: list<u8>
+    }
+
+    record interface-directive-site {
+        interface-name: string,
+        arguments: list<u8>
+    }
```
pimeys, line 283: Please add some docs to all of these. Even if they're internal, but for us.

crates/grafbase-sdk/wit/world.wit
```diff
@@ -255,33 +247,89 @@ interface types {
...
+        query-to-resp: list<tuple<u32, u32, u32>>,
+        // Represents an element of the response associated with the directive, the data is the part of the directive arguments that relied on response data.
+        elements: list<list<u8>>
+    }
+
+    record authorization-decisions {
+        granted-bitset: list<u8>,
+        element-ix-to-error-ix: list<tuple<u32, u32>>,
+        errors: list<error>
+    }
```
pimeys, line 305: Docs.

crates/grafbase-sdk/src/types/elements.rs
```diff
@@ -0,0 +1,174 @@
+use serde::Deserialize;
+
+use crate::{
+    types::DirectiveSite,
+    wit::{self},
+    SdkError,
+};
+
+/// A list of elements present in the query on which one of the extension's directive was applied on their definition.
```
pimeys, line 9: You should elaborate a bit more on this. Maybe even with an example.

crates/grafbase-sdk/src/types/elements.rs
```diff
@@ -0,0 +1,174 @@
...
+                                ix: resp_start + i as u32,
+                            });
+                        (query_element, resp_items)
+                    }),
+            )
+        })
+    }
+}
+
+/// A response element containing a directive site and associated items
```
pimeys, line 159: An example here would be helpful.

crates/grafbase-sdk/src/types/elements.rs
```diff
@@ -0,0 +1,174 @@
...
+#[derive(Clone, Copy)]
+pub struct ResponseElement<'a> {
+    data: &'a [u8],
+    pub(super) ix: u32,
+}
+
+impl<'a> ResponseElement<'a> {
+    /// Get the response items
```
pimeys, line 167: Is the plural correct here? Item or items?

crates/grafbase-sdk/src/types/elements.rs
pimeys: Like write it like you would have a sock puppet in your hand, explaining this to me.

pimeys (approved):

crates/grafbase-sdk/src/types/elements.rs
```diff
@@ -0,0 +1,174 @@
...
+        Item = (
+            &'a str,
+            impl Iterator<
+                    Item = (
+                        QueryElement<'a>,
+                        impl ExactSizeIterator<Item = ResponseElement<'a>> + 'a,
+                    ),
+                > + 'a,
+        ),
+    > + 'a {
```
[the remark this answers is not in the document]
Finistere: Can't make aliases with `impl`, could make a trait but not really easier

PR #2761, opened 2025-02-27
fix(extension): Pro-actively mark instances as poisoned to handle future cancelation

Finistere:


crates/wasi-component-loader/src/extension/instance.rs
```diff
@@ -33,22 +33,21 @@ impl ExtensionInstance {
         directive: wit::FieldDefinitionDirective<'_>,
         inputs: InputList,
     ) -> crate::Result<FieldOutput> {
+        // Futures may be canceled, so we pro-actively mark the instance as poisoned until proven
+        // otherwise.
+        self.poisoned = true;
+
```
pimeys, line 39: Would be great to have some helper to do this and not DRY so much. But it might be annoying to write, so...

pimeys (approved):
Important thing to have. Nice.

PR #2763, opened 2025-02-28
cli: add grafbase extension publish

tomhoule:
closes GB-8582

### DIFF
--- a/cli/src/extension/publish.rs
+++ b/cli/src/extension/publish.rs
@@ -1,0 +1,55 @@
+use crate::{backend::api, cli_input::ExtensionPublishCommand};
+use extension::VersionedManifest;
+use std::fs;
+
+use super::EXTENSION_WASM_MODULE_FILE_NAME;
+
+#[tokio::main]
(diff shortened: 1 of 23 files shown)


cli/src/extension/publish.rs
```diff
@@ -0,0 +1,55 @@
...
+    if !wasm_blob_path.exists() {
+        return Err(anyhow::anyhow!(
+            "Failed to find extension WASM module at `{}`",
+            wasm_blob_path.display()
+        ));
+    }
+
+    match api::extension_publish::extension_publish(manifest, &wasm_blob_path).await? {
+        api::extension_publish::ExtensionPublishOutcome::Success { name, version } => {
+            println!("🌟 Extension `{name}@{version}` published successfully");
```
pimeys, line 41: Aren't we usually doing these cheerful lines of text from the report module?

pimeys (approved):
aw esome

cli/src/extension/publish.rs
tomhoule: Yeah, shouldn't have broken the convention for expediency. Will move these over next PR. We should revisit the report module later.

PR #2765, opened 2025-02-28
feat: extension permission management

pimeys:
I noticed a few issues when testing:

- The host network connections were allowed even if we set networking to false, only the guest network was sandboxed. Now we check the setting in the host.
- The extension should define the default permissions it needs, and the user will respect those if using the normal `rest = "0.1"` syntax. The user can manually change these permissions, and the user config will override the defaults.


cli/src/extension/build.rs
```diff
@@ -262,9 +294,10 @@ fn parse_manifest(source_dir: &Path, wasm_path: &Path) -> anyhow::Result<Manifes
         homepage_url: extension_toml.extension.homepage_url,
         repository_url: extension_toml.extension.repository_url,
         license: extension_toml.extension.license,
+        permissions,
     };
 
-    Ok(manifest)
+    Ok(dbg!(manifest))
```
tomhoule, line 300: stray dbg

crates/extension/src/manifest/v1.rs
```diff
@@ -62,6 +85,47 @@ mod tests {
...
+                "FieldResolver": {
+                    "resolver_directives": ["custom"]
+                }
+            },
+            "sdl": "directive @custom on FIELD_DEFINITION",
+            "sdk_version": "0.1.0",
+            "minimum_gateway_version": "0.1.0",
+            "description": "Mandatory description",
+            "homepage_url": "http://example.com/my-extension",
+            "permissions": ["stdout", "stderr", "environment_variables", "network"]
```
tomhoule, line 102: I'm wondering, will we need parameters inside some of these in the future, like "only these env vars", or "only this domain, network wise"

tomhoule: We can worry about that in manifest v2 maybe

tomhoule (approved):

crates/extension/src/manifest/v1.rs
pimeys: Maybe, we don't know yet.

PR #2769, opened 2025-02-28
gateway: extension registry integration 

tomhoule:
closes GB-8495


cli/src/extension/install.rs
```diff
@@ -51,49 +49,3 @@ pub(super) async fn execute() -> anyhow::Result<()> {
 
     Ok(())
 }
-
-async fn download_extension(
```
tomhoule, line 55: Moved to the extension_catalog crate

crates/extension-catalog/src/load.rs
```diff
@@ -22,9 +27,116 @@ pub async fn load_manifest(mut url: Url) -> Result<Manifest, String> {
             .await
             .map_err(|err| err.to_string())?
     };
+
     Ok(manifest.into_latest())
 }
 
+pub async fn download_extension_from_registry_if_needed(
```
tomhoule, line 34: Could be cleaner, but there's time pressure right now

crates/federated-server/src/server/gateway/create_extension_catalog.rs
```diff
@@ -0,0 +1,368 @@
...
+                name: "test_one",
+                version: Version {
+                    major: 0,
+                    minor: 1,
+                    patch: 0,
+                },
+            },
+        ]
+        "#);
+    }
```
tomhoule, line 367: We need more tests, but time is too short.

PR #2770, opened 2025-02-28
chore(extension): Host wasi part of authorization

Finistere:
Had to separate the directive related stuff to a separate interface. Having the arguments as a `&[u8]` is just painful to manage and we don't want `String` for the names. Bindgen can't mix owned/borrowed versions, so had to write them myself.


crates/runtime/src/extension.rs
```diff
@@ -56,8 +68,23 @@ pub trait ExtensionRuntime: Send + Sync + 'static {
...
+        context: &'ctx Self::SharedContext,
+        extension_id: ExtensionId,
+        // (directive name, (definition, arguments))
+        elements: impl IntoIterator<
+            Item = (
+                &'ctx str,
+                impl IntoIterator<Item = DirectiveSite<'ctx, impl Anything<'ctx>>> + Send + 'ctx,
+            ),
+        > + Send
+        + 'ctx,
```
pimeys, line 83: This would look better as a generic parameter, so the long type signature is not polluting the param type defs.

crates/wasi-component-loader/src/extension/runtime.rs
```diff
@@ -96,56 +112,35 @@ impl ExtensionRuntime for ExtensionsWasiRuntime {
...
-            let mut instance = pool.get().await;
-
-            let arguments = minicbor_serde::to_vec(arguments).unwrap();
             let directive = wit::FieldDefinitionDirective {
                 name,
                 site: wit::FieldDefinitionDirectiveSite {
                     parent_type_name: field.parent_entity().name(),
                     field_name: field.name(),
-                    arguments: &arguments,
+                    arguments: minicbor_serde::to_vec(arguments).unwrap(),
```
pimeys, line 122: You should not use `minicbor_serde::to_vec` anymore. You did the fix in the crate for null values, so we now have our own `to_vec` that uses the new setting:

https://github.com/grafbase/grafbase/blob/main/crates/wasi-component-loader/src/cbor.rs

crates/wasi-component-loader/src/extension/runtime.rs
```diff
@@ -216,81 +176,131 @@ impl ExtensionRuntime for ExtensionsWasiRuntime {
...
+    ) -> Result<AuthorizationDecisions, ErrorResponse> {
+        let mut instance = self.get(ExtensionPoolId::Authorization(extension_id)).await?;
+        let mut directive_names = Vec::<(&str, u32, u32)>::new();
+        let mut query_elements = Vec::new();
+        for (name, sites) in elements {
+            let start = query_elements.len();
+            for site in sites {
+                // Some help for rust-analyzer who struggles for some reason.
+                let site: DirectiveSite<'_, _> = site;
+                let arguments = minicbor_serde::to_vec(site.arguments).unwrap();
```
pimeys, line 253: Same here, use

https://github.com/grafbase/grafbase/blob/main/crates/wasi-component-loader/src/cbor.rs

crates/wasi-component-loader/src/tests/extensions.rs
```diff
@@ -56,7 +56,7 @@ async fn simple_resolver() {
         site: wit::FieldDefinitionDirectiveSite {
             parent_type_name: "Query",
             field_name: "cats",
-            arguments: &minicbor_serde::to_vec(&FieldArgs { name: "cat" }).unwrap(),
+            arguments: minicbor_serde::to_vec(&FieldArgs { name: "cat" }).unwrap(),
```
pimeys, line 59: https://github.com/grafbase/grafbase/blob/main/crates/wasi-component-loader/src/cbor.rs

pimeys (approved):
A few fixes to be made, then plz merge.

crates/wasi-component-loader/src/extension/runtime.rs
Finistere: Good point, it should never happen in practice since I'm not using serde_json::Value but we can't know that here.

crates/runtime/src/extension.rs
Finistere: well ok, will change in another pr

PR #2779, opened 2025-03-03
fix: subgraph url can be nullable

pimeys:
Add --virtual to the call to define the subgraph as virtual, e.g. to be managed by an extension.

### DIFF
--- a/cli/src/cli_input/publish.rs
+++ b/cli/src/cli_input/publish.rs
@@ -1,5 +1,5 @@
 use super::FullGraphRef;
-use clap::Parser;
+use clap::{Args, Parser};
 use url::Url;
 
 /// Publish a subgraph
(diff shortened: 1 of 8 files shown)


cli/src/cli_input/publish.rs
```diff
@@ -17,11 +17,23 @@ pub struct PublishCommand {
...
+#[group(required = true, multiple = false)]
+pub(crate) struct SubgraphSource {
+    /// The URL to the GraphQL endpoint. Can be omitted if the subgraph is virtual and completely defined
+    /// by an extension.
+    #[arg(long)]
+    pub(crate) url: Option<Url>,
+
+    /// Subgraph does not exist, but is handled by an extension.
+    #[arg(long)]
+    pub(crate) r#virtual: bool,
```
tomhoule, line 38: We need a `if source.url.is_none() && !source.virtual { return Error(...) }` to enforce that

pimeys: No we do not, look a bit up and you see we do it with clap. It's a group. You have to have one of them defined, but not both.

PR #2781, opened 2025-03-03
Parameterize the username used to login to the helm registry

cristifalcas:
Parameterize the username used to login to the helm registry.

### DIFF
--- a/.github/workflows/build-gateway-chart.yml
+++ b/.github/workflows/build-gateway-chart.yml
@@ -20,18 +20,8 @@ env:
   CHART_PATH: gateway/helm
 
 jobs:
-  clone-repo:
-    name: Clone Repository
-    runs-on: ubicloud-standard-2
-    outputs:
-      build: ${{ steps.filter.outputs.build }}
-    steps:
-      - name: Checkout Repository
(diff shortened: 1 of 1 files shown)


.github/workflows/build-gateway-chart.yml
```diff
@@ -16,11 +16,22 @@ concurrency:
 
 env:
   REGISTRY_REPOSITORY: ghcr.io/grafbase/helm-charts
+  REGISTRY_USERNAME: grafbase
   CHART_PATH: gateway/helm
 
 jobs:
+  clone-repo:
+    name: Clone Repository
```
fbjork, line 24: why is this needed?

cristifalcas: This clones the current repository in order to be accessible to the commands we want to run

fbjork: We do it here already no? https://github.com/grafbase/grafbase/pull/2781/files#diff-e9fa7d2d7c793a00500d3f196767edba4a5b0124d747ddc714dd45b442243914L43

fbjork (approved):

PR #2790, opened 2025-03-04
 feat(engine): add authorization extension support in the engine

Finistere:


crates/engine/src/prepare/cached/builder/mod.rs
```diff
@@ -291,221 +292,6 @@ impl<'a> Solver<'a> {
             })
     }
 
-    fn populate_modifiers_after_partition_generation(&mut self) -> SolveResult<()> {
```
Finistere, line 294: Moved to separate file

PR #2794, opened 2025-03-04
improve gateway resilience

cristifalcas:
improve gateway resilience


gateway/helm/values.yaml
```diff
@@ -5,7 +5,7 @@ replicaCount: 1
 image:
   # Kubernetes image pull policy. Options: Always, IfNotPresent, Never.
   # https://kubernetes.io/docs/concepts/containers/images/#image-pull-policy
-  pullPolicy: IfNotPresent
+  pullPolicy: Always
```
fbjork, line 8: Why?

cristifalcas: it's better if the image tag is `latest`. Without `Always`, it will use whatever container is already pulled, which may not be the "latest".

gateway/helm/templates/deployment.yaml
```diff
@@ -89,6 +87,12 @@ spec:
             - {{- . | toYaml | trim | nindent 14 }}
             {{- end }}
             {{- end }}
+          lifecycle:
+            preStop:
+              exec:
+                command:
+                  - sleep
+                  - 15
```
fbjork, line 95: Not a huge fan of sleeping to fix the underlying cause. Let's understand the reason for the timeouts before we apply this strategy.

PR #2802, opened 2025-03-04
chore(engine): Add authorization extension tests

Finistere:


crates/engine/codegen/domain/schema.graphql
```diff
@@ -26,8 +26,6 @@ union DirectiveSite @id @meta(module: "directive_site") @variants(remove_suffix:
   | InputObjectDefinition
   | ScalarDefinition
   | FieldDefinition
-  | InputValueDefinition
-  | EnumValue
```
tomhoule, line 30: Removed because unused for now?

crates/engine/src/prepare/operation_plan/query_modifications.rs
```diff
@@ -195,6 +195,7 @@ where
                 let offset = self.modifications.errors.len();
                 self.modifications.errors.extend(errors.into_iter().map(Into::into));
                 for (element_ix, error_ix) in element_to_error {
+                    println!("{} {}", modifier_ids.len(), element_ix);
```
tomhoule, line 198: intentional?

tomhoule (approved):
Nice!

crates/engine/src/prepare/operation_plan/query_modifications.rs
Finistere: ah right thx

PR #2806, opened 2025-03-05
chore: install aws cli manually on release

pimeys:
This broke randomly in the last few days. Might be that our github runner has now too old aws, without --checksum-algorithm available. Oh well, let's try to install it manually!

PR #2807, opened 2025-03-05
Install latest version of AWS CLI in release workflows

tomhoule:
There has been a regression in the github actions default image version. We need the algorithm argument to publish to R2.


tomhoule:
Debugging successful, merging the fix in https://github.com/grafbase/grafbase/pull/2806

PR #2813, opened 2025-03-05
chore(engine): Ground work for requiresScopes/authenticated extensions

Finistere:
Again PR that does too much, just takes so much time to iterate on SDK. The goal was to have `@requiresScopes`-like extension in integration-tests to iterate on the types and check the engine parts work.

- the `kind` in the extension.toml for authentication is now `authentication` rather than `auth` which will be confusing with `authorization.
- I've renamed `FIeldResolver` to `Resolver` at various places, but maintained backwards compatibility in the manifest.
- I made the `directives.definitions` path optional, we use `definitions.graphql` by default if it exists. As `field_resolver` are also optional now, removed them from the init template. Not sure if we should even keep them now to be honest.
- Renamed `Authenticator` to `AuthenticationExtension` for consistency and renamed `Resolver` and `Authorization` to `ResolverExtension` and `AuthorizationExtension` respectfully. This avoids an import as the name can be shared with the proc macro.
- I've wrapped additional wit types in the SDK.

I've changed the `Token` to be an enum to include the `anonymous` variant. Feels a bit specific to `@authenticated`, but it's also a common pattern in frameworks to special case the anonymous case so well.

The `ExtensionRuntime` will now receive the `engine::RequestContext` which is now public and gives access to the headers & access token. This makes it easy to share it to authorization extension with little copies.


crates/engine/src/engine/runtime.rs
```diff
@@ -3,15 +3,18 @@ use std::{future::Future, sync::Arc};
...
 
 pub trait Runtime: Send + Sync + 'static {
     type Hooks: runtime::hooks::Hooks;
     type Fetcher: runtime::fetch::Fetcher;
     type OperationCache: runtime::operation_cache::OperationCache<Arc<CachedOperation>>;
-    type Extensions: runtime::extension::ExtensionRuntime<SharedContext = <Self::Hooks as runtime::hooks::Hooks>::Context>;
+    type Extensions: runtime::extension::ExtensionRuntime<
+            Arc<RequestContext>,
+            SharedContext = <Self::Hooks as runtime::hooks::Hooks>::Context,
+        >;
```
pimeys, line 17: This could be a type alias...

crates/grafbase-sdk/src/types/token.rs
```diff
@@ -0,0 +1,41 @@
...
+
+    /// Token with claims serialized in JSON.
+    pub fn json_claims(claims: Vec<u8>) -> Self {
+        Self(wit::RawToken::Claims(wit::Data::Json(claims)))
+    }
+
+    /// Token with claims serialized in CBOR.
+    pub fn cbor_claims(claims: Vec<u8>) -> Self {
+        Self(wit::RawToken::Claims(wit::Data::Cbor(claims)))
+    }
```
pimeys: Should these be public?

extensions/jwt/src/lib.rs
```diff
@@ -63,19 +64,14 @@ impl Authenticator for Jwt {
...
-
-        Ok(Token::new(claims))
+        // We have validated this token, so we're good. We're just re-using the serialized claims data as is.
+        let mut parts = token_str.split('.');
+        parts.next().expect("headers");
+        Ok(Token::from_bytes(
+            base64::engine::general_purpose::URL_SAFE_NO_PAD
+                .decode(parts.next().expect("body"))
+                .unwrap(),
+        ))
```
pimeys, line 74: Maybe clean up this a bit. The jwt extension is an example for our users how to build an authentication extension. So the general style what we had before was much nicer than what we have here. Would not nag about it if this wouldn't be something the users implementing auth extensions would read.

pimeys (approved):

crates/engine/src/engine/runtime.rs
Finistere: what part? This is literally only a type definition.

crates/grafbase-sdk/src/types/token.rs
Finistere: removed those, went back to a simpler variant with only bytes & anonymous case

extensions/jwt/src/lib.rs
Finistere: fair enough, done

crates/engine/src/engine/runtime.rs
pimeys: I have some brain rot...

PR #2815, opened 2025-03-05
feat(gb-8632): check gateway and sdk version on extension load

pimeys:


crates/wasi-component-loader/src/lib.rs
```diff
@@ -6,6 +6,10 @@
 //! It is important the compiled WebAssembly code implements at least the minimal required types and interfaces.
 //! More on those on the crate README.
 
+/// The minimum required SDK version for the WebAssembly component. The engine must
+/// refuse to load the component if the SDK version is lower than this.
+pub static MINIMUM_SDK_VERSION: semver::Version = semver::Version::new(0, 8, 0);
```
Finistere, line 11: Can't we add it automatically with with an include / builds.rs? The only way this can be maintainable IMHO is for it to be automatic

Finistere: We do this in the schema for example:
https://github.com/grafbase/grafbase/blob/main/crates/engine/schema/build.rs

pimeys: No. You have to change this only if the interface changes. Otherwise it stays the same.

PR #2817, opened 2025-03-05
chore: Remove Extension trait

Finistere:
The `Extension` trait is redundant with the other `ResolverExtension` etc. trait. So moved the `new` method in there. This also cleans up the `new` for authentication & authorization as they will never receive any schema directives.

This plus the previous `Resolver` -> `ResolverExtension` rename, makes the extension definition simpler as a simple import is necessary the `XExtension` trait & macro.

I've changed the few uses I've seen of `Box<dyn std::error::Error>` also. We're returning GraphQL errors which can have extensions, so IMHO it's a better default type and makes this more consistent overall.

The SDK namespace is not very consistent though. `Error` and `Headers` are part of the root namespace but others like `QueryElements` for authorization aren't. Should we add all the types in the root namespace at least?

### DIFF
--- a/cli/templates/extension/src/auth.rs.template
+++ b/cli/templates/extension/src/auth.rs.template
@@ -1,21 +1,17 @@
 use grafbase_sdk::{
     types::{Configuration, SchemaDirective, ErrorResponse, Token},
-    AuthenticationExtension, Extension, Headers,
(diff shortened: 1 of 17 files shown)


tomhoule (approved):
Nice cleanup!

PR #2825, opened 2025-03-07
Snowflake extension init

tomhoule:
See the README for details. It has limitations.

One limitation we should lift before we merge this PR is taking the env vars for the private and public key from the configuration instead of hardcoded environment variables.

Additionally, with a bit more work, we could returned rows with named columns instead of `[[JSON!]!]` from the resolver.

The tests mock the response instead of querying against the live snowflake API.

closes GB-8456


extensions/snowflake/extension.toml
```diff
@@ -0,0 +1,17 @@
+[extension]
+name = "snowflake"
+version = "0.1.0"
+kind = "resolver"
+description = "Resolve GraphQL fields with snowflake queries"
```
fbjork, line 5:
```suggestion
description = "Resolve GraphQL fields with Snowflake queries"
```

Finistere (approved):

PR #2827, opened 2025-03-07
gateway-config: env var expansion in extensions config

tomhoule:
Treat every string value inside `[extensions.<name>.config]` as a `DynamicString<String>`, meaning environment variables will be expanded:

```toml
[extensions.snowflake.config]
warehouse = "Test"

[extensions.snowflake.config.authentication]
private_key = "{{ env.SNOWFLAKE_PRIVATE_KEY }}"
```

closes GB-8647

### DIFF
--- a/crates/gateway-config/src/extensions.rs
+++ b/crates/gateway-config/src/extensions.rs
@@ -1,6 +1,10 @@
-use std::path::{Path, PathBuf};
+use std::{
+    path::{Path, PathBuf},
+    str::FromStr as _,
+};
 
 use semver::VersionReq;
+use serde::Deserialize as _;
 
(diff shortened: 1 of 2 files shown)


crates/gateway-config/src/extensions.rs
```diff
@@ -103,3 +147,44 @@ impl ExtensionsConfig {
...
+        "#;
+
+        let err = toml::from_str::<StructuredExtensionsConfig>(toml)
+            .unwrap_err()
+            .to_string();
+
+        insta::assert_snapshot!(err, @r#"
+        TOML parse error at line 4, column 14
+          |
+        4 |             [config.test]
```
tomhoule, line 172: Not ideal.

PR #2829, opened 2025-03-07
cli: grafbase compose command

tomhoule:
This is long overdue. The options and the logic to fetch subgraph schemas are taken from `grafbase dev`. If `--graph-ref` is passed, the subgraphs from the given remote graph are taken into account, in addition to the subgraphs from the local configuration toml (if present). One of `--graph-ref` or the local paths must be present.

closes GB-6934


cli/src/cli_input/sub_command.rs
```diff
@@ -21,6 +22,9 @@ pub enum SubCommand {
     Logout,
     /// Set up and deploy a new graph
     Create(CreateCommand),
+    /// Compose a federated graph from subgraph schemas
+    #[clap(hide = true)]
```
tomhoule, line 26: Hide first, or just ship it?

Finistere: up to you

PR #2849, opened 2025-03-12
fix: aws upload fix for good

pimeys:
So... The issue was we had a crc command for the s3 ls, which does not exist. We can remove that and custom aws cli installations.


pimeys:
> I don't see the removal, it was added back, no?

Yeah the previous PR removed both crc checks. This adds it back to the upload part.

tomhoule (approved):
Ok get it now

fbjork (approved):

PR #2851, opened 2025-03-12
chore(gb-8679): test extensions from their own repo

pimeys:


Cargo.toml
```diff
@@ -23,12 +23,7 @@ exclude = [
     "crates/wasi-component-loader/examples",
     "examples/hooks-template",
     "examples/query_plan",
-    "extensions/rest",
-    "extensions/jwt",
-    "extensions/nats",
-    "extensions/authenticated",
-    "extensions/requires-scopes",
-    "extensions/snowflake",
+    "extensions",
```
fbjork, line 26: why keep this folder?

pimeys: It's a bit like Rust will test all the existing crates whenever they are about to release a new version to see if there's any regressions. We are doing the same here.

fbjork: Short term it's probably ok, but when the SDK is more stable and we have many extensions we need to remove this test or they will slow things down a lot.

pimeys: Let's reconsider then, right now this is not an issue and really prevents nasty things from happening we only realize after releasing a version. And then it's the release manager's job to fix the code at that point.

PR #2857, opened 2025-03-13
chore(gb-8682): use depot and sccache to speed up builds

pimeys:
Adds a few niceties

- sccache everywhere, so build most of the time just our own crates
- use the depot actions with a bit faster CPU and IO, much faster network
- use their docker setup, which caches heavily. so docker builds are extremely fast (when they most of the time just do nothing)

### DIFF
--- a/.github/workflows/binaries.yml
+++ b/.github/workflows/binaries.yml
@@ -35,7 +35,7 @@ concurrency:
 
 jobs:
   detect-change-type:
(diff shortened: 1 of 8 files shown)


fbjork (approved):

fbjork (approved):

tomhoule (approved):
Nice!

PR #2873, opened 2025-03-16
chore(extensions): Add Lease object to temporarily provide objects to the guest

Finistere:
Allows providing data in either owned, shared or shared mutable (rw lock) with guest in a consistent manner.
Not perfect, but does allow the engine to provide data in whatever way
seems best and for wasi-component-loader to treat it all the same way.


crates/engine/auth/src/jwt.rs
```diff
@@ -165,7 +166,17 @@ impl JwtProvider {
         // but 'iss' is the only one that I can think of that might be useful.
         claims.insert("iss".to_string(), issuer.into());
 
-        Some(LegacyToken::Jwt(JwtToken { claims }))
+        // We just validated the JWT token. Instead of de-serializing and re-serializing the
```
Finistere, line 169: Imitate jwt extension, makes it easier to share the bytes with the guest.

crates/engine/src/engine/execute.rs
```diff
@@ -23,7 +23,7 @@ use super::{Engine, Runtime, RuntimeExt, errors, runtime::HooksContext};
 
 pub(crate) use stream::StreamResponse;
 
-pub struct RequestContext {
+pub(crate) struct RequestContext {
```
Finistere, line 26: Doesn't need to be public anymore

crates/engine/src/prepare/operation_plan/query_modifications.rs
```diff
@@ -164,12 +164,14 @@ where
         ctx.extensions()
             .authorize_query(
                 extension_id,
-                ctx.request_context,
                 &ctx.hooks_context,
+                Lease::Owned(ctx.request_context.subgraph_default_headers.clone()),
```
Finistere, line 168: Temporary, changing it later to an actual mutable variant

crates/wasi-component-loader/src/resources/headers.rs
```diff
@@ -0,0 +1,103 @@
+use crate::extension::api::wit::HeaderError;
+
+use super::WasmOwnedOrLease;
+
+pub type Headers = WasmOwnedOrLease<http::HeaderMap>;
+
+impl Headers {
+    pub async fn get(&self, name: &str) -> Vec<Vec<u8>> {
```
Finistere, line 8: Copy pasted from before and adapted to be re-usable.

crates/integration-tests/src/federation/runtime/extension.rs
```diff
@@ -256,16 +263,25 @@ impl runtime::extension::ExtensionRuntime<Arc<engine::RequestContext>> for TestE
             .collect();
         async move {
             let instance = self.get_global_instance(extension_id).await;
-            instance
-                .authorize_query(ctx.clone(), wasm_context, elements_grouped_by_directive_name)
+            headers
```
Finistere, line 266: yeah ugly, will be changed today

crates/runtime/src/extension.rs
```diff
@@ -63,8 +63,66 @@ impl Token {
     }
 }
 
+/// It's not possible to provide a reference to wasmtime, it must be static and there are too many
+/// layers to have good control over what's happening to use a transmute to get a &'static.
+/// So this struct represents a lease that the engine grants on some value T that we expect to have
+/// back. Depending on circumstances it may be one of the three possibilities.
+pub enum Lease<T> {
```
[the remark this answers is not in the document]
Finistere: Had them at some point, the problem is that you usually need to clone the shared/shared mut version to share it obviously. but the owned/singleton variant is not meant to be cloned. So had an ugly `unreachable!` in the Clone impl. So in the end you end up creating the inner Arc / Arc<Rw> and clone it outside and only then wrapping it inside a lease. Still a bit figuring out the API for this

PR #2874, opened 2025-03-16
chore(extension): Add header modification support in authorize_query

Finistere:
Moving some files around for clarity and adding a `GraphqlContext` for the lack of a better name that holds the context relevant for a single operation (within batch request / websocket session) which is the wasm_context (provided by wasi-component-loader) and the subgraph_default_headers if there is any authorize_query extensions that modifies it.


crates/engine/src/execution/request/mod.rs
```diff
@@ -0,0 +1,169 @@
+mod context;
```
Finistere, line 1: pretty much just moved code

crates/engine/src/execution/request/response_extension.rs
```diff
@@ -0,0 +1,45 @@
+use grafbase_telemetry::otel::{opentelemetry::trace::TraceContextExt, tracing_opentelemetry::OpenTelemetrySpanExt};
```
Finistere, line 1: Moved code, changed from Engine methods to standalone ones

crates/engine/src/execution/request/well_formed_graphql_request.rs
```diff
@@ -0,0 +1,116 @@
+use std::sync::Arc;
```
Finistere, line 1: moved code

pimeys (approved):

PR #2878, opened 2025-03-17
chore(extension): Moving lease logic into wasi-component-loader

Finistere:
Has two benefits:
- engine doesn't depend on tokio anymore and the lease logic which is
  pecular to wasm is kept in wasi-component-loader. (can't pass references so we provide a lease on owned data we want back).
- the logic of whether to create a lock or not can be kept in
  wasi-component-loader. If there is single authorization extension,
  not only can we skip the lock for the headers but also for the
  authorization state later. Not sure how much of a perf impact this
  will make though.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2610,7 +2610,6 @@ dependencies = [
  "sha2 0.10.8",
  "sonic-rs",
  "thiserror 2.0.12",
- "tokio",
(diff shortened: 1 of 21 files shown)


tomhoule (approved):
A win for separation of concerns, nice!

PR #2892, opened 2025-03-18
chore(integration-tests): Add support for wasm extensions in integration-tests

Finistere:
Adding the ability to run proper wasm extensions in integration-tests mixed with pure Rust ones. Extensions will be in `integration-tests/data/extensions` which has a `build.sh` script to build them all.

In tests they can be trivially added with just `with_extension("authenticated")` and mixed with pure Rust ones like this:

```rust
.with_extension(AuthenticationExt::new(StaticToken::bytes(Vec::new())))
.with_extension("authenticated")
```

This makes it easy to test `wasi-component-loader` and wasm extensions in close to real conditions with all the tooling we have in integration-tests.


crates/engine/src/execution/request/mod.rs
```diff
@@ -98,6 +98,7 @@ impl<R: Runtime> Engine<R> {
             },
             None => (headers, access_token),
         };
+        println!("{access_token:?}");
```
pimeys, line 101: Uh, are you sure?

pimeys (approved):

pimeys:
Hey, one thing I would do here before merging. See the build graph:

image

We should maybe run the extension tests together with the builds, not make the builds depend on that. Only in the rust-prs workflow. So the workflow would be:

1) what-changed
2) wasm-extensions
3) everything else

This would make a full workflow to take about 8 minutes instead of 13.

crates/engine/src/execution/request/mod.rs
Finistere: thanks

PR #2909, opened 2025-03-20
fix: Add wasm authorization test case + fixes

Finistere:
Adding a real wasm extension test case.

- I'm now hiding the internal wasm error and exposing them as error
  logs. Showing a wasm stacktrace in a graphql error is pretty useless.
- I wasn't handling default values correctly. Really need to replace
federated-graph by cynic-parser to avoid logic duplication.
- And there was an issue with multiple directives for a single
authorization extension.

### DIFF
--- a/crates/integration-tests/data/extensions/crates/authorization-010/src/lib.rs
+++ b/crates/integration-tests/data/extensions/crates/authorization-010/src/lib.rs
@@ -1,0 +1,111 @@
+use grafbase_sdk::{
+    AuthorizationExtension, Error, IntoQueryAuthorization, SubgraphHeaders, Token,
+    types::{AuthorizationDecisions, Configuration, ErrorResponse, QueryElements, ResponseElements},
+};
+
+#[derive(AuthorizationExtension)]
+struct CustomAuthorization;
+
+#[derive(serde::Deserialize)]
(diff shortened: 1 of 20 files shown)


pimeys (approved):

crates/integration-tests/data/extensions/crates/authorization-010/src/lib.rs
```diff
@@ -0,0 +1,111 @@
...
+        headers: &mut SubgraphHeaders,
+        token: Token,
+        elements: QueryElements<'_>,
+    ) -> Result<impl IntoQueryAuthorization, ErrorResponse> {
+        let mut state = State { denied_ids: Vec::new() };
+        let mut builder = AuthorizationDecisions::deny_some_builder();
+        let error_id = builder.push_error(Error::new("Not authorized, query auth SDK010"));
+
+        for (name, elements) in elements.iter_grouped_by_directive_name() {
+            println!("{name}");
```
yanns, line 56: Do you want to keep the `println!`?

Finistere: ah no not at all, think I've removed it in a further PR but not sure anymore. will double check, thanks!

PR #2925, opened 2025-03-24
Protobuf compiler plugin to generate gRPC virtual subgraphs

tomhoule:
For use in combination with the upcoming grpc extension.

One invocation corresponds to the schema of one virtual GraphQL subgraph, however many protocol buffer IDL files are given as input.

In that GraphQL schema, we have:

- A translation of the definitions in the protobuf IDL file in `@protoMessages`, `@protoEnums` and `@protoServices` directives on the schema definition.
- Root mutation and subscription fields (we should also support query fields via a protobuf option soon) corresponding to service methods, with an `@grpcMethod` directive.
- The input and output type definitions required for these methods.

Protocol buffer definitions tend to use a lot of namespacing, but that is not a thing in GraphQL schemas. Instead, here I followed a convention of replacing dots with underscores in fully qualified paths to use as type names. That should avoid collisions and be relatively easy to discover.

closes GB-8846


crates/protoc-gen-grafbase-subgraph/src/schema/records.rs
```diff
@@ -0,0 +1,211 @@
...
+    pub(crate) fn render_graphql_type(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.write_str(match self {
+            ScalarType::Double | ScalarType::Float => "Float",
+            ScalarType::Int32
+            | ScalarType::Int64
+            | ScalarType::UInt32
+            | ScalarType::UInt64
+            | ScalarType::Sint32
+            | ScalarType::Sint64
+            | ScalarType::Fixed32
```
Finistere, line 136: You should have dedicated support for the 64bits variants. The gateway will not accept anything that doesn't fit in a `i32` for `Int`

Finistere: By dedicated, I mean creating a scalar for it. I'll drop `BigInt` in the gateway, but you can create your own

tomhoule: Indeed — would two custom scalars, `I64` and `U64` do the trick?

Finistere: Up to you really, the problem that comes is that it's the same namespace as other scalars. So you're sending back `i64` today afaik but another app might send them as strings

Finistere: so might be worth making this configurable one day. But `I64` and `U64` sound sensible to me

tomhoule: We can make sure we accept strings for i64/u64 in the extension, making a note to add that.

tomhoule: (done and pushed)

crates/protoc-gen-grafbase-subgraph/src/display_utils.rs
```diff
@@ -0,0 +1,35 @@
...
+        fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+            (self.0)(f)
+        }
+    }
+
+    DisplayFn(f)
+}
+
+pub(crate) fn grpc_path_to_graphql_name(path: &str) -> impl fmt::Display {
+    display_fn(|f| {
```
Finistere, line 19: Seems like you're re-inventing: https://docs.rs/itertools/latest/itertools/trait.Itertools.html#method.format_with

crates/protoc-gen-grafbase-subgraph/src/main.rs
```diff
@@ -0,0 +1,74 @@
...
+    let request = match CodeGeneratorRequest::decode(raw_request) {
+        Ok(request) => request,
+        Err(decode_err) => {
+            return bail(format!(
+                "Failed to decode CodeGeneratorRequest from stdin: {decode_err}",
+            ));
+        }
+    };
+
+    let mut files = Vec::with_capacity(1);
```
Finistere, line 34: don't think `with_capacity(1)` is really useful imho, just `vec![file]` would be clearer

crates/protoc-gen-grafbase-subgraph/src/render_graphql_sdl/graphql_types.rs
```diff
@@ -0,0 +1,139 @@
...
+    }
+
+    match ty {
+        FieldType::Scalar(scalar_type) => scalar_type.render_graphql_type(f)?,
+        FieldType::Message(proto_message_id) => schema.view(*proto_message_id).graphql_output_name().fmt(f)?,
+        FieldType::Enum(proto_enum_id) => schema.view(*proto_enum_id).graphql_name().fmt(f)?,
+    }
+
+    if repeated {
+        f.write_str("!]")?;
```
Finistere, line 84: inner type is never optional?

Finistere (approved):

crates/protoc-gen-grafbase-subgraph/src/render_graphql_sdl/graphql_types.rs
tomhoule: No, there can't be nulls, only messages in repeated fields. The messages can be empty, but they're always there.

crates/protoc-gen-grafbase-subgraph/src/display_utils.rs
tomhoule: true

crates/protoc-gen-grafbase-subgraph/src/render_graphql_sdl/graphql_types.rs
Finistere: ah right indeed

PR #2939, opened 2025-03-26
feat: sdk DirectiveSite method simplication

Finistere:


crates/grafbase-sdk/changelog/0.13.0.md
```diff
@@ -0,0 +1,3 @@
+# Breaking changes
+
+- Renamed `FieldDirectiveSite.field_name()` to `FieldDirectiveSite.name()` and similarly for all `*DirectiveSite` structs.
```
tomhoule, line 3: Should the `impl IntoQueryAuthorization` also be in there?

tomhoule (approved):

crates/grafbase-sdk/changelog/0.13.0.md
Finistere: It's only the template that is changed to avoid a Rust warning so not really necessary, but thanks!

PR #2950, opened 2025-03-31
feat: gateway mcp server

pimeys:
Something to play with. We miss things in the protocol, e.g. header passing is not there yet. But we can do a few things already...

grafbase.toml:

```toml
[graph]
introspection = true

[mcp]
enabled = true
enable_mutations = false
name = "Grafbase Enterprise Platform API"
instructions = "This server provides information on our GraphQL API. You can find details on graphs, branches of the graphs, statistics on a specific graph and extensions."

[[headers]]
rule = "insert"
name = "authorization"
value = "Bearer {{ env.GRAFBASE_ACCESS_TOKEN }}"
```

NOTE: only header insertion works right now with the MCP, forwarding does nothing. This blocks us from using it on production until the mcp protocol gets extension support (should be in the next couple of days).

run this branch with `-g api -c grafbase.toml`. Remember to have an access token defined in the environment.

In the cursor MCP config, add:

```json
{
  "mcpServers": {
    "grafbase-api": {
      "url": "http://127.0.0.1:5000/mcp"
    }
  }
}
```

Now, you might need to restart cursor or do a few tricks, but you should see this view if things work out:

image

Then in the chat, try something like:

```
what is the latest version of the nats extension in the grafbase-api
```

you should get something like this as the response:

image

Some other successful queries:

image

image


crates/federated-server/src/server/mcp.rs
```diff
@@ -0,0 +1,40 @@
...
+use axum::Router;
+use engine::Runtime;
+use gateway_config::ModelControlProtocolConfig;
+use rmcp::transport::{SseServer, sse_server::SseServerConfig};
+use server::McpServer;
+use tokio_util::sync::CancellationToken;
+
+use super::gateway::EngineWatcher;
+
+pub(super) fn router<S, R: Runtime>(
```
Finistere, line 17: You don't need `S`, it expresses your router state _requirements_. So if you `Router` is self-sufficient it should return a `Router<()>`

Finistere: You just need to move the `merge` after the `.with_state()`, would avoid the generic here which is confusing imho

crates/federated-server/src/server/mcp/server.rs
```diff
@@ -0,0 +1,672 @@
...
+) {
+    match definition {
+        Definition::Scalar(scalar) => {
+            let r#type = match scalar.ty {
+                ScalarType::String => "string",
+                ScalarType::Float => "number",
+                ScalarType::Int => "integer",
+                ScalarType::BigInt => "integer",
+                ScalarType::Boolean => "boolean",
+                ScalarType::Unknown => "string",
```
Finistere, line 642: don't know if you have that choice, but `Unknown` is basically arbitrary json rather than a string

crates/federated-server/src/server/mcp/server.rs
```diff
@@ -0,0 +1,672 @@
...
+
+                Err(ErrorData::invalid_request(message, None))
+            }
+            (Some(data), _) => Ok(data),
+            (None, _) => Err(ErrorData::internal_error("Invalid query response", None)),
+        }
+    }
+}
+
+fn variables_to_string(variables: Map<String, Value>) -> String {
```
Finistere, line 333: Not needed for this PR, but you should rather send those as proper variables rather than nesting them within the query string. So `field(arg: $var1, arg2: $var2)` etc.

crates/gateway-config/src/mcp.rs
```diff
@@ -0,0 +1,27 @@
...
+    /// The service path in the gateway.
+    #[serde(default = "default_mcp_path")]
+    pub path: String,
+    /// Whether mutations are enabled for the MCP service.
+    #[serde(default)]
+    pub enable_mutations: bool,
+    /// Pass the downstream auth header here until the MCP service
+    /// supports passing headers.
+    #[serde(default)]
+    pub authentication: Option<DynamicString<String>>,
```
Finistere, line 22: We have header rules that can insert headers, seems clearer and more versatile than this option IMHO.

crates/federated-server/src/server/state.rs
```diff
@@ -12,6 +14,9 @@ pub struct ServerStateInner<R: engine::Runtime, SR> {
     /// The server runtime, defining how to trigger IO depending on the platform.
     #[cfg_attr(not(feature = "lambda"), allow(unused))]
     pub server_runtime: SR,
+
+    /// For MCP connection management.
+    pub sse_txs: Arc<DashMap<Arc<str>, mpsc::Sender<ClientJsonRpcMessage>>>,
```
Finistere, line 19: Would move it in a dedicated `McpState` or equivalent struct that itself is part of the `ServerState` to separate concerns.

crates/federated-server/src/server.rs
```diff
@@ -173,17 +179,27 @@ pub async fn router<R: engine::Runtime, SR: ServerRuntime>(
...
     );
 
     let mut router = server_runtime
         .base_router()
         .unwrap_or_default()
-        .route(path, get(engine_execute).post(engine_execute))
+        .route(path, get(graphql_execute).post(graphql_execute))
         .route_service(websocket_path, WebsocketService::new(websocket_sender));
 
+    let ct = match config.mcp {
```
Finistere, line 193: Seems like  we should re-use the cancellation token to handle graceful shutdown everywhere in the future, no? Doesn't seem like you'll be gracefully shutting down MCP here. (doesn't matter for this PR)

crates/gateway-config/src/mcp.rs
```diff
@@ -0,0 +1,27 @@
+use serde_dynamic_string::DynamicString;
+
+#[derive(Clone, Debug, serde::Deserialize)]
+#[serde(deny_unknown_fields)]
```
Finistere, line 4: You could have added `#[serde(default)]` and implemented `Default`, find it clearer imho.

Finistere (approved):

crates/gateway-config/src/mcp.rs
Finistere: you're also literrally using a `forward` rule when you could just use an `insert` rule, I wonder who wrote the header rules :D /troll

crates/federated-server/src/server/mcp/server.rs
pimeys: The issue is then we have nested variables from fields. These are really hard to get right if we want to parameterize...

crates/gateway-config/src/mcp.rs
pimeys: This is not how it's going to be done eventually. We will use the header rules, but right now the rmcp crate or any other mcp library cannot pass headers. It's coming, so we can keep this ugly hack until we actually can transfer headers.

crates/federated-server/src/server/mcp/server.rs
Finistere: wdym? `field1(argA) { field2(argB) }`?

crates/gateway-config/src/mcp.rs
Finistere: What I'm suggesting is that you can simply remove this hack and use an `insert` rule today, instead of using both a `forward` rule AND this config option

crates/gateway-config/src/mcp.rs
pimeys: There is no default for things like name tho...

Finistere: ah ok. Would have used a default name like `Grafbase GraphQL MCP`or something like this imho. Feels like a nice-to-have to have a good name, rather than something absolutely required to get it up and running.

crates/federated-server/src/server.rs
pimeys: Yeah... it's to stop the spawned SSE servers. These will be gone when the crate implements the new protocol, and we don't need to keep an SSE alive anymore.

crates/gateway-config/src/mcp.rs
fbjork: good defaults is always a good idea to reduce friction

crates/federated-server/src/server/mcp/server.rs
pimeys: I mean the selection set can also have variables. If we do the capture here, we should also enforce it to the selection and that's kind of tough because the LLM controls the creation of the selection.

pimeys: E.g.

```graphql
foo { bar(arg: $x) { id } }
```

Finistere: I don't understand what you mean. My point is that you could pass to variable identifiers in the query string and then pass the data as variables, that would avoid the need to serialize by hand in the query string. That's what we're doing for subgraph queries. Here you're writing them by hand, it's not the LLM writing those arguments as string for you. Anyway

PR #2957, opened 2025-04-01
sdk: deserialize_seed for arguments

tomhoule:
Small change, but it's standalone, so let's get it in independently.

### DIFF
--- a/crates/grafbase-sdk/src/types/directive.rs
+++ b/crates/grafbase-sdk/src/types/directive.rs
@@ -58,6 +58,15 @@ impl<'a> FieldDefinitionDirective<'a> {
         minicbor_serde::from_slice(&self.0.arguments).map_err(Into::into)
     }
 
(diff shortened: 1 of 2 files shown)


Finistere (approved):

crates/grafbase-sdk/src/types/directive.rs
```diff
@@ -58,6 +58,15 @@ impl<'a> FieldDefinitionDirective<'a> {
         minicbor_serde::from_slice(&self.0.arguments).map_err(Into::into)
     }
 
+    /// Deserialize the arguments of the directive using a `DeserializeSeed`.
+    pub fn deserialize_arguments_seed<T>(&self, seed: T) -> Result<T::Value, SdkError>
```
Finistere, line 62: Thinking about it I think you could have just used `arguments_seed` for consistency.

tomhoule: Good point, will change next PR

PR #2980, opened 2025-04-04
feat(gb-8837): postgres wit guest side

pimeys:
This was fun.

Misses a lot of transformations, and probably some Postgres connection options too. Will get back to those when implementing the extension.


crates/grafbase-sdk/src/host_io/postgres/query.rs
```diff
@@ -0,0 +1,232 @@
...
+
+    /// Executes the SQL query with the bound parameters.
+    ///
+    /// This method sends the query to the database server and returns the number of rows affected.
+    /// For INSERT, UPDATE, or DELETE statements, this represents the number of rows modified.
+    /// For other statements, the meaning of the return value depends on the specific operation.
+    ///
+    /// # Returns
+    /// The number of rows affected by the query, or an error message if the execution failed
+    pub fn execute(self) -> Result<u64, String> {
```
Finistere, line 96: Would consider using `Result<..., SdkError>` everywhere to hide the fact it's a `String` and leave your some margin the future for changes.

crates/grafbase-sdk/src/host_io/postgres/query.rs
```diff
@@ -0,0 +1,232 @@
...
+    /// # Type Parameters
+    /// * `T` - The type to deserialize the JSON into
+    ///
+    /// # Returns
+    /// * `Ok(Some(T))` if the value is not NULL and was successfully deserialized
+    /// * `Ok(None)` if the value is NULL
+    /// * `Err` with a message if deserialization failed
+    pub fn as_json<T>(&self) -> Result<Option<T>, String>
+    where
+        T: DeserializeOwned,
```
Finistere, line 225: could just be `Deseriliaze<'_>` imho, doesn't need to be owned

crates/grafbase-sdk/src/host_io/postgres/types.rs
```diff
@@ -0,0 +1,378 @@
...
+use uuid::Uuid;
+
+use crate::wit;
+
+/// A trait for types that can be converted to database values.
+///
+/// This trait allows Rust types to be converted into database values that can be
+/// used in SQL queries. It provides methods to convert values with or without
+/// explicit type hints.
+pub trait DatabaseType {
```
Finistere, line 23: Do you want to expose that type publicly or should it be kept private, as in sealed/name not accessible?
You're using raw wit types, so doesn't seem like something users should use, but 🤷

Finistere (approved):

crates/grafbase-sdk/src/host_io/postgres/types.rs
Finistere: yeah just wanted to point it out just in case^^

pimeys: I know. Kind of think this could be public, but then the API needs to change a bit. Will figure it out later.

PR #2989, opened 2025-04-08
feat(sdk): Add root types for schema

Finistere:


crates/wasi-component-loader/src/extension/api/since_0_14_0/instance/utils.rs
```diff
@@ -55,6 +60,21 @@ pub fn create_complete_subgraph_schemas(schema: &Schema, extension_id: Extension
                 ws::Schema {
                     directives: Vec::new(),
                     type_definitions: Vec::new(),
+                    root_types: ws::RootTypes {
+                        query_id: if schema.query().exists_in_subgraph(&id) {
```
pimeys, line 64: Wait, do I understand this correctly? So schema has one query type, and the exact same type either exists in a subgraph?

pimeys (approved):

crates/wasi-component-loader/src/extension/api/since_0_14_0/instance/utils.rs
Finistere: Everything is remapped to the same Query type in composition

PR #3024, opened 2025-04-14
feat(gb-8838): sdk fixes for postgres extension

pimeys:
A few things I needed when building the extension side of things.

### DIFF
--- a/crates/grafbase-sdk/src/types/schema.rs
+++ b/crates/grafbase-sdk/src/types/schema.rs
@@ -66,6 +66,12 @@ impl SubgraphSchema<'_> {
         self.0.type_definitions.values().map(move |def| (schema, def).into())
     }
 
+    /// Retrieves a specific type definition by its unique identifier.
+    pub fn type_definition(&self, id: &DefinitionId) -> Option<TypeDefinition<'_>> {
+        let schema = self.0;
+        self.0.type_definitions.get(id).map(move |def| (schema, def).into())
+    }
+
     /// Iterator over the directives applied to this schema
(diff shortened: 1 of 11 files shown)


crates/grafbase-sdk/src/types/schema.rs
```diff
@@ -647,6 +653,18 @@ pub enum WrappingType {
     List,
 }
 
+impl WrappingType {
```
Finistere, line 656: Not sure if relevant, but I did add those on `Type`

Finistere (approved):

crates/grafbase-sdk/src/types/schema.rs
```diff
@@ -647,6 +653,18 @@ pub enum WrappingType {
...
+    /// Returns true if this wrapping type is a List
+    pub fn is_list(self) -> bool {
+        matches!(self, Self::List)
+    }
+
+    /// Returns true if this wrapping type is NonNull (required)
+    pub fn is_required(self) -> bool {
```
Finistere, line 663: Would use `is_non_null` to be consistent with naming and GraphQL spec

crates/grafbase-sdk/src/types/selection_set.rs
```diff
@@ -134,6 +134,14 @@ impl<'a> SelectionSet<'a> {
         self.fields_ordered_by_parent_entity()
     }
 
+    /// Returns the field at the given index in this selection set.
+    pub fn field(&self, index: usize) -> Option<Field<'a>> {
```
Finistere, line 138: Feels like a bit weird API to me, either you need the position with `enumerate()` or you don't and `Field` is `Copy`. So curious to see how you use it, but nothing against it either 🤷

PR #3028, opened 2025-04-14
feat(composition): Add initial support for composite schema directives

Finistere:
Adding very basic support for composite schema directives, treating them as extension directives to let them pass through composition.
That's enough for me to get started in the engine at least.


tomhoule (approved):
Nice way to integrate this right without baking it in :+1:

PR #3070, opened 2025-04-22
cli: de-nest backend module and delete obsolete clippy warnings

tomhoule:
The nesting was an artifact of `backend` being its own crate in the past.

Let's simplify. Also delete clippy annotations that do not apply anymore.


fbjork (approved):
Nice!

PR #3073, opened 2025-04-23
composition: fix handling of extra directives on enums

tomhoule:
Directives from extensions on enum definitions would sometimes fail to appear in the federated SDL.

Extra directives are sorted by directive site id. For every top level type definition, we ingest the type and its directives in a first pass, then the fields / union members in a second pass. Enums are the exception: we ingest the enum values right away. That lead the enum value's directives to come before the enum's directives although their directive side ids are greater. A binary search would then fail to find the enum's directives.

The fix is to ingest directives right after we generate the directive site id.

closes GB-8935


pimeys (approved):
cool, can we get a new version of the crate too?

tomhoule:
Next PR.

PR #3074, opened 2025-04-23
cli: make SubgraphCache the source of truth for subgraphs

tomhoule:
`SubgraphCache` is a type used by the `grafbase dev` command.

The original motivation of this PR is facilitating the integration of the new web UI for grafbase dev, where we will need to gather subgraph and composed schemas in one place to write them to a file.

Before this PR:

- Subgraphs with schemas loaded from files never touched the subgraph cache
- A new subgraph cache was instantiated every time the config was reloaded, but without replacing the existing subgraph cache. This is a bug, because new versions of schemas from files were not getting taken into account.
- Composition logic was in multiple places

This PR centralizes the logic of loading and composing subgraph schemas into the SubgraphCache type.

- `SubgraphCache::new()` instantiates subgraphs at the start.
- `SubgraphCache::compose()` is the entry point for composition in `grafbase dev`.
- `SubgraphCache::reload_local_subgraphs()` reloads local subgraphs from file or introspection. This is useful e.g. after a configuration change.

And the rest is private helpers.

### DIFF
--- a/cli/src/dev/hot_reload/subgraph_watcher.rs
+++ b/cli/src/dev/hot_reload/subgraph_watcher.rs
@@ -1,0 +1,235 @@
+use super::{super::subgraphs::SubgraphCache, WATCHER_DEBOUNCE_DURATION, reload_subgraphs};
+use crate::{
+    dev::subgraphs::{CachedIntrospectedSubgraph, CachedSubgraph},
+    errors::BackendError,
+};
+use gateway_config::Config;
+use grafbase_graphql_introspection::introspect;
+use notify_debouncer_full::{
+    DebounceEventResult, Debouncer, RecommendedCache, new_debouncer,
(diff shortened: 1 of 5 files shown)


cli/src/dev/hot_reload/subgraph_watcher.rs
```diff
@@ -0,0 +1,235 @@
...
+use std::{collections::HashSet, sync::Arc};
+use tokio::{runtime::Handle, sync::mpsc, time::MissedTickBehavior};
+use tokio_util::sync::CancellationToken;
+
+pub(super) struct SubgraphWatcher {
+    watcher: Option<Debouncer<RecommendedWatcher, RecommendedCache>>,
+    cancellation_token: Option<CancellationToken>,
+}
+
+impl SubgraphWatcher {
```
tomhoule, line 21: The code in this file is mostly just moved.

PR #3077, opened 2025-04-23
cli: simplify and expose `grafbase compose` command

tomhoule:
Use the new SubgraphCache mechanism to share logic with `grafbase dev`. Now that the configuration files are merged, with an implicit default, the API feels much cleaner, so let's un-hide the command and release it.

closes GB-8936


cli/src/backend.rs
tomhoule, line 1: Dead code

PR #3081, opened 2025-04-24
cli: integrate new assets for grafbase dev

tomhoule:
The main difference with the previous iteration of the app is in the data requirements. The new app expects an `/app/data.json` endpoint that contains

I am stopping here in the interest of time, but there are weaknesses we should address in the future. In particular, `data.json` is polled, and it can be large. We should use SSE or similar here to avoid unnecessary redownloads. Or at least e-tag based caching.

closes GB-8903


cli/src/dev/hot_reload.rs
```diff
@@ -107,7 +107,10 @@ pub(crate) async fn hot_reload(
         subgraph_watcher.stop();
 
         let config = match Config::load(&config_path) {
-            Ok(config) => config,
+            Ok(mut config) => {
+                config.graph.introspection = Some(true);
```
tomhoule, line 111: This was a bug. The first gateway config would always get enabled introspection, but not anymore as the config gets reloaded.

tomhoule: Will add that to the changelog

PR #3082, opened 2025-04-24
chore: bump cli to 0.93.0

pimeys:
We should release now to get postgres working in dev.


fbjork (approved):

tomhoule (changes requested):
https://github.com/grafbase/grafbase/pull/3081 must be part of this now (new assets), requesting changes to prevent an accidental merge

PR #3089, opened 2025-04-27
feat(engine): Advanced composite batch key/lookup matching

Finistere:
Adding more complex lookup/key matching logic to take into account `@oneOf` and add test cases for batch variants. Single lookup isn't working yet.

Fixing `@oneOf` handling in composition (not being part of federation spec) and adding `@shareable` to composite schema link


crates/graphql-wrapping-types/src/lib.rs
```diff
@@ -75,7 +75,7 @@ impl Wrapping {
 
     const fn set_list_length(&mut self, len: u8) {
         assert!((len as u32) < MAX_LIST_WRAPINGS, "list wrapper overflow");
-        self.0 = (self.0 & !LIST_WRAPPER_LENGTH_MASK) | ((len as u16) << LIST_WRAPPER_SHIFT);
+        self.0 = (self.0 & INNER_IS_REQUIRED_FLAG) | ((len as u16) << LIST_WRAPPER_SHIFT) | (self.0 & ((1 << len) - 1));
```
tomhoule, line 78: I'm not sure I get this change. Was there a bug?

tomhoule (approved):

crates/graphql-wrapping-types/src/lib.rs
Finistere: ah yes, the problem was that if you changed the list length, it would not clean the bits for higher list wrapping (null/non-null). So equality would fail if you compared a wrapping you modified with a freshly created one. They represented the same thing, but the underlying `u16` wasn't the same.

PR #3090, opened 2025-04-27
update gateway version to force a perl pkg upgrade

cristifalcas:
update gateway version to force a perl pkg upgrade


gateway/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "grafbase-gateway"
-version = "0.36.0"
+version = "0.36.1"
```
fbjork, line 3: Lets update the version in a separate PR

PR #3094, opened 2025-04-28
composition: fix scoping of imports

tomhoule:
The map holding imports by named was not scoped by subgraphs, leading imports to sometimes exceed the scope they were intended for.

The bug was probably introduced as the import map was moved from the ingest_subgraph() context to Subgraphs.


Finistere (approved):
Thanks!

PR #3096, opened 2025-04-28
composition: no auto-importing federation directives in virtual subgraphs

tomhoule:
That behaviour is for backwards compatibility. Virtual subgraphs are new, let's start clean there.


fbjork (approved):

Finistere (approved):
nice

PR #3105, opened 2025-04-29
chore(engine): Avoid cloning errros everywhere

Finistere:
Today each ResponsePart has its own errors vec that will appended to
the `ResponseBuilder` one. This move is useless as their only purpose is
to be serialized later.

Furthermore, before I used to add an error for every field which had a problem, even if it was the same one. I think it's pretty much useless. The primary purpose of the error is to help a human to understand what went wrong.

So now I only add unique errors, avoiding duplicates. Less data to serialize and clone.

Have to update extensions repo after.

### DIFF
--- a/.github/workflows/rust-prs.yml
+++ b/.github/workflows/rust-prs.yml
@@ -619,6 +619,8 @@ jobs:
           cargo build -p test-matrix
 
       - name: Test all extensions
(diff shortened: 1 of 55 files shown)


Finistere:
And merged another PR in this one..., now it's big sorry :/

tomhoule (approved):

PR #3111, opened 2025-04-30
graphql-composition: allow federated graphs without Query root

tomhoule:
As long as either Mutation or Subscription is defined.

closes GB-8919


crates/graphql-composition/tests/composition/empty_query_type/api.graphql.snap
```diff
@@ -0,0 +1,6 @@
+---
+source: crates/graphql-composition/tests/composition_tests.rs
+expression: actual_api_sdl
+input_file: crates/graphql-composition/tests/composition/empty_query_type/test.md
+---
+
```
tomhoule, line 6: Missing the mutation

crates/graphql-composition/tests/composition/null_value/api.graphql.snap
```diff
@@ -0,0 +1,21 @@
+---
```
tomhoule, line 1: This test is better if it doesn't end up in an unrelated error.

PR #3132, opened 2025-05-05
feat(engine): Add FieldSelectionMap parser

Finistere:
Parser for [FieldSelectionMap](https://github.com/graphql/composite-schemas-spec/blob/main/spec/Appendix%20A%20--%20Field%20Selection.md).
We don't do anything with it yet though.

Fixes: GB-8914

### DIFF
--- a/crates/engine/field-selection-map/Cargo.toml
+++ b/crates/engine/field-selection-map/Cargo.toml
@@ -1,0 +1,16 @@
+[package]
+name = "engine-field-selection-map"
+version = "0.1.0"
+edition = "2021"
+description = "Field selection map implementation from the Composite Schema specification."
+publish = false
+
+[dependencies]
+grafbase-workspace-hack.workspace = true
+itertools.workspace = true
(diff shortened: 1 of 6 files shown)


Finistere:
Fixes: GB-8914

crates/engine/field-selection-map/Cargo.toml
```diff
@@ -0,0 +1,16 @@
...
+
+[dependencies]
+grafbase-workspace-hack.workspace = true
+itertools.workspace = true
+winnow.workspace = true
+
+[dev-dependencies]
+insta.workspace = true
+# Not sure how to make it work with Hakari, it'll add debug feature everywhere with this...
+# winnow = { workspace = true, features = ["debug"] }
```
tomhoule, line 16: but only for dev-dependencies, no? At least I hope

Finistere: nop, everywhere...

PR #3133, opened 2025-05-05
cli: add third-party plugins support

tomhoule:
Like cargo and git.

For example, if you have a `grafbase-postgres` binary installed, you can call it with `grafbase postgres`.

This PR includes:

- Executing third-party plugins when available
- Advertising them in the help text
- A `grafbase list-third-plugins` command to list them
- Handling of bad arguments, unknown third party plugins, etc.

The gateway (`grafbase-gateway`) is given special treatment and excluded from plugins, since it is its own thing, and it would be confusing if `grafbase gateway` worked locally, but not in an environment where the CLI is not installed.

closes GB-8998

### DIFF
--- a/cli/src/cli_input/sub_command.rs
+++ b/cli/src/cli_input/sub_command.rs
@@ -10,12 +10,14 @@ use super::{
 
 #[derive(Debug, Parser, strum::AsRefStr, strum::Display)]
 #[strum(serialize_all = "lowercase")]
-pub enum SubCommand {
+pub(crate) enum SubCommand {
     /// Graph branch management
     Branch(BranchCommand),
     /// Output completions for the chosen shell to use, write the output to the
     /// appropriate location for your shell
     Completions(CompletionsCommand),
+    /// List all plugin subcommands. To list first-party subcommands, run `grafbase help`.
+    ListPlugins,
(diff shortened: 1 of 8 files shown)


cli/src/cli_input/sub_command.rs
```diff
@@ -10,12 +10,14 @@ use super::{
...
 #[strum(serialize_all = "lowercase")]
-pub enum SubCommand {
+pub(crate) enum SubCommand {
     /// Graph branch management
     Branch(BranchCommand),
     /// Output completions for the chosen shell to use, write the output to the
     /// appropriate location for your shell
     Completions(CompletionsCommand),
+    /// List all third-party subcommands. To list first-party subcommands, run `grafbase help`.
+    ListThirdParty,
```
fbjork, line 20: is this needed? Why not show plugin commands by default when you run grafbase help?

tomhoule: It's not possible because with clap, the help text template has to be known at compile time. Cargo has the same issue.

Finistere (approved):

cli/src/cli_input/sub_command.rs
pimeys: And git.

PR #3148, opened 2025-05-11
feat(gateway): Shape/de-serialization support of derived fields

Finistere:
With skip/include, authorization support.

Need to add the `@derived` directive with automatic detection and better
handling of nullability, and should be done with it.


crates/integration-tests/tests/gateway/composite/derived/mod.rs
```diff
@@ -0,0 +1,151 @@
...
+
+                type Query {
+                    post: Post!
+                }
+
+                type Post {
+                    id: ID!
+                    author_id: ID!
+                    author_x: ID!
+                    author: User! @is(field: "{ id: author_id x: author_x}")
```
pimeys, line 73: Would this work as:

```suggestion
                    author: User! @is(field: "{ id: author_id, x: author_x}")
```

Looks a bit cleaner if you can add a comma here...

pimeys (approved):

crates/integration-tests/tests/gateway/composite/derived/mod.rs
Finistere: made an issue in their repo: https://github.com/graphql/composite-schemas-spec/issues/187 We'll see what they see

PR #3162, opened 2025-05-19
feat(gb-8894): add config option for per-subgraph mtls

pimeys:
Config example:

```toml
[subgraphs.products.mtls]
root.certificate = "/path/to/ca.pem"
root.is_bundle = false
identity = "/path/to/ca.pem"
allow_invalid_certs = false
```

- `root.certificate`: path to either a root CA certificate, or a certificate chain
- `root.is_bundle`: if using a chain of certs, set this to true
- `identity`: is the client identity file. this file has PEM cert and key in the same file, pkcs#8 format
- `allow_invalid_certs`: set this to true only on development

Certain things to consider:

- If the subgraph is running with a self-signed certificate, the CA must have the hostname in the SAN section
- The private key must be in RSA, SEC1 Elliptic Curve or PKCS#8 format.


crates/integration-tests/data/mtls-subgraph/Dockerfile
```diff
@@ -0,0 +1,37 @@
...
+WORKDIR /app
+
+# Copy the built binary from the builder stage
+COPY --from=builder /usr/src/app/target/release/mtls-subgraph /app/mtls-subgraph
+
+# Create certificates directory
+RUN mkdir -p /app/certs
+
+# Copy TLS certificates
+COPY certs/*.pem /app/certs/
```
tomhoule, line 31: Failing in CI

tomhoule (approved):
Nice!

crates/integration-tests/data/mtls-subgraph/Dockerfile
pimeys: Genau.

pimeys: image

PR #3165, opened 2025-05-20
chore(gb-9090): remove sdk 0.8, update to wasmtime 32

pimeys:
The issue why we could not update was how we had some types defined that were not used anywhere in the WIT. Also our 0.8.0 support was so different and so broken, that I just deleted it. Nothing is compiled against that right now, it was pre-pre-pre extensions and we can throw it away.


crates/grafbase-sdk/wit/since_0_14_0/directive.wit
```diff
@@ -26,16 +26,10 @@ interface directive {
...
-        subgraph-name: string,
-        name: string,
-        arguments: list<u8>
-    }
-
     record field-definition-directive {
         name: string,
         site: field-definition-directive-site,
-        // serialized in CBOR
+
```
tomhoule, line 32: not accurate anymore?

tomhoule (approved):

PR #3166, opened 2025-05-20
chore(gb-9091): update to rust 1.87.0

pimeys:


fbjork (approved):

crates/engine/src/execution/request/mod.rs
```diff
@@ -33,27 +33,27 @@ impl<R: Runtime> Engine<R> {
     pub(crate) fn unpack_http_request<B>(
         &self,
         request: http::Request<B>,
-    ) -> Result<(EarlyHttpContext, http::HeaderMap, B), http::Response<Body>> {
+    ) -> Result<(EarlyHttpContext, http::HeaderMap, B), Box<http::Response<Body>>> {
```
tomhoule, line 36: Why this change?

tomhoule: If it's because the error variant is larger than the ok variant, it's generally not true, it's not taking `B` into account, which will pretty much always be the same size as the default Body.

pimeys: Yeah it's clippy not letting us to do that. Should we just say clippy to go somewhere else?

PR #3172, opened 2025-05-26
Add a way to assume all fields are used

praguevara:
- Introduce trait `UsageProvider`
- Add public function `check_assuming_all_used`

### DIFF
--- a/crates/operation-checks/src/aggregate_field_usage.rs
+++ b/crates/operation-checks/src/aggregate_field_usage.rs
@@ -20,6 +20,11 @@ pub trait UsageProvider {
 
     /// Get all used argument IDs for finding used input types.
     fn used_argument_ids(&self) -> Box<dyn Iterator<Item = schema::ArgumentId> + '_>;
+
+    /// Returns true if this provider assumes all input types are used.
+    fn assume_all_input_types_used(&self) -> bool {
+        false
+    }
 }
 
 /// Usage count of fields in a set of operations.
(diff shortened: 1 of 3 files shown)


crates/operation-checks/src/aggregate_field_usage.rs
```diff
@@ -19,6 +40,64 @@ pub struct FieldUsage {
...
+    fn argument_is_left_out(&self, _argument_id: schema::ArgumentId) -> bool {
+        // When assuming all usage, we also assume arguments with defaults might be left out
+        true
+    }
+
+    fn type_condition_is_used(&self, _type_condition: &str) -> bool {
+        true
+    }
+
+    fn used_argument_ids(&self) -> Box<dyn Iterator<Item = schema::ArgumentId> + '_> {
```
tomhoule, line 96: Let's see how it works out in practice, but I think we may have to change the signature for this one (is_argument_used()).

tomhoule: Ah that's new, I didn't get that at first sight. Checking it out in more depth.

tomhoule: I think this will cause breaking changes in input types to be missed, because if the iterator is empty, the input types used in field arguments will not be considered used, and will not trigger breaking change diagnostics as a consequence.

I can see at least two ways to fix that: 

1. An extra `assume_all_input_types_used() -> bool` method on the trait to short-circuit the whole find_used_input_types() implementation
2. Or we move find_used_input_types to the trait

PR #3176, opened 2025-05-28
feat(engine): @is support for batch lookup & single key

Finistere:
Inject a single key for `@lookup` based on the provided `@is` directive. The logic is more or less hacky currently. But now I think I see how to deal with it. It's a complex mix of logic to ensure we only inject key fields, while supporting multiple keys, auto-detection with `@is`, batching, validating everything etc.

Changed `Wrapping` API to be a bit closer to GraphQL nomenclature.

Also fixed a problem with argument default values being put in a list.


crates/graphql-wrapping-types/src/lib.rs
```diff
@@ -23,15 +23,9 @@ const INNER_IS_REQUIRED_FLAG: u16 = 0b1000_0000_0000_0000;
 /// The list_wrapping is stored from innermost to outermost and use the start and end
 /// as the positions within the list_wrapping bits. Acting like a simplified fixed capacity VecDeque.
```
tomhoule, line 24: I like this change :+1:

PR #3180, opened 2025-05-30
feat(engine): @is support for the remaining @lookup batch cases

Finistere:
Pretty big PR, with a ton of tests cases. Now `@lookup` has two modes basically:

- Either there is no `@is` directive on the arguments and do the auto-detection of possible injections that match a key. That's what the current code is already doing.
- Or there is, and we don't auto-detect anything at all. Everything must be explicitly specified. Here I'm also restricting the patterns to the same we auto-detect (`@oneOf` for multiple keys, batch-only, nested keys, composite keys). The detected requirements must match exactly one key or a `@oneOf` input object must be used to separate them.

In practice we might need a bit more flexibility, but it's complex enough for now^^ I'll have to write extensive documentation on this


crates/engine/schema/src/builder/graph/directives/composite/require.rs
```diff
@@ -0,0 +1 @@
+
```
tomhoule, line 1: Intentionally empty?

tomhoule (approved):
:100:

PR #3183, opened 2025-06-02
feat(engine): Parse composite require directive

Finistere:
Adding it to the schema, but not doing anything with it yet.


crates/engine/schema/src/builder/graph/directives/composite/require.rs
```diff
@@ -1 +1,111 @@
...
+                .requires_records
+                .push(FieldRequiresRecord {
+                    subgraph_id,
+                    field_set_record: requires,
+                    injection_ids,
+                });
+        }
+    }
+
+    todo!()
```
pimeys, line 110: Is this intended?

crates/engine/schema/src/builder/graph/directives/composite/mod.rs
```diff
@@ -146,13 +132,31 @@ fn ingest_before_federation_directives<'sdl>(
     Ok(())
 }
 
-fn ingest_after_federation_directives<'sdl>(
-    ingester: &mut DirectivesIngester<'_, 'sdl>,
-    def: sdl::SdlDefinition<'sdl>,
-    dir: sdl::Directive<'sdl>,
+pub(crate) fn ingest_composite_field_directives_after_federation(
```
pimeys, line 135: Just a sidenote, if we are iterating over all the field definitions in multiple places, would it be possible to combine them all into one loop? I haven't checked if this is the case, just feels a bit expensive to iterate all the field definitions for one directive here.

gateway/tests/integration_tests.rs
```diff
@@ -129,9 +129,11 @@ impl<Response> GqlRequestBuilder<Response> {
         .send()
         .await
         .unwrap()
-        .json::<Response>()
+        .json::<serde_json::Value>()
         .await
-        .unwrap()
+        .unwrap();
+        println!("Received response:\n{}", serde_json::to_string_pretty(&value).unwrap());
```
pimeys, line 135: Do we want this for every test?

pimeys (approved):

crates/engine/schema/src/builder/graph/directives/composite/require.rs
Finistere: not really, but fixed in next pr

crates/engine/schema/src/builder/graph/directives/composite/mod.rs
Finistere: A few things need to be done in different phases. Some things could surely be optimized, but we'll see once we have a schema that takes a long time to build.

PR #3187, opened 2025-06-03
Make all workspace crates stop depending on graphql-federated-graph

tomhoule:
Except for graphql-composition. In the next PR, the graphql-federated-graph will be deleted and become a module inside the graphql-composition crate. That will improve encapsulation, make the API clearer and more evolvable, and enable implementation simplifications.


crates/engine/schema/src/builder/extension/link_directive.rs
```diff
@@ -0,0 +1,145 @@
+use std::collections::HashMap;
```
tomhoule, line 1: This module is just copy-pasted and pared down to just what the schema builder needs.

fbjork (approved):

PR #3188, opened 2025-06-03
graphql-composition: absorb federated-graph crate

tomhoule:
The FederatedGraph data structure has no reason to be in a separate crate anymore. Absorb it into graphql-composition for more clarity, simplicity and encapsulation.

The graphql-federated-graph dependency, which was always an implementation detail, is now deprecated. Its functionality has been folded into graphql-composition, as a module. As a result, the crate does not re-export the `graphql_federated_graph` crate anymore, just a select few items. Most fields of `FederatedGraph` have also become private. Please get in touch if this is something you were relying on.


crates/graphql-composition/src/federated_graph/directives/is.rs
```diff
@@ -0,0 +1,58 @@
+use cynic_parser_deser::{ConstDeserializer as _, ValueDeserialize};
```
tomhoule, line 1: New file, we were using RequireDirective for `@is`, which works but is just weird.

Finistere: used a type alias for this in engine-schema

PR #3197, opened 2025-06-04
feat(engine): require support & resolver rework

Finistere:
In this PR I'm starting to rename `SelectionSetResolver` to `Resolver`. There will be only one extension. It'll mostly be a superset of `SelectionSetResolver` but with the directive rules from `FIeldResolver` (no explicit resolver directives in manifest => all are resolvers, instead of the current more custom logic). I'll remove the previous two once I've migrated the current extensions.

I've added a `is_internal` field for `InputValueDefinition` to keep track of `@require` arguments as they're implicit internal for a given subgraph. I need to add a bunch of tests for this though. Today I'm mostly marking them as `inaccessible` and inject them when I see they're resolved by the corresponding subgraph.

For resolvers I'm adding the notion of `guest_batch`. It's detected if `@require` expects a list, so the following:
```
dummy(ids: [ID!] @require(field:"[id]"))
```

In that case we assume the resolver to batch internally. If not, we call the resolver multiple times for each parent object. Today we don't put any rate limits, but we should think about it at some point.

I've fully reworked the deserialization of resolver responses. Before I was doing a bunch of tricks to expose a list of fields resolved by an extension as a single object. This kinda worked for single objects, but fell apart with lists. Now I've considerably simplified it by processing fields one by one and calling a final "post processing" method from the object deserializer. The latter adds typename fields, derived fields, authorization errors etc.


crates/engine/schema/src/builder/graph/directives/resolvers.rs
```diff
@@ -54,30 +55,111 @@ fn create_root_graphql_resolvers(ingester: &mut DirectivesIngester<'_, '_>) {
...
+                    field
+                        .resolver_ids
+                        .push(ResolverDefinitionId::from(graph.resolver_definitions.len() - 1))
+                }
+                _ => {}
+            }
+        }
+    }
+
+    let builder = ingester.deref_mut();
```
Finistere, line 103: same logic as before for SelectionSetResolver which will be removed.

crates/engine/src/response/write/deserialize/object/fields.rs
```diff
@@ -0,0 +1,565 @@
...
+        let value = result?;
+
+        response_fields.push(ResponseObjectField { key, value });
+
+        Ok(())
+    }
+}
+
+impl<'ctx> ConcreteShapeFieldsContext<'ctx, '_, '_> {
+    pub fn finalize_deserialized_object_fields(
```
pimeys, line 425: Maybe split this one up a bit.

crates/engine/schema/src/generated/resolver.rs
```diff
@@ -23,44 +23,43 @@ use walker::{Iter, Walk};
...
 ///   | GraphqlRootFieldResolverDefinition
 ///   | GraphqlFederationEntityResolverDefinition
 ///   | FieldResolverExtensionDefinition
-///   | SelectionSetResolverExtensionDefinition
+///   | ExtensionResolverDefinition
 ///   | LookupResolverDefinition
 /// ```
 #[derive(serde::Serialize, serde::Deserialize)]
 pub enum ResolverDefinitionRecord {
+    Extension(ExtensionResolverDefinitionRecord),
```
pimeys, line 37: Could this be a breaking change? Remember to test the postgres extension with a new CLI if it builds with that before merging.

pimeys (approved):

crates/engine/schema/src/generated/resolver.rs
Finistere: nothing breaking yet and yes still works

PR #3210, opened 2025-06-09
mcp: align JSON schema and implementation for variables

tomhoule:
The JSON schema says `variables` is optional in the request, but the implementation requires it. That leads to errors in Cursor.

We can either advertise the `variables` parameter as required, or make it optional to align schema and implementation.

This commit does both, because I couldn't persuade Cursor that it's OK for variables to be optional. So make the variables parameter explicitly required, so Cursor always sends it, but in case it doesn't, it is actually optional.

closes GB-9184


Finistere (approved):

Finistere:
Would be worth mentioning this in a comment imho

PR #3212, opened 2025-06-10
feat(gb-9166): audit logs guest side

pimeys:
Let's do this in parts. Now, we discussed with commercetools, and it kind of makes sense to audit in the on_response hook. This PR implements the wit and SDK side, the host parts are just a stub for now.


crates/grafbase-sdk/src/component/state.rs
```diff
@@ -14,19 +14,32 @@ type InitFn =
 static mut INIT_FN: Option<InitFn> = None;
 static mut EXTENSION: Option<Box<dyn AnyExtension>> = None;
 static mut SUBSCRIPTION: Option<Box<dyn Subscription>> = None;
+static mut AUDIT_LOGS: Option<wit::AuditLogs> = None;
```
pimeys, line 17: The idea, what we had in mind with @Finistere and what I'm going to try out is that:

- ok, audit logs are per request
- we have them as an atomic pointer in the host
- whenever we call an extension, we swap the pointer first
- unsafe cowboy things

The alternative idea is to pass them in every extension call, but I'm trying this approach first.

Finistere (approved):

crates/grafbase-sdk/src/component/state.rs
pimeys: I'll consider in the next PRs.

PR #3214, opened 2025-06-11
feat(gb-9166): event queue

pimeys:
Being a nice citizen this week and sending small PRs. The next step of the audit is the event queue, which fits into a nice library. This will only do anything if there is a hooks extension defined, and the extension defines a filter for events they want to see in the on_response hook.


crates/event-queue/src/lib.rs
```diff
@@ -0,0 +1,169 @@
...
+    ///
+    /// # Arguments
+    ///
+    /// * `builder` - The operation execution event builder
+    pub fn push_operation(&self, builder: ExecutedOperationBuilder<'_>) {
+        if !self.event_allowed(EventFilterType::Operation) {
+            return;
+        }
+
+        let Some(queue) = self.inner.as_ref() else {
```
Finistere, line 70: This condition seems a bit unnecessary now as the `filter` will be `None`, it's probably cheap enough to just create a queue each time. Or `filter` being `None` is redundant.

PR #3218, opened 2025-06-11
feat(gb-9190): shared-context to hooks with event queue

pimeys:
Let's start with the axum layer. The context is now available, but does not really do anything. Tomorrow I'll trickle this down like Ronald Reagan to the engine in the request extensions.


crates/wasi-component-loader/src/context.rs
```diff
@@ -28,26 +29,29 @@ pub struct SharedContext {
     pub(crate) authorization_state: AuthorizationState,
     /// A log channel for access logs.
     pub(crate) trace_id: TraceId,
+    pub(crate) event_queue: EventQueue,
 }
 
 // FIXME: Remove me once hooks & extensions context are merged.
 impl Default for SharedContext {
```
Finistere, line 36: I wouldn't have re-used the legacy hook struct if you could avoid it

crates/wasi-component-loader/src/extension/api/since_0_10_0/instance/hooks.rs
```diff
@@ -1,16 +1,21 @@
 use futures::future::BoxFuture;
 
-use crate::extension::HooksInstance;
+use crate::{SharedContext, extension::HooksInstance};
 
 impl HooksInstance for super::ExtensionInstanceSince0_10_0 {
     fn on_request(
```
Finistere, line 6: for resolver I just ended up having default implementation of the methods, just too painful to update everytime

crates/grafbase-sdk/wit/since_0_17_0/shared-context.wit
```diff
@@ -0,0 +1,11 @@
+interface shared-context {
+    use event-queue.{event-queue};
+
+    /// A shared context resource that provides access to tracing and event handling capabilities
+    resource shared-context {
+        /// Returns the request trace identifier as a string
+        trace-id: func() -> string;
+        /// Returns the request event queue
+        event-queue: func() -> event-queue;
```
Finistere, line 9: I don't think I would have exposed the event queue as is. This makes it look like you can also pop from anywhere. Would have added a `register-event` function and for the `on_response` would have actually provided a list of events. Since you need to explicitly register the events you're interested in, no need to pop them one by one. Well maybe not the most efficient design to allocate a vec all the time though, so well don't know for the `on_response` side

Finistere (approved):

crates/grafbase-sdk/wit/since_0_17_0/shared-context.wit
pimeys: It is not public for the user.

Finistere: The WIT is as public as the SDK itself. We just don't offer support for it

PR #3222, opened 2025-06-13
feat(engine): Add support for single entity lookup for resolver extension

Finistere:
Mostly a ton of tests. Realized all the `@oneOf` batch cases weren't relying on the `@is` directive, silently ignored, but only on the auto-detection, so fixed that too.

Added a "strict" mode function for composition to warn for unknown directive and surprise, surprise, found an issue :D

### DIFF
--- a/crates/graphql-composition/src/compose/directives.rs
+++ b/crates/graphql-composition/src/compose/directives.rs
@@ -86,7 +86,13 @@ pub(super) fn collect_composed_directives<'a>(
                         })
                     }
                     (None, true) => Some(ir::DirectiveProvenance::ComposeDirective),
-                    (None, false) => None,
+                    (None, false) => {
+                        ctx.diagnostics.push_warning(format!(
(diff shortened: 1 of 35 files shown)


crates/graphql-composition/src/compose/directives.rs
```diff
@@ -86,7 +86,20 @@ pub(super) fn collect_composed_directives<'a>(
                         })
                     }
                     (None, true) => Some(ir::DirectiveProvenance::ComposeDirective),
-                    (None, false) => None,
+                    (None, false) => {
+                        if ctx.strict {
+                            ctx.diagnostics.push_fatal(format!(
```
tomhoule, line 91: At this point in the composition process, I think the directives you don't want to be ignored are already filtered out. Have you tried doing this in `ingest_subgraphs::directives.rs` in the NoMatch branch? (there's already a warning being pushed there)

tomhoule (approved):
Impressive amount of tests :100:

PR #3226, opened 2025-06-18
feat(wit): Improve http client to re-use the http headers resource.

Finistere:
Makes it easier to re-use subgraph headers and error happen earlier in the guest
rather than host side for invalid name/value.


tomhoule (approved):
Looks good. Just wondering, isn't that a breaking change that would warrant a new wit version?

Finistere:
yes, but we never released sdk 0.17^^ Will do for the blogpost so I can release the extensions

PR #3227, opened 2025-06-18
feat(examples): Add REST with composite schemas

Finistere:


Cargo.toml
```diff
@@ -308,6 +309,7 @@ debug = true
 # by telling cargo to optimize at the link stage (in addition to the
 # normal optimizations during the compilation stage)
 lto = "thin"
+debug = true
```
tomhoule, line 312: Won't that enable debug symbols also for gateway and cli builds?

PR #3241, opened 2025-06-24
sdk: expose the running gateway's URL

tomhoule:
This is for more end to end tests, and in particular tests for the public metadata endpoints.


crates/grafbase-sdk/src/test/gateway.rs
```diff
@@ -47,6 +47,11 @@ impl TestGateway {
         TestGatewayBuilder::new()
     }
 
+    /// Full url of the GraphQL endpoint on the gateway.
+    pub fn gateway_endpoint(&self) -> &Url {
```
Finistere, line 51: would call it `url`

Finistere: `gateway.gateway_endpoint` is redundant at least imho

Finistere (approved):

PR #3261, opened 2025-07-02
feat(gb-9263): more granular log field types

pimeys:
Also cuts a new breaking SDK 0.19.0. You're welcome.

### DIFF
--- a/crates/grafbase-sdk/wit/since_0_19_0/logger.wit
+++ b/crates/grafbase-sdk/wit/since_0_19_0/logger.wit
@@ -1,0 +1,87 @@
+interface logger {
+    /// The severity level of a log entry, ordered from most severe to least severe.
+    enum log-level {
+        /// Critical errors that may cause the application to terminate.
+        error,
+        /// Warning messages for potentially harmful situations.
+        warn,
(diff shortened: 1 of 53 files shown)


fbjork (approved):

crates/grafbase-sdk/wit/since_0_19_0/logger.wit
```diff
@@ -0,0 +1,87 @@
...
+        /// Rotate every month.
+        monthly,
+        /// Rotate every year.
+        yearly,
+    }
+
+    /// Represents different types of values that can be stored in log entry fields.
+    variant log-field {
+        /// A string value.
+        field-string(string),
```
Finistere, line 37: you can escape names with`%` so `%string` for example

crates/wasi-component-loader/src/extension/api/since_0_19_0/wit/mod.rs
```diff
@@ -0,0 +1,34 @@
+pub mod authorization_types;
```
Finistere, line 1: a lot of copy pasted files, not the end of the world but would be better to just reference the existing ones. like authoriation types aren't new since a while

PR #3264, opened 2025-07-02
chore(gb-9270): simplify federated-server functions with ctx

pimeys:
Long overdue, but the amount of methods with a bunch of arguments you need to keep on piling up every time you want to pass something was never nice. Hopefully this is better. Also refactored the config and graph watcher channels into a single channel, which simplified things even further.


crates/federated-server/src/server/gateway/graph.rs
```diff
@@ -0,0 +1,97 @@
...
+
+impl GraphDefinition {
+    /// Converts a `GraphDefinition` into a `Graph`.
+    ///
+    /// This method processes the graph definition based on its type:
+    ///
+    /// - For `ObjectStorage` variants, it returns a graph with potential trusted documents support
+    /// - For `Sdl` variants, it returns a graph without trusted documents
+    ///
+    /// Trusted documents will only be enabled for `ObjectStorage` variants when both an access
```
tomhoule, line 38: I don't know if that was the logic before this refactoring, but the logic we would want to land on is making trusted documents work for graphs defined from SDL, as long as the graph ref is present (just like for telemetry)

crates/federated-server/src/server/events.rs
```diff
@@ -0,0 +1,61 @@
...
+/// all update sources into a single stream.
+#[derive(Clone)]
+pub enum UpdateEvent {
+    /// A graph definition update event
+    Graph(GraphDefinition),
+    /// A configuration update event
+    Config(Box<Config>),
+}
+
+#[allow(dead_code)]
```
tomhoule, line 15: Why is this necessary?

tomhoule (approved):
:100:

crates/federated-server/src/server/gateway/graph.rs
tomhoule: Discussed on slack. This is not a problem introduced by this PR, let's fix separately. Issue: GB-9274

PR #3271, opened 2025-07-07
fix(gb-9292): validate resource/header values better

pimeys:


crates/federated-server/src/server/graph_fetch_method/access_token.rs
```diff
@@ -0,0 +1 @@
+
```
tomhoule, line 1: Intentionally empty?

crates/access-token/Cargo.toml
```diff
@@ -0,0 +1,14 @@
+[package]
```
tomhoule, line 1: Maybe this could be a module in the federated server crate?

tomhoule (approved):

PR #3275, opened 2025-07-09
chore(federated-server): regroup engine logic

Finistere:


crates/federated-server/src/engine/runtime.rs
```diff
@@ -101,27 +96,61 @@ impl GatewayRuntime {
...
-        let operation_cache = operation_cache(gateway_config, &mut redis_factory)?;
+        let operation_cache = operation_cache(ctx.gateway_config, &mut redis_factory)?;
 
         tracing::debug!("Building extensions");
 
-        let extensions = WasmExtensions::new(extension_catalog, gateway_config, schema, logging_filter)
-            .await
-            .map_err(|e| crate::Error::InternalError(format!("Error building an extension: {e}")))?;
+        let extensions = WasmExtensions::new(
+            extension_catalog,
```
pimeys, line 104: Maybe we should have a context/config object in this too? I see us adding more things here...

PR #3291, opened 2025-07-17
feat(auth): Differentiate between /mcp & /graphql authentication

Finistere:
With the following configuration:
```toml
[authentication.protected_resources.graphql]
extensions = ["auth1"]
default = "anonymouns"


[authentication.protected_resources.mcp]
extensions = []
default = "deny"
```

If nothing is specified, all authentication extensions are called. If
the list is empty, it behaves as if no authentication extensions was
specified.


tomhoule:
> Maybe we should use authentication.protected_resources to follow OAuth
terminology?

I like the idea

tomhoule (approved):

tomhoule:
I was just rewriting the configuration, and we have

```toml
    [authentication]
    default = "deny"
```

We definitely want to allow by default for the graphql endpoint, but deny would be nice for /mcp. Is it trivial to allow having a different default per resource?

PR #3330, opened 2025-07-30
chore: Speed up integration-tests

Finistere:
Currently, even on my 32-cores machines, integration tests are extremely slow because of the wasm compilation. What I didn't realize while adding the wasm cache is that wasmtime adds the mtime of the current executable by default in the path in debug builds... (by checking if debug assertions are enabled or not).
Wasmtime uses a `GIT_REV` var to distinguish between versions, but doesn't work for me. So now disabled the mtime and added a hash of the main `Cargo.lock` file in the path to prevent problems.

In wasi-component-loader I've the structure to use a single engine, instead of one per extension, as the documentation suggests. Before, I was using a nested `cache` dir in the extension folder. That doesn't work anymore now that the engine is shared. So added a:

```toml
[wasm]
cache_path = ".."
```

in the configuration, mostly for us than anything else. If not present it defaults to `<current working dir>/.grafbase/wasm-cache`, and if that doesn't work to `<tmp dir>/grafbase-wasm-cache`.

`integration-tests` & `gateway` tests override it to be `.grafbase/wasm-cache` at the project root level to share the cache as the gateway test re-uses some of the compiled extensions.

Similar to Nexus, adding caching for `~/.cargo` & `target` directories.

Now, on my machine, the `integration-tests` run in 8s warm, 12s cold. It took forever before, especially because of contracts tests that use extensions extensively. We still have cli extension build test that take forever also, but don't know how to solve them yet...

### DIFF
--- a/.github/workflows/rust-prs.yml
+++ b/.github/workflows/rust-prs.yml
@@ -40,6 +40,19 @@ jobs:
           # Seems like if it's not zero you dont get branches
           fetch-depth: 0
 
+      - name: Run sccache-cache
+        uses: mozilla-actions/sccache-action@7d986dd989559c6ecdb630a3fd2557667be217ad # v0.0.9
+
+      - name: Cache cargo registry and build
+        uses: actions/cache@v4
+        with:
+          path: |
+            ~/.cargo/registry/index/
+            ~/.cargo/registry/cache/
+            ~/.cargo/git/db/
+            target/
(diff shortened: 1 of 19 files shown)


pimeys (approved):
interesting!

Finistere:
Oh and forgot an important part, I'm compiling all dependencies with `opt-level = 2`, not sure if it's better for CI, but considering we're going to cache more aggressively, might worth it. Locally it seems ok for now

PR #3336, opened 2025-07-31
gateway: url-encode the path for object storage schema requests

tomhoule:
closes GB-9436

### DIFF
--- a/crates/federated-server/src/graph/loader/object_storage.rs
+++ b/crates/federated-server/src/graph/loader/object_storage.rs
@@ -119,26 +119,30 @@ impl ObjectStorageUpdater {
             .build()
             .map_err(|e| crate::Error::InternalError(e.to_string()))?;
 
-        let object_storage_host = object_storage_host();
+        let mut object_storage_url: url::Url = object_storage_host().parse().map_err(|err| {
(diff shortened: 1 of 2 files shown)


crates/federated-server/src/graph/loader/object_storage.rs
```diff
@@ -119,26 +119,30 @@ impl ObjectStorageUpdater {
...
             GraphRef::LatestProductionVersion { graph_slug } => {
-                format!("{object_storage_host}/graphs/{graph_slug}")
+                format!("/graphs/{graph_slug}")
             }
             GraphRef::LatestVersion {
                 graph_slug,
                 branch_name,
-            } => format!("{object_storage_host}/graphs/{graph_slug}/branch/{branch_name}"),
+            } => {
+                let branch_name = url::form_urlencoded::byte_serialize(branch_name.as_bytes()).collect::<String>();
```
Finistere, line 134: Won't we have the same problem with graph names?

tomhoule: Thought so too, but we validate that these only have alphanumeric characters and hyphens in the api, so they should be fine: https://github.com/grafbase/api/blob/47a6be29e3e5e76912dd603ccfc784ce3b810bb3/crates/domain/src/validation.rs

PR #3347, opened 2025-08-04
Claude setup

Finistere:


pimeys (approved):
A good start. Now, add new CLAUDE.md files to different crates when the time comes (you know when you know), and add modifications if it does something weird repeatedly.

PR #3349, opened 2025-08-05
gateway: fix panic when subgraph with resolver extension has URL

tomhoule:
Return an error instead.

### DIFF
--- a/crates/engine/schema/src/builder/graph/directives/resolvers.rs
+++ b/crates/engine/schema/src/builder/graph/directives/resolvers.rs
@@ -65,6 +65,13 @@ fn create_extension_resolvers(ingester: &mut DirectivesIngester<'_, '_>) -> Resu
             match directive.ty {
                 ExtensionDirectiveType::FieldResolver => {
                     let subgraph_id = directive.subgraph_id;
+                    if !directive.subgraph_id.is_virtual() {
+                        return Err(
+                            "Field resolver extensions can only be used with virtual subgraphs (subgraphs without a URL)."
+                                .to_string()
+                                .into(),
+                        );
+                    }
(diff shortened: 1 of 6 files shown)


crates/integration-tests/tests/gateway/extensions/field_resolver/url_subgraph_error.rs
```diff
@@ -0,0 +1,46 @@
...
+            .with_extension(StaticFieldResolverExt::json(
+                r#"{"id": "1", "name": "Test Product", "price": 99.99}"#.into(),
+            ))
+            .try_build()
+            .await;
+
+        let error = result.unwrap_err();
+        let error_message = error.to_string();
+
+        assert!(
```
Finistere, line 41: you could have just used `insta::assert_snapshot(result.unwrap_err()` here, that's what I do at least

crates/engine/schema/src/builder/graph/directives/resolvers.rs
```diff
@@ -79,13 +86,23 @@ fn create_extension_resolvers(ingester: &mut DirectivesIngester<'_, '_>) -> Resu
                 }
                 ExtensionDirectiveType::Resolver => {
```
Finistere, line 87: You add a test for `field_resolver` (legacy), but you would need to add a another for `resolver` case. Can do it afterwards if you want

Finistere (approved):
nice, thought it was handled more gracefully

crates/integration-tests/tests/gateway/extensions/field_resolver/url_subgraph_error.rs
tomhoule: fixed

PR #3350, opened 2025-08-05
feat(composite/lookup): Accept namespaces for the output.

Finistere:
In gRPC we can't return a list directly, it's always returned within a message which ends up looking like:

```graphql
type Query {
  products(ids: [ID!]!): ProductsResponse! @lookup
}

type ProductsResponse {
    products: [Product!]!
}

type Product @key(fields: "id") {
  id: ID!
}
```

Today, adding `@lookup` on `products` would fail since the output isn't an entity. So added support for namespaces, where we automatically retrieve from a nested field the entities.


tomhoule (approved):
Perfect, thank you!

PR #3385, opened 2025-08-15
graphql-composition: compose with mixed fedv1 and entity interfaces

tomhoule:
closes GB-9530


crates/graphql-composition/tests/composition/federation_v2/entity_interfaces/entity_interface_basic/subgraphs/forest.graphql
```diff
@@ -0,0 +1,14 @@
+extend schema
+  @link(
+    url: "https://specs.apollo.dev/federation/v2.3"
+    import: ["@key", "@external", "@requires", "@provides", "@interfaceObject"]
+  )
```
tomhoule, line 5: These just moved, with added imports, since they're federation v2. We don't want to treat them like federation v1 subgraphs.

PR #3461, opened 2025-08-27
Adding a de-duplication id in the query-solver

Finistere:
First step of the deduplication algorithm, adding an id that uniquely identifies equal fields/resolvers

Also reverting the query position to be a u16, wanted to separate but somehow got mixed. The GraphQL spec only specifies that fields _should_ be in the right order. So to maximize our chances, I'm resetting the query position at every new depth level when I'm generating those traversing the query in BFS order. So we'll only return incorrect order if there is more than `u16::MAX` fields at the same depth.

And I'm reducing the size of `ResponseValueId` used in response object paths even further.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2963,6 +2963,7 @@ dependencies = [
  "petgraph",
  "pretty_assertions",
  "priority-queue",
+ "rapidhash",
  "serde",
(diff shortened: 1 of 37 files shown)


tomhoule (approved):
Solid!

PR #3493, opened 2025-09-04
Traffic shaping

Finistere:
Adding:
```toml
[traffic_shaping]
inflight_deduplication = true
```


tomhoule (approved):
Nice! I didn't expect it to be this straightforward.

PR #3544, opened 2025-09-25
graphql-composition: deprecate ingest_loaded_extension()

tomhoule:
The `grafbase-extensions` cargo feature and the `Subgraphs::ingest_loaded_extensions()` method are removed, or more accurately hidden from the public API. Adjustments are made so composition of Grafbase extensions is now transparent based on the `@link` url, without hints from the caller of composition.


crates/graphql-composition/src/subgraphs/extensions.rs
```diff
@@ -9,10 +9,6 @@ pub(crate) struct ExtensionRecord {
 }
 
 impl Subgraphs {
-    pub(crate) fn extension_is_defined(&self, name: StringId) -> bool {
```
Finistere, line 12: Why removed?

crates/graphql-composition/src/ingest_subgraph/schema_definitions/link.rs
```diff
@@ -154,9 +150,20 @@ fn ingest_grafbase_extension_from_link(
         let Some(name) = r#as.or_else(|| {
             let mut segments = link_url.url.path_segments()?;
 
-            segments.next_back()?;
+            let name = segments.next_back()?;
```
Finistere, line 153: Just in case, I'm changing this to use the same name as the link

Finistere (approved):

crates/graphql-composition/src/ingest_subgraph/schema_definitions/link.rs
tomhoule: Yeah, I can remove that change, doesn't matter.

crates/graphql-composition/src/subgraphs/extensions.rs
tomhoule: The duplicate extensions check is pushed back further. I'm still not 100% happy and settled with a solution for deduplicating extensions from the subgraphs.

PR #3552, opened 2025-09-26
rework link url handling

Finistere:


crates/graphql-composition/src/ingest_subgraph/schema_definitions/link.rs
```diff
@@ -124,17 +124,28 @@ fn ingest_grafbase_extension_from_link(
...
-                match file_name {
-                    // This is going to be the name for locally built extension. In that case, the parent directory has a more descriptive name.
-                    "build" => path
-                        .parent()
-                        .and_then(|parent| parent.file_name()?.to_str())
-                        .map(|s| subgraphs.strings.intern(s)),
-                    _ => Some(subgraphs.strings.intern(file_name)),
-                }
-            })
+            let mut segments = link_url.url.path_segments()?;
```
tomhoule, line 127: Does that work for file urls? I guess so, if the tests are ok

Finistere: yeah same thing

PR #3553, opened 2025-09-26
graphql-composition: handle namespace prefixes in composeDirective

tomhoule:
Previously, something like this would not result on the `@reports__observation` directive to be in the federated graph:

```graphql
extend schema
  @link(url: "https://specs.apollo.dev/federation/v2.3", import: ["@composeDirective"])
  @link(url: "https://specs.mydomain.com/specs/reports/v1.0")
  @composeDirective(name: "@reports__observation")
  @reports__observation(test: "argument")
```

This commit fixes that.


Finistere (approved):
thanks

tomhoule:
Pushed some improvements in how linked schemas are handled in the composed schema.

